Issue #5, opened 2020-04-22
Fix Process.Start of url in about tab in settings


jjw24:
![image](https://user-images.githubusercontent.com/26427004/79972695-b2c13e80-84d9-11ea-995b-e3981113d773.png)

PR #13, opened 2020-04-23
Fix Process.Start on urls and directories

jjw24:
Previous .NetFramework Process.Start code needs to be changed now that it's upgraded to Core when opening URLs and directories

Fix:
1. Change to open using SearchWeb browser command.
2. Change to open FilesFolders command

Closes https://github.com/Flow-Launcher/Flow.Launcher/issues/5


JsonRPC/wox.py
```diff
@@ -4,9 +4,9 @@
 import sys
 import inspect
 
-class Wox(object):
+class Flow.Launcher(object):
```
SysC0mp, line 7: I do not know a lot about python, but is it valid to use a dot in a class name?

jjw24: good pick up. will fix that up in the other PR.

PR #15, opened 2020-04-25
Fix substring scoring

jjw24:
Problem context:
Previously search results such as searching 'code' for 'Visual Studio Codes' does not make the regular precision cut as it does not get the bonus, where as both 'cod' and 'codes' do make the cut.

Solution:
Less than 5 chars scoring is 10 per char, else onward 5 per char


jjw24:
From JJW24/Wox fork

image

PR #23, opened 2020-04-29
Folder plugin- Recursive delete sub folders and files

jjw24:
Current folder delete function in Folder plugin does not delete folders that contain files. 

Fix by adding true to recursive delete

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Folder/ContextMenuLoader.cs
+++ b/Plugins/Flow.Launcher.Plugin.Folder/ContextMenuLoader.cs
@@ -89,7 +89,7 @@ namespace Flow.Launcher.Plugin.Folder
                                 if (record.Type == ResultType.File)
                                     File.Delete(record.FullPath);
                                 else
-                                    Directory.Delete(record.FullPath);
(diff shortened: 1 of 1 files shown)


Plugins/Flow.Launcher.Plugin.Folder/ContextMenuLoader.cs
```diff
@@ -89,7 +89,7 @@ public List<Result> LoadContextMenus(Result selectedResult)
                                 if (record.Type == ResultType.File)
                                     File.Delete(record.FullPath);
                                 else
-                                    Directory.Delete(record.FullPath);
+                                    Directory.Delete(record.FullPath, true);
```
NickSeagull, line 92: Perhaps adding an explanation like:

```suggestion
                                    Directory.Delete(record.FullPath, true); // `true` so it deletes recursively
```

jjw24:
thanks guys. @NickSeagull I also see comments as explaining something the the code does not itself explain. While on commit like this it is not obvious, but when others see the code in IDE it will be obvious. If ok I will not make the suggested change and proceed to merge.

PR #24, opened 2020-04-30
fix typo in IPublicAPI

JohnTheGr8:
Replace `IPublicAPI.RestarApp` with `IPublicAPI.RestartApp` and mark `IPublicAPI.RestarApp` obsolete.


jjw24:
@JohnTheGr8 thanks for the change.  Would it be feasible to conclude by searching for the type in github and if no results, that this is not being used by plugins? Or have you already done that?

JohnTheGr8:
@jjw24 The old method is still there but marked obsolete, this shouldn't be a problem. A quick Github code search gives back forks of Wox...

PR #25, opened 2020-04-30
Add support for F# plugins

JohnTheGr8:
It is trivial to write plugins in F# for Wox already, and I have written a few in the past ([example](https://github.com/JohnTheGr8/Wox.Plugin.Bang)). The only trick was to specify `csharp` in the plugin.json file.

This PR:
* adds F# to the supported languages in `PluginsLoader`
* adds `fsharp` as a supported value of the `Language` field in `plugin.json`
* adds a sample hello-world plugin project written in F#

Cheers


jjw24:
This is awesome add. 

Soon we will be able to add VB .Net to the supported list :P :P

PR #28, opened 2020-05-01
Fix thumbnail image loading

SysC0mp:
Should fix the missing thumbnail images.
ImageLoader class is not thread-safe at it's current state and the icons cannot be loaded in parallel.


JohnTheGr8:
lgtm

theClueless:
@SysC0mp what can happen if it is run in parallel? what is not thread safe? I agree that it can load some images more than once but the amount of time it will invest in doing it serially seems worse to me.

jjw24:
I have the same concern as @theClueless. Might need to investigate a bit more to see if there is an improved solution, otherwise it might take longer to load if user has a lot of images.

jjw24:
@SysC0mp could you share the error in the logs for those images not loaded please.

SysC0mp:
Following error is thrown for a couple of thumbnails on Flow startup: 
image
image
The images which throw that always differ. But I am curious because those errors should occur on your machines too directly when you start Flow using the Debug config from the dev branch. 

Call stack: 
image

JohnTheGr8:
Was just testing this @SysC0mp and I am getting this error as well. The error goes away for me if I set `ThumbnailOptions` to `ThumbnailOnly`.

I'll make a suggestion commit now so you can check it out yourself.

JohnTheGr8:
@SysC0mp committed, give it a try

SysC0mp:
Thanks for testing @JohnTheGr8 but unfortunately I still get the errors with that option provided.

JohnTheGr8:
Well I just tried again and got the error myself, what the hell. I tried a few times before and the change to `ThumbnailOptions` did the trick, but now it fails again. 
Sorry for wasting your time.

SysC0mp:
All good, any testing & help is appreciated 👍  Maybe we could try to add some kind of "retry" logic to it because it seems that this error occurs only the first time trying for some of the thumbnails. I'll give that a try.

SysC0mp:
I changed the image loading back to parallel and added a retry logic to the whole `LoadInternal` method. In that way, the error is fixed for me and we do not need to sequentially load the thumbnails.

PR #29, opened 2020-05-01
Fix fsharp plugin icon path

SysC0mp:
The plugin.json file was missing an IcoPath which led to errors on Flow startup.


JohnTheGr8:
@SysC0mp there is no `app.png` file in the plugin's folder...

SysC0mp:
You are right... 
It defaults currently to the missing icon image. We can leave the IcoPath either empty for now or i'll add the default app.png that is used in the HelloWorldCSharp sample, but any IcoPath is needed. Otherwise, PluginMetadata.PluginDirectory will throw an error.

JohnTheGr8:
@SysC0mp Alright. I didn't know `PluginMetadata.PluginDirectory` throws an error when `IcoPath` is missing. Perhaps we should fix that...

SysC0mp:
Yes, we should definitely investigate why that error appears and if it should throw that. I think we need to make sure in some way that every plugin provides a unique icon, but i am not yet sure on how to do that the best way.

PR #30, opened 2020-05-02
Dependency updates

JohnTheGr8:


jjw24:
I am not near a computer at the moment, so will have to have a look tomorrow. 

Can you test the release build and publishing self contained that packages can be built. I remember I had an issue during the .Net core migration I wasn't able output package when I upgraded propertychang.fody. see if that's still an issue please.

jjw24:
still getting some languages through when publishing:
Maybe we need to add SatelliteResourceLanguages en to all other projects?

image

jjw24:
> still getting some languages through when publishing:
> Maybe we need to add SatelliteResourceLanguages en to all other projects?

other than this everything else lgtm

JohnTheGr8:
@jjw24 these are normal when publishing, and not related to `FSharp.Core`. You can publish from the dev branch and see that they're getting created there as well...

jjw24:
> @jjw24 these are normal when publishing, and not related to `FSharp.Core`. You can publish from the dev branch and see that they're getting created there as well...

Yeah I mean maybe we can add it in for other projects because my understanding of its use is able to exclude publishing of localized resources (cs, de, es, fr, etc.) Let me try and see if it works.

jjw24:
ok they are excluded. I will put that through as separate PR so we can keep this one clean.

PR #32, opened 2020-05-03
Folder plugin: handle environment string paths

### LINKED ISSUE #20: Support %appdata% in Folder plugin
As an example, I type I %appdata% Folder plugin will switch that text to the appdata director + \\ so i can continue to type the folder I want to go to:
Type %appdata%\Microsot it will change the text to c:\users\blahuser\appdata\microsoft\

JohnTheGr8:
Closes #20 

The folder plugin now recognizes and handles environment string paths.

### Behavior

1. Simply typing `%` will list all supported environment string paths (listed below).
2. Typing after the first `%` will filter the results
3. Typing a full path containing an environment string will replace it with the actual path and perform a normal path search

### Screenshots

env_string_search

appdata_search

### Supported environment string paths
```
%appdata%
%localappdata%
%programfiles%
%programfiles%
%programdata(x86)%
%userprofile%
%temp%
```

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Folder/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Folder/Main.cs
@@ -27,6 +27,8 @@ namespace Flow.Launcher.Plugin.Folder
         private readonly PluginJsonStorage<Settings> _storage;
         private IContextMenu _contextMenuLoader;
 
+        private static Dictionary<string, string> _envStringPaths;
+
         public Main()
         {
             _storage = new PluginJsonStorage<Settings>();
@@ -48,6 +50,7 @@ namespace Flow.Launcher.Plugin.Folder
             _context = context;
             _contextMenuLoader = new ContextMenuLoader(context);
             InitialDriverList();
+            LoadEnvironmentStringPaths();
         }
 
         public List<Result> Query(Query query)
@@ -58,7 +61,14 @@ namespace Flow.Launcher.Plugin.Folder
             if (!IsDriveOrSharedFolder(search))
                 return results;
 
-            results.AddRange(QueryInternal_Directory_Exists(query));
+            if (search.StartsWith("%"))
+            {
+                results.AddRange(GetEnvironmentStringPathResults(search, query));
+            }
+            else
+            {
+                results.AddRange(QueryInternal_Directory_Exists(query.Search, query));
+            }
 
             // todo why was this hack here?
             foreach (var result in results)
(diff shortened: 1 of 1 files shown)


Plugins/Flow.Launcher.Plugin.Folder/Main.cs
```diff
@@ -146,14 +161,37 @@ private void InitialDriverList()
             }
         }
 
+        private void LoadEnvironmentStringPaths()
+        {
+            _envStringPaths = new Dictionary<string, string>();
+            
+            var specialPaths =
```
SysC0mp, line 168: Maybe we could use `System.Environment.GetEnvironmentVariables()` here to get them dynamically loaded on each system?

SysC0mp: Apart from that I just tested it and it is working great! :)

JohnTheGr8: `GetEnvironmentVariables()` will return all sorts of variables, way more than the valid environment string paths like `%appdata%`. 

I considered loading them dynamically, but it seems there is no good way of doing it. `Environment.SpecialFolder` has some of them but not all (for example `temp` is missing) and it also has a lot of other special folders that do not correspond to valid environment-string paths.

JohnTheGr8: That said, we can add a couple more common ones like `%public%` and `%windir%` and cover 99% of use-cases...

SysC0mp: I am currently testing the usage of `System.Environment.GetEnvironmentVariables()`. We can check the results of that with `Directory.Exists()` and only add those to the dictionary. This also includes `%temp%` for example. In that way, we would also get user-defined environment variables, if I'm not mistaking.

SysC0mp: I could check-in my current test code for you to expand on it if you like.

JohnTheGr8: We could merge this and discuss your changes in a subsequent PR.

SysC0mp: Alright, I'll open another PR after this if you want.

JohnTheGr8: Thanks. We'll probably also need some reload logic if we go the env-var way, so let's keep them separate....

jjw24:
Lgtm.

This plugin's main is getting large. We shouldn't chuck everything in main for new plugins going forward, and refactor existing ones later.

PR #34, opened 2020-05-03
Enhance modifier + num launching

gdziedzic:
# Enhance modifier + num launching

* Adding the ability to set custom modifiers in settings
image
* as a result user see this
image

# Next steps
* in the future we might show the exact hotkey on the list (instead of _Mod+<num>_ . I haven't added it right now because it's a bit more complex and I wanted to get feedback from the users whether it doesn't make it too cluttered. 
* alternatively we could step back and take a more minimalistic approach and display just the number
image

### DIFF
--- a/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
+++ b/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
@@ -139,7 +139,7 @@ namespace Flow.Launcher.ViewModel
             }
         }
 
-        public List<string> OpenResultModifiersList => new List<string> { "Alt", "Ctrl", "Ctrl+Alt" };
+        public List<string> OpenResultModifiersList => new List<string> { KeyConstant.Alt, KeyConstant.Ctrl, $"{KeyConstant.Ctrl}+{KeyConstant.Alt}" };
         private Internationalization _translater => InternationalizationManager.Instance;
         public List<Language> Languages => _translater.LoadAvailableLanguages();
         public IEnumerable<int> MaxResultsRange => Enumerable.Range(2, 16);
--- a/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
+++ b/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
@@ -139,6 +139,7 @@ namespace Flow.Launcher.ViewModel
             }
         }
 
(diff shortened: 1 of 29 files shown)


JohnTheGr8:
Haven't looked at the code yet, but this looks nice.

> in the future we might show the exact hotkey

Agreed.

> alternatively we could step back and take a more minimalistic approach and display just the number

Suggestion: we should let users completely hide this if they are comfortable with the shortcuts.

SysC0mp:
Great addition, love it. I was also able to test it without any issues. 👍 

> in the future we might show the exact hotkey

That would be great because the Mod+[1-9] could confuse users. We could also make this a part of each theme so that users could make it larger / smaller like they want to. At the moment I think it is a bit too big, but that is completely personal preference.

> Suggestion: we should let users completely hide this if they are comfortable with the shortcuts.

I also agree with that because I think it is always the best option to let the user decide how they want it to behave.

JohnTheGr8:
> At the moment I think it is a bit too big

Agreed, perhaps it should be same font size as the subtitle and aligned with it by default?

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
```diff
@@ -139,6 +139,7 @@ public List<string> QuerySearchPrecisionStrings
             }
         }
 
+        public List<string> OpenResultModifiersList => new List<string> { "Alt", "Ctrl", "Ctrl+Alt" };
```
jjw24, line 142: great addition @gdziedzic.  Nitpick here, could we put the strings in Constant? I see them referenced more than once.

Also as @JohnTheGr8 mentioned, could we add an option to hide the modifier key display on main window please

gdziedzic:
> Also as mentioned, could we add an option to hide the modifier key display on main window please
image

* Added ability to hide
* Displaying full hotkey
image
* changed font style to subitem as suggested
* extracted common consts

SysC0mp:
Looking really nice now, great work! 👍 

One thing I stumbled upon while testing: 
If I set the "Maximum results shown" to something higher than 9, the modifier will still be shown but cannot be used. I guess the maximum shown modifier should be 9. 

image

gdziedzic:
> , the modifier will still be shown but cannot be used. I guess the maximum shown modifier should be 9.

Good spot thank you! Fixed now

image

SysC0mp:
Thank you, looking fine to me now 🥳

PR #36, opened 2020-05-04
New look

jjw24:
New look for:
- Main search window
- Settings window and tabs

Styling applied at App.xaml and carries through for plugins downloaded


jjw24:
wrong branch

Issue #11, opened 2020-04-22
New look for the launcher


jjw24:
New UI for:

- main query window

- settings page for the launcher and plugins

Suggestions and mock-ups are very welcome

PR #37, opened 2020-05-04
Launcher appearance uplift

jjw24:
New look for:

- Main search window
- Settings window and tabs

Styling applied at App.xaml and carries through for plugins downloaded

update

Closing https://github.com/Flow-Launcher/Flow.Launcher/issues/11


jjw24:
@Flow-Launcher/team

JohnTheGr8:
idea: when the results are all from a single plugin, the magnifying glass icon could be replaced with said plugin's icon.

JohnTheGr8:
another minor idea: since the plugin description text is almost always too long to fit in the plugin listing (left side in plugins tab), perhaps we should replace it with something shorter... the version of the plugin, maybe?

SysC0mp:
Looking really good, @jjw24! 💯

SysC0mp:
> another minor idea: since the plugin description text is almost always too long to fit in the plugin listing (left side in plugins tab), perhaps we should replace it with something shorter... the version of the plugin, maybe?
That is a good point. I think we could also make the settings window a bit bigger as default, at the moment it is rather small per default and not resizable.

SysC0mp:
One thing I noticed while testing: The whole UI design depends on the color mode chosen in Windows 10 i guess. I have the dark mode activated, and the settings look like this: 
image
image
Maybe you already noticed that, please ignore me then. 😄

jjw24:
Shadow effect implementation

With shadow effect(left) vs without
shadow


I noticed that adding shadow effect on main search window has a drain on GPU usage. I quite like having the effect on, do you guys know if there is a better approach to implement it?



With shadow effect:
with shadow

without:
without shadow 3

jjw24:
> One thing I noticed while testing: The whole UI design depends on the color mode chosen in Windows 10 i guess. I have the dark mode activated, and the settings look like this:

> It looks alright using the light mode. Maybe you already noticed that, please ignore me then. 😄

Good pick up @SysC0mp I did not noticed that. 

Very strange though I did not know dark mode would have such effect on the new UI.

jjw24:
> That is a good point. I think we could also make the settings window a bit bigger, at the moment it is rather small per default and not resizable. That would also allow us to move the tab pages to the left of the screen later on and get some tab design like VSCode or Alfred does it, which I personally really like. What do you think?

I agree, settings window needs to be bigger, it can get really cluttered. I like Alfred's approach personally. I think we can do this as a seperate PR so this change can establish a good baseline first, also so I dont spend more time on resizing and adjusting the layout :P

jjw24:
> another minor idea: since the plugin description text is almost always too long to fit in the plugin listing (left side in plugins tab), perhaps we should replace it with something shorter... the version of the plugin, maybe?

Plugin version, action key word or just blank to reduce the clutter?

JohnTheGr8:
> Agree. Plugin version, action key word or just blank to reduce the clutter?

We could go with keyword there, but I think the version should be mentioned somewhere as well. It is currently not displayed everywhere...

jjw24:
> idea: when the results are all from a single plugin, the magnifying glass icon could be replaced with said plugin's icon.

Like this?
image

SysC0mp:
> > another minor idea: since the plugin description text is almost always too long to fit in the plugin listing (left side in plugins tab), perhaps we should replace it with something shorter... the version of the plugin, maybe?
> 
> Agree. Plugin version, action key word or just blank to reduce the clutter?

I would prefer the version there atm as the action keyword can be the same for a lot of plugins (*) and we could show that on the side of every plugin in a later change, but I am happy with every option. 👍

jjw24:
> > That is a good point. I think we could also make the settings window a bit bigger, at the moment it is rather small per default and not resizable. That would also allow us to move the tab pages to the left of the screen later on and get some tab design like VSCode or Alfred does it, which I personally really like. What do you think?
> 
> I agree, settings window needs to be bigger, it can get really cluttered. I like Alfred's approach personally. I think we can do this as a seperate PR so this change can establish a good baseline first, also so I dont spend more time on resizing and adjusting the layout :P

Come to think of it, if we resize the settings window, would that impact existing plugins' layout? @JohnTheGr8

JohnTheGr8:
> I would prefer the version there atm as the action keyword can be the same for a lot of plugins (*) and we could show that on the side of every plugin in a later change, but I am happy with every option. 👍

Ok, let's go with version.

> Come to think of it, if we resize the settings window, would that impact existing plugins' layout? @JohnTheGr8 

Honestly, I have not worked on any plugins with UI for Wox in the past, so I don't have experience. We could certainly test it out and see what happens. 

That said, I think that if we're going to break the layout of plugins, now is the best time to do it.

SysC0mp:
> > > That is a good point. I think we could also make the settings window a bit bigger, at the moment it is rather small per default and not resizable. That would also allow us to move the tab pages to the left of the screen later on and get some tab design like VSCode or Alfred does it, which I personally really like. What do you think?
> > 
> > 
> > I agree, settings window needs to be bigger, it can get really cluttered. I like Alfred's approach personally. I think we can do this as a seperate PR so this change can establish a good baseline first, also so I dont spend more time on resizing and adjusting the layout :P
> 
> Come to think of it, if we resize the settings window, would that impact existing plugins' layout? @JohnTheGr8

That is hard to say i guess as every plugin does it's own styling. If some of the controls on the settings page are right-aligned for example, the change could create some structural problems. But as @JohnTheGr8 says, now is the time 💪 😄

jjw24:
> > > > That is a good point. I think we could also make the settings window a bit bigger, at the moment it is rather small per default and not resizable. That would also allow us to move the tab pages to the left of the screen later on and get some tab design like VSCode or Alfred does it, which I personally really like. What do you think?
> > > 
> > > 
> > > I agree, settings window needs to be bigger, it can get really cluttered. I like Alfred's approach personally. I think we can do this as a seperate PR so this change can establish a good baseline first, also so I dont spend more time on resizing and adjusting the layout :P
> > 
> > 
> > Come to think of it, if we resize the settings window, would that impact existing plugins' layout? @JohnTheGr8
> 
> That is hard to say i guess as every plugin does it's own styling. If some of the controls on the settings page are right-aligned for example, the change could create some structural problems. But as @JohnTheGr8 says, now is the time 💪 😄

Could we break this convo into a separate issue and you guys provide some examples like from Alfred (I dont have Mac OS) and others on how large the size should be and how the layout could look please.

JohnTheGr8:
Agreed. Let's focus this PR on fixing the dark-mode issue and getting it approved.

jjw24:
> Agreed. Let's focus this PR on fixing the dark-mode issue and getting it approved.

So @JohnTheGr8 @SysC0mp for this PR are we good to:
- Focus on fixing of dark-mode issue
- Change side bar plugin description to version
- Change the magnifying glass icon to plugin icon when all plugins the same
- Change the magnifying glass icon to spinning progress bar when query runs longer, and remove existing progress bar line (I added this in as I have always want to to do this and now since we are adding the above to magnifying glass icon placeholder)
- Create a separate issue for resizing settings window, gather some examples of layout then PR seperately

JohnTheGr8:
> * Change the magnifying glass icon to plugin icon when all plugins the same
> * Change the magnifying glass icon to spinning progress bar when query runs longer, and remove existing progress bar line (I added this in as I have always want to to do this and now since we are adding the above to magnifying glass icon placeholder)

Great idea as well. Would it be too much trouble to split these to a separate PR as well? Not a big deal, either way.

jjw24:
> > * Change the magnifying glass icon to plugin icon when all plugins the same
> > * Change the magnifying glass icon to spinning progress bar when query runs longer, and remove existing progress bar line (I added this in as I have always want to to do this and now since we are adding the above to magnifying glass icon placeholder)
> 
> Great idea as well. Would it be too much trouble to split these two to a separate PR as well? Not a big deal, either way.

Good suggestion, lets separate these two to separate PR as well.

jjw24:
Anyone @Flow-Launcher/team got a win 7 machine/vm that can check of the new UI is OK on the OS?

jjw24:
Build for testing the UI
https://github.com/Flow-Launcher/Flow.Launcher/releases/download/0.0.1/Flow.Launcher-0.0.1.exe

SysC0mp:
Had no chance to look at the code yet, but I was using the new UI / version since yesterday and it is working flawlessly so far.

jjw24:
@Flow-Launcher/team this PR good to go or we still need time to test?

Flow.Launcher/ResultListBox.xaml
```diff
@@ -91,6 +91,7 @@
                     Value="True">
                     <Setter TargetName="Title" Property="Style" Value="{DynamicResource ItemTitleSelectedStyle}" />
                     <Setter TargetName="SubTitle" Property="Style" Value="{DynamicResource ItemSubTitleSelectedStyle}" />
+                    <Setter TargetName="imgIco" Property="Style" Value="{DynamicResource ItemImageSelectedStyle}" />
```
SysC0mp, line 94: Small nitpick: Maybe use the same naming style here as Title / Subtitle. ImageIcon instead of imgIco

Flow.Launcher/SettingWindow.xaml
```diff
@@ -212,20 +216,40 @@
...
                     </Grid.RowDefinitions>
-                    <StackPanel Background="{Binding PreviewBackground}" Grid.Row="0" Margin="0">
+                    <StackPanel Grid.Row="0">
+                        <Grid>
+                            <Grid.RowDefinitions>
+                                <RowDefinition />
+                                <RowDefinition Height="15"/>
+                                <RowDefinition Height="30"/>
+                            </Grid.RowDefinitions>
+                            <TextBlock Grid.Row="0" Text="Query window shadow effect" Margin="0 30 0 0" FontSize="14" />
```
SysC0mp, line 230: We should use resources here instead of fixed text

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Views/SettingsControl.xaml
```diff
@@ -6,19 +6,26 @@
...
-            <RadioButton Name="NewWindowBrowser" GroupName="OpenSearchBehaviour" Content="New window" Click="OnNewBrowserWindowClick" Margin="10" />
-            <RadioButton Name="NewTabInBrowser" GroupName="OpenSearchBehaviour" Content="New tab" Click="OnNewTabClick" Margin="10" />
+        <StackPanel>
+            <Grid Grid.Row="0" Margin="40 40 0 0">
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="160" />
+                    <ColumnDefinition Width="140"/>
+                    <ColumnDefinition Width="100"/>
+                </Grid.ColumnDefinitions>
+                <Label Content="Open bookmarks in:" FontSize="15" Grid.Column="0" Margin="0 5 0 0"/>
```
SysC0mp, line 21: Resource instead of static text

Plugins/Flow.Launcher.Plugin.WebSearch/SettingsControl.xaml
```diff
@@ -55,12 +70,12 @@
...
             </ListView.View>
         </ListView>
-        <StackPanel Grid.Row="3" HorizontalAlignment="Right" Orientation="Horizontal">
+        <StackPanel Grid.Row="4" HorizontalAlignment="Right" Orientation="Horizontal">
             <Button Click="OnDeleteSearchSearchClick" Width="100" Margin="10"
                     Content="{DynamicResource flowlauncher_plugin_websearch_delete}" />
             <Button Click="OnEditSearchSourceClick" Width="100" Margin="10"
                     Content="{DynamicResource flowlauncher_plugin_websearch_edit}" />
-            <Button Click="OnAddSearchSearchClick" Width="100" Margin="10"
+            <Button Click="OnAddSearchSearchClick" Width="100" Margin="10 10 0 10"
```
SysC0mp, line 78: Maybe add a margin to the right as well so that the button is not completely right aligned. 
```suggestion
            <Button Click="OnAddSearchSearchClick" Width="100" Margin="10 10 10 10"
```

jjw24:
> Apart from those little things, I used the new UI the last couple days and it is still working without any flaws so far.

Good pickup. Fixed

jjw24:
Cheers. I will create those other to do as issues and work on them once I am done with other pr

Flow.Launcher/SettingWindow.xaml
```diff
@@ -32,70 +33,72 @@
...
         <TabItem Header="{DynamicResource general}">
-            <StackPanel Orientation="Vertical">
-                <CheckBox Margin="10" IsChecked="{Binding PortableMode}">
-                    <TextBlock Text="Portable Mode" />
-                </CheckBox>
-                <CheckBox Margin="10" IsChecked="{Binding Settings.StartFlowLauncherOnSystemStartup}"
+            <ScrollViewer ui:ScrollViewerHelper.AutoHideScrollBars="True" Margin="60,30,0,30">
+                <StackPanel Orientation="Vertical">
+                    <ui:ToggleSwitch Margin="10" IsOn="{Binding PortableMode}">
+                        <TextBlock Text="Portable Mode" />
```
JohnTheGr8, line 39: small nitpick that I spotted after approving: this should be a resource too

jjw24: Good pick up. I will make an issue and fix tonight. I am not able to right now.

PR #48, opened 2020-05-14
Fix translations for the main project

JohnTheGr8:
The translation files from `Flow.Launcher/Languages` are currently not included in the build output and are therefore not available at runtime. 
Changing the language from settings only affects the plugin-specific fields in the `Plugin` tab, because the plugin projects are copying their language files correctly.


jjw24:
Good pick up, thank you

PR #35, opened 2020-05-04
Packaging updates and AppVeyor integration

JohnTheGr8:
### Changes

**Plugin project packing:**
* moved the package properties (id, authors, etc) from nuspec to the project file
* the nuget package is produced on release builds automatically (along with a symbols file)

**AppVeyor**:
* the main project files are patched with the build version
* the nuget packages for the plugin are packed as artifacts
* updated the custom post-build script

### Considerations for discussion

**Versioning:**
How do we want to deal with versioning? The current way used by Wox is to use `x.x.{build}` as a version that applies to all projects (the launcher itself, the plugin project, and all default plugins). This leads to subsequent releases having a big jump in the version number (for example `v1.3.981` -> `v1.3.996`). If we want to change to a different versioning scheme, we should probably decide this now.

**Squirrel**:
Are we going to use Squirrel for installation/updates? I have not looked into it much, but it seems to not behave too well with .net core.

### TODO

- [x] integrate sourcelink for the plugin project
- [ ] remove AssemblyInfo files?

(related to #31)

PR #52, opened 2020-05-18
Lower PropertyChange.Fody version to fix runtime failure

jjw24:
Dev build runtime start up failure when `%userprofile%\.nuget\packages\propertychanged.fody` folder is deleted:
"
Description: A .NET Core application failed.
Application: Flow.Launcher.exe
Path: C:\Users\Blah\AppData\Local\Flow\app-1.0.0\Flow.Launcher.exe
Message: Error:
An assembly specified in the application dependencies manifest (Flow.Launcher.deps.json) was not found:
package: 'PropertyChanged.Fody', version: '3.2.8'
path: 'lib/netstandard1.0/PropertyChanged.dll'
"

Reduce the PropertyChange.Fody version.

### DIFF
--- a/Flow.Launcher.Core/Flow.Launcher.Core.csproj
+++ b/Flow.Launcher.Core/Flow.Launcher.Core.csproj
@@ -57,7 +57,7 @@
     <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="squirrel.windows" Version="1.5.2" />
-    <PackageReference Include="PropertyChanged.Fody" Version="3.2.8">
+    <PackageReference Include="PropertyChanged.Fody" Version="2.2.4">
       <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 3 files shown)


jjw24:
Just putting out a temporary prerelease so it's not broken, while this is under review.

JohnTheGr8:
@jjw24 do we actually use `PropertyChanged.Fody`?

I don't see any uses of `INotifyPropertyChanged` or the [`AddINotifyPropertyChangedInterface Attribute`](AddINotifyPropertyChangedInterfaceAttribute) anywhere, except in `Flow.Launcher.Plugin/BaseModel.cs`... but this file is modified because of `JetBrains.Annotations`. 

I have looked into the compiled `dll` with and without `PropertyChanged.Fody` and cannot notice any difference.

JohnTheGr8:
That said, I have been trying the various versions of `PropertyChanged.Fody` and noticed that version `2.5.13` is the latest version that does not cause runtime errors...

jjw24:
@JohnTheGr8 happy to remove if not used, there are bits of code that are left overs which I am keen to get rid off. I will have a play and put in a commit if via to do so.

JohnTheGr8:
the package has  been downgraded to 2.5.13 in #35

Issue #57, opened 2020-05-20
Flow doesn't start when loading plugin fails


DonKonfetti:
### Are you submitting a bug report?

Yes

### Steps to reproduce

1. Install Wox processkiller plugin
2. Restart Flow
3. Nothing happens, Flow silently crashes while starting

### Detailed logs

I've used dnSpy to look into this issue. I see that loading the processkiller plugin fails and some unhandled exceptions occurs. This is the relevant content of the output window:

```
14:10:31.671 Flow.Launcher.exe (CoreCLR: clrhost): 'C:\Users\schwarz\AppData\Roaming\FlowLauncher\Plugins\Wox.Plugin.ProcessKiller-952489db-c7fc-4f54-b718-81a1f74467e0\Wox.Plugin.ProcessKiller.dll' geladen.
14:10:31.717 Ausnahme ausgelöst: "System.Reflection.ReflectionTypeLoadException" in System.Private.CoreLib.dll
14:10:31.719 Zustatzinformation: Unable to load one or more of the requested types.
14:10:31.795 Flow.Launcher.exe (CoreCLR: clrhost): 'C:\Users\schwarz\AppData\Local\FlowLauncher\app-0.9.1\System.Reflection.Metadata.dll' geladen.
14:10:31.798 Flow.Launcher.exe (CoreCLR: clrhost): 'C:\Users\schwarz\AppData\Local\FlowLauncher\app-0.9.1\System.Collections.Immutable.dll' geladen.
14:10:31.969 Flow.Launcher.exe (CoreCLR: clrhost): 'C:\Users\schwarz\AppData\Local\FlowLauncher\app-0.9.1\de\PresentationFramework.resources.dll' geladen.
14:10:32.013 Ausnahme ausgelöst: "System.IO.IOException" in PresentationFramework.dll
14:10:32.016 Zustatzinformation: Die Ressource "images/app_error.png" kann nicht gefunden werden.
14:10:32.062 Ausnahme ausgelöst: "System.IO.IOException" in PresentationCore.dll
14:10:32.064 Zustatzinformation: Die Ressource "images/app_error.png" kann nicht gefunden werden.
14:10:32.104 Flow.Launcher.exe (CoreCLR: clrhost): 'C:\Users\schwarz\AppData\Local\FlowLauncher\app-0.9.1\de\System.Xaml.resources.dll' geladen.
[...]

PR #58, opened 2020-05-20
Add handling of reflection type error

jjw24:
Make plugin load failure user friendly. When a plugin loading fails, inform the user.

https://github.com/Flow-Launcher/Flow.Launcher/issues/57


Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -50,27 +53,43 @@ public static IEnumerable<PluginPair> DotNetPlugins(List<PluginMetadata> source)
...
                     try
                     {
+                        var types = assembly.GetTypes();
+                        
                         type = types.First(o => o.IsClass && !o.IsAbstract && o.GetInterfaces().Contains(typeof(IPlugin)));
                     }
                     catch (InvalidOperationException e)
                     {
+                        erroredPlugins.Add(metadata.Name);
+
```
NickSeagull, line 72: Love these lines, so simple!

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -50,27 +53,43 @@ public static IEnumerable<PluginPair> DotNetPlugins(List<PluginMetadata> source)
...
                         Log.Exception($"|PluginsLoader.DotNetPlugins|Can't find class implement IPlugin for <{metadata.Name}>", e);
                         return;
                     }
+                    catch (ReflectionTypeLoadException e)
+                    {
+                        erroredPlugins.Add(metadata.Name);
+
+                        Log.Exception($"|PluginsLoader.DotNetPlugins|The GetTypes method was unable to load assembly types for <{metadata.Name}>", e);
```
NickSeagull, line 80: Where this error will be seen? If the user will see it, perhaps we could try making it to look as a solution, like pointing the user where to fix it?

jjw24: All those errors were logged previously and not visible to the user. And this uncaught exception would just fail the app completely.

For this PR, it simply puts out a message to let user know the failed plugins, and continue to start up, I think it's the simplest approach.

Failure of loading could be due to a number of reasons , but none a non dev user would not know how to fix it. I think reporting it might be the best approach?

NickSeagull: We could add something like "This is an error caused by the plugin X" so the user knows that it is something wrong with a plugin, rather than with their installation

jjw24: This is ok?

image

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -85,6 +104,19 @@ public static IEnumerable<PluginPair> DotNetPlugins(List<PluginMetadata> source)
...
+
+            if (erroredPlugins.Count > 0)
+            {
+                var errorPluginString = "";
+
+                erroredPlugins.ForEach(x => errorPluginString += x + Environment.NewLine);
+
+                Task.Run(() =>
+                {
+                    MessageBox.Show($"Uh oh, unable to load plugins:{Environment.NewLine}{errorPluginString}");
```
gdziedzic: I like it! 
Maybe we could also hint the user about taking a look into logs? Like:
`MessageBox.Show($"Uh oh, unable to load plugins:{Environment.NewLine}{errorPluginString}{Environment.NewLine}Please refer to the logs for more information.");`
Potentially you could also include some styling for the message box like: ```MessageBoxIcon.Warning```

Flow.Launcher.Core/Plugin/PluginsLoader.cs
gdziedzic: I like it! 
Maybe we could also hint the user about taking a look into logs? Like:
<<your current message>> Please refer to the logs for more information 
Potentially you could also include some styling for the message box like: ```MessageBoxIcon.Warning```

jjw24: Updated:

image

jjw24: @gdziedzic please re-review

PR #59, opened 2020-05-21
Create a portable pdb file in release builds of Flow.Launcher.Plugin

JohnTheGr8:
This should fix the `Symbols package publishing failed.` error when uploading the .snupkg file to nuget.


JohnTheGr8:
@jjw24 I have updated the version so that a new package can be uploaded to nuget that includes symbols (can't replace the same version).

Also, I think the `continuous-integration/appveyor/branch` check does not fire here because I am pushing from a fork branch... Perhaps that status should not be required.

jjw24:
> @jjw24 I have updated the version so that a new package can be uploaded to nuget that includes symbols (can't replace the same version).
> 
> Also, I think the `continuous-integration/appveyor/branch` check does not fire here because I am pushing from a fork branch... Perhaps that status should not be required.

removed the check.

Issue #56, opened 2020-05-18
Update UIs missed


jjw24:
1. Folder plugin change action keyword window layout needs to be updated

2. Settings' Hotkey tab add hotkey window layout needs update

PR #60, opened 2020-05-23
Update ActionKeyword window

jjw24:
1. Folder plugin change action keyword window layout needs to be updated

2. Settings' Hotkey tab add hotkey window layout needs update

https://github.com/Flow-Launcher/Flow.Launcher/issues/56


JohnTheGr8:
@jjw24 I'm seeing a couple issues that could be addressed as part of this PR:

1. The `Current action keyword` text is cut off. Example:



2. The current keyword value is not updated in UI unless I change to another view and back. Example:



3. I think the "Success" message after submitting the form can be removed.

4. Closing the form by hitting Esc would also be nice

JohnTheGr8:
@jjw24 is this ready?

jjw24:
> @jjw24 is this ready?

I havent done add hot key tab window and press Esc to cancel per your comment. 

I will request review once pr is ready, otherwise it's still wip

JohnTheGr8:
@jjw24 also, I tested the appveyor build and using Escape to exit the change-action-keyword window does nothing.

jjw24:
> @jjw24 also, I tested the appveyor build and using Escape to exit the change-action-keyword window does nothing.

That's added to the change key in Hotkey tab. I will add it also for change ActionKeyword window too

jjw24:
> I think we should remove the margins and the `MaxWidth="100"` from the line under the plugin description text. The max-width does not play well with translations.
> 
> Also, `Query time` and `Init time` do not update when you switch language unless you switch to a different plugin tab...
> 
> Here's a screenshot that demonstrates both:
> 
> image

@JohnTheGr8 Good pick up. Sorted

JohnTheGr8:
> 
> It was added to the change key in Hotkey tab. I will add it also for change ActionKeyword window too

Gotcha, thank you... I realize I was vague in my first comment about this :|

JohnTheGr8:
> 
> @JohnTheGr8 Good pick up. Sorted

(Note to self:)
There's still a few quirks when using translations. It will be easier if I address them in a separate PR instead of taking the time to explain them here.

jjw24:
> the `success` translation key should not have been removed from the translation files, as it is being used in other places. Example:
> image

readded, good review @JohnTheGr8

PR #61, opened 2020-05-25
docs: Add spectrum badge

NickSeagull:
This PR adds a badge to make joining the community easier.


JohnTheGr8:
@NickSeagull could you re-open this PR and target the dev branch, please?

Also, before we offer more choices to users, I think we need to write down some community guidelines so people are not confused about where to post questions, issues, etc. 

We have not decided on these afaik so perhaps we should start with that.

NickSeagull:
Will re-open when stuff is decided :)

PR #63, opened 2020-05-26
Plugin enhancements

JohnTheGr8:
This PR solves the issue we had with publishing symbols to nuget: `The checksum does not match for the dll(s) and corresponding pdb(s).`

After a bit of digging, I figured out that the problem was caused by Fody. Switching the PDB format to embedded solves this issue.

I have also:
* enabled deterministic builds for the plugin project in AppVeyor
* removed the `FodyWeavers.xml` file from the plugin project's nupkg
* removed `SolutionAssemblyInfo.cs` from the plugin project, which was causing the plugin dll to have the wrong `AssemblyFileVersion`, `AssemblyFileVersion` and `AssemblyInformationalVersion` in CI builds

Opening the nupkg file with NugetPackageExplorer verifies that SourceLink is valid now and the URLs are correctly embedded:

image

I have also decompiled the dll with dotPeek and verified that Fody has done its magic correctly.

### DIFF
--- a/Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
+++ b/Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
@@ -15,7 +15,7 @@
 
   <PropertyGroup>
     <Version>1.0.0</Version>
-    <PackageVersion>1.0.0-beta2</PackageVersion>
+    <PackageVersion>1.0.0-beta3</PackageVersion>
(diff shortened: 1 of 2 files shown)


jjw24:
Looks really good! Thank you for sorting that issue out.

Tested in clean VM Win 10, all good using self-contained publish, all good.

JohnTheGr8:
Thank you. Just submitted it to nuget, should be available shortly...

PR #79, opened 2020-06-24
Fix UWP loading

jjw24:
- Remove AppxPackagingTlb.dll


taooceros:
Is this branch simply remove the dll and use ComImport class to do the same job?

jjw24:
yep

taooceros:
LGTM, I just change the declear of out variable to inline deslearation so that it will be clearer.

taooceros:
Let me take a test before approving it.

taooceros (approved):

Issue #10, opened 2020-04-22
Update text message in launcher update


jjw24:
Update text message in launcher update

![image](https://user-images.githubusercontent.com/26427004/80039958-5f340c80-853c-11ea-94aa-de500d797f6d.png)

https://github.com/jjw24/Wox/issues/184

PR #80, opened 2020-06-25
Refactoring in Flow.Launcher.Core

JohnTheGr8:
Wanted to work on some new features related to the core project, and thought it was a good opportunity to refactor the code beforehand...


Flow.Launcher.Core/Configuration/Portable.cs
```diff
@@ -126,13 +124,16 @@ public void CreateUninstallerEntry()
         {
             var uninstallRegSubKey = @"Software\Microsoft\Windows\CurrentVersion\Uninstall";
             // NB: Sometimes the Uninstall key doesn't exist
-            using (var parentKey =
-                RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Default)
-                    .CreateSubKey("Uninstall", RegistryKeyPermissionCheck.ReadWriteSubTree)) {; }
+            RegistryKey
+                .OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Default)
+                .CreateSubKey("Uninstall", RegistryKeyPermissionCheck.ReadWriteSubTree)
+                .Dispose();
```
jjw24, line 130: We should stick to using ()

Flow.Launcher.Core/Configuration/Portable.cs
```diff
@@ -126,13 +124,16 @@ public void CreateUninstallerEntry()
...
 
-            var key = RegistryKey.OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Default)
-                .CreateSubKey(uninstallRegSubKey + "\\" + Constant.FlowLauncher, RegistryKeyPermissionCheck.ReadWriteSubTree);
-            key.SetValue("DisplayIcon", Constant.ApplicationDirectory + "\\app.ico", RegistryValueKind.String);
+            var key = RegistryKey
+                        .OpenBaseKey(RegistryHive.CurrentUser, RegistryView.Default)
```
jjw24, line 133: this should be wrapped in using

Flow.Launcher.Core/Configuration/Portable.cs
```diff
@@ -142,7 +143,8 @@ public void CreateUninstallerEntry()
 
         internal void IndicateDeletion(string filePathTodelete)
         {
-            using (StreamWriter sw = File.CreateText(filePathTodelete + "\\" + DataLocation.DeletionIndicatorFile)){}
+            var deleteFilePath = Path.Combine(filePathTodelete, DataLocation.DeletionIndicatorFile);
+            File.CreateText(deleteFilePath).Close();
```
jjw24, line 147: why are we not using using(){} for this?

Flow.Launcher.Core/Configuration/Portable.cs
```diff
@@ -152,21 +154,17 @@ internal void IndicateDeletion(string filePathTodelete)
...
-                DataLocationRoamingDeleteRequired = true;
+            // Get full path to the .dead files for each case
+            var portableDataDeleteFilePath = Path.Combine(portableDataDir, DataLocation.DeletionIndicatorFile);
+            var roamingDataDeleteFilePath = Path.Combine(roamingDataDir, DataLocation.DeletionIndicatorFile);
 
-            if ((portableDataPath + "\\" + DataLocation.DeletionIndicatorFile).FileExits())
-                DataLocationPortableDeleteRequired = true;
-
-            if (DataLocationRoamingDeleteRequired)
+            // Should we switch from %AppData% to portable mode?
```
jjw24, line 164: What do you mean? This call is after the switch from portable to roaming or vice versa is done, to clean up either location leftover.

Flow.Launcher.Core/Configuration/Portable.cs
```diff
@@ -176,18 +174,14 @@ public void PreStartCleanUpAfterPortabilityUpdate()
 
                     Environment.Exit(0);
                 }
-
-                return;
             }
-
-            if(DataLocationPortableDeleteRequired)
+            // Should we switch from portable mode to %AppData%?
```
jjw24, line 178: see previous comment please

Flow.Launcher.Core/Plugin/PluginInstaller.cs
```diff
@@ -71,8 +75,7 @@ internal static void Install(string path)
                         File.Create(Path.Combine(existingPlugin.Metadata.PluginDirectory, "NeedDelete.txt")).Close();
                     }
 
-                    UnZip(path, newPluginPath, true);
```
jjw24, line 74: this is not needed?

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -32,27 +28,18 @@ public static class PluginManager
         // todo happlebao, this should not be public, the indicator function should be embeded 
         public static PluginsSettings Settings;
         private static List<PluginMetadata> _metadatas;
-        private static readonly string[] Directories = { Constant.PreinstalledDirectory, DataLocation.PluginsDirectory };
 
-        private static void ValidateUserDirectory()
```
jjw24, line 37: this is not needed? we would want to if exits right

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -127,65 +110,74 @@ public static IEnumerable<PluginPair> DotNetPlugins(List<PluginMetadata> source)
...
+            else
+            {
+                return source
+                    .Where(o => o.Language.ToUpper() == AllowedLanguage.Python)
+                    .Select(metadata => new PluginPair
+                    {
+                        Plugin = new PythonPlugin(Constant.PythonPath),
+                        Metadata = metadata
+                    });
+            }
```
jjw24, line 169: Probably just matter of personal preference. Would 
```
if (!String.IsNullOrEmpty(Constant.PythonPath))
     return true condition

return false condition
```
be more readable?

jjw24:
Also fixes issue https://github.com/Flow-Launcher/Flow.Launcher/issues/10

Flow.Launcher.Core/Plugin/PluginInstaller.cs
JohnTheGr8: The old code was unzipping the same file twice. New code unzips it once and then moves the directory in this step. You can test it, it works.

That said, perhaps the initial unzip should use a random directory inside the temp folder...

Flow.Launcher.Core/Plugin/PluginManager.cs
JohnTheGr8: Do you mean the removed `Directory.Exists` check? If so, the [documentation of `Directory.CreateDirectory`](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=netcore-3.1) implies that this check is redundant...

Flow.Launcher.Core/Plugin/PluginsLoader.cs
JohnTheGr8: would be the same to me, honestly. Should I invert it?

Flow.Launcher.Core/Configuration/Portable.cs
JohnTheGr8: It's a matter of preference, but I don't really like the empty using statements... the empty `{ }` feel like noise to me, always catching my eye when I'm scrolling through the file. But if the team prefers them that way, it's ok with me.

Flow.Launcher.Core/Configuration/Portable.cs
JohnTheGr8: again, matter of preference; this is equivalent to empty using statement but nicer, IMO...

Flow.Launcher.Core/Configuration/Portable.cs
JohnTheGr8: You mean the comment is confusing, correct? I'll update it to clarify.

Flow.Launcher.Core/Plugin/PluginInstaller.cs
jjw24: oh good pick up, shouldnt unzip twice.

how does it handle unzipping same plugin to the same locations ?

Flow.Launcher.Core/Plugin/PluginsLoader.cs
jjw24: yeah let's just leave it. i am not that fuss either way.

Flow.Launcher.Core/Configuration/Portable.cs
jjw24: maybe we should wrap it in using please, just so we cater for some unforeseen exception down the  track

Flow.Launcher.Core/Configuration/Portable.cs
jjw24: yeah, just a little confusing, not sure what the comment is for. Is it something i can answer? (i wrote the code for it)

Flow.Launcher.Core/Configuration/Portable.cs
JohnTheGr8: BTW, this specific step (lines 127-130) is probably here by accident... it creates a key named "Uninstall" in `HKEY_CURRENT_USER`...

Flow.Launcher.Core/Configuration/Portable.cs
JohnTheGr8: `using` does not suppress exceptions, it just disposes the object in case of exception... and `IndicateDeletion` is used in two places in the same file, `DisablePortableMode` and `EnablePortableMode`, both of which do not handle exceptions (we should probably change this behavior).

But okay, I'll switch back to a `using` statement here.

Issue #95, opened 2020-06-30
Error Loading All Plugins


robertoleonardo:
Yes

- 

I am switching to Flow from Wox, but I can't seem to get any third party plugins to load on restart after a successful installation message.  Is that just how it is for now, with flow in this early stage, or should they be working?  I've tried installing--in flow--numerous of the plugins i use in wox and upon restarting the application i get a plugin load error for each and every plugin.

Apologies if i'm dumb and the plugins aren't supposed to work just because they haven't been modified to work on flow, but the content on github here suggests they should be working even now.  

ERROR MESSAGE (using spotify as an example, but i've tried numerous plugins and none work):  "The following plugin has errored and cannot be loaded....Wox.Spotify Premium...Please refer to the logs for more information"

LOG FILE: 

[2020-06-30.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/4850218/2020-06-30.txt)

SCREENSHOT:

![Snag_1b31d48](https://user-images.githubusercontent.com/13153043/86099402-4aa84f00-ba6c-11ea-91fa-28da97074fa1.png)

jjw24:
Hey how's it going. We are still working on integrating with Wox plugins, currently Wox plugins are still erroring out. What's some of the plugins that you would like to use, I can see if I am able to update them to Flow Plugins for now.

robertoleonardo:
Ah - thank you! i figured that was likely to be the explanation. please take your time and don't go out of your way to prioritize my preferences, but the only three i use other than the built ins are:
1) wox.processkiller
2) wox.spotifypremium
3) google search plus

Thank you all for your work on this - it's making my work life a lot easier. and the windows search/environmental variables search is HUGE. i appreciate you.


________________________________

JohnTheGr8:
I was thinking that the process-killer plugin could be one of the defaults, @jjw24 what do you think?

robertoleonardo:
Agree - I have always thought it kind of odd that it wasn't


________________________________

theClueless:
@jjw24 @JohnTheGr8 In the past I took process killer plugin and made some updates and upgrades to it (much faster and doesn't crush as much).
It was in a private repo until now but maybe you could use it as a baseline :)
https://github.com/theClueless/Wox.Plugins

JohnTheGr8:
@theClueless thanks for pointing this out, I'll try to find some time to ~~steal~~ port it tomorrow.

[...]

PR #103, opened 2020-07-02
Add plugin: Process Killer

JohnTheGr8:
As discussed in #95, I have ported [Wox.Plugin.ProcessKiller](https://github.com/theClueless/Wox.Plugins/tree/master/Wox.Plugin.ProcessKiller) as a default plugin for the Flow Launcher.

I have also made a few minor enhancements:
- highlighting how the process name matched the result title
- set it up for i18n to prepare for translations
- added a context-menu option to kill all instances of a given executable
- removed the `kill all X processes` when the filtered process results are not all from the same path, to avoid killing processes by mistake.
- support spaces in process search

I think we should find a better icon but other than that it works well...

Screenshots of how it looks in action:

image

image


jjw24:
Could we break main up, at the very minimum the code handling processes should be in a separate file/class?

JohnTheGr8:
> 
> 
> Could we break Main's code up, at the very minimum the code handling processes should be in a separate file/class?

I agree, I wanted to split the code but it was getting late yesterday and I left it for later.

Done now, and also added a couple minor enhancements.

Issue #118, opened 2020-07-09
Allow Explorer Plugin to Search File Contents


Kalaschnik:
_Everything_ is not able to do this. However, since Flow uses the Explorer Plugin (to replace Everything) this should be a welcoming enhancement.

PR #119, opened 2020-07-12
Add file content search for Explorer plugin

jjw24:
Add file content search for Explorer plugin using Windows Index's own indexed content.

Customisable Action Keyword. Default is 'doc:'

Closing #118


Kalaschnik:
I have not tried you changes; however, in my experience one cannot set _Everything_ to index file contents which results in a very bad performance (i.e., naive search), compared to the instant results of Windows explorer search (either through start menu or file explorer search (starting in 2004 update). I demonstrated this by searching a string within a PDF in Windows Start Menu search and _Everything_ (not even sure if _Everything_ can search within PDFs):

GIF 13 07 2020 09-10-32


Is there no way to use Windows Search for file contents? Would be cool to finally get rid of Everything..

jjw24:
Oh, sorry i wasnt clear in the PR summary. This pr actually adds the windows index's content search. Explorer will use it to search for file content.

I was just asking only specifically how we want to trigger a file content search with explorer, at the moment I built the trigger like Everything plugin where you need to type 'content:' then your search to trigger it.

Kalaschnik:
Ahh Sorry! Thanks for clarifying. I thought we re-introduce _Everything_ here!

- Would it be possible to trigger content search without any keyword, or would this decrease the performance too much?
- Windows Search is using `documents: {search string}`
- Windows Explorer is using `{search string} kind:document`, when explicitly selected by the user. Implicitly it searches any kind.

Anyway, I also would have a preference for `content {search string}`, since it is shorter. However, it would be neat (for later release) if the content keyword is customizable.

jjw24:
> Ahh Sorry! Thanks for clarifying. I thought we re-introduce _Everything_ here!
> 
> - Would it be possible to trigger content search without any keyword, or would this decrease the performance too much?
> - Windows Search is using `documents: {search string}`
> - Windows Explorer is using `{search string} kind:document`, when explicitly selected by the user. Implicitly it searches any kind.
> 
> Anyway, I also would have a preference for `content {search string}`, since it is shorter. However, it would be neat (for later release) if the content keyword is customizable.

All good. 

I would want to trigger content search without a keyword, this doesn't have any drastic impact on performance as it is just pulling data already in the windows index, but it would make the results very messy. That's why I didn't do too much work on how we want to trigger it until if there is any feedback.

jjw24:
Updated PR summary to make it clearer

JohnTheGr8:
we can use a shorter keyword like `doc:` for convenience

Kalaschnik:
> we can use a shorter keyword like `doc:` for convenience

+1 for `doc:`, yet in a later release this should be user adjustable.

Kalaschnik:
I would propose `in:`, this is also what Mac’s Alfred is using, and it is really short :)

> When you need to expand your search scope beyond these core file types included in Alfred's results, the File Search keywords are here to help you.
> 
> Use the "open" keyword to open a file (or press the spacebar as a shortcut!)
> Use the "find" keyword to reveal the file in Finder
> **Use the "in" keyword to search inside your files' content**
_https://www.alfredapp.com/help/features/file-search/_

jjw24:
I am making the action keyword customizable.

It's going to be triggered like how you would trigger a plugin.

I think content makes more sense but is too long,  doc or in kind seems a bit confusing for an action keyword ?

Flow.Launcher.Infrastructure/UserSettings/PluginSettings.cs
```diff
@@ -15,6 +15,13 @@ public void UpdatePluginSettings(List<PluginMetadata> metadatas)
                 if (Plugins.ContainsKey(metadata.ID))
                 {
                     var settings = Plugins[metadata.ID];
+
+                    // TODO: Remove. This is one off for 1.2.0 release.
```
jjw24, line 19: Just a note here, this is a one off, and will be deleted in the next release. It is needed as we introduced an additional action keyword.

I have also introduced version in plugin settings in the following commit to lay the ground work for handling future plugin meta updates.

jjw24:
As part of this change, i am also thinking about changing the Explorer trigger hotkey currently global "*" to "e". This will only trigger Explorer search when users chooses to with "e". This will reduce the numerous results from Explorer currently with any searches that the users performs, reduce the unnecessary clutter. @Flow-Launcher/team 

What do you think?

Kalaschnik:
> As part of this change, I am also thinking about changing the Explorer trigger hotkey currently global "*" to "e". This will only trigger Explorer search when the users choose to do so with "e". This will reduce the numerous results from Explorer currently with any searches that the users perform, reduce the unnecessary clutter of results. @Flow-Launcher/team
> 
> What do you think?

Would that mean searching files and folders only gets triggered when typing e before?

jjw24:
> Would that mean searching files and folders only gets triggered when typing e before?
> 

Yep

Kalaschnik:
I’m not sure about that actually. Right now I do not need to enter keyword to search for files and folders. More recently I was using Alfred (changed work), and it bugged me a lot that its default settings it would only search preferences and contacts. However, most of the time I want to find stuff (i.e., files and folders). So for Alfred, I clicked the checkboxes under Extras:

image

I would propose that Flow Launcher should search for files and folders without a keyword by default. What do the others think @Flow-Launcher/team ?

JohnTheGr8:
Either way is okay with me, it's simple enough to change to one's liking...

jjw24:
Fair enough, I will just leave it how it is.

PR #125, opened 2020-07-19
Move Everything plugin to its own repo

jjw24:
Now that Explorer plugin supports file and file content search, which are the core functionalities provided by Everything plugin, we can move Everything to its own repo and be maintained as a external plugin if users still want to download and use.

Repo: https://github.com/Flow-Launcher/Flow.Launcher.Plugin.Everything


jjw24:
> let's also remove `Constant.EverythingSDKPath`

Great pick up 👍

Issue #128, opened 2020-07-27
High CPU Usage


giggio:
**Describe the bug/issue**
Flow is sometimes using 1 to 5% CPU when it is not being used. It is not clear why this is happening. It stopped after a while.

I have taken a full memory dump from the process and will try to analyze it, if I find more info I'll comment here.

**To Reproduce**
I can't find a way to replicate the behavior, it happens intermittently.

**Screenshots**
N/A

**Your System**
```
Windows build number: Version 10.0.19041.388
Flow Launcher version: 1.1.2
```

**Flow Launcher Error Log**

I have found exceptions when that was happening. It seems it was indexing a file at the OneDrive for business folder. I have changed the folder in the logs to `<filename>`, as the file name and path are sensitive. I don't think this is related (see the dumps later on the comments).

````
2020-07-27 11:10:15.2840|ERROR|ImageLoader.Load|-------------------------- Begin exception --------------------------
2020-07-27 11:10:15.2840|ERROR|ImageLoader.Load|Failed to get thumbnail for <filename> on second try
2020-07-27 11:10:15.3018|ERROR|ImageLoader.Load|Exception full name:
 <System.Runtime.InteropServices.COMException>
2020-07-27 11:10:15.3018|ERROR|ImageLoader.Load|Exception message:
 <Error while extracting thumbnail for <filename>>
2020-07-27 11:10:15.3212|ERROR|ImageLoader.Load|Exception stack trace:
 <   at Flow.Launcher.Infrastructure.Image.WindowsThumbnailProvider.GetHBitmap(String fileName, Int32 width, Int32 height, ThumbnailOptions options) in C:\Workbench\Git\Flow.Launcher\Flow.Launcher.Infrastructure\Image\ThumbnailReader.cs:line 150
   at Flow.Launcher.Infrastructure.Image.WindowsThumbnailProvider.GetThumbnail(String fileName, Int32 width, Int32 height, ThumbnailOptions options) in C:\Workbench\Git\Flow.Launcher\Flow.Launcher.Infrastructure\Image\ThumbnailReader.cs:line 109
   at Flow.Launcher.Infrastructure.Image.ImageLoader.GetThumbnail(String path, ThumbnailOptions option) in C:\Workbench\Git\Flow.Launcher\Flow.Launcher.Infrastructure\Image\ImageLoader.cs:line 209
[...]

PR #133, opened 2020-08-05
Limiting the number of ImageSources cached to address memory issue

jjw24:
**Problem:**
Memory usage can run up quite high for Flow, and can hit 300 mbs for myself if constantly performing searches. It does reduce back down to around 130 when querying is stopped.

_Note:_
This is a partial fix to the memory issue, it does not address it complete, and other factors that contribute to it are still been investigated.

This PR does not mean memory will reduce significantly, it is more focused on addressing the unnecessary high memory caused by improperly handling of image caching. So reduce image cache (ImageSources)'s memory's footprint is the main focus here. 

**How to reproduce:**
Open query, type 'a' (especially if Explorer hotkey is set as global '*') and you should get numerous results. Keep pressing down arrow on the keyboard and should see Flow.Launcher.exe in task manager grow to as high as 300mbs for myself. It may vary depending on the user and the application images that need to be displayed on the machine.

If unable to repro, change the power mode on Windows to 'Recommended' then try.

**Partial solution:**
Limiting the number of ImageSources cached to reduced the memory usage going over too high.
- Set maximum number of images cached to 50 from 5000.

- Once cached images go over permissible factor (2) X number of images (50), perform clean up by removing infrequently used ones from dictionary.

**Testing**
1. Take snapshots of the memory usage by using the Diagnostic tools of Visual studio (Debug -> Windows -> Diagnostic tools).
2. Take the first snapshot after all the dlls are loaded and the memory utilization is stabilized.
3. Take the next memory snapshot after many queries (don't have to execute an application). The aim is to just load many icons.
4. Compare the second snapshot by setting the first one as the baseline.
5. Search for memory size diff in bytes for Concurrent Dictionary with value ImageSource.
6. Compare this value to that obtained without the changes in this PR, to get the results as shown in the following images.

_Result_
- Before this PR change:
image

- With this PR:
image

**Related**: #128


taooceros:
@bao-qian has introduced a way to lazy load image for resultlistview, should I add that to this pr?

Flow.Launcher.Infrastructure/Image/ImageCache.cs
```diff
@@ -2,17 +2,18 @@
...
 
 namespace Flow.Launcher.Infrastructure.Image
 {
     [Serializable]
     public class ImageCache
     {
-        private const int MaxCached = 5000;
+        private const int MaxCached = 50;
         public ConcurrentDictionary<string, int> Usage = new ConcurrentDictionary<string, int>();
         private readonly ConcurrentDictionary<string, ImageSource> _data = new ConcurrentDictionary<string, ImageSource>();
```
taooceros, line 15: Maybe we should merge this two dictionary to one by creating another utility class, which will allow us to remove element easier?

jjw24:
> @bao-qian has introduced a way to lazy load image for resultlistview, which is a quite simple change, should I add that to this pr?

Are you referring to the change in the other PR ?

taooceros:
> > @bao-qian has introduced a way to lazy load image for resultlistview, which is a quite simple change, should I add that to this pr?
> 
> Are you referring to the change in the other PR ?

Yes, it is part of commit in #195

jjw24:
> > > @bao-qian has introduced a way to lazy load image for resultlistview, which is a quite simple change, should I add that to this pr?
> > 
> > 
> > Are you referring to the change in the other PR ?
> 
> Yes, it is part of commit in #195

I think we can keep them separate and then just resolve any conflict depending which goes in first

taooceros:
Sure

Flow.Launcher.Infrastructure/Image/ImageCache.cs
jjw24: yeah, let me have a look

taooceros: I have finished that, if you want, I can add that to the pr.

jjw24: haha ok cool, yes please

jjw24:
@taooceros new changes look good. Do you want to approve and merge it in?

taooceros:
Sure!

taooceros (approved):

PR #145, opened 2020-08-25
Fix Quick Folder Access in Explorer plugin not triggering, incorrect search behaviour, and plus minor tweaks

jjw24:
- Fixed Quick Folder Access not triggering

- Fixed behaviour to continue searching for files and folders using index search even when a match is found in Quick Folder Access list. This helps if user wants file/folder that has the same name in Quick Folder Access list but in different location

- Prevent user from using global action keyword for content search

- Exclude from Index Search when only '_' is typed

- Fixed file extension search using '*' failing/not returning results


Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -34,17 +34,20 @@ internal List<Result> Search(Query query)
...
+
+            if (quickFolderLinks.Count > 0)
+                results.AddRange(quickFolderLinks);
+            
+            // This allows the user to type the assigned action keyword and only see the list of quick folder links
+            if (settings.QuickFolderAccessLinks.Count > 0
+                && query.ActionKeyword == settings.SearchActionKeyword
+                && string.IsNullOrEmpty(query.Search))
+                    return quickFolderAccess.FolderListAll(query, settings.QuickFolderAccessLinks, context);
+
```
JohnTheGr8, line 50: the second if statement can be moved up before the definition of `quickFolderLinks`

JohnTheGr8 (dismissed):
Looking good, just one minor nitpick. Apologies for not reviewing earlier, time has been limited...

Issue #84, opened 2020-06-28
Different Flow Icon for missing App Icon


Kalaschnik:
I would propose a different and more neutral Icon, if an app icon was not found. I actually thought that `app_error.png` was used for more critical conditions.
![image](https://user-images.githubusercontent.com/9831678/85940507-1ffbb080-b91d-11ea-87ff-e3e736b70f79.png)

This indicated that something is wrong with Spotify, maybe a question mark is the better choice here?!

jjw24:
Good idea

ArttuLai:
For example PowerToys Run uses a generic window icon for missing apps and I think it works well — not that it means you should use one too. I also think the red sad face looks like shouldn't press that button, and it looks like warning.

<img width="541" alt="ptrunGenericAppIcon" src="https://user-images.githubusercontent.com/42007282/91642331-36b0a780-ea33-11ea-9207-1278ae4f9674.png">

Maybe it could be something like that? I even made a quick mockup just to show what I was thinking 🙂

![genericApp (128x128)](https://user-images.githubusercontent.com/42007282/91642450-31a02800-ea34-11ea-9b07-fa0afcb1f506.png)

ArttuLai:
<img width="641" alt="genericAppMockup" src="https://user-images.githubusercontent.com/42007282/91642794-e63b4900-ea36-11ea-9fe8-f6c1ea3a0d17.png">

Kalaschnik:
I prefer your version over the MDL2 Icons, which are here for reference:
https://github.com/microsoft/PowerToys/tree/master/src/modules/launcher/PowerLauncher/Images

However, it looks a bit macOSish — is this your asset/can you modify it to look more Windows-like?

ArttuLai:
> However, it looks a bit macOSish — is this your asset/can you modify it to look more Windows-like?

I agree, it looks kinda Mac-like now that you said it 😅 I was going for something like the new Fluent app icons by Microsoft. But it's mine and I can modify it / see what I can do.

Kalaschnik:
> @Kalaschnik I agree, it looks kinda Mac-like now that you said it 😅 I was going for something like the new Fluent app icons by Microsoft. 

That’s cool. I had the same intension for the Flow logo :)

ArttuLai:
@Kalaschnik OK I made some variants:

https://1drv.ms/u/s!Aj_Hiawo_tA7iMJxSZdKcKN2RPwhGg?e=DKrWQb

[...]

PR #151, opened 2020-09-01
Add 'app_missing_img.png' icon

ArttuLai:
Adds a new icon I designed to be used as the placeholder icon when Flow can't find the real app icon. Reference to issue [#84](https://github.com/Flow-Launcher/Flow.Launcher/issues/84)


jjw24:
Hey @ArttuLai thanks so much for contributing. Do you think you could include all the different version of the icon from the issue in the PR as well? It would be good to have a few variations, please also include the raw files so in case we want to update it in the future.

jjw24:
Thanks @ArttuLai . I just made a commit to move the files to its own folder.

jjw24 (approved):

jjw24:
Will make a PR later to utilise one of the images

PR #160, opened 2020-09-21
Fix plugin dependency loading and remove excess duplicate package references

jjw24:
**Problem:**
Currently in PluginLoader method, we are using AssemblyLoadContext.Default.LoadFromAssemblyPath to load plugin assembly dlls. The Default.LoadFromAssemblyPath loads assemblies into the default assembly load context and as a result the program looks for assembly dlls in the main application directory not in the plugin's own directory. 

Consequently for an external plugin, in order for its dependent assembly to work you have to duplicate your assembly's PackageReference to the Flow.Launcher.Plugin library, or for default plugins you have to duplicate it with the main project:

image

**Solution:**
1. Add a custom assembly loader for plugins and resolve any of the plugin's dependent assemblies to load from the plugin's own directory, or if the assembly exists with Flow.Launcher.Plugin then use the assembly there.

2. Add Directory.Build.targets so default plugins will automatically make project reference to Flow.Launcher.Plugin not available for build.

3. Use `CopyLocalLockFileAssemblies=true` in project file for both default and future external plugins so all referenced assemblies in the plugin are outputted and available for the main application to point to. 

**Result**
Plugins no longer need to have the same package reference in Flow.Launcher.Plugin, duplicates can be removed and plugin assemblies are resolving correctly.


jjw24:
TODO:
Version bump

JohnTheGr8 (approved):
Awesome. Apologies this took so long to look into... ;(

jjw24:
> Awesome. Apologies this took so long to look into... ;(

All good bud :)

PR #172, opened 2020-10-13
Move Old HttpWebRequest to HttpClient Instance

taooceros:


jjw24:
Hey I just saw the PR is going against Master. Can you change the base to Dev instead please, you can do this by clicking on 'Edit'. We develop against the Dev branch.

image

taooceros:
Yes, thank you for telling me that I can do this.

jjw24:
Hey @taooceros , it would be better to separate out the two change set please, otherwise  one may deplay the other due to review/discussion. Let me know if you need help to do this.

Once the improvements from bao-qian is in separate pr, i can review it, as I have a few questions about the change set such as why remove translate as it sets the future use for translating other languages, and that particular line removed is currently used to translate user's Chinese into Pinyin so the string matcher can compare.

taooceros:
Sorry, I do want to separate them into two branches... . However, it seems that they have automatically merge each other. I will take a look on them and try to seperate them into two pull requests later.
Thanks for telling me that they are being together!

taooceros:
I will create a new pull request from another branch regarding that this pull request is based on the master branch, while I would like to keep it the same as the main repository of Flow.

taooceros:
Check out #179 and #178

PR #178, opened 2020-10-17
Move Old HttpWebRequest to HttpClient Instance

taooceros:
This pull request changes the Get & Download method implementation in Flow.Launcher.Infrastructure.Http by moving the old HttpWebRequest API to new HttpClient.

This has fixed my issue of connection to Github while checking for new Update.


Flow.Launcher.Infrastructure/Http/Http.cs
```diff
@@ -6,85 +6,112 @@
...
+                ProxyProperty.Port => (new Uri($"http://{Proxy.Server}:{Proxy.Port}"), WebProxy.Credentials),
+                ProxyProperty.UserName => (WebProxy.Address, new NetworkCredential(Proxy.UserName, Proxy.Password)),
+                ProxyProperty.Password => (WebProxy.Address, new NetworkCredential(Proxy.UserName, Proxy.Password)),
+                _ => throw new ArgumentOutOfRangeException()
+            };
+        }
+
+        public static async Task Download([NotNull] string url, [NotNull] string filePath)
+        {
+            using var response = await client.GetAsync(url);
```
jjw24, line 80: client.GetAsync crashes the entire app for me when i try to install plugin. Cant see any reason why though, no exception is thrown

taooceros: Let me take a check.

taooceros: Oh that's because before the Http.Download is not async, so no await is added there.

jjw24:
getting this error also when i check for update:

image

taooceros:
> getting this error also when i check for update:
> 
> image

fix it, and the previous one. Forget to commit seperately, so both is on last commit.

taooceros:
Hold on, update also call the method Http.Download so need to change that as well. Though, since we have InstallOrUpdate method, why not simply call that one instead of downloading newest release and install.

taooceros:
image
emmm a question. Why do we need the mainthread to exercute this method?

taooceros:
Done, please review. @jjw24 
and it seems that we don't need to invoke Install method into main thread......

Plugins/Flow.Launcher.Plugin.WebSearch/SuggestionSources/Baidu.cs
```diff
@@ -22,9 +23,9 @@ public override async Task<List<string>> Suggestions(string query)
             try
             {
                 const string api = "http://suggestion.baidu.com/su?json=1&wd=";
-                result = await Http.Get(api + Uri.EscapeUriString(query), "GB2312");
```
jjw24, line 25: removal of "GB2312" is intentional?

jjw24:
good work

jjw24:
for reference on why use ConfigureAwait(false)

https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f

Plugins/Flow.Launcher.Plugin.WebSearch/SuggestionSources/Baidu.cs
taooceros: Yes, I have checked, the ReadAsString doesn't need the encoding to create the result.

jjw24: do we need to keep  "GB2312"?

jjw24 (approved):

PR #183, opened 2020-10-18
Pinyin library change

taooceros:
Moved from #179 

Wox has changed the Pinyin Library to ToolGood.Words. [#2962](https://github.com/Wox-launcher/Wox/pull/2962) Thank you for the change @bao-qian .
This pull request has done the same thing, but optimize a few.

It keeps the full pinyin query, which seems perform the same as current pinyin library.
It keeps using the Concurrent Dictionary as cache, instead of Wox's Memory Cache. I think changing the cache implementation should be done in separately. However, the cache for Pinyin is changed from store single chinese character to whole word.
IAlphabet interface is kept in this pull request.
I also merge the query of win32 and uwp in Program plugin into one query by adding enabled property to the IProgram interface and cast win32 and uwp to IProgram.
I have checked the previous cache storage and it actually makes the loading time quite long, so I remove the Save and Loading cache in this pull request because it seems that the library is fast enough.

### DIFF
--- a/Flow.Launcher.Infrastructure/Alphabet.cs
+++ b/Flow.Launcher.Infrastructure/Alphabet.cs
@@ -19,7 +19,7 @@ namespace Flow.Launcher.Infrastructure
 
     public class Alphabet : IAlphabet
     {
-        private ConcurrentDictionary<string, string> _pinyinCache;
+        private ConcurrentDictionary<string, string> _pinyinCache = new ConcurrentDictionary<string, string>();
         private Settings _settings;
 
         public void Initialize([NotNull] Settings settings)
--- a/Flow.Launcher.Infrastructure/Alphabet.cs
+++ b/Flow.Launcher.Infrastructure/Alphabet.cs
@@ -3,12 +3,12 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using hyjiacan.util.p4n;
-using hyjiacan.util.p4n.format;
 using JetBrains.Annotations;
 using Flow.Launcher.Infrastructure.Logger;
 using Flow.Launcher.Infrastructure.Storage;
 using Flow.Launcher.Infrastructure.UserSettings;
+using ToolGood.Words.Pinyin;
+using System.Threading.Tasks;
 
 namespace Flow.Launcher.Infrastructure
 {
@@ -19,160 +19,44 @@ namespace Flow.Launcher.Infrastructure
 
     public class Alphabet : IAlphabet
     {
-        private readonly HanyuPinyinOutputFormat Format = new HanyuPinyinOutputFormat();
-        private ConcurrentDictionary<string, string[][]> PinyinCache;
-        private BinaryStorage<Dictionary<string, string[][]>> _pinyinStorage;
+        private ConcurrentDictionary<string, string> _pinyinCache;
         private Settings _settings;
-         
+
         public void Initialize([NotNull] Settings settings)
         {
             _settings = settings ?? throw new ArgumentNullException(nameof(settings));
-            InitializePinyinHelpers();
         }
 
-        private void InitializePinyinHelpers()
-        {
-            Format.setToneType(HanyuPinyinToneType.WITHOUT_TONE);
 
-            Stopwatch.Normal("|Flow Launcher.Infrastructure.Alphabet.Initialize|Preload pinyin cache", () =>
+        public string Translate(string content)
+        {
+            if (_settings.ShouldUsePinyin)
             {
-                _pinyinStorage = new BinaryStorage<Dictionary<string, string[][]>>("Pinyin");
-                
-                lock(_pinyinStorage)
+                string result = _pinyinCache.GetValueOrDefault(content);
+                if (result == null)
                 {
-                    var loaded = _pinyinStorage.TryLoad(new Dictionary<string, string[][]>());
-
-                    PinyinCache = new ConcurrentDictionary<string, string[][]>(loaded);
-                }
-
-                // force pinyin library static constructor initialize
-                PinyinHelper.toHanyuPinyinStringArray('T', Format);
-            });
-            Log.Info($"|Flow Launcher.Infrastructure.Alphabet.Initialize|Number of preload pinyin combination<{PinyinCache.Count}>");
-        }
-
-        public string Translate(string str)
-        {
-            return ConvertChineseCharactersToPinyin(str);
-        }
-
-        public string ConvertChineseCharactersToPinyin(string source)
-        {
-            if (!_settings.ShouldUsePinyin)
-                return source;
-
-            if (string.IsNullOrEmpty(source))
-                return source;
-
-            if (!ContainsChinese(source))
-                return source;
-                
-            var combination = PinyinCombination(source);
-            
-            var pinyinArray=combination.Select(x => string.Join("", x));
-            var acronymArray = combination.Select(Acronym).Distinct();
-
-            var joinedSingleStringCombination = new StringBuilder();
-            var all = acronymArray.Concat(pinyinArray);
-            all.ToList().ForEach(x => joinedSingleStringCombination.Append(x));
-
-            return joinedSingleStringCombination.ToString();
-        }
-
-        public void Save()
-        {
-            if (!_settings.ShouldUsePinyin)
-            {
-                return; 
-            }
-
-            lock(_pinyinStorage)
-            {
-                _pinyinStorage.Save(PinyinCache.ToDictionary(i => i.Key, i => i.Value));
-            }            
-        }
-
-        private static string[] EmptyStringArray = new string[0];
-        private static string[][] Empty2DStringArray = new string[0][];
-
-        /// <summmary>
-        /// replace chinese character with pinyin, non chinese character won't be modified
-        /// Because we don't have words dictionary, so we can only return all possiblie pinyin combination
-        /// e.g. 音乐 will return yinyue and yinle
-        /// <param name="characters"> should be word or sentence, instead of single character. e.g. 微软 </param>
-        /// </summmary>
-        public string[][] PinyinCombination(string characters)
-        {
-            if (!_settings.ShouldUsePinyin || string.IsNullOrEmpty(characters))
-            {
-                return Empty2DStringArray;
-            }
-
-            if (!PinyinCache.ContainsKey(characters))
-            {
-                var allPinyins = new List<string[]>();
-                foreach (var c in characters)
-                {
-                    var pinyins = PinyinHelper.toHanyuPinyinStringArray(c, Format);
-                    if (pinyins != null)
+                    if (WordsHelper.HasChinese(content))
                     {
-                        var r = pinyins.Distinct().ToArray();
-                        allPinyins.Add(r);
+                        result = WordsHelper.GetPinyin(content,";");
+                        result = GetFirstPinyinChar(result) + result.Replace(";","");
+                        _pinyinCache[content] = result;
                     }
                     else
                     {
-                        var r = new[] { c.ToString() };
-                        allPinyins.Add(r);
+                        result = content;
                     }
                 }
-
-                var combination = allPinyins.Aggregate(Combination).Select(c => c.Split(';')).ToArray();
-                PinyinCache[characters] = combination;
-                return combination;
+                return result;
             }
             else
             {
-                return PinyinCache[characters];
+                return content;
             }
         }
 
-        public string Acronym(string[] pinyin)
+        private string GetFirstPinyinChar(string content)
         {
-            var acronym = string.Join("", pinyin.Select(p => p[0]));
-            return acronym;
-        }
-
-        public bool ContainsChinese(string word)
-        {
-            if (!_settings.ShouldUsePinyin)
-            {
-                return false;
-            }
-
-            if (word.Length > 40)
-            {
-                //Skip strings that are too long string for Pinyin conversion.
-                return false;
-            }
-
-            var chinese = word.Select(PinyinHelper.toHanyuPinyinStringArray)
-                              .Any(p => p != null);
-            return chinese;
-        }
-
-        private string[] Combination(string[] array1, string[] array2)
-        {
-            if (!_settings.ShouldUsePinyin)
-            {
-                return EmptyStringArray;
-            }
-
-            var combination = (
-                from a1 in array1
-                from a2 in array2
-                select $"{a1};{a2}"
-            ).ToArray();
-            return combination;
+            return new string(content.Split(";").Select(c => c.First()).ToArray());
         }
     }
 }
--- a/Flow.Launcher.Infrastructure/Alphabet.cs
+++ b/Flow.Launcher.Infrastructure/Alphabet.cs
@@ -36,8 +36,8 @@ namespace Flow.Launcher.Infrastructure
                 {
                     if (WordsHelper.HasChinese(content))
                     {
-                        var result = WordsHelper.GetPinyin(content,";");
-                        result = GetFirstPinyinChar(result) + result.Replace(";","");
+                        var result = WordsHelper.GetPinyin(content, ";");
+                        result = GetFirstPinyinChar(result) + result.Replace(";", "");
                         _pinyinCache[content] = result;
                         return result;
                     }
@@ -46,6 +46,8 @@ namespace Flow.Launcher.Infrastructure
                         return content;
                     }
                 }
+                else
+                    return _pinyinCache[content];
             }
             else
             {
--- a/Flow.Launcher.Infrastructure/Alphabet.cs
+++ b/Flow.Launcher.Infrastructure/Alphabet.cs
@@ -47,7 +47,9 @@ namespace Flow.Launcher.Infrastructure
                     }
                 }
                 else
+                {
                     return _pinyinCache[content];
+                }
             }
             else
             {
--- a/Flow.Launcher.Infrastructure/Alphabet.cs
+++ b/Flow.Launcher.Infrastructure/Alphabet.cs
@@ -32,7 +32,7 @@ namespace Flow.Launcher.Infrastructure
         {
             if (_settings.ShouldUsePinyin)
             {
-                if (_pinyinCache.ContainsKey(content))
+                if (!_pinyinCache.ContainsKey(content))
                 {
                     if (WordsHelper.HasChinese(content))
                     {
--- a/Flow.Launcher.Infrastructure/Alphabet.cs
+++ b/Flow.Launcher.Infrastructure/Alphabet.cs
@@ -32,21 +32,20 @@ namespace Flow.Launcher.Infrastructure
         {
             if (_settings.ShouldUsePinyin)
             {
-                string result = _pinyinCache.GetValueOrDefault(content);
-                if (result == null)
+                if (_pinyinCache.ContainsKey(content))
                 {
                     if (WordsHelper.HasChinese(content))
                     {
-                        result = WordsHelper.GetPinyin(content,";");
+                        var result = WordsHelper.GetPinyin(content,";");
                         result = GetFirstPinyinChar(result) + result.Replace(";","");
                         _pinyinCache[content] = result;
+                        return result;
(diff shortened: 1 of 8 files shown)


jjw24:
Thanks @taooceros! If PR is ready, let us know via request review :)

taooceros:
I think this should be ready now.

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -50,7 +50,6 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
 
             if (_alphabet != null)
             {
-                query = _alphabet.Translate(query);
```
jjw24, line 53: what's the reason for removing this? This should cater for the translation for Chinese to Pinyin.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -77,13 +78,15 @@ public List<Result> Query(Query query)
                 uwps = _uwps;
             }
 
+            var searchText = WordsHelper.HasChinese(query.Search) ? WordsHelper.GetPinyin(query.Search) : query.Search;
```
jjw24, line 81: This should be part of Translate so other plugins could benefit from as well right?

Flow.Launcher.Infrastructure/Alphabet.cs
```diff
@@ -19,160 +19,45 @@ public interface IAlphabet
...
-            {
-                return Empty2DStringArray;
-            }
-
-            if (!PinyinCache.ContainsKey(characters))
+            if (_settings.ShouldUsePinyin)
             {
-                var allPinyins = new List<string[]>();
-                foreach (var c in characters)
+                if (_pinyinCache.ContainsKey(content))
```
jjw24, line 35: I think this should be !_pinyinCache.ContainsKey(content) instead

Flow.Launcher.Infrastructure/Alphabet.cs
```diff
@@ -19,160 +19,45 @@ public interface IAlphabet
...
-                // force pinyin library static constructor initialize
-                PinyinHelper.toHanyuPinyinStringArray('T', Format);
-            });
-            Log.Info($"|Flow Launcher.Infrastructure.Alphabet.Initialize|Number of preload pinyin combination<{PinyinCache.Count}>");
-        }
-
-        public string Translate(string str)
-        {
-            return ConvertChineseCharactersToPinyin(str);
-        }
```
jjw24, line 57: The Translate method, could we keep the `return ConvertChineseCharactersToPinyin(str);` so we know the code is in relation to pinyin conversion. And the actual code that does the conversion can stay in ConvertChineseCharactersToPinyin method.

In the future the Translate method can expand to facilitating other language conversions.

Flow.Launcher.Infrastructure/Alphabet.cs
```diff
@@ -19,160 +19,45 @@ public interface IAlphabet
...
-                        allPinyins.Add(r);
+                        return content;
                     }
                 }
-
-                var combination = allPinyins.Aggregate(Combination).Select(c => c.Split(';')).ToArray();
-                PinyinCache[characters] = combination;
-                return combination;
+                else
+                    return _pinyinCache[content];
```
jjw24, line 50: curly brackets please

Flow.Launcher.Infrastructure/Alphabet.cs
```diff
@@ -19,160 +19,45 @@ public interface IAlphabet
...
-                return EmptyStringArray;
-            }
-
-            var combination = (
-                from a1 in array1
-                from a2 in array2
-                select $"{a1};{a2}"
-            ).ToArray();
-            return combination;
+            return string.Concat(content.Split(';').Select(x => x.First()));
```
jjw24, line 60: just a question, why is this code only get the first char, do you know what its used for?

Flow.Launcher.Infrastructure/Alphabet.cs
taooceros: I think since bao has implement the ialphabet interface, if in the future we want to add more translation, we should add a new alphabet class that contains another language translation.
Bao has removed that in wox but I think it can be kept for future expansion.

Flow.Launcher.Infrastructure/Alphabet.cs
taooceros: Oh that's true, such a silly mistake 😅, thank you for indicating the error! I will change it later.

Flow.Launcher.Infrastructure/Alphabet.cs
taooceros: This is for getting the acronym for pinyin, just like our discussion in the acronym search. That should be removed once we have done the acronym search in fuzzy search.

Flow.Launcher.Infrastructure/StringMatcher.cs
taooceros: I moved that from using alphabet to program because I think query may not need to be in the cache. Also, Bao has assumed that query should be pinyin or English, so he removed it without adding it somewhere else. 
I agree with that but I think occasionally input chinese but results nothing may seems weird, so I moved it to program. 
It is also for removing extra translation since query may be translated in every fuzzy search, but if we moved it there, there will be only one translation. 

This seems a little bit ugly though😂.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
taooceros: As mentioned in my last response, this is for performance. Maybe we can move it to a more general place to translate query if using pinyin so that other plugin can benefit from it, too.

taooceros:
Thank you for the reviews! @jjw24

Flow.Launcher.Infrastructure/Alphabet.cs
taooceros: ok😉

JohnTheGr8: hey @taooceros, what @jjw24 meant here was to change to this:

```csharp
else
{
    return _pinyinCache[content];
}
```

taooceros: Sorry for misunderstanding, have edited.

Flow.Launcher.Infrastructure/StringMatcher.cs
jjw24: I would like to keep the ability to search using Chinese as well as Pinyin. In terms of it being called every time, only when user has Use Pinyin switched on. In terms of it being in memory cache, you can add a bool parameter in Translate that determines whether it should be stored.

What do you think?

Plugins/Flow.Launcher.Plugin.Program/Main.cs
jjw24: Please see my comment above. I am sure users also use Pinyin search for control panel and other plugins right?

Flow.Launcher.Infrastructure/Alphabet.cs
jjw24: Yeah i agree. We should really use inheritance/polymorphism here, but I dont know what languages we can translate. I guess my concern here is that Translate is not clear enough to indicate it's intended behaviour, although i do want to keep it generic naming so later on translate can take on different forms for different language translations. What do you think adding comments in here to explain what sort of translation it is right now? At least the user can get an idea since this is also a public method.

Flow.Launcher.Infrastructure/Alphabet.cs
jjw24: Hmm if this is for acronym search, could we remove this and add it in the other acronym search PR, I have a feeling we going to forget to remove this code.

Flow.Launcher.Infrastructure/StringMatcher.cs
taooceros: Well I mean that we will always do a fuzzy search between query and every single program in the program plugin, so that the translation will be called n times (number of program) once query change. However, if we put that in here, or somewhere the query is created, it will be called only one time per query. If we don't put it in memory cache, it may slow down the query if Chinese character is in the query.
I think the best way is to put it to the plugin manager or even earlier when the query is created and translate that if pinyin is enabled. I will take a look on that. That will enable every plugin to search with Chinese character.
Actually, I think that there won't be so much chance for us to search things with Chinese, because that is quite messy to type that in.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
taooceros: Actually whether translate the query won't affect people search with pinyin, but will affect people search with raw Chinese character if pinyin is enabled. As my last post has suggested, I would like to put it to somewhere the query is created.

Flow.Launcher.Infrastructure/Alphabet.cs
taooceros: I have done the part of removing this in the acronym branch. I put it here because I think this will be merged earlier so that I won't like to change the behavior since on the past pinyin acronym search is done with this method.

Flow.Launcher.Infrastructure/Alphabet.cs
taooceros: I think we can change the name of the `Alphabet` that implement the `IAlphabet` to indicate that which alphabet it is. For example, for the alphabet that for pinyin, we may change it to pinyinAlphabet? Although the name is a little bit long hhh.
That shall tell users what the alphabet is doing.

Flow.Launcher.Infrastructure/StringMatcher.cs
taooceros: I have taken a look on where can we put the translation for query. However, if we put it to when query is generated, whether that may break some characteristics of plugin that utilize Chinese search, and that's not what we want.

Therefore, I think maybe we should just keep it here, even though suffer a bit performance loss (though should not be significant). However, that seems not elegant hhhh. Let me think whether there may be a better way to solve that issue, because I think simply removing it or translating query separately is also not elegant.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -72,22 +72,17 @@ public List<Result> Query(Query query)
             Win32[] win32;
             UWP.Application[] uwps;
 
-            lock (IndexLock)
-            { // just take the reference inside the lock to eliminate query time issues.
-                win32 = _win32s;
-                uwps = _uwps;
-            }
```
taooceros, line 78: Bao has removed the lock in Wox, so I think removing it won't be an issue in Flow, too.

Plugins/Flow.Launcher.Plugin.Program/Programs/IProgram.cs
```diff
@@ -9,5 +9,6 @@ public interface IProgram
         string UniqueIdentifier { get; set; }
         string Name { get; }
         string Location { get; }
+        bool Enabled { get; }
```
taooceros, line 12: I added a property to the generic interface of all program to check whether they are enabled.  Both win32 and uwp who implement this interface has this property.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -72,22 +72,17 @@ public List<Result> Query(Query query)
...
-                .Select(p => p.Result(query.Search, _context.API));
+            var result = win32.Cast<IProgram>()
+                 .Concat(uwps)
+                 .AsParallel()
+                 .Where(p => p.Enabled)
+                 .Select(p => p.Result(query.Search, _context.API))
+                 .Where(r => r != null && r.Score > 0)
+                 .OrderBy(r=>r.Score).ToList();
 
-            var result = results1.Concat(results2).Where(r => r != null && r.Score > 0).ToList();
```
taooceros, line 88: Since we have the IProgram interface, why not use it to make the query done in one parallel linq selection.

Flow.Launcher.Infrastructure/Alphabet.cs
jjw24: Sounds good!

Flow.Launcher.Infrastructure/Alphabet.cs
jjw24: Sure as long as this acronym search code is only specific to Pingyin

Plugins/Flow.Launcher.Plugin.Program/Programs/IProgram.cs
jjw24: this is for where user decides to disable the program via the Program plugin context menu or settings right?

jjw24:
Pinyin changes are looking good.

I think if we can rename the Alphabet class that would be sweet.

Can you do me a favour, can we pull out the changes related to Programs plugin - locking and program query optimization. Let's stick to one change set per PR so it's easier to review and test.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
jjw24: I dont suppose those two will be updated often unless a reindex is run. Can the user reindex and also query at the same time?

jjw24:
also please bump the version in Plugin.Program's plugin.json file

taooceros:
Sure, I will rollback the changes for programs plug-ins and create a new one for it.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
taooceros: I would consider even if they query and reindex at the same time, it won't affect each other since index only change the memory index after finishing the whole index process (which is time consuming), but assigning values should not be an issue. On the time that Flow is reindexing, the variable won't change until it finish.

Plugins/Flow.Launcher.Plugin.Program/Programs/IProgram.cs
taooceros: Yes, and it is also for when program is not loaded correctly. This change won't change anything since the property is already there for win32 and uwp.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
taooceros: I have provide a small test and it seems that won't fine of querying when reindexing.

jjw24:
Just done some testing around the usage, I think the acronym search caused the Chinese to Pinyin search to fail:

lets so i have two entries as below.

searching pinyin is good
image

using acronym is good
image

but Chinese to Pinyin, hmmm:
image

The acronym search has changed the pinyin translation
image

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -8,6 +8,7 @@
 using Flow.Launcher.Infrastructure.Storage;
 using Flow.Launcher.Plugin.Program.Programs;
 using Flow.Launcher.Plugin.Program.Views;
+using ToolGood.Words.Pinyin;
```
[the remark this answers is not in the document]
taooceros: No, this is needed. This is the new Pinyin Library that replaces the original one.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -77,6 +78,7 @@ public List<Result> Query(Query query)
                 uwps = _uwps;
             }
 
+
```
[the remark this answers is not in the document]
taooceros: Sure

Flow.Launcher.Infrastructure/Alphabet.cs
jjw24: Is it possible to rename Alphabet to PinyinAlphabet?

taooceros:
Hmmmm you are right, I think the original implementation of the acronym search in the pinyin search is done by adding the acronym character to the front of the translated query, which will lead to the weird query translation.
···
var combination = PinyinCombination(source);
            
            var pinyinArray=combination.Select(x => string.Join("", x));
            var acronymArray = combination.Select(Acronym).Distinct();

            var joinedSingleStringCombination = new StringBuilder();
            var all = acronymArray.Concat(pinyinArray);
            all.ToList().ForEach(x => joinedSingleStringCombination.Append(x));

            return joinedSingleStringCombination.ToString();
···
You can see that the current implementation is joining all the combination together including acronym into a large string. I think it doesn't allow we use Chinese query to search for pinyin result.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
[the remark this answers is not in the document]
jjw24: yeah but it's not used here

Flow.Launcher.Infrastructure/Alphabet.cs
taooceros: Sure

Plugins/Flow.Launcher.Plugin.Program/Main.cs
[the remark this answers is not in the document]
taooceros: Oh, that is the Program plugin file. That's not needed since I have added the query translation back. I will remove it. Thank you for pointing out.

jjw24:
> The acronym part will be removed after the native acronym search has been finished, which means that the translation of query will be clean.

yeah that's what I was thinking, once the acronym search is implemented in string matcher, this should be needed here. Is that PR ready to go?

taooceros:
> > The acronym part will be removed after the native acronym search has been finished, which means that the translation of query will be clean.
> 
> yeah that's what I was thinking, once the acronym search is implemented in string matcher, this shouldnt be needed here. Is that PR ready to go?

It should be ready to go except the highlighted data is hard to control due to the translation. Do you think it is ok when sometimes the highlighted data is not on the right place?

jjw24:
> > > The acronym part will be removed after the native acronym search has been finished, which means that the translation of query will be clean.
> > 
> > 
> > yeah that's what I was thinking, once the acronym search is implemented in string matcher, this shouldnt be needed here. Is that PR ready to go?
> 
> It should be ready to go except the highlighted data is hard to control due to the translation. Do you think it is ok when sometimes the highlighted data is not on the right place? Or maybe if it is translated, we don't include the highlight data.

is this when translating Pinyin and matching a Chinese character program?

taooceros:
> > > > The acronym part will be removed after the native acronym search has been finished, which means that the translation of query will be clean.
> > > 
> > > 
> > > yeah that's what I was thinking, once the acronym search is implemented in string matcher, this shouldnt be needed here. Is that PR ready to go?
> > 
> > 
> > It should be ready to go except the highlighted data is hard to control due to the translation. Do you think it is ok when sometimes the highlighted data is not on the right place? Or maybe if it is translated, we don't include the highlight data.
> 
> is this when translating Pinyin and matching a Chinese character program? I think the original code doesnt highlight Chinese

Yes, it only highlights if search with acronym since they are the character that lies one the first few characters and match the chinese character. We can discuss this on #185 .

jjw24:
> > > > > The acronym part will be removed after the native acronym search has been finished, which means that the translation of query will be clean.
> > > > 
> > > > 
> > > > yeah that's what I was thinking, once the acronym search is implemented in string matcher, this shouldnt be needed here. Is that PR ready to go?
> > > 
> > > 
> > > It should be ready to go except the highlighted data is hard to control due to the translation. Do you think it is ok when sometimes the highlighted data is not on the right place? Or maybe if it is translated, we don't include the highlight data.
> > 
> > 
> > is this when translating Pinyin and matching a Chinese character program? I think the original code doesnt highlight Chinese
> 
> Yes, it only highlights if search with acronym since they are the character that lies one the first few characters and match the chinese character. We can discuss this on #185 .

Ok sounds good. 

I havent had a chance to look at 185 yet, will do so soon. Lets not forget to remove this acronym search once 185 is finished.

taooceros:
Of course, the integration has been finished, except for the highlight, and once those two pr has been merged, I will create a new pr to do the job.

jjw24:
Lets get rid of that white space and then we can merge this in and get a release going :)

jjw24 (approved):

Issue #136, opened 2020-08-18
Acronym search (or more feature) for programs


taooceros:
Does Flow have a plan to implement the search with acronym? I have tried the latest preview in Wox, which contains fascinating implementation of the indexing and searching using the acronym of programs name.

For example, when I want to open the program like Visual Studio Code, I can simply type vsc and select the first query result, which will be much faster and possibly more precious than typing "visual...." in order to open the same program.

On the other hand, I have met a significant issue in wox. When I use more than 3 or 4 letters to search the result, it will cause huge latency for the result to appear and may cause high CPU usage and stunt of wox in 1196 version ([#3093](https://github.com/Wox-launcher/Wox/issues/3093)). I don't know whether this issue is due to the implementation of acronym search. This cause me whenever I want to search a program with wox, I need to find a shortcut of the program so that it won't crash wox and my computer🤣.

Besides, I will be hugely pleased if Flow can implement the pinyin feature in Wox, since there are quite a lot of programs in my computer named with Chinese. 😃😃😃

jjw24:
Hey how's it going, pinyin feature is already there, you just need to turn on 'Should Use Pinyin' in settings. Let me know if this is what you are after.

In terms of acronym search, should already be there as well, let me take a look with the 'vsc' example and get back to you.

taooceros:
> Hey how's it going, pinyin feature is already there, you just need to turn on 'Should Use Pinyin' in settings. Let me know if this is what you are after.
> 
> In terms of acronym search, should already be there as well, let me take a look with the 'vsc' example and get back to you.

Thank you so much for responding so quickly. I think the situation for now is quite similar to the query situation in wox 1129, which require "space" between the acronym, otherwise it will not be considered in the fuzzy search.

taooceros:
I have checked wox for version 1.4.1129, which provides the same query result when I try to search query with acronym. I think wox in 1.4.1196 or higher has fixed the issues if you are using the same algorithm as wox. However, these versions also brough up an issue as I mentioned before.

jjw24:
Just to clarify, you were able to find and use the Pinyin search feature?

When you say needs 'space', do you mean needing to type 'v s c' for visual studio code?

taooceros:
Yes, I found that selection for pinyin search. Thank you!
Yes, "v s c" is working, though in a weird way which match the 'v' and 's' in visual, with 'c' in code, while "v s s c" doesn't work here.

[...]

PR #185, opened 2020-10-18
Add Acronym Support for Fuzzy Search

taooceros:
1. This pull request add acronym support for fuzzysearch. Link to #136 
2. This pr adds mapping from translated text to original text, allowing better highlighted text display for people using pinyin search (or potential future translation).


taooceros:
I originally want to do it as the way I have done before, but then I change my mind and decide to write a straight-forward search for the acronyms and query, which will make retrieve metadata easier and support the case capital letter and space both exist.

Because it is a straightforward loop, the performance should be good.

taooceros:
I think the test case need update

taooceros:
If this pull request has been approved, #183 should remove the part that create acronym before the pinyin. However, in order to achieve a correct matchdata highlight, IAlphabet should return an extra bool value suggesting whether the string is translated, since one Chinese character will project to multiple letter in pinyin form.

taooceros:
I am not sure how we should implement the data match for translated data, since the length change in translation...... should we simply don't highlight data when the string has been translated? Or return an extra string for matching data, which make it the same length as the original string.
Or create an extra class for mapping the index of translated string to original string, and create highlighted index based on that.

taooceros:
@jjw24 please take a try on the new mapping class, and it allow us to even highlight text when using full pinyin. However, it will take another loop to do the mapping every time, not sure if it will be possible to optimize it more.

jjw24:
just need some clarification:
- is this acronym support for cases where 'vsc' gets 'visual studio code'?
- essentially it is for identifying first letter of every word?

taooceros:
> just need some clarification:
> - is this acronym support for cases where 'vsc' gets 'visual studio code'?
> - essentially it identifies first letter of every word?
> - should 'vs' for 'visual studio code' match?
> - currently this pr even match 'v' for 'visual studio code' which is a bit excessive

It matches the first letter of a word seperated from whitespace or upper letter. So vsc will match both "visual studio code" or "Visual Studio Code" or "visualStudioCode" or "VisualStudioCode".
Not sure whether we should make vs match for visual studio code. I think you are right that a small panalty should be added for unmatched after query finish, while a larger panalty has been added for unmatches before query end.

jjw24:
When would there be a scenario where "visualStudioCode" or "VisualStudioCode" need to be acronym matched? program names are usually seperated by space, and if not the case it think should leave it to the fuzzy search to match up those kind of strings, that's what fuzzy matching is for.

I am concerned with performance when you have that loop at the beginning to do acronym search and you iterate through the whole 'stringtocompare' if that does not get a match, the rest of the code does another loop through the 'stringtocompare'.

i suggest we work it into the current loop. Take a look at the branch i pushed up:
https://github.com/Flow-Launcher/Flow.Launcher/compare/add_acronym_search?expand=1
https://github.com/Flow-Launcher/Flow.Launcher/tree/add_acronym_search

this has worked acronym search into the loop without another extra loop, it basically just projects forward to see the next substring if there is a match in case the loop is satisfied with the fuzzy search match and jumps to the next char.

I think it's a start, although it doesnt do the "visualStudioCode" or "VisualStudioCode" (which i think we should improve the actual fuzzy search code to do this) or where the the substring is numerical.

taooceros:
> When would there be a scenario where "visualStudioCode" or "VisualStudioCode" need to be acronym matched? program names are usually seperated by space, and if not the case it think should leave it to the fuzzy search to match up those kind of strings, that's what fuzzy matching is for.
> 
> I am concerned with performance when you have that loop at the beginning to do acronym search and you iterate through the whole 'stringtocompare' if that does not get a match, the rest of the code does another loop through the 'stringtocompare'.
> 
> i suggest we work it into the current loop. Take a look at the branch i pushed up:
> https://github.com/Flow-Launcher/Flow.Launcher/compare/add_acronym_search?expand=1
> (https://github.com/Flow-Launcher/Flow.Launcher/tree/add_acronym_search)
> 
> this has worked acronym search into the loop without another extra loop, it basically just projects forward to see if there is a match in the next substring in case the loop is satisfied with the fuzzy search match and jumps to the next char.
> 
> I think it's a start, although it doesnt do the "visualStudioCode" or "VisualStudioCode" (which i think we should improve the actual fuzzy search code to do this) or where the the substring is numerical.
> 
> Let me know what you think of this change

I added that because in my computer sometimes programs split there name like CamelCase or Pascal Case, for example "PostmanCanary" or "JetBrains". So I added that feature.

taooceros:
I will take a look on the branch, soon. However, I have encountered an issue that after merging dev branch, the setting windows is unable to get the dynamic resource, which most of the text in textblock disappears. I am not sure what cause that.....

jjw24:
> I will take a look on the branch, soon. However, I have encountered an issue that after merging dev branch, the setting windows is unable to get the dynamic resource, which most of the text in textblock disappears. I am not sure what cause that.....

try delete output folder and rebuild

>I added that because in my computer sometimes programs split there name like CamelCase or Pascal Case, for example "PostmanCanary" or "JetBrains". So I added that feature.

hmm fair point, let me have a think how we can build it in.

taooceros:
> > I will take a look on the branch, soon. However, I have encountered an issue that after merging dev branch, the setting windows is unable to get the dynamic resource, which most of the text in textblock disappears. I am not sure what cause that.....
> 
> try delete output folder and rebuild

Let me take a try.

> 
> > I added that because in my computer sometimes programs split there name like CamelCase or Pascal Case, for example "PostmanCanary" or "JetBrains". So I added that feature.
> 
> hmm fair point, let me have a think how we can build it in.

I have checked the code for split right now, and it require an extra for loop to do that. So it will be the same for us to use our for loop to do the same jobs. However, I do think we can merge fuzzy search and acronym search since fuzzy search do perform a for loop. I don't do that because I haven't fully understand the full fuzzy search, so it make me a few scare to touch some important things. Let me take a try on that.

taooceros:
By the way, Marry Christmas! ❤

taooceros:
> try delete output folder and rebuild

That seems fix it. Thank you

jjw24:
> By the way, Marry Christmas! ❤

Haha thank you. Merry Christmas to you too :)

jjw24:
> I don't do that because I haven't fully understand the full fuzzy search, so it make me a few scare to touch some important things. Let me take a try on that.

No stress, we have a lot of tests around it.

>I have checked the code for split right now, and it require an extra for loop to do that.

Yeah I wonder if there is a way to do it without the split

jjw24:
please resolve conflict @taooceros

taooceros:
Sorry I merge the wrong one, I will reset it soon.

taooceros:
ok, this should be great now...😅 @jjw24

jjw24:
No problem will take it for a spin

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -126,27 +177,39 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
...
 
                     currentQuerySubstringIndex++;
 
-                    allQuerySubstringsMatched = AllQuerySubstringsMatched(currentQuerySubstringIndex, querySubstrings.Length);
+                    allQuerySubstringsMatched =
+                        AllQuerySubstringsMatched(currentQuerySubstringIndex, querySubstrings.Length);
+
                     if (allQuerySubstringsMatched)
-                        break;
+                        continue;
```
jjw24, line 189: this change is correct?

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -43,21 +43,23 @@ public MatchResult FuzzyMatch(string query, string stringToCompare)
         /// </summary>
```
jjw24, line 43: please update the FuzzyMatch summary to include the Acronym search steps

Flow.Launcher.Infrastructure/StringMatcher.cs
taooceros: Em yes, that is used to before early return of full match to check more acronym.

Flow.Launcher.Infrastructure/StringMatcher.cs
taooceros: 👌, and I have a question, should we add an option to open acronym search?

jjw24: sorry what do you mean open acronym search?

taooceros: I mean we can have a parameters that indicates whether the search should include acronym search (maybe default is true).

jjw24: any reason why someone wouldnt use it or have the search method included?

taooceros: Emmm I am not quite sure. For example, maybe we don't want acronym search used to matching for explorer plugin because the index search doesn't include acronym search.
Maybe for some description matching (generally in subtitle matching), we don't want acronym search.

jjw24: Good point, though Explorer's search can't utilize fuzzy search at all due to usage of DirectiryInfoSearch and Index, so no need to worry here. 

If we don't want description search might as well not use fuzzy on it either.

taooceros: Well, I mean the Explorer utilize fuzzy search to create highlight data for results in result manager, so that may be a place that shouldn't use acronym search. ```

```

Comment (jjw24) [reply Flow.Launcher.Infrastructure/StringMatcher.cs:43]: but you can use acronym search with Explorer, so you cant type 'gk' and hit GitKraken, you have to type 'git' so highlight data would still be correct right?

File: Flow.Launcher.Test/FuzzyMatcherTest.cs [hunk 3]
```diff
@@ -142,20 +144,20 @@ public void GivenQueryString_WhenAppliedPrecisionFiltering_ThenShouldReturnGreat
...
             // Should
-            Assert.AreEqual(expectedScore, rawScore, 
+            Assert.AreEqual(expectedScore, rawScore,
                 $"Expected score for compare string '{compareString}': {expectedScore}, Actual: {rawScore}");
         }
 
         [TestCase("goo", "Google Chrome", SearchPrecisionScore.Regular, true)]
         [TestCase("chr", "Google Chrome", SearchPrecisionScore.Low, true)]
         [TestCase("chr", "Chrome", SearchPrecisionScore.Regular, true)]
-        [TestCase("chr", "Help cure hope raise on mind entity Chrome", SearchPrecisionScore.Regular, false)]
```
jjw24, line 152: why is this test case removed?

Flow.Launcher.Test/FuzzyMatcherTest.cs
```diff
@@ -129,11 +131,11 @@ public void GivenQueryString_WhenAppliedPrecisionFiltering_ThenShouldReturnGreat
 
         [TestCase(Chrome, Chrome, 157)]
         [TestCase(Chrome, LastIsChrome, 147)]
-        [TestCase(Chrome, HelpCureHopeRaiseOnMindEntityChrome, 25)]
+        [TestCase(Chrome, HelpCureHopeRaiseOnMindEntityChrome, 90)]
```
jjw24, line 134: this shouldnt have such a high score that it shows, acronym search matches it despite 'Help' at the start?

Flow.Launcher.Test/FuzzyMatcherTest.cs
```diff
@@ -129,11 +131,11 @@ public void GivenQueryString_WhenAppliedPrecisionFiltering_ThenShouldReturnGreat
...
         [TestCase(Chrome, UninstallOrChangeProgramsOnYourComputer, 21)]
         [TestCase(Chrome, CandyCrushSagaFromKing, 0)]
-        [TestCase("sql", MicrosoftSqlServerManagementStudio, 110)]
-        [TestCase("sql  manag", MicrosoftSqlServerManagementStudio, 121)]//double spacing intended
+        [TestCase("sql", MicrosoftSqlServerManagementStudio, 90)]
```
jjw24, line 137: this score has dipped due to acronym penalty?

Flow.Launcher.Test/FuzzyMatcherTest.cs
```diff
@@ -142,20 +144,20 @@ public void GivenQueryString_WhenAppliedPrecisionFiltering_ThenShouldReturnGreat
...
         [TestCase("chr", "Google Chrome", SearchPrecisionScore.Low, true)]
         [TestCase("chr", "Chrome", SearchPrecisionScore.Regular, true)]
-        [TestCase("chr", "Help cure hope raise on mind entity Chrome", SearchPrecisionScore.Regular, false)]
         [TestCase("chr", "Help cure hope raise on mind entity Chrome", SearchPrecisionScore.Low, true)]
         [TestCase("chr", "Candy Crush Saga from King", SearchPrecisionScore.Regular, false)]
         [TestCase("chr", "Candy Crush Saga from King", SearchPrecisionScore.None, true)]
-        [TestCase("ccs", "Candy Crush Saga from King", SearchPrecisionScore.Low, true)]
-        [TestCase("cand", "Candy Crush Saga from King",SearchPrecisionScore.Regular, true)]
-        [TestCase("cand", "Help cure hope raise on mind entity Chrome", SearchPrecisionScore.Regular, false)]
+        [TestCase("ccs", "Candy Crush Saga from King", SearchPrecisionScore.Regular, true)]
```
jjw24, line 157: why has this precision changed to regular?

Flow.Launcher.Infrastructure/StringMatcher.cs
taooceros: Yeah I agree. I think there is some conner case but it is not a lot.

Flow.Launcher.Test/FuzzyMatcherTest.cs
```diff
@@ -196,31 +201,38 @@ public void GivenQueryString_WhenAppliedPrecisionFiltering_ThenShouldReturnGreat
         [TestCase("sql serv man", MicrosoftSqlServerManagementStudio, SearchPrecisionScore.Regular, true)]
         [TestCase("sql studio", MicrosoftSqlServerManagementStudio, SearchPrecisionScore.Regular, true)]
         [TestCase("mic", MicrosoftSqlServerManagementStudio, SearchPrecisionScore.Regular, true)]
+        [TestCase("mssms", MicrosoftSqlServerManagementStudio, SearchPrecisionScore.Regular, true)]
+        [TestCase("msms", MicrosoftSqlServerManagementStudio, SearchPrecisionScore.Regular, true)]
```
jjw24, line 205: hmm ok, so acronym match doesnt have to match entire string

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -136,6 +121,22 @@ void continueAction(Task t)
...
+                {
+                    if (e.Query.RawQuery == QueryText) // TODO: allow cancellation
+                    {
+                        PluginManager.UpdatePluginMetadata(e.Results, pair.Metadata, e.Query);
+                        _resultsUpdateQueue.Post(new ResultsForUpdate(e.Results, pair.Metadata, e.Query, _updateToken));
+                    }
+                };
+            }
+        }
+
```
jjw24, line 139: what's this change?

jjw24 (changes requested):
please update summary

Flow.Launcher.Test/FuzzyMatcherTest.cs
taooceros: Well, yes, for now. Because we include description in searching for now, which means that there will be quite a lot of unmatched acronym if we match the entire string.

Flow.Launcher.Test/FuzzyMatcherTest.cs
jjw24: hmm despite the penalty, i wonder if this is will bring up irrelevant results

Flow.Launcher/ViewModel/MainViewModel.cs
taooceros: That's due to unintended merge. It seems just swap the posible of the two registration.

Flow.Launcher.Test/FuzzyMatcherTest.cs
taooceros: Because the acronym search will make this case matched, through `cure hope raise`. I think it is not the original idea for this test?

Flow.Launcher.Test/FuzzyMatcherTest.cs
taooceros: Because the acronym search will allow that to be matched with SearchPrecisionScore.Regular

Flow.Launcher.Test/FuzzyMatcherTest.cs
jjw24: This was to prevent fuzzy match bring up this kind of unrelated result. Maybe scoring needs adjusting so three acronyms shouldn't being up this long string?

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: There is a bit of difference, can you please double check including it is ok

Flow.Launcher.Test/FuzzyMatcherTest.cs
taooceros: yes, it match the middle "SQL" but without the M

jjw24:
Feels like we should adjust the scoring. With the current code, is it possible to score based on percentage of acronyms matched? So three characters is not going to bring up an entire long phrase

taooceros:
> Feels like we should adjust the scoring. With the current code, is it possible to score based on percentage of acronyms matched? So three characters is not going to bring up an entire long phrase

Yes, I agree with you. However, I think we still match the description in Program plugin, which means some description is long? Maybe we shall seperate them, with matching title with acronym and description without acronym, and add panelty for mismatch after the query is finished for acronym.

Flow.Launcher.Test/FuzzyMatcherTest.cs
taooceros: I am not quite sure, but there shouldn't be few strings that will act like that. Maybe we can release a pre-release and check how people respond to it.

Flow.Launcher.Test/FuzzyMatcherTest.cs
taooceros: emmm agree with you.

taooceros:
> please update summary

Will do tmr, or on the weekend.

jjw24:
> > Feels like we should adjust the scoring. With the current code, is it possible to score based on percentage of acronyms matched? So three characters is not going to bring up an entire long phrase
> 
> Yes, I agree with you. However, I think we still match the description in Program plugin, which means some description is long? Maybe we shall seperate them, with matching title with acronym and description without acronym, and add panelty for mismatch after the query is finished for acronym.
> 
> # e.g.
> image

i dont think the problem is whether we match description or not, in general acronym shouldnt bring up irrelevant long strings. That's why i was thinking maybe scoring needs to be percentage based. I dont think we should specifically exclude acronym from description because this will start to get confusing, not to mention the string matcher class is getting dangerously unmaintainable. 

What you have with penalty may work as well, and for long strings penalty could be so large that the acronym score will go into negative, which is fine as users can still use rest of fuzzy search method to get the results if they were to look for a long string.

jjw24:
Just reworking scoring to be based on percentage not total acronyms so code doesn't bring up long strings. Will push up the branch for you to take a look

taooceros:
Thank you!

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -72,19 +87,63 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
...
-                if (fullStringToCompareWithoutCase[compareStringIndex].Equals(' ') && currentQuerySubstringIndex == 0)
-                {
+                if (fullStringToCompareWithoutCase[compareStringIndex] == ' ' && currentQuerySubstringIndex == 0)
                     spaceIndices.Add(compareStringIndex);
+
+                // Acronym check
+                if (char.IsUpper(stringToCompare[compareStringIndex]) ||
+                    char.IsNumber(stringToCompare[compareStringIndex]) ||
+                    char.IsWhiteSpace(stringToCompare[compareStringIndex]) ||
+                    spaceMet)
```
jjw24, line 107: @taooceros what's this 'spaceMet' variable trying to solve and is this needed for translation?

taooceros: I think it is used to mark character after space to be acronym even when it is lower case.

jjw24: Ok cool, I thought so as well. I am thinking of using substring index = 0 to replace this, since the start of substring char is always acronym

taooceros: Have we split the string to compare? I think we only split the query string.

jjw24: Oh we didn't, lol ok nvm then

jjw24: Alternatively then, we can go back an index to check if it is space

taooceros: Yeah, that's feasible.

jjw24:
please look at the updates and let me know if good to merge in here:
https://github.com/taooceros/Flow.Launcher/pull/4/files

taooceros:
> please look at the updates and let me know if good to merge in here:
> https://github.com/taooceros/Flow.Launcher/pull/4/files

Great! I will take a look soon.

taooceros:
@jjw24 any further change we need to make for this PR?

jjw24:
Let me take a final look in 2hrs

jjw24 (approved):
Great work @taooceros 👍

PR #189, opened 2020-10-25
Program plugin improvement

taooceros:
This commit port some change from Wox authored by @bao-qian, which greatly improved the indexing time of win32 programs.
It also optimized a few code, and remove lock that seems insignificant.

1. Optimize the query of Program plugin by using generic to concat win32 and uwp apps.
2. Remove extra duplicate fuzzy search.
3. Greatly improve index speed by changing the way of Directory enumeration.


taooceros:
Question: Whether program plugin should query description?

taooceros:
The change should be ready for reviewing, and after using it for a while, I haven't found issues.

jjw24:
> Question: Whether program plugin should query both description and name? 

Would there be a significant benefit not querying description?

I suppose in theory it helps where you dont know the name of the program. 

>Should description only be something that needs to displayed or people would like to query with it?

In terms of displaying the description, I am personally quite ok with it added to the name as it is. My own usage, i dont really query the description unless i absolutely dont know what the program is called. 

I don't really know the impact, maybe put out a separate PR to test it out to get a better understanding?

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -254,10 +235,7 @@ private static Win32 ExeProgram(string path)
             {
                 var program = Win32Program(path);
                 var info = FileVersionInfo.GetVersionInfo(path);
-                if (!string.IsNullOrEmpty(info.FileDescription))
```
jjw24, line 257: whats the reason for removing this check?

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -273,47 +251,23 @@ private static IEnumerable<string> ProgramPaths(string directory, string[] suffi
...
-                                                "|The directory trying to load the program from does not exist", e);
-                        }
-                    }
-                }
-                catch (Exception e) when (e is SecurityException || e is UnauthorizedAccessException)
-                {
-                    ProgramLogger.LogException($"|Win32|ProgramPaths|{currentDirectory}" +
-                                                $"|Permission denied when trying to load programs from {currentDirectory}", e);
-                }
+                var paths = Directory.EnumerateFiles(directory, "*", SearchOption.AllDirectories)
```
jjw24, line 256: We still need to handle exceptions possibly thrown from Directory.EnumerateFiles, Exception + DirectoryNotFoundException +  SecurityException + UnauthorizedAccessException pretty sure are all also thrown by the method

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -265,27 +265,30 @@ public class Application : IProgram
...
+                var title = (Name, Description) switch
+                {
+                    (var n, null) => n,
+                    (var n, var d) when d.StartsWith(n) => d,
+                    (var n, var d) when n.StartsWith(d) => n,
+                    (var n, var d) when !string.IsNullOrEmpty(d) => $"{n}: {d}",
+                    _ => Name
+                };
+
+                var matchResult = StringMatcher.FuzzySearch(query, title);
```
jjw24, line 280: 👍 👍 👍

taooceros:
Well there isn't a significant benefit for removing the behavior of search from description except the trivial less time of search, but generally I think few people will use descriptions to search the program, and @bao-qian has removed the entire description part in wox, so just propose that here. Because in my view, flow should be a launcher that help us launcher program fast, so the long description may not be useful for search fast.
It is just a proposal, and wait for other to provide ideas.

jjw24:
> Well there isn't a significant benefit for removing the behavior of search from description except the trivial less time of search, but generally I think few people will use descriptions to search the program, and @bao-qian has removed the entire description part in wox, so just propose that here. Because in my view, flow should be a launcher that help us launcher program fast, so the long description may not be useful for search fast.
> It is just a proposal, and wait for other to provide ideas. 

May be we can break this question out to an issue and see if anyone else or the team has input.

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
jjw24: @taooceros

taooceros:
Sure!

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
jjw24: @taooceros whats the reason for removing this check?

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -254,10 +235,7 @@ private static Win32 ExeProgram(string path)
             {
                 var program = Win32Program(path);
                 var info = FileVersionInfo.GetVersionInfo(path);
-                if (!string.IsNullOrEmpty(info.FileDescription))
-                {
-                    program.Description = info.FileDescription;
-                }
+                program.Description = info.FileDescription;
```
jjw24, line 238: hey @taooceros how come we removed the check here

jjw24:
Got a question above and also need to resolve conflict, after that this is good to go in 👍

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
taooceros: Because this is the only place Description has been initialized, so no matter whether it is empty or null, it won't affect result, i think. Also, the creation of title has checked about the null or empty for description.

taooceros:
> Got a question above and also need to resolve conflict, after that this is good to go in 👍

😀😀😉

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
jjw24: yes you are right 👍  cool thanks

taooceros: Sorry I have checked again, and it seems that my statement is not right. I will add that check back to void null reference.

jjw24 (dismissed):
👍  thanks @taooceros , merging in

jjw24 (approved):

PR #190, opened 2020-10-26
Bookmark Plugin fix

taooceros:
1. Add Edge Dev Bookmark to the cache list
2. Use the fuzzy search score as the result score

PR #191, opened 2020-11-01
Release 1.4.0 | Plugin 1.2.2

jjw24:
Release Notes
=========

## Features

- Upgraded Pinyin library to ToolGood.Words: https://github.com/Flow-Launcher/Flow.Launcher/pull/183
  - Store whole Chinese words instead of individual characters  
  
  - No longer saves cache on disk

- For Program plugin, added ability to open containing folder using custom app + args, eg. Total Commander https://github.com/Flow-Launcher/Flow.Launcher/pull/194 

- Improved Program plugin: https://github.com/Flow-Launcher/Flow.Launcher/pull/189
  - Optimized the querying of programs in Program plugin by using generics to concatenate Win32 and UWP apps. 
  
  - Removed duplicate fuzzy search calls. 
  
  - Greatly improved indexing speed by changing how Directory enumerates.

## Bug fixes

- Added loading of Edge Dev bookmarks and fixed scoring based on fuzzy search https://github.com/Flow-Launcher/Flow.Launcher/pull/190

- Improve memory consumption by limiting number of images preloaded in memory: https://github.com/Flow-Launcher/Flow.Launcher/pull/133
  - Addressing the unnecessary high memory caused by improperly handling of image caching.
  
  - Set maximum number of images cached to 50 from 5000.
  
  -  Clean up by removing infrequently used images from dictionary once they go over permissible factor (2) X number of images (50).

- Fix plugin dependency loading and remove excess duplicate package references: #160
  - Added a custom assembly loader for plugins and resolve any of the plugin's dependent assemblies to load from the plugin's own directory, or if the assembly exists with Flow.Launcher.Plugin then use the assembly there.
  
  - Added Directory.Build.targets so default plugins will automatically look to Flow.Launcher.Plugin's referenced packages if not available in the local plugin folder.

  - Used CopyLocalLockFileAssemblies=true in project file for both default and future external plugins so all referenced assemblies in the plugin are outputted and available for the main application to point to.


JohnTheGr8 (approved):
let's do it!

taooceros (approved):
Let's go!

Issue #193, opened 2020-11-03
Plugin Program - open containing folder in custom file explorer


kubalav:
I don't use Windows Explorer. I would like to open containing folder in Total Commander. Plugin Everything works like a charm.

Can you implement this feature?

taooceros:
Which plugin are you using? Everything or the Explorer plugin?

kubalav:
I mean Program plugin. Everything plugin works with folders (single click), but not with Open parent folder (right click on item).
Program plugin doesn't work to open containing folder (right click on item).

taooceros:
Got it, and I will try to implement that soon. Thank you for the suggestion.

jjw24:
Hi @kubalav could you please test this build to see if it is as intended: https://ci.appveyor.com/api/buildjobs/2dcxc0jwdsg7x2ta/artifacts/Output%2FPackages%2FFlow-Launcher-v1.3.1.zip

kubalav:
I entered path to the Total Commander executable and it is works. It opens new instance of Total Commander.
Can you adjust it to open in existing instance?

taooceros:
Check out here http://www.ghisler.ch/wiki/index.php/Command_line_parameters. Take a try on add /O to the path? If that doesn't work, I will create another textbox to allow adding customized arguments.

kubalav:
It unable to start with `"C:\Program Files\Total Commander\TOTALCMD64.EXE" /O`

taooceros:
OK, I will add another textbox to do the job. Thank you for testing!

taooceros:
@kubalav take a try on this one. [](https://ci.appveyor.com/api/buildjobs/dr2p3y5na9mubina/artifacts/Output%2FPackages%2FFlow-Launcher-v1.3.1.zip).
I don't have an explorer that is useful with arguments, so I am not sure whether this works. But I have tested it with explorer and it seems works.

kubalav:
Wow! It's working. Thank you!

Can you make it for Everything plugin too?

taooceros:
Sure, I will do that tomorrow.

taooceros:
Just to make sure, are you using the everything plugin, which is built in on wox, or the explorer plugin, which is built-in on Flow.

kubalav:
From Wox.

taooceros:
ok

taooceros:
[...]

PR #194, opened 2020-11-04
Add Exploere Customization

taooceros:
Link to #193 
This is only for program plugin for now.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
+++ b/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
@@ -44,7 +44,7 @@
     <system:String x:Key="flowlauncher_plugin_program_customizedexplorer">Customized Explorer</system:String>
     <system:String x:Key="flowlauncher_plugin_program_args">Args</system:String>
     <system:String x:Key="flowlauncher_plugin_program_tooltip_customizedexplorer">You can customized the explorer used for opening the container folder by inputing the Environmental Variable of the explorer you want to use. It will be useful to use CMD to test whether the Environmental Variable is avaliable.</system:String>
-    <system:String x:Key="flowlauncher_plugin_program_tooltip_args">Enter the customized args you want to add for your customized explorer. Check the explorer's website for details on additional args.</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_tooltip_args">Enter the customized args you want to add for your customized explorer. %s for parent directory, %f for full path (which only works for win32). Check the explorer's website for details.</system:String>
 
     <!--Dialogs-->
     <system:String x:Key="flowlauncher_plugin_program_disable_dlgtitle_success">Success</system:String>
--- a/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
+++ b/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
@@ -40,7 +40,12 @@
     <system:String x:Key="flowlauncher_plugin_program_plugin_description">Search programs in Flow Launcher</system:String>
 
     <system:String x:Key="flowlauncher_plugin_program_invalid_path">Invalid Path</system:String>
-    
+
+    <system:String x:Key="flowlauncher_plugin_program_customizedexplorer">Customized Explorer</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_args">Args</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_tooltip_customizedexplorer">You can customized the explorer used for opening the container folder by inputing the Environmental Variable of the explorer you want to use. It will be useful to use CMD to test whether the Environmental Variable is avaliable.</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_tooltip_args">Enter the customized args you want to add for your customized explorer. Check the explorer's website for details on additional args.</system:String>
+
     <!--Dialogs-->
     <system:String x:Key="flowlauncher_plugin_program_disable_dlgtitle_success">Success</system:String>
     <system:String x:Key="flowlauncher_plugin_program_disable_dlgtitle_success_message">Successfully disabled this program from displaying in your query</system:String>
--- a/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
@@ -324,7 +324,9 @@ namespace Flow.Launcher.Plugin.Program.Programs
 
                         Action = _ =>
                         {
-                            Main.StartProcess(Process.Start, new ProcessStartInfo(Package.Location));
+                            Main.StartProcess(Process.Start, new ProcessStartInfo(
+                            !string.IsNullOrEmpty(Main._settings.CustomizedExploere)?Main._settings.CustomizedExploere:"exploere"
+                            , Package.Location));
 
                             return true;
                         },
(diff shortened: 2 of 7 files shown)


taooceros:
Question: Maybe this should be a global setting?

jjw24:
> Question: Maybe this should be a global setting?

I think this can be a global setting. This basically allows the user to open the containing folder using another program other than Windows explorer right?

jjw24:
hey in the code 'Explorer' has a typo in several places

jjw24:
> > Question: Maybe this should be a global setting?
> 
> I think this can be a global setting. This basically allows the user to open the containing folder using another program other than Windows explorer right? This can be useful for Explorer plugin as well.

On second thought, maybe leave this to the plugin to implement, because not all plugins do this.

taooceros:
True, but I think program plugin, explorer, and potential everything plugin may use the same setting? Though I think setting them seperately is not hard.

jjw24:
hey @taooceros just made a couple changes, let me know if it is all good

Plugins/Flow.Launcher.Plugin.Program/Settings.cs
```diff
@@ -14,9 +14,13 @@ public class Settings
         public bool EnableStartMenuSource { get; set; } = true;
 
         public bool EnableRegistrySource { get; set; } = true;
+        public string CustomizedExplorer { get; set; } = Explorer;
+        public string CustomizedArgs { get; set; }
```
taooceros, line 18: All the other things are good, except I am not sure whether it would be ok to not give this one an initail value?

taooceros (commented):
I plan to add the customized args aligning with what I have done in the Everything plugin, by adding the default args with select,[path], and allows users to use %s and %f to customize their selection for the containing folder.

Plugins/Flow.Launcher.Plugin.Program/Settings.cs
taooceros: Oh since I am going to change the default value, there is nothing to concern. Thank you for those changes!

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -324,7 +324,9 @@ public List<Result> ContextMenus(IPublicAPI api)
 
                         Action = _ =>
                         {
-                            Main.StartProcess(Process.Start, new ProcessStartInfo(Package.Location));
+                            Main.StartProcess(Process.Start, new ProcessStartInfo(
+                            !string.IsNullOrEmpty(Main._settings.CustomizedExplorer) ? Main._settings.CustomizedExplorer:Settings.Explorer,
+                            $"{Main._settings.CustomizedArgs} \"{Package.Location}\"".Trim()));
```
jjw24, line 329: does UWP work with Args? Since "/select,%f" is now the default init

Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
```diff
@@ -40,7 +40,12 @@
     <system:String x:Key="flowlauncher_plugin_program_plugin_description">Search programs in Flow Launcher</system:String>
 
     <system:String x:Key="flowlauncher_plugin_program_invalid_path">Invalid Path</system:String>
-    
+
+    <system:String x:Key="flowlauncher_plugin_program_customizedexplorer">Customized Explorer</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_args">Args</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_tooltip_customizedexplorer">You can customized the explorer used for opening the container folder by inputing the Environmental Variable of the explorer you want to use. It will be useful to use CMD to test whether the Environmental Variable is avaliable.</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_tooltip_args">Enter the customized args you want to add for your customized explorer. %s for parent directory, %f for full path (which only works for win32). Check the explorer's website for details.</system:String>
```
jjw24, line 47: sorry i am not too across how the %s and %f option can be used. Could you give me an example?

taooceros: %s is the parent directory and %f is the file path (.exe, .lnk, etc.) I get the idea from the everything plugin which can select the file while opening the parent directory. However, that seems may work not good for uwp.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
taooceros: Well...That's an issue... Let me take a look on solution.

Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
jjw24: oh ok so this option allows for win32 programs to open both file and parent directory when using 'open containing folder' ?

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
jjw24: i got it working by removing CustomizedAgs, if it doesnt work at all for UWP, perhaps remove it?

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -140,10 +140,10 @@ public List<Result> ContextMenus(IPublicAPI api)
...
                     Action = _ =>
                     {
-
-
-                        Main.StartProcess(Process.Start, new ProcessStartInfo("explorer", ParentDirectory));
-
+                        Main.StartProcess(Process.Start, new ProcessStartInfo(
+                            !string.IsNullOrWhiteSpace(Main._settings.CustomizedExplorer) ? Main._settings.CustomizedExplorer:Settings.Explorer,
+                            !string.IsNullOrWhiteSpace(Main._settings.CustomizedArgs)?Main._settings.CustomizedArgs.Replace("%s",$"\"{ParentDirectory}\"").Replace("%f",$"\"{FullPath}\""):
+                            Settings.ExplorerArgs));
```
jjw24, line 146: For readability, what do you think if we do the if statement layout like this:
image

If you like it I can push this change up

Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
taooceros: Yes, let me think how I can fix that. If I can't, I will remove that

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
taooceros: Sure

taooceros (commented):
It seems that for me, the layout is a bit exceed the border

jjw24:
> 
> image
> 
> It seems that for me, the layout is a bit exceed the border

Oh that's strange. Maybe you can reduce the width of the CustomizeExplorer textbox a little bit. I was trying to align them up with the others.

jjw24:
Once uwp is sorted we can get this in

taooceros:
> Once uwp is sorted we can get this in

Oh sorry I forget to mention, it has been solved by moving the default to %s.

jjw24:
> > Once uwp is sorted we can get this in
> 
> Oh sorry I forget to mention, it has been solved by moving the default to %s.

hey I just tested it out, UWP still not working...

taooceros:
> hey I just tested it out, UWP still not working...
Hey I have tested, too, it works....... Which explorer and args are you using?

jjw24:
> > hey I just tested it out, UWP still not working...
> 
> Hey I have tested, too, it works....... Which explorer and args are you using?

Strange. I just remove roaming profile then start, using the default, which is %s for args

taooceros:
> Strange. I just remove roaming profile then start, using the default, which is %s for args

Sorry I have found the issue, I will fix that soon.

jjw24 (approved):
👍

Issue #42, opened 2020-05-08
Icon flickering when searching


jjw24:
Icons flickering due to results constantly being updated for query result text highlighting.

How to reproduce:
Search 'settings' letter by letter, focus on Wox's Settings query result and you will notice the icon flickers despite the result not changing.

Possible root cause:
The compare result method will return true most of the time as highlight data for both title and subtitle changes. This will mean same results always get updated in ResultsViewModel as user types, thus the flickering as same icons get reloaded constantly.
![image](https://user-images.githubusercontent.com/26427004/77400858-f738bc00-6dff-11ea-8145-a4f4a83b14bc.png)

From https://github.com/jjw24/Wox/issues/171

PR #195, opened 2020-11-10
Render improvement

taooceros:
Change UI rendering logic to fix image flickering (the major fix is by removing the isAsync for image binding)
This comes from @bao-qian 's change in Wox, [#2899](https://github.com/Wox-launcher/Wox/pull/2899). Thank you for introducing the fix for image flickering.

The other change is majoring makes the code simpler and more readable, and remove duplicate logic.

Also fix that sometimes the selected index is not first when updating result views. https://social.msdn.microsoft.com/Forums/vstudio/en-US/5ff71969-f183-4744-909d-50f7cd414954/binding-a-tabcontrols-selectedindex-not-working?forum=wpf


jjw24:
Hmmm, getting this odd behaviour...
image

jjw24:
Closes https://github.com/Flow-Launcher/Flow.Launcher/issues/42

taooceros:
Here should work better.

jjw24:
When I am query, i am missing results. It seems it only gets the top portion of the results list, evident when i use Explorer plugin to browse folders.

taooceros:
> When I am query, i am missing results. It seems it only gets the top portion of the results list, evident when i use Explorer plugin to browse folders.

That's because I use the MaxResult to limit the number of results to be added into result list. If that's not intended, I can remove it.
image

jjw24:
> > When I am query, i am missing results. It seems it only gets the top portion of the results list, evident when i use Explorer plugin to browse folders.
> 
> That's because I use the MaxResult to limit the number of results to be added into result list. If that's not intended, I can remove it.
> image
> 
> 

Oh I see. Let's keep the original behaviour, let it flow :)

taooceros:
> Oh I see. Let's keep the original behaviour, let it flow :)

ok!

jjw24:
so far I came across intermittent issues:
- in the result list, sometimes highlighting does not show, so I couldnt just type the result and then press enter on the 1st result
- the highlighting sometimes jumped to 2nd or 3rd result

taooceros:
> in the result list, sometimes highlighting does not show, so I couldnt just type the query and then press enter on the 1st result

This may because of the delay I added for batch processing, let's make it shorter and take a try.

> the highlighting sometimes jumped to 2nd or 3rd result

This is quite weird....and I haven't met that situation. Let's take a look after I reduce the delay time.

taooceros:
UPDATE
> the highlighting sometimes jumped to 2nd or 3rd result

That's because some of the plugin query is slower, so that their result view is updated later than others, and the selection has not been updated, so that the selection is not on them even though their score is higher. I will try to update selection every time result view is change.

jjw24:
just used the latest changes, hmm still getting this incorrect result highlighting occasionally:
image

taooceros:
Weird......I haven't met that situation after the commit. Let me take a look again later,

jjw24:
also i am experiencing random delays occasionally.. strange

taooceros:
I have met that it will stunt at few second at the first query, and then the query is fine. Well....I don't understand why that is happen.

taooceros:
I added a null check when processing current result view, which seems solve some stunt for logging.
At the same time of testing, I found that the explorer plugin seems sometimes fail at query some of the query with special character.
image

jjw24:
> I added a null check when processing current result view, which seems solve some stunt for logging.

Cool, I will update to the latest and test

> At the same time of testing, I found that the explorer plugin seems sometimes fail at query some of the query with special character.
> image
> 

Can you put this into a quick issue please so I can remmeber to get around to fixing it.

taooceros:
> Can you put this into a quick issue please so I can remmeber to get around to fixing it.

Sure

jjw24:
> I added a null check when processing current result view, which seems solve some stunt for logging.
> At the same time of testing, I found that the explorer plugin seems sometimes fail at query some of the query with special character.
> image

sorry mate, it seems the stunt/delay is still there when i query, both query and result list freezes, quite reproducible now with queries that return longer list of results. use Explorer to test

taooceros:
Thank you for pointing that out, let me try on large set of query result. That may be a good way for me to debug the performance of the change.

taooceros:
> sorry mate, it seems the stunt/delay is still there when i query, both query and result list freezes, quite reproducible now with queries that return longer list of results. use Explorer to test

Weird, why doesn't I get the stunt when the resultview is large. Could you tell me is it the first-time query? I do encounter stunt when query at the first time.

image
You can see from the scroll bar that the result count is large, but I haven't encountered the stunt. It appears around 20ms, I think? (I think the explorer plugin has set a limit of 100 results so this is the largest result set I can get).

Let me check what can lead the listview stunt.

taooceros:
I found out the reason of stunting for me, which is the image loading hasn't been done, and because the image binding `isAsync` is removed for fixing image flickering, the ui stunt at the first time query since the image hasn't been done.

jjw24:
Awesome, good find

jjw24:
> Weird, why doesn't I get the stunt when the resultview is large. Could you tell me is it the first-time query? I do encounter stunt when query at the first time.

I'm getting it after doing the same query several times... But not always consistently

taooceros:
Sorry, the last commit isn't behaving like what it has described. Please Ignore it.
> I'm getting it after doing the same query several times... But not always consistently

Let me try if I can add the default image first, and then change the image to the image that it should like.

taooceros:
Let's try this one!

taooceros:
Weird....why do the build fail again? Is anything in the building environment change?

jjw24:
So just tested latest, i get the stunt/delay + issues, all happen occasionally:
1. on start up after one character typed
2. still getting the same stunt/delay after typing a certain query, for me it's using Explorer with shortcut + "docker"
3. and on rare occasions result list doesnt seem to be updating correctly:
image

taooceros:
The last commit is quite magic for me toward multitasking. It seems that I originally use global token assign to each result update, which results unintended token assignment so that the result update doesn't behave properly.

This fixs my issue of unintended update.

taooceros:
The last commit seems reduce my startup stunts. But require more testing whether it is effective.
The reason for adding that is WPF hasn't supported multi-item operation in listview binding, but simply notice reset at startup seems cause stunt issue? I am not quite sure about that.
https://github.com/dotnet/runtime/issues/18087

jjw24:
> So just tested latest, i get the stunt/delay + issues, all happen occasionally:
> 
> 1. on start up after one character typed
> 2. still getting the same stunt/delay after typing a certain query, for me it's using Explorer with shortcut + "docker"
> 3. and on rare occasions result list doesnt seem to be updating correctly, (e is my shortcut to trigger Explorer):

Using the latest commit 8f6f2fb, no longer getting the above problems.

However result seems to jump around, especially evident when using WebSearch, actionkeyword set to * and with suggestions enabled. The original behaviour is the suggestions result is updated while the search stays atop. The WebSearch suggestions result has a bit of delay so in the code it is set to update results list view once the api call is done, maybe (have not looked at the code yet) this bit of behaviour is not honored?

taooceros:
> However result seems to jump around occasionally, evident when using WebSearch, actionkeyword set to * and with suggestions enabled. The original behaviour is the suggestions result is updated while the search stays atop.

hmmm not sure what you mean? Are you mean the selected result jump around? I haven't got issue when only assigning one search source to * as action keyword. However, when multiple search source are bounded together, it seems that it works unintended.

> The original behaviour is the suggestions result is updated while the search stays atop.

That's true, I have checked the code, and it has used the event driven callback to update the later result.

taooceros:
Multiple Search Source may due to the design issue of WebSearch, I will check that out later.

taooceros:
UPDATE: It seems that may due to the unintended assignment of global token, which is assigned later, so that it is not cancelled, which leads to the results flow. Let me check how to solve that issue.

taooceros:
@jjw24 fix that.

jjw24:
Amazing, will take it for a spin now.

Also i havent looked at the code, and just as a suggestion only, is it possible to pull out the fix for image flicking issue as a standalone pr while this result list improvement is worked on? or are they tied up together

taooceros:
> Amazing, will take it for a spin now.
> 
> Also i havent looked at the code, and just as a suggestion only(totally up to you), is it possible to pull out the fix for image flicking issue as a standalone pr while this result list improvement is worked on? or are they tied up together

Ok, that's possible, I believe that the flickering issue is only caused by the IsAsync property in xaml image tag, so the fix toward that only include removing async loading and adding customized LazyAsync class for loading default image before the image is loaded in cache.

jjw24:
> > However result seems to jump around occasionally, evident when using WebSearch, actionkeyword set to * and with suggestions enabled. The original behaviour is the suggestions result is updated while the search stays atop.

this is sort of what i mean, can now reliably reproduce with 'linux academ'. it should be the first result in the list rather than 'linux academy'

image

taooceros:
> > > However result seems to jump around occasionally, evident when using WebSearch, actionkeyword set to * and with suggestions enabled. The original behaviour is the suggestions result is updated while the search stays atop.
> 
> this is what i mean, can now reliably reproduce with 'linux academ'. it should be the first result in the list rather than 'linux academy'
> 
> image
> 
> repro:
> websearch => *, enable suggestions
> 
> type 'linux academ', first time it displays first result 'linux academ' and highlights the first result (which is correct). Select second result 'linux academy', then retype 'linux academ' and you get the screenshot above.
> 
> Correct behaviour should be: highlighting on first result 'linux academy', as is consistent with dev branch behaviour.

Oh, I misunderstand your problem. I will check that today.

jjw24:
just merged dev back into this PR. Please check if the merge changes are ok

taooceros:
> just merged dev back into this PR. Please check if the merge changes are ok

Seems great.

taooceros:
@jjw24 fix it by also sync selected item with selected index when adding results.

jjw24:
Let me give it a spin

taooceros:
Found out that the context menu cannot be displayed.....not sure what cause that. Convert to draft until I fix it.

taooceros:
Found out that the context menu cannot be displayed.....not sure what cause that. Convert to draft until I fix it.

jjw24:
no problem.

I had fair bit of issues with the commits a640ead. i found 6b96228 to be quite stable with only one issue with Explorer plugin where if it doesnt find a result, flow will then not display any plugin's results, no error in logs either.

taooceros:
> no problem.
> 
> I had fair bit of issues with the commit a640ead. i found 6b96228 to be quite stable with only one issue with Explorer plugin where if it doesnt find a result, flow will then not display any plugin's results, no error in logs either.

Yes you are right, it seems that observable collection has done something special to make sure binding works better. I do not fully understand why, but I will move it back, and it seems much more stable.
I have met an issue that I want to ask whether you have met when trying this branch. Sometime after quick typing, the resultlist won't be able to display, and the whole data thread block, which seems due to invoke collection change event. Not sure whether it is possible to false reset when invoking stunt.

taooceros:
Change back to ObservableCollection works well.

jjw24:
> I have met an issue that I want to ask whether you have met when trying this branch. Sometime after quick typing, the resultlist won't be able to display, and the whole data thread block, which seems due to invoke collection change event. Not sure whether that only appear when I debug or it is an issue. 

Yep I had that with commit 6b96228  I was able to consistently replicate with explorer plugin for some reason when it doesn't find a result

taooceros:
> Yep I had that with commit 6b96228 I was able to consistently replicate with explorer plugin for some reason when it doesn't find a result

Are you compiling it as debug? Is any error throw or simply no results show? Let me investigate it more. Would you please tell me is it anyway to replicate it consistently?

> Does this solve the data thread block?

I wrap the data update event inside a Task.Run, and wait it only for at most 300ms. This should solve the block issue. It seems that I haven't encounter the same issue after changing that. But what's weird is that the timeout exception also hasn't been thrown in those testing.

taooceros:
Take Back my word. The thread block isn't caused by updating the observable collection, but due to the popped-out exception for checking whether unacceptable cancellation token appear in the update queue. This cause the Task stop but not throwing exception since it is wrap in the task.run. I change that to only throw it on debug and recreate the task if any exception is throw.

jjw24:
>Are you compiling it as debug? Is any error throw or simply no results show? Let me investigate it more. Would you please tell me is it anyway to replicate it consistently?

i test with a release build actually. 

Should i test again now after the updates?

taooceros:
> >Are you compiling it as debug? Is any error throw or simply no results show? Let me investigate it more. Would you please tell me is it anyway to replicate it consistently?
> 
> i test with a locally compiled release build actually. 
> 
> Should i test again now after the updates?

Well, I don't think the commit above seems able to fix the empty result when explorer plugin cannot find result. Howeve, you can take a try on debugging it and seems what's happening when explorer plugin cannot find result since I haven't successfully reproduced that. The error handling has been added so that debugging may be useful to find out what happens. I will take a try on using release version and see what happens if error throw, since debug version will simply throw the error and not able to see what happen after error.

taooceros:
Well....I think this possibly caused by the long query time of explorer plugin in some special case, and it is not able to cancel itself since no CancellationToken is passed to it, which is causing thread deadlock in thread pool……
image
You can see from the picture that there is a large number of works that is queued to wait to be done, and most of them are in Explorer plugin.

jjw24:
hmm ok, seems different. Repro my one:

1. Was using  commit 6b96228.
2. Using locally compiled release build
3. Assign an action key to explorer, mine was 'e'
4. Trigger action key word, and type random letters till explorer no longer show result
5. Delete entire query and retype anything and no results show up

Side note using the above steps if Explorer is able to get a result, then everything is still ok

taooceros:
> hmm ok, seems different. Repro my one:
> 
> 1. Was using  commit [6b96228](https://github.com/Flow-Launcher/Flow.Launcher/commit/6b9622844c3beda0a32c564f906787f2a30d0df4).
> 2. Using locally compiled release build
> 3. Assign an action key to explorer, mine was 'e'
> 4. Trigger action key word, and type random letters till explorer no longer show result
> 5. Delete entire query and retype anything and no results show up
> 
> Side note using the above steps if Explorer is able to get a result, then everything is still ok

I believe we are encountering the same issue. When encounter complex query, the explorer takes too long to query, which is unable to stop when next query comes, and finally cause too much threads in threadpool to wait previous search, so that other task won't be able to run even though queued.

Is it possible to add cancellation to explorer plugin, and what do you think about #233 ?

taooceros:
Sorry, incidentally close the pr.

jjw24:
@taooceros is this good for another round of testing?

taooceros:
> @taooceros is this good for another round of testing?

Let me check whether merging dev create some issue.

taooceros:
ready for another round of testing now

jjw24:
hmmmmm

image

image

taooceros:
Hmmm that's weird.... Haven't encountered before, will take a look now.

taooceros:
fix, this is due to merge issue




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "Flow-Launcher/Flow.Launcher"                                                                                    <notifications@github.com&gt;;
发送时间:&nbsp;2021年1月10日(星期天) 上午10:47
收件人:&nbsp;"Flow-Launcher/Flow.Launcher"<Flow.Launcher@noreply.github.com&gt;;
抄送:&nbsp;"天煞"<1569418337@qq.com&gt;;"Mention"<mention@noreply.github.com&gt;;
主题:&nbsp;Re: [Flow-Launcher/Flow.Launcher] Render improvement (#195)





 
hmmmmm

jjw24:
dups appearing for web search:
image

taooceros:
> dups appearing for web search:
> image

hmmm this is really hard to reproduce....... I have only encountered once unintentionally.

taooceros:
Are they appearing at the same time or the second one appearing later?

jjw24:
they are appearing at the same time when using websearch plugin

jjw24:
i am running it directly from the branch in visual studio

jjw24:
remove output, remove roaming data location, then run from vs, search 'g a' and i get it instantly

taooceros:
hmmm still cannot reproduce.... Are you using Google as search suggestion or none? I have tried Baidu and Bing, and both are fine.

jjw24:
None. Basically all default settings

taooceros:
> None. Basically all default settings

Fix it. It is due to result update event create duplicate result.

jjw24:
@taooceros could you merge dev in and resolve conflict please

taooceros:
> @taooceros could you merge dev in and resolve conflict please

Sure

Flow.Launcher/ViewModel/ResultViewModel.cs
```diff
@@ -131,6 +128,7 @@ public override bool Equals(object obj)
             }
         }
 
+
```
jjw24, line 131: lol how come ur prs always have random additional white space

jjw24 (approved):

Issue #201, opened 2020-11-14
Make it easier to add a custom theme


giggio:
**Describe the solution you'd like**

Currently the theme settings tab already has a link to "Browse for more themes" and that is super useful. But it is not clear where to put the theme you created.

I found out reading the code that it is supposed to be at `%APPDATA%\FlowLauncher\Themes`, so I added it there and it worked. If it is a portable installation it will be different.

It would be nice if there was another link to that folder, on the same settings tab (or the apropriate folder if it is a portable instalation).

**Describe alternatives you've considered**

The alternative is searching in the docs (I couldn't find it, hopefully this issue will help others find it) or reading the code. Docs could help, but the link would make it more intuitive and simpler to use.

**Additional context**

The code is this:

https://github.com/Flow-Launcher/Flow.Launcher/blob/fd56af77dc96d0e2e10a5b43254c4981d915afe4/Flow.Launcher.Core/Resource/Theme.cs#L26-L32

And this:

https://github.com/Flow-Launcher/Flow.Launcher/blob/fd56af77dc96d0e2e10a5b43254c4981d915afe4/Flow.Launcher.Infrastructure/UserSettings/DataLocation.cs#L15-L21

From the second code file you can see that if the installation is portable the directory will not be on `%APPDATA%`.

jjw24:
This new enhancement will go out in the next release :)

PR #202, opened 2020-11-15
Add quicklink to customized theme folder

taooceros:
Link to #201 
Add a quick link to access user theme folder.
image

If anyone would like to make it looks better, feel free to do it!

### DIFF
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -21,7 +21,7 @@ namespace Flow.Launcher.Core.Resource
         private ResourceDictionary _oldResource;
         private string _oldTheme;
         public Settings Settings { get; set; }
-        private const string Folder = "Themes";
+        private const string Folder = Constant.Themes;
         private const string Extension = ".xaml";
         private string DirectoryPath => Path.Combine(Constant.ProgramDirectory, Folder);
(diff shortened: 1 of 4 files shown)


taooceros:
Why do the check fail.....I think there isn't any significant change that may affect the check.

jjw24:
> Why do the check fail.....I think there isn't any significant change that may affect the check.

 Error from CI:
C:\Program Files\dotnet\sdk\5.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(241,5): error NETSDK1005: Assets file 'C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.WebSearch\obj\project.assets.json' doesn't have a target for 'netcoreapp3.1'. Ensure that restore has run and that you have included 'netcoreapp3.1

I am afk, can u check if the project websearch has the target framework in project file, pretty sure it does

taooceros:
image
I think it does have the target framework that is as intended. This branch should not contains things about .net 5. Weird.....

taooceros:
Oh maybe it is because that I push too fast that the restoration of project file hasn't been restored successfully. Let me push something it later and check whether that can solve this.

JohnTheGr8:
@jjw24 @taooceros build failures should be resolved with #206

jjw24:
Thanks for the fix both 👍 

@taooceros i changed the code to open path using FilesFolders class

jjw24 (approved):

Issue #203, opened 2020-11-15
URL's in Settings don't respect default system browser and just use Google Chrome


deefrawley:
**Describe the bug/issue**
Settings - Theme, Plugin and About pages. All URL's start a Google Chrome instance despite Brave Browser being my default browser.

**To Reproduce**
1. Set a browser default to something other than Chrome
2. Open Settings either from Flow or from the tray menu
3. Click on any hyperlink

**Your System**
```
Windows build number: 10.0.19042.630
Flow Launcher version: 1.4.0
```

jjw24:
Hey how's it going.

I haven't looked into opening with default browser yet, but can you open brave browser by setting the path to brave's exe in the plugin?

deefrawley:
Hi Jeremy,

I added the explicit Brave path to the URL and the Web Search plugin and restarted Flow. It still uses Chrome.

taooceros:
I think the brower used to open the link in setting windows won't link to the brower path in web search plugin and url. I think Jeremy has misunderstood your question.
I have taken a look on the implementation of the hyperlink, and the default is chrome.
https://github.com/Flow-Launcher/Flow.Launcher/blob/98630d1af58998e76b24bf8f5c2423d1b1f747a0/Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs#L14-L20
Let me check if it is possible to change that to the default brower.

deefrawley:
Thanks @taooceros. I did temporarily switch my default browser to Edge and restarted Flow to try that (and make sure it wasn't something specific to Brave) but it still used Chrome. I could try uninstalling Chrome altogether to see what happens if that would help?

taooceros:
I found out that why in my computer it works, because I don't have chrome installed😅. If that is not too troublesome for you, you can try that. However, let me take a try to find out the way to use default browsers to open url if no specific browser path has provided. I think using chrome as default is not working as intended.

taooceros:
[Flow.Launcher.Plugin.zip](https://github.com/Flow-Launcher/Flow.Launcher/files/5557361/Flow.Launcher.Plugin.zip)
@deefrawley I am not sure whether it is possible to take the change via changing one specific dll, but take a try to replace the dll with the one in the zip. For now, the behavior should open the default browser in a new window.

deefrawley:
Yes, that new DLL worked thanks.

PR #214, opened 2020-11-18
Use default system browser

taooceros:
#203 Use the Default Browsers in the computer instead of simply using chrome as default.

### DIFF
--- a/Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
@@ -1,18 +1,45 @@
+using Microsoft.Win32;
 using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 
+
 namespace Flow.Launcher.Plugin.SharedCommands
 {
     public static class SearchWeb
     {
+        private static string GetDefaultBrowserPath()
+        {
+            string name = string.Empty;
+            try
+            {
+                var regDefault = Registry.CurrentUser.OpenSubKey("Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\FileExts\\.htm\\UserChoice", false);
+                var stringDefault = regDefault.GetValue("ProgId");
+
+                using var regKey = Registry.ClassesRoot.OpenSubKey(stringDefault + "\\shell\\open\\command", false);
+                name = regKey.GetValue(null).ToString().ToLower().Replace("" + (char)34, "");
+
+                if (!name.EndsWith("exe"))
+                    name = name.Substring(0, name.LastIndexOf(".exe") + 4);
+
+            }
+            catch
+            {
+                return string.Empty;
+            }
+
+            return name;
+        }
+
         /// <summary> 
         /// Opens search in a new browser. If no browser path is passed in then Chrome is used. 
         /// Leave browser path blank to use Chrome.
         /// </summary>
 		public static void NewBrowserWindow(this string url, string browserPath = "")
         {
+            browserPath = string.IsNullOrEmpty(browserPath) ? GetDefaultBrowserPath() : browserPath;
+
             var browserExecutableName = browserPath?
                                         .Split(new[] { Path.DirectorySeparatorChar }, StringSplitOptions.None)
                                         .Last();
@@ -44,7 +71,10 @@ namespace Flow.Launcher.Plugin.SharedCommands
         /// </summary>
         public static void NewTabInBrowser(this string url, string browserPath = "")
         {
-            var psi = new ProcessStartInfo() { UseShellExecute = true};
+            browserPath = string.IsNullOrEmpty(browserPath) ? GetDefaultBrowserPath() : browserPath;
+
+
+            var psi = new ProcessStartInfo() { UseShellExecute = true };
             try
             {
                 if (!string.IsNullOrEmpty(browserPath))
--- a/Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
@@ -4,7 +4,6 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 
-
 namespace Flow.Launcher.Plugin.SharedCommands
 {
     public static class SearchWeb
@@ -72,7 +71,6 @@ namespace Flow.Launcher.Plugin.SharedCommands
         public static void NewTabInBrowser(this string url, string browserPath = "")
         {
             browserPath = string.IsNullOrEmpty(browserPath) ? GetDefaultBrowserPath() : browserPath;
-
 
             var psi = new ProcessStartInfo() { UseShellExecute = true };
             try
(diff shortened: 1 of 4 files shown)


Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
```diff
@@ -1,18 +1,45 @@
...
 namespace Flow.Launcher.Plugin.SharedCommands
 {
     public static class SearchWeb
     {
+        private static string GetDefaultBrowserPath()
+        {
+            string name = string.Empty;
+            try
+            {
+                var regDefault = Registry.CurrentUser.OpenSubKey("Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\FileExts\\.htm\\UserChoice", false);
```
JohnTheGr8, line 17: why
`Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\FileExts\\.htm\\UserChoice`
and not
`Software\\Microsoft\\Windows\\Shell\\Associations\\UrlAssociations\\http\\UserChoice`
?

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
```diff
@@ -1,18 +1,45 @@
...
+                var stringDefault = regDefault.GetValue("ProgId");
+
+                using var regKey = Registry.ClassesRoot.OpenSubKey(stringDefault + "\\shell\\open\\command", false);
+                name = regKey.GetValue(null).ToString().ToLower().Replace("" + (char)34, "");
```
JohnTheGr8, line 21: This can be simplified to:
```csharp
name = regKey.GetValue("").ToString().Replace("\"", "");
```

Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -263,7 +263,7 @@ private async void OnCheckUpdates(object sender, RoutedEventArgs e)
 
         private void OnRequestNavigate(object sender, RequestNavigateEventArgs e)
         {
-            SearchWeb.NewBrowserWindow(e.Uri.AbsoluteUri);
+            SearchWeb.NewTabInBrowser(e.Uri.AbsoluteUri);
```
JohnTheGr8, line 266: Can you also replace the use of `SearchWeb.NewBrowserWindow` in other places in settings?

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
JohnTheGr8: pleas also add a `using` here

JohnTheGr8 (changes requested):

Flow.Launcher/SettingWindow.xaml.cs
taooceros: Sure.

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
taooceros: OK, I am not sure about the use of registry since I am not familier with it, so I just copy one and it seems it works. I will test the one you suggested and it seems it will be more reasonable to use that one instead of the htm default link.

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
taooceros: Thanks for suggestion.

Flow.Launcher/ReportWindow.xaml.cs
```diff
@@ -52,8 +52,8 @@ private Paragraph Hyperlink(string textBeforeUrl, string url)
             var link = new Hyperlink { IsEnabled = true };
             link.Inlines.Add(url);
             link.NavigateUri = new Uri(url);
-            link.RequestNavigate += (s, e) => SearchWeb.NewBrowserWindow(e.Uri.ToString());
-            link.Click += (s, e) => SearchWeb.NewBrowserWindow(url);
+            link.RequestNavigate += (s, e) => SearchWeb.NewTabInBrowser(e.Uri.ToString());
+            link.Click += (s, e) => SearchWeb.NewTabInBrowser(url);
```
jjw24, line 56: why is this changed to open as a tab?

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
```diff
@@ -1,18 +1,45 @@
...
+            return name;
+        }
+
         /// <summary> 
         /// Opens search in a new browser. If no browser path is passed in then Chrome is used. 
         /// Leave browser path blank to use Chrome.
         /// </summary>
 		public static void NewBrowserWindow(this string url, string browserPath = "")
         {
+            browserPath = string.IsNullOrEmpty(browserPath) ? GetDefaultBrowserPath() : browserPath;
```
jjw24, line 41: with getting the default browser from user, which could be any browser right, do we know if it will support opening in new window?

jjw24:
Why are we changing the behaviour of opening hyperlinks in Flow from new window to as a tab? I think the core of the issue is supporting default browser right?

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
JohnTheGr8: It works on firefox and chromium-based browsers, and there is a fix for IE a few lines later.... I think this covers pretty much every browser.

JohnTheGr8 (commented):
> 
> 
> Why are we changing the behaviour of opening hyperlinks in Flow from as a new window to as a tab? I think the core of the issue is supporting default browser right?

opening in a new tab is expected behavior I think, no app opens links in a new window AFAIK...

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
```diff
@@ -1,18 +1,45 @@
+using Microsoft.Win32;
 using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 
+
```
jjw24, line 7: lets remove this spacing

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
```diff
@@ -44,7 +71,10 @@ public static void NewBrowserWindow(this string url, string browserPath = "")
         /// </summary>
         public static void NewTabInBrowser(this string url, string browserPath = "")
         {
-            var psi = new ProcessStartInfo() { UseShellExecute = true};
+            browserPath = string.IsNullOrEmpty(browserPath) ? GetDefaultBrowserPath() : browserPath;
+
```
jjw24, line 75: remove spacing please

taooceros:
Yes, and also we didn't provide the option for user to choose open in new window or tap. Besides, the default system behavior for browsers except is opening websites in new tab.

jjw24:
Looks good to me. Good to go from you @JohnTheGr8 ?

jjw24:
I will add Flow.Launcher.Plugin version bump to the 1.5.0 version bump pr once this goes in.

JohnTheGr8 (approved):

PR #207, opened 2020-11-15
Release 1.5.0 | Plugin 1.3.0

jjw24:
Release Notes
=========

## Features

- Use default system browser #214 

- Added Slovak translations for plugins #205

- Added a quick link to open custom theme folder #202

## Bug Fixes

- Fixed unauthorised access to directory causing Flow to quit loading Win32 apps altogether in Program plugin #210

PR #215, opened 2020-11-19
Release 1.4.1

jjw24:
Release Notes
=========

## Bug fixes

- Fixed unauthorised access to directory causing Flow to quit loading Win32 apps altogether  #210


JohnTheGr8:
@jjw24 why don't we go with #207 instead?

taooceros:
> @jjw24 why don't we go with #207 instead?

I think @jjw24 wants to publish the large bug fix faster, since #207 will include more feature enhancement.

taooceros (approved):

JohnTheGr8:
I see no reason not to release everything included in #207 if we're going to make a release...

JohnTheGr8 (changes requested):
IMO we can release all changes from #207. No good reason to selectively pick these changes...

Disapproving this PR because merging it to master will cause a CI failure (as I have mentioned to @jjw24 before).

PR #223, opened 2020-11-29
fix tray menu translation

### LINKED ISSUE #211: Translations doesn't apply to context menu in notification area
Context menu in notification area contains 3 string (Open, Settings, Exit) in English only.

taooceros:
fix #211

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -53,11 +53,12 @@ namespace Flow.Launcher
         private void OnInitialized(object sender, EventArgs e)
         {
             // show notify icon when flowlauncher is hided
-            InitializeNotifyIcon();
         }
 
         private void OnLoaded(object sender, RoutedEventArgs _)
         {
+            InitializeNotifyIcon();
+
             // todo is there a way to set blur only once?
             ThemeManager.Instance.SetBlurForWindow();
             WindowsInteropHelper.DisableControlBox(this);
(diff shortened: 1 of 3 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -53,11 +53,12 @@ private void OnClosing(object sender, CancelEventArgs e)
         private void OnInitialized(object sender, EventArgs e)
         {
             // show notify icon when flowlauncher is hided
```
jjw24, line 55: this comment goes with the method right?

jjw24: also while you are at it please change word from 'hided' to 'hidden' please

jjw24:
putting it there requires a restart, is it possible to put it somewhere where restart is not required like the settings text?

Flow.Launcher/MainWindow.xaml.cs
taooceros: Sure.

taooceros:
> putting it there requires a restart, is it possible to put it somewhere where restart is not required like the settings text?

Not sure what you mean......

jjw24:
The InitializeNotificationIcon call, that will only get called when program is started. So when you change the language, the translation on the tray menu will not happen untill next start. 

Am wondering if there is a way to have it so when language is changed that method is called so when user changes the language the tray menu changes straight away as well

taooceros:
Understand, maybe we can create a public method used to update tray text when language has changed. I will take a look on that later.

taooceros:
> The InitializeNotificationIcon call, that will only get called when program is started. So when you change the language, the translation on the tray menu will not happen untill next start.

> Am wondering if there is a way to have it so when language is changed that method is called so when user changes the language the tray menu changes straight away as we

Done it

jjw24:
👍

jjw24 (approved):

PR #228, opened 2020-11-30
Update to .Net5

taooceros:
.Net 5 Upgrade.
close #228 

- Net 5 has added several improvement to System.Text.Json, and serialization with private getter and setter has been added.


JohnTheGr8:
@taooceros looking good thus far. 

Remember to also rename the publish profile: `Flow.Launcher/Properties/PublishProfiles/NetCore3.1-SelfContained.pubxml`

taooceros:
> @taooceros looking good thus far.
> 
> Remember to also rename the publish profile: `Flow.Launcher/Properties/PublishProfiles/NetCore3.1-SelfContained.pubxml`

ok

JohnTheGr8:
@taooceros you forgot to check in the renamed publish-profile file.

Plugins/Flow.Launcher.Plugin.WebSearch/Flow.Launcher.Plugin.WebSearch.csproj
```diff
@@ -119,14 +119,6 @@
...
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Page Include="SettingsControl.xaml">
-      <Generator>MSBuild:Compile</Generator>
-      <SubType>Designer</SubType>
-    </Page>
-    <Page Include="SearchSourceSetting.xaml">
-      <Generator>MSBuild:Compile</Generator>
-      <SubType>Designer</SubType>
-    </Page>
```
jjw24, line 129: is this removal intended?

Plugins/Flow.Launcher.Plugin.Program/Flow.Launcher.Plugin.Program.csproj
```diff
@@ -109,7 +109,6 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Microsoft.Windows.SDK.Contracts" Version="10.0.18362.2005" />
```
jjw24, line 112: is this removal intended?

jjw24:
apologies if you have already done so but can we also test to double check external plugins still works fine with this upgrade (just to make sure)

Plugins/Flow.Launcher.Plugin.WebSearch/Flow.Launcher.Plugin.WebSearch.csproj
taooceros: Yes, it said duplicate page file, since the default page file has included them.

taooceros:
> apologies if you have already done so, can we test to double check external plugins still works fine with this upgrade (i know it's just core to 5, but just to make sure)

Sure, no need for hurry, just put it here for checking.

Plugins/Flow.Launcher.Plugin.Program/Flow.Launcher.Plugin.Program.csproj
taooceros: Sorry, forget to response. Yes, because .Net 5 change the way of using RT. You can see on the TargetFramework of Program plugin is different from simply net5-windows but adding the version number. That replace the nuget package.

taooceros:
> @taooceros you forgot to check in the renamed publish-profile file.

Thank you for reminding!

taooceros:
Updating to .Net 5 and using WinRT seems cause the PackageManager doesn't work in Program plugin. https://github.com/dotnet/runtime/issues/45632
Not sure why that throw.

jjw24:
is this still draft or ready for review

taooceros:
This should be ready now

jjw24:
Could you please list all the tests that have been done so we can cover all basis

taooceros:
> Could you please list all the tests that have been done so we can cover all basis

Will do that after I have my computer. 😉

taooceros:
Test Checked:
1. All internal plugins work as intended. 
    - Bug for `new PackageManager()` hasn't been fixed, but it only happens on the first time, so a try catch can fix it.
2. External plugins work as intended, even when their Target Framework is netcoreapp 3.1
3. Python plugin works as intended.
4. Have used it for a while and haven't encountered other bugs.

taooceros:
Another things is that .Net 5 obsolete BinaryFormatter because of security issue (though I think it will be secure for us). Should we find an alternative for that?

jjw24:
> Another things is that .Net 5 obsolete BinaryFormatter because of security issue (though I think it will be secure for us). Should we find an alternative for that?

Where is it used?

taooceros:
> 
> > Another things is that .Net 5 obsolete BinaryFormatter because of security issue (though I think it will be secure for us). Should we find an alternative for that?
> 
> Where is it used?

Only in a few place about the cache. Win32 and UWP, and image usage storage.

jjw24:
if we can change it, lets do it. Are there any preferred alternatives?

taooceros:
> if we can change it, lets do it. Are there any preferred alternatives?

I am not quite sure. Maybe it is quite possible to simply use JsonStorage, but not sure about whether it will be slower. Another alternative is ProtoBuf-net, and I have taken a try, v2 works fine while v3 doesn't work because of the cyclic reference for UWP.

jjw24:
Remember to update readme as well

jjw24:
can we reference the actual file so people can download straight away

taooceros:
> can we reference the actual file so people can download straight away

ok

jjw24:
>External plugins work as intended, even when their Target Framework is netcoreapp 3.1

Just double checking, this means you have used and tested an external plugin with the upgraded .Net 5 Plugin api

taooceros:
> >External plugins work as intended, even when their Target Framework is netcoreapp 3.1
> 
> Just double checking, this means you have used and tested an external plugin with the upgraded .Net 5 Plugin api

Yes, I have several installed external Plugins and they work fine.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
```diff
@@ -54,22 +55,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Content Include="Languages\*.xaml">
-      <Generator>MSBuild:Compile</Generator>
-      <SubType>Designer</SubType>
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </Content>
```
jjw24, line 61: why are we removing this, does bookmark plugin not have any language files?

taooceros: Because xaml is included by default for wpf program

jjw24: cool thanks for letting me know. How come we dont do this for other projects?

README.md
```diff
@@ -78,11 +78,11 @@ Yes please, submit an issue to let us know.
 
 ## Developing/Debugging
 
-Flow Launcher's target framework is .Net Core 3.1
+Flow Launcher's target framework is .Net Core 5
```
jjw24, line 81: .Net 5, no Core any more

README.md
```diff
@@ -78,11 +78,11 @@ Yes please, submit an issue to let us know.
...
 
 Install Visual Studio 2019
 
-Install .Net Core 3.1 SDK via Visual Studio installer or manually from [here](https://dotnet.microsoft.com/download/dotnet-core/thank-you/sdk-3.1.201-windows-x64-installer)
+Install .Net Core 5 SDK via Visual Studio installer or manually from [here](https://dotnet.microsoft.com/download/dotnet/thank-you/sdk-5.0.103-windows-x64-installer)
```
jjw24, line 85: same as above

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
taooceros: Because these two are error when building🤣

taooceros: Some of the plugin doesn't have a setting page, so doesn't have to enable wpf

README.md
taooceros: ok will change that

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
jjw24: I think a couple more plugins do have settings and use wpf, could you make the change so it's consistent while you are making the others please

jjw24:
> Bug for new PackageManager() hasn't been fixed, but it only happens on the first time, so a try catch can fix it.

is this a .Net 5 bug or something we havent noticed in Core

taooceros:
> > Bug for new PackageManager() hasn't been fixed, but it only happens on the first time, so a try catch can fix it.
> 
> is this a .Net 5 bug or something we havent noticed in Core

I think it is bug for CS/WinRT, which in .net core we don't use it.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
jjw24: hold on, are you sure it's included by default?
image

jjw24:
hmm Everything plugin is not happy:
image

I suspect it's because of the infra dll too old?

taooceros:
> hmm Everything plugin is not happy:
> image
> 
> I suspect it's because of the infra dll too old?

I think that's for the whole 1.8.0 because of the migration of ISavable.

taooceros:
> > Bug for new PackageManager() hasn't been fixed, but it only happens on the first time, so a try catch can fix it.
> 
> is this a .Net 5 bug or something we havent noticed in Core

I have contacted cswinrt, and they said that they have fixed it, but it still exist though.

jjw24:
ok so we need to sort Everything out before 1.8.0 goes out then right?

taooceros:
> ok so we need to sort Everything out before 1.8.0 goes out then right?

yes, because Everything plugin use an old Flow.Infracturature, but that hasn't been loaded to Flow because Flow exist a new one. Actually, due to the new plugin api changed, everything plugin maybe able to get rid of it.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
taooceros: sure

taooceros:
Let me take a look on whether the xaml do have loaded by default.

jjw24:
bookmark plugin got the same issue...

taooceros:
> bookmark plugin got the same issue...

Will take a look on both. Maybe resource xaml is not included by default, but page xaml does😅

jjw24:
I tested adding .Net 5 Flow.Launcher.Plugin dll to WindowWalker plugin project which is .Net Core, getting this error:
image

taooceros:
> I tested adding .Net 5 Flow.Launcher.Plugin dll to WindowWalker plugin project which is .Net Core, getting this error:
> image

Yeah they cannot add it since it is higher version. However, if they keep the current Flow.Launcher.Plugin.dll, they can be loaded successfully by flow.

taooceros:
> > bookmark plugin got the same issue...
> 
> Will take a look on both. Maybe resource xaml is not included by default, but page xaml does😅

I add the Content back, and they work. Thank you for the discovery!

jjw24:
>  if they keep the current Flow.Launcher.Plugin.dll, they can be loaded successfully by flow.

but that will mean these plugins will not get the new updates to Plugin project

taooceros:
> > if they keep the current Flow.Launcher.Plugin.dll, they can be loaded successfully by flow.
> 
> but that will mean these plugins will not get the new updates to Plugin project

If they want to get the updated version, they just need to upgrade to .Net 5

jjw24:
Plugin project will need to be a major version bump then

taooceros:
> Plugin project will need to be a major version bump then

Isn't we bump that from 1.4.0 to 1.5.0? Or you mean to bump it to 2.0.0

jjw24:
bump to 2.0.0, it's a breaking change for plugins

taooceros:
Would you please do that for me? I am not quite sure how to bump version😂

jjw24:
> Would you please do that for me? I am not quite sure how to bump version😂

yeah no problem, leave it with me. I will do that when we release

jjw24:
are these formatting differences because you are using VSCode?
image

taooceros:
> are these formatting differences because you are using VSCode?
> image

I think so.

taooceros:
Well my VS will format it like this ..... 
image

jjw24:
maybe when you have time have a look at how we can sync the format. I dont want to end up with too many 'fix format' commits 😛  

When I get some time I will see if VSCode is a better tool than Visual Studio and switch if so

taooceros:
> maybe when you have time have a look at how we can sync the format. I dont want to end up with too many 'fix format' commits 😛

Yeah sure🤣🤣

> When I get some time I will see if VSCode is a better tool than Visual Studio and switch if so, because i use it for other programming stuff.


I think VSCode is quite great for us to use, even though it will not be able to us to track CPU usage and Memory usage with VSCode.
Omnisharp provides similar programing experience with VS, except that it cannot give intellisence for unusing namespace.
Omnisharp will need similar time to load project as VS. However, it won't block the main thread so that if we only want to edit a few files, it would be really fast for us to do that. 
For VS, we have to wait for the long startup time.... Generally I won't let VS open until I use it, so VSCode makes me feel really great when I need to only edit one or two line of code comparing to VS.

jjw24:
> except that it cannot give intellisence for unusing namespace (so I have to manually using System.Linq before using Select, Where, etc.).

No wonder occasionally you leave unused namespaces around 😆 

PR LGTM, couple more testing and will merge in with 1.8.0

jjw24 (approved):

Issue #220, opened 2020-11-25
Question about Plugin Development


taooceros:
A few question about the development of plugin.

**Question**
1. Is it possible for plugin to utilize some method in Flow.Infructure such as the string matcher?
2. How to debug the plugin without copying the whole folder to the Flow source project?

jjw24:
hey how's it going.

1. Definitely, and this can be exposed through the IPublicAPI interface. Also another function from Infrastructure that can be useful is the JsonStorage class for saving settings. 

2.  I can see this being useful, maybe publish Flow as a nuget as well so can be called and run by individual plugins?

taooceros:
> Definitely, and this can be exposed through the IPublicAPI interface. Also another function from Infrastructure that can be useful is the JsonStorage class for saving settings.

Great, I will work around on it and take a look on how to expose that method to IPluginAPI.

> I can see this being useful, maybe publish Flow as a nuget as well so can be called and run by individual plugins?

Interesting, but is that means that plugin may need to depend on Flow to debug, which is a bit large, and remove it when publishing? Not familiar with that.

taooceros:
There is a question when expanding the plugin api. Since IPublicAPI is under Flow.Launcher.Plugin namespace, it is unable to reach back to the class created in Flow.Launcher.Infrastructure namesapce. So if I want to reference to the class from Infrastructure, I may need to create a similar copy of that in Plugin namespace, which is weird....

jjw24:
Flow.Launcher.Plugin.PluginInitContext class contains IPublicAPI, so I think you can do something similar like ReloadAllPluginData() right?

jjw24:
> > I can see this being useful, maybe publish Flow as a nuget as well so can be called and run by individual plugins?
> 
> Interesting, but is that means that plugin may need to depend on Flow to debug, which is a bit large, and remove it when publishing? Not familiar with that.

yeah good point, will be large. Maybe a slim down version of it? i guess the goal is allow the plugin to load, start up and run on its own.

taooceros:
> Flow.Launcher.Plugin.PluginInitContext class contains IPublicAPI, so I think you can do something similar like ReloadAllPluginData() right?

[...]

PR #230, opened 2020-12-05
Expand Plugin API

taooceros:
#220 

Expand IPublicAPI for plugin users to use.

1. StringMatcher.FuzzySearch()
2. Http.GetAsync & Http.GetStreamAsync & Http.DownloadAsync
3. Add & remove ActionKeyword

Detail Implementation require further discussion.


taooceros:
@Flow-Launcher/team any idea?

Zeroto521:
add some hooks, see https://github.com/jjw24/Wox/issues/194

> I want to be hook master.

taooceros:
> add some hooks, see [jjw24/Wox#194](https://github.com/jjw24/Wox/issues/194)
> 
> > I want to be hook master.

emmm I am not sure what do you mean by hooking...... Do you mean that we should allow plugin to do more things other than querying?

jjw24:
why we exposing action key word methods?

taooceros:
> why we exposing action key word methods?

To allow plugins to have more than one actionkeyword. Currently, although we allow mutiple keywords, only internal plugins can utilize it because external plugin has no access to their actionkeywords.

taooceros:
Also, plugins won't be available to change action keywords for other plugin since they don't know the id of other plugin.

Zeroto521:
Exposed the system calling function process.

There would have a hook called `before_user_query`. 
It means to do something before the `query`.  
A simple example is that it could be used to log querying words.

taooceros:
> Exposed the system calling function process.
> 
> There would have a hook called before_user_query.
> It means to do something before the query.
> A simple example is that it could be used to log querying words.

Like registering an event that will be triggered before querying? But why don't we do the job when querying?

jjw24:
> > why we exposing action key word methods?
> 
> To allow plugins to have more than one actionkeyword. Currently, although we allow mutiple keywords, only internal plugins can utilize it because external plugin has no access to their actionkeywords.

more than one action keyword like in websearch plugin?

taooceros:
> > > why we exposing action key word methods?
> > 
> > 
> > To allow plugins to have more than one actionkeyword. Currently, although we allow mutiple keywords, only internal plugins can utilize it because external plugin has no access to their actionkeywords.
> 
> more than one action keyword like in websearch plugin?

Yes, and in Explorer plugin.

Zeroto521:
You should not do a stack of things in one place. And some hooks better could have global authority. before_all_user_query could do something before each plugin.

jjw24:
i think i have also done some changes on this stringmatcher migration but got distracted. I might merge them in if  feasible and ok

taooceros:
> You should not do a stack of things in one place. And some hooks better could have global authority. before_all_user_query could do something before each plugin.

Do you want it to be activated for plugins? Like an event, allowing plugin to do something here, that will be triggered before plugin querying every time user is querying. Or it is simply something that we just do in Flow.

taooceros:
@Zeroto521 Let's move that discussion to #267 , since this pr is majorly focusing on the APIs lie in IPublicAPI.

jjw24 (approved):

PR #231, opened 2020-12-06
Add PluginsManager for managing PluginsManifest repo

jjw24:
This PR adds the new plugin PluginsManager to manage installing, and uninstalling plugins.

- downloads plugins from PluginsManifest repo https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest
- uninstalling default and downloaded plugins

New PR will be added for the update plugins feature & context menu


Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
```diff
@@ -0,0 +1,29 @@
...
+            Settings = settings;
+        }
+
+        public List<Result> LoadContextMenus(Result selectedResult)
+        {
+            // Open website
+            // Go to source code
+            // Report an issue?
+            // Request a feature?
+            return new List<Result>();
```
taooceros, line 26: So the implementation of this method will be on next PR?

Plugins/Flow.Launcher.Plugin.PluginsManager/Models/PluginsManifest.cs
```diff
@@ -0,0 +1,35 @@
...
+        internal List<UserPlugin> UserPlugins { get; private set; }
+        internal PluginsManifest()
+        {
+            var json = string.Empty;
+
+            using (var wc = new WebClient { Proxy = Http.WebProxy() })
+            {
+                try
+                {
+                    json = wc.DownloadString("https://raw.githubusercontent.com/Flow-Launcher/Flow.Launcher.PluginsManifest/main/plugins.json");
```
taooceros, line 22: Why not simply use Http.Get since it is a text format file.

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        internal List<Result> Search(List<Result> results, string searchName)
+        {
+            if (string.IsNullOrEmpty(searchName))
+                return results;
+
+            return results
+                    .Where(x => StringMatcher.FuzzySearch(searchName, x.Title).IsSearchPrecisionScoreMet())
+                    .Select(x => x)
+                    .ToList();
```
taooceros, line 87: Why not use the score to order it?
Like 
```c#
.Where(x=>{
x.Score = StringMatcher.FuzzySearch(searchName,x.Title));
return x.Score>0;
})
```

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -0,0 +1,198 @@
...
+                            Action = e =>
+                            {
+                                Application.Current.MainWindow.Hide();
+                                InstallOrUpdate(x);
+
+                                return true;
+                            }
+                        }));
+
+            return Search(results, searchName);
```
taooceros, line 111: Well I do think it will be more elegent to simply use the `Select` method to get the resultlist but this is fine, too.

Plugins/Flow.Launcher.Plugin.PluginsManager/Utilities.cs
```diff
@@ -0,0 +1,68 @@
...
+                zipStream.CopyTo(streamWriter);
+            }
+        }
+
+        internal static string GetContainingFolderPathAfterUnzip(string unzippedParentFolderPath)
+        {
+            var unzippedFolderCount = Directory.GetDirectories(unzippedParentFolderPath).Length;
+            var unzippedFilesCount = Directory.GetFiles(unzippedParentFolderPath).Length;
+
+            // addjust path depending on how the plugin is zipped up
```
taooceros, line 48: addjust 
is this typo hhh?

Plugins/Flow.Launcher.Plugin.PluginsManager/Utilities.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            return string.Empty;
+        }
+
+        internal static void Download(string downloadUrl, string toFilePath)
+        {
+            using var wc = new WebClient { Proxy = Http.WebProxy() };
+
+            wc.DownloadFile(downloadUrl, toFilePath);
+        }
```
taooceros, line 66: Since we have referenced to Flow.Launcher.Infruastructure,  why not simply use the one lies in Http.cs?

taooceros (commented):
Great work!

Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
jjw24: yeah next pr :)

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
jjw24: score > 0 will return a lot of things that have a score, that's why we are using IsSearchPrecisionScoreMet(), it matches the precision level that the user selected, default is 50, so only results with a score equal or higher will return.

taooceros: Well, you are not right here. The score property will make every raw score lower than search precision become 0, so no worry for more results. The one you suggested that can be lower than search precision is called rawScore.

jjw24: yeah you are right score.

I dont see a benefit of doing x.score > 0 as oppose to just calling the function which has the logic embedded without having to hardcode any numbers

taooceros: I mean by changing x.score to the score from the fuzzy search can make the result list sort since the view model will sort it by its score.

jjw24: can you give me a code snippet please, not sure what you mean

taooceros: I mean by setting the score of each Result can help the result list to sort it based on the score.
Check out this https://github.com/Flow-Launcher/Flow.Launcher/blob/920668db0042460b8d08a86e89f44e2385159012/Flow.Launcher/ViewModel/ResultsViewModel.cs#L156-L201

Although I don't believe this part of codes are clear enough, but it do provides the ability to sorts the results to display on the results list view based on their score.

jjw24: Ohh gotcha, I forgot to set the score for each of the result.

jjw24: sorted. good catch, thank you

taooceros (approved):
Let's go!

Issue #233, opened 2020-12-07
Allow Plugins to utilize the CancellationToken from MainThread


taooceros:
**Discussion**

Should we allow plugins to utilize the CancellationToken from Flow so that they can cancel their works properly when the query is updated.
This can be done by adding another method to IPlugin like QueryAsync with an default implementation pointing back to the Original Query method since c# 8 has added that feature so that we can expand the interface without potential compatibility issue.
Also, should we change Parallel.Foreach(plugins,=>) to Task.run for each plugin so that every plugin can query at the same time theoretically instead of waiting potential previous one due to potential query before it on work splitting for Parallel.Foreach. Or we can simply make all things async that change the result type of the potential new QueryAsync method to Task<Result> and make everything async.

jjw24:
@taooceros is there anymore to do on this one?

taooceros:
> @taooceros is there anymore to do on this one?

That's done!

jjw24:
Legend!

PR #234, opened 2020-12-08
Use CancellationToken to cancel Explorer index search task

taooceros:
Easy fix but compromised. I think it would be better if we can use full async instead of `GetAwaitor().GetResult`.
Link to #195 Explorer plugin cause threadpool task shortage.

#233 may allow this to become better if we don't need to create a new CancellationTokenSource but use the one in the main thread, and if we can make plugin use Task<Result> as result type, this can become even better that don't need to wait the result.


jjw24:
hey @taooceros whats the issue this is fixing? and how do i repro it?

jjw24:
> hey @taooceros whats the issue this is fixing? and how do i repro it?
all good i saw the comment in the othe PR:
image

jjw24:
is this problem happening in dev?

taooceros:
> is this problem happening in dev?

Not sure, but I believe it could happen.

jjw24:
is this still needed now the async model is in?

taooceros:
> is this still needed now the async model is in?

No, I will close this one.

PR #235, opened 2020-12-08
Add space for pinyin translation

### LINKED ISSUE #229: Space between Pinyin translation
Should we add space as the separation of words translated to Pinyin? So that it won't be possible to use `gd` to search something like `DingDing`, which should not be the target.

taooceros:
close #229

### DIFF
--- a/Flow.Launcher.Infrastructure/PinyinAlphabet.cs
+++ b/Flow.Launcher.Infrastructure/PinyinAlphabet.cs
@@ -39,37 +39,32 @@ namespace Flow.Launcher.Infrastructure
                     {
                         var resultList = WordsHelper.GetPinyinList(content);
 
-                        List<int> chineseIndexs = new List<int>();
-
-                        for (int i = 0; i < content.Length; i++)
-                        {
-                            if (resultList[i].Length != 1 || !(resultList[i][0] == content[i]))
-                                chineseIndexs.Add(i);
-                        }
                         StringBuilder resultBuilder = new StringBuilder();
+                        
 
-                        foreach (var chineseIndex in chineseIndexs)
-                        {
-                            resultBuilder.Append(resultList[chineseIndex].First());
-                        }
-                        resultBuilder.Append(' ');
-
-                        int currentChineseIndex = 0;
-                        int lastChineseIndex = -1;
                         for (int i = 0; i < resultList.Length; i++)
                         {
-                            if (currentChineseIndex < chineseIndexs.Count && chineseIndexs[currentChineseIndex] == i)
+                            if (content[i] >= 0x3400 && content[i] <= 0x9FD5)
+                                resultBuilder.Append(resultList[i].First());
+                        }
+
+                        resultBuilder.Append(' ');
+
+                        bool pre = false;
+
+                        for (int i = 0; i < resultList.Length; i++)
+                        {
+                            if (content[i] >= 0x3400 && content[i] <= 0x9FD5)
                             {
                                 resultBuilder.Append(' ');
-
                                 resultBuilder.Append(resultList[i]);
-                                currentChineseIndex++;
-                                lastChineseIndex = i;
+                                pre = true;
                             }
                             else
                             {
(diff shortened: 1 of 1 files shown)


Flow.Launcher.Infrastructure/PinyinAlphabet.cs
```diff
@@ -36,8 +36,8 @@ public string Translate(string content)
                 {
                     if (WordsHelper.HasChinese(content))
                     {
-                        var result = WordsHelper.GetPinyin(content, ";");
-                        result = GetFirstPinyinChar(result) + result.Replace(";", "");
+                        var result = WordsHelper.GetPinyin(content, " ");
```
jjw24, line 39: with this change, when i pinyin search 你好 for nihao it doesnt hit as it return splitted up 'ni hao'

Flow.Launcher.Infrastructure/PinyinAlphabet.cs
```diff
@@ -59,7 +59,7 @@ public string Translate(string content)
 
         private string GetFirstPinyinChar(string content)
         {
-            return string.Concat(content.Split(';').Select(x => x.First()));
+            return string.Concat(content.Split(' ').Select(x => x.First()));
```
jjw24, line 62: need to add remove empty entries otherwise it will throw because you are doing First():
return string.Concat(content.Split(' ', StringSplitOptions.RemoveEmptyEntries).Select(x => x.First()));

Flow.Launcher.Infrastructure/PinyinAlphabet.cs
taooceros: I don't think anybody would use Chinese to search for pinyin like characters.... If both are chinese, it will be fine because the string to compare will also contain space.

taooceros: I don't think anybody would use Chinese to search for pinyin like characters.... If both are chinese, it will be fine because the string to compare will also contain space.

Flow.Launcher.Infrastructure/PinyinAlphabet.cs
taooceros: Well it isn't an issue before because the translated query won't contain something like ;;, but since we now use space to split. I will take a look on the translated query when the original query contain both Chinese and letter and space.

Flow.Launcher.Infrastructure/PinyinAlphabet.cs
jjw24: ok so we dont need to worry about chinese to pinyin.

jjw24:
hmm i think this change solves acronym pinyin search but due to the way result could be constructed, it's not hitting something like 'ni hao' or 'ni hao' for my test case, the score is too low for my test sample below:
image

although i dont know if what i am using is a valid test example. I dont use a Chinese system, can you give me an example to test

Flow.Launcher.Infrastructure/PinyinAlphabet.cs
taooceros: No, because the reason using pinyin is to make searching faster by reducing the process of typing Chinese, since typing Chinese require extra typing.

taooceros:
> hmm i think this change solves acronym pinyin search but due to the way result could be constructed, it's not hitting query like 'nihao' or 'ni hao' for my test case, the score is too low for my test example below:
> image
> 
> although i dont know if what i am using is a valid test example. I dont use a Chinese OS system, can you give me an example to test

I believe the weird situation of your testing is due to that the space will be added to all character used for translation due to the Pinyin library characteristic. The situation also exist in current situation, though not significantly which only affect the acronym part. I will contact the author to ask whether he is able to add the split character only to Chinese character instead of all character. Before his response, I will try to find alternative solution.

taooceros:
@jjw24 I manually set the space instead of using the method provided by the library. It seems works fine, now.

jjw24 (approved):

PR #239, opened 2020-12-15
Plugin installation and uninstallation restart without prompt

jjw24:
Directly restart flow without another prompt after plugin install/uninstall. This helps make the experience more seamless

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/Languages/en.xaml
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/Languages/en.xaml
@@ -6,13 +6,11 @@
     <system:String x:Key="plugin_pluginsmanager_downloading_plugin">Downloading plugin</system:String>
     <system:String x:Key="plugin_pluginsmanager_please_wait">Please wait...</system:String>
     <system:String x:Key="plugin_pluginsmanager_download_success">Successfully downloaded</system:String>
-    <system:String x:Key="plugin_pluginsmanager_uninstall_prompt">Do you want to uninstall the following plugin?{0}{1}{2} by {3}</system:String>
-    <system:String x:Key="plugin_pluginsmanager_install_prompt">Do you want to install the following plugin?{0}{1}{2} by {3}</system:String>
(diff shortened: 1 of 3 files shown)


taooceros:
Do we have to restart Flow when installing plugin? Is it possible to install Plugins without restart?

JohnTheGr8:
> 
> 
> Do we have to restart Flow when installing plugin? Is it possible to install Plugins without restart?

Well, I've looked into this in the past, so let me answer.... we could use [DotNetCorePlugins](https://github.com/natemcmaster/DotNetCorePlugins) to load plugins dynamically without the need to reload...

**but**, there is sadly [an outstanding issue](https://github.com/natemcmaster/DotNetCorePlugins/issues/141) with this library when loading plugins that have `UseWPF`.

taooceros:
> > Do we have to restart Flow when installing plugin? Is it possible to install Plugins without restart?
> 
> Well, I've looked into this in the past, so let me answer.... we could use [DotNetCorePlugins](https://github.com/natemcmaster/DotNetCorePlugins) to load plugins dynamically without the need to reload...
> 
> **but**, there is sadly [an outstanding issue](https://github.com/natemcmaster/DotNetCorePlugins/issues/141) with this library when loading plugins that have `UseWPF`.

Seems good, but is it possible to load dynamically with our current infrastructure in PluginManager?

Current we load plugin like this.
```c#
public static void LoadPlugins(PluginsSettings settings)
        {
            _metadatas = PluginConfig.Parse(Directories);
            Settings = settings;
            Settings.UpdatePluginSettings(_metadatas);
            AllPlugins = PluginsLoader.Plugins(_metadatas, Settings);
        }
```

So is it possible when installing plugin by adding one plugin in AllPlugins and update Settings, or we just simply reload all plugin instead of restarting flow, or this is unallowed?

jjw24:
> So is it possible when installing plugin by adding one plugin in AllPlugins and update Settings, or we just simply reload all plugin instead of restarting flow,

To be honest i havent really investigated this option. Either way if we can it would be great.

> or this is unallowed?

definately allowed. I dont see why we wouldnt allow this

jjw24:
in the mean time though, lets get this easy change out please

jjw24:
created an issue to capture this https://github.com/Flow-Launcher/Flow.Launcher/issues/241

taooceros (approved):

PR #243, opened 2020-12-17
Update plugins via Pluginsmanager

jjw24:
- use ```pm update``` to see a list of available plugins to update
- prompt to go to update section when user tries to install an already existing plugin but has an update pending
- update plugin method call


Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -62,25 +92,89 @@ internal void InstallOrUpdate(UserPlugin plugin)
...
+
+                                                Context.API.RestartApp();
+
+                                                return true;
+                                            }
+
+                                            return false;
+                                        }
+                                    })
+                            .ToList();
```
taooceros, line 168: Actaully, we can remove the ToList() here to keep it as an IEnumerable to make Linq calculate the where and select method together, which should be faster (though not significant), since this result is only calculated once else where.

taooceros (approved):
Seems Great

jjw24:
done please rereview

jjw24:
done please rereview @taooceros

taooceros (approved):
Great work!

jjw24:
awesome thanks. 

Do you have the ability to merge as well?

PR #246, opened 2020-12-20
Update url website in Settings

jjw24:
Update the website url in Settings' About tab

### DIFF
--- a/Flow.Launcher.Infrastructure/Constant.cs
+++ b/Flow.Launcher.Infrastructure/Constant.cs
@@ -35,5 +35,7 @@ namespace Flow.Launcher.Infrastructure
         public const string DefaultTheme = "Darker";
 
         public const string Themes = "Themes";
+
+        public const string Website = "https://flow-launcher.github.io";
     }
 }
--- a/Flow.Launcher/SettingWindow.xaml
+++ b/Flow.Launcher/SettingWindow.xaml
@@ -429,8 +429,8 @@
                 <TextBlock Grid.Row="0" Grid.ColumnSpan="2" Text="{Binding ActivatedTimes, Mode=OneWay}" FontSize="12" />
                 <TextBlock Grid.Row="1" Grid.Column="0" Text="{DynamicResource website}"/>
                 <TextBlock Grid.Row="1" Grid.Column="1" HorizontalAlignment="Left">
-                    <Hyperlink NavigateUri="{Binding Github, Mode=OneWay}" RequestNavigate="OnRequestNavigate">
-                        <Run Text="{Binding Github, Mode=OneWay}" />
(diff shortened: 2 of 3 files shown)


JohnTheGr8:
@jjw24 let's also update the `Find more plugins` link. We could make it point to the manifest repo and have a (categorized) list of plugins in the readme, perhaps?

jjw24:
> @jjw24 let's also update the `Find more plugins` link. We could make it point to the manifest repo and have a (categorized) list of plugins in the readme, perhaps?

good pickup.

>We could make it point to the manifest repo

hmm maybe instead just change the query to list the plugins? Ideally even load the plugins in the Settings page, this maybe something to come in the next iteration.

> have a (categorized) list of plugins in the readme, perhaps?

This means we also need to maintain the list in the readme right? Add a new plugin in the manifest json and then need to also add to the list in the readme.

Ideally users dont need to visit the repo and plugin devs only need to do so to add or update their plugin

JohnTheGr8:
> hmm maybe instead just change the query to list the plugins? Ideally even load the plugins in the Settings page, this maybe something to come in the next iteration.

I dont know, people _would_ expect a link to open when they click on this. If the query window opened instead, it could be unexpected/awkward...

> This means we also need to maintain the list in the readme right? Add a new plugin in the manifest json and then need to also add to the list in the readme.

Not a big maintenance cost, is it? Plugin devs update the readme once in their first PR, just an extra line. This can include name, short description, link to source code and issue tracker.

> Ideally users dont need to visit the repo and plugin devs only need to do so to add or update their plugin

Having all plugins listed and organized somewhere would help, especially for potential users that are trying to decide if Flow is worth a try. But anyway, it's just an idea...

jjw24:
> > hmm maybe instead just change the query to list the plugins? Ideally even load the plugins in the Settings page, this maybe something to come in the next iteration.
> 
> I dont know, people _would_ expect a link to open when they click on this. If the query window opened instead, it could be unexpected/awkward...
> 

yeah true. What about just removing it and implement a category system when more plugins are submitted? When we got more plugins we can use the category system when loading in Setting. 

> > This means we also need to maintain the list in the readme right? Add a new plugin in the manifest json and then need to also add to the list in the readme.
> 
> Not a big maintenance cost, is it? Plugin devs update the readme once in their first PR, just an extra line. This can include name, short description, link to source code and issue tracker.
> 
> > Ideally users dont need to visit the repo and plugin devs only need to do so to add or update their plugin
> 
> Having all plugins listed and organized somewhere would help, especially for potential users that are trying to decide if Flow is worth a try. But anyway, it's just an idea...

jjw24:
Updated the url. Please re-review

JohnTheGr8 (approved):

PR #249, opened 2020-12-27
Plugin manager improve

taooceros:
1. Don't create PluginsManager instance every time query, which is really slow due to internet issue. Instead, update plugin manifest every 12 hours, triggered by query.
2. Don't display potential plugins when typing "pm", but instead display hotkeys. (Just personal idea, since I don't think people will know it can be used to update or uninstall if displaying plugin list on "pm")


Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
```diff
@@ -3,20 +3,28 @@
...
+using Flow.Launcher.Infrastructure;
+using System;
+using System.Threading.Tasks;
 
 namespace Flow.Launcher.Plugin.PluginsManager
 {
     public class Main : ISettingProvider, IPlugin, ISavable, IContextMenu, IPluginI18n
     {
-        internal PluginInitContext Context { get; set; }
+        internal static PluginInitContext Context { get; set; }
```
jjw24, line 16: does this need to be static, since we are passing it through constructors already

jjw24:
thanks for making this change, really appreciate it. 

I might make a couple tweaks just more around formatting and push to this PR if ok

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -300,6 +324,7 @@ private void Uninstall(PluginMetadata plugin)
             using var _ = File.CreateText(Path.Combine(plugin.PluginDirectory, "NeedDelete.txt"));
         }
 
+
```
jjw24, line 327: how come some times your changes contain extra space?

taooceros: hmmm, I don't know…… Maybe due to auto formatting?

Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
taooceros: Oh we shouldn't make it static, though I think it is singleton. That has been changed to because I want to pass Context.API to the delegate to the default hotkey list.

taooceros:
> thanks for making this change, really appreciate it.
> 
> I might make a couple tweaks just more around formatting and push to this PR if ok

Sure!

Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
taooceros: Let me check if there is a better way of passing variable.

taooceros:
A quick question, do we want the hotkeys here to be able to be modified? If not, why not make the strings to be const? @jjw24

jjw24:
> A quick question, do we want the hotkeys here to be able to be modified? If not, why not make the strings to be const? @jjw24

yeah that is something i thought would be nice later down the track to allow it to be modifiable by users, although it just wasnt a priority to impment for now

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -66,105 +113,115 @@ internal void InstallOrUpdate(UserPlugin plugin)
...
-                          NewVersion = pluginFromManifest.Version,
-                          existingPlugin.Metadata.IcoPath,
-                          PluginExistingMetadata = existingPlugin.Metadata,
-                          PluginNewUserPlugin = pluginFromManifest
-                      };
+            var resultsForUpdate =
+                from existingPlugin in Context.API.GetAllPlugins()
+                join pluginFromManifest in pluginsManifest.UserPlugins
+                    on existingPlugin.Metadata.ID equals pluginFromManifest.ID
+                where existingPlugin.Metadata.Version.CompareTo(pluginFromManifest.Version) > 0
```
jjw24, line 167: just going to make a change here, dont use compareto, use equals or ==

use compareto for sorting

taooceros: Shouldn't we only update plugin if the version is higher on manifest instead of they are not equal?
Sorry, the comparison should be less than 0.😖

jjw24: my apologies, you are right actually. I was coming from the perception that we dont bother actually compare the version strings, since new updates should have version greater than the previous, but using compareto is better code.

I have changed it to less than 0

jjw24:
looks good. Will merge and lets get the release rolling 👍

jjw24 (approved):

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
taooceros: Haha thank you. I found that due to my computer has an unpublished window walker, which has version higher than the one in our manifest. 😝

PR #253, opened 2020-12-30
Replace Json.Net with System.Text.Json

### LINKED ISSUE #252: Replace Newtonsoft.Json with System.Text.Json
Since we are on .net core, why not try the new System.Text.Json? No need to reference to Json.Net anymore. It should be faster and more light-weighted.

taooceros:
close #252


jjw24:
hey @taooceros have you tested this new serializer and working with settings and our default plugins?

taooceros:
Yes, I have tested with default plugin and the setting, and found one issue. Originally, Json.Net will use ignore null value to load setting.json, while System.Text.Json doesn't successfully load that with the setting, but successfully load it without ignore null value.
Also, it doesn't support manually property ordering, so I put the plugins setting to the later part of the Settings.cs

taooceros:
It seems work fine.

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
```diff
@@ -100,8 +98,12 @@ public bool HideNotifyIcon
 
         public HttpProxy Proxy { get; set; } = new HttpProxy();
 
-        [JsonConverter(typeof(StringEnumConverter))]
+        [JsonConverter(typeof(JsonStringEnumConverter))]
         public LastQueryMode LastQueryMode { get; set; } = LastQueryMode.Selected;
+
+
+        // Order defaults to 0 or -1, so 1 will let this property appear last
```
jjw24, line 105: since this comment is no longer relevant. lets replace it with 
'This needs to be loaded last by staying at the bottom'

taooceros: ok

Flow.Launcher.Infrastructure/Storage/JsonStorage.cs
```diff
@@ -24,10 +24,9 @@ internal JsonStrorage()
         {
             // use property initialization instead of DefaultValueAttribute
             // easier and flexible for default value of object
-            _serializerSettings = new JsonSerializerSettings
+            _serializerSettings = new JsonSerializerOptions
             {
-                ObjectCreationHandling = ObjectCreationHandling.Replace,
-                NullValueHandling = NullValueHandling.Ignore
+                IgnoreNullValues = false
```
jjw24, line 29: I am assuming these options and comment are no longer applicable?

taooceros: yes, the default behavior of System.Text.Json is replace, and ignore doesn't work well as my comment before.

jjw24 (approved):

Issue #256, opened 2020-12-31
'pm install' does not install packages


deefrawley:
**Describe the bug/issue**
The new package manager doesn't install packages. I can choose the package, I get the mesasage box with the details, and then the two toaster popups saying the package has been downloaded but Flow doesn't restart and the packages are not available or located in the %APPDATA%/Roaming/FlowLauncher directory. The log files logs nothing about the attempt.

Note I have completely uninstalled Flow and deleted all %APPDATA% directories afterwards for a clean install and issue persists. Running explicitly as Administrator doesn't help (Flow crashed the first time I did this and start typing 'pm' but I haven't been able to reproduced that).

**To Reproduce**
1. Activate Flow
2. Type 'pm install' and choose any package
3. Click 'yes' on the message box showing the plugin details and confirming you want to install
4. Toaster popup saying 'downloading...''
5. Toaster popup saying downloading has finished.
6. Nothing else happens. Enter Flow Settings - Plugins tab and nothing new listed.

**Your System**
```
Windows build number: 10.0.19042.6851
Flow Launcher version: 1.6 
```
**Flow Launcher Error Log**
[2021-01-01.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/5758428/2021-01-01.txt)

taooceros:
@jjw24 
![image](https://user-images.githubusercontent.com/45326534/103431818-46264400-4c11-11eb-976a-f46f8433a985.png)
Seems the translation in catch is false.

taooceros:
> (Flow crashed the first time I did this and start typing 'pm' but I haven't been able to reproduced that)

That may because you access the downloaded manifest before it has downloaded successfully. Will add a default List for access, since plugins are initialized async.

taooceros:
Haven't been able to reproduce the error...... Is it possible for me to provide you with an edited version of plugin manager with more error handling so that you can log the error so we can look up?

taooceros:
[Flow.Launcher.Plugin.PluginsManager.zip](https://github.com/Flow-Launcher/Flow.Launcher/files/5758536/Flow.Launcher.Plugin.PluginsManager.zip)

Take a try on replacing the Flow.Launcher.Plugin.PluginsManager.dll with this one and check whether Logger can record the error?

deefrawley:
Replaced the file and Flow won't even launch now, no log created at all. 
[...]

PR #257, opened 2021-01-01
Fix error handling in PluginsManager and plugins query, fixed postbuild script duplicate deletion during build

jjw24:
- Fix wrong error message for PluginsManager
- Added try catch for install plugin call
- Added try catch for querying plugin and plugins tasks
- Fixed postbuild script to remove duplicate with the condition that the versions are the same. Specific to issue 256, the direct  package reference (ICSharpCode.SharpZipLib 1.2.0) used by PluginsManager has a different version to the output of other packages' ICSharpCode.SharpZipLib (0.86) dependency.

Closing #256

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -422,7 +422,7 @@ namespace Flow.Launcher.ViewModel
                                     }
                                     catch(Exception e)
                                     {
-                                        Log.Exception($"|MainViewModel|Exception when querying {plugin.Metadata.Name}", e);
+                                        Log.Exception("MainViewModel", $"Exception when querying the plugin {plugin.Metadata.Name}", e, "QueryResults");
                                     }
                                 }
                             });
@@ -442,7 +442,7 @@ namespace Flow.Launcher.ViewModel
                         }
                     }, currentCancellationToken).ContinueWith(t =>
                     {
-                        Log.Exception("|MainViewModel|Error when querying plugin", t.Exception?.InnerException);
+                        Log.Exception("MainViewModel", "Error when querying plugins", t.Exception?.InnerException, "QueryResults");
                     }, TaskContinuationOptions.OnlyOnFaulted);
                 }
             }
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -21,6 +21,7 @@ using Flow.Launcher.Plugin.SharedCommands;
 using Flow.Launcher.Storage;
 using System.Windows.Media;
 using Flow.Launcher.Infrastructure.Image;
+using Flow.Launcher.Infrastructure.Logger;
 
 namespace Flow.Launcher.ViewModel
 {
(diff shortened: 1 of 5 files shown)


taooceros:
Would you please add a continue with for the task that do the query jobs of Plugins to catch the exception thrown by Plugins? There is an option that will only run the continue with on fault, which means error thrown.

taooceros:
> Would you please add a continue with for the task that do the query jobs of Plugins to catch the exception thrown by Plugins? There is an option that will only run the continue with on fault, which means error thrown.

Or maybe a try catch within the parallel. Foreach to catch the error and throw them to log.

taooceros:
@jjw24 I added the exception handling for plugin querying, please check.

jjw24:
> @jjw24 I added the exception handling for plugin querying, please check.

looks good, thank you

taooceros:
Just a question before approval, why should we have different version of the same dll reference? I don't believe we use sharpziplib in other places than our PluginsManager, so why should that fix it?

jjw24:
not as a direct reference, but some assemblies possibly in Infrastrucutre project use it as an indirect reference as far as i understand,  which results in the dll being outputted in the parent folder. Additionally our default plugins still reference Infrastructure, and because we need plugins to output all dlls, the indirect referenced sharpziplib is also being outputted in those plugins.

taooceros:
Just a question before approval, why should we have different version of the same dll reference? I don't believe we use sharpziplib in other places than our PluginsManager, so why should that fix it?

> not as a direct reference, but some assemblies possibly in Infrastrucutre project use it as an indirect reference (different version) as far as i understand, which results in the dll being outputted in the parent folder. Additionally our default plugins still reference Infrastructure, and because we need plugins to output all dlls, the indirect referenced sharpziplib is also being outputted in those plugins.

Oh you are right, but why don't we make those version the same? Since the only reason we reference to Infrastructure is that this plugin is native.

jjw24:
> Just a question before approval, why should we have different version of the same dll reference? I don't believe we use sharpziplib in other places than our PluginsManager, so why should that fix it?
> 
> > not as a direct reference, but some assemblies possibly in Infrastrucutre project use it as an indirect reference (different version) as far as i understand, which results in the dll being outputted in the parent folder. Additionally our default plugins still reference Infrastructure, and because we need plugins to output all dlls, the indirect referenced sharpziplib is also being outputted in those plugins.
> 
> Oh you are right, but why don't we make those version the same? Since the only reason we reference to Infrastructure is that this plugin is native.

well our direct reference is 1.2.0, the indirect one is 0.89. Dont think there is a easy way to lift the other assemblies' referenced version

taooceros:
> > Just a question before approval, why should we have different version of the same dll reference? I don't believe we use sharpziplib in other places than our PluginsManager, so why should that fix it?
> > > not as a direct reference, but some assemblies possibly in Infrastrucutre project use it as an indirect reference (different version) as far as i understand, which results in the dll being outputted in the parent folder. Additionally our default plugins still reference Infrastructure, and because we need plugins to output all dlls, the indirect referenced sharpziplib is also being outputted in those plugins.
> > 
> > 
> > Oh you are right, but why don't we make those version the same? Since the only reason we reference to Infrastructure is that this plugin is native.
> 
> well our direct reference is 1.2.0, the indirect one is 0.89. Dont think there is a easy way to lift the other assemblies' referenced version

hmmm you are right. Maybe we should take a look later, but merging this now.

taooceros (approved):

PR #272, opened 2021-01-04
Use ValueTask instead of Task

taooceros:
Since most of the image will be stored on cache, so synchronous action will be done quite significant for SetImage method in ResultViewModel, using ValueTask instead of Task can avoid those cost.

### DIFF
--- a/Flow.Launcher/ViewModel/ResultViewModel.cs
+++ b/Flow.Launcher/ViewModel/ResultViewModel.cs
@@ -15,7 +15,7 @@ namespace Flow.Launcher.ViewModel
     {
         public class LazyAsync<T> : Lazy<ValueTask<T>>
         {
-            private T defaultValue;
+            private readonly T defaultValue;
 
             private readonly Action _updateCallback;
             public new T Value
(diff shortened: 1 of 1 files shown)


jjw24:
do we know if this will bring a benefit? (https://stackoverflow.com/questions/43000520/why-would-one-use-taskt-over-valuetaskt-in-c#:~:text=Use%20Task%20when%20something%20is,you%20can't%20know%20the)

whats the improvement here using ValueTask and do we know for sure is worth it?

taooceros:
It can avoid the task allocation since most of our image processing is sync, although wrapped in an async action. If the image is in the cache, it will run synchronizely. So there will be some waste when using task.
The copy issue isn't large because the value task in lazy async will only be assessed as a field, so it won't be copied. It will only be copied at the first time using the factory to create the task.
I don't believe it will bring significant performance boost or memery reduce, but value task should be a better usage here.

taooceros:
> Use Task when something is usually or always going to be genuinely asynchronous, i.e. not immediately complete; use ValueTask when something is usually or always going to be synchronous, i.e. the value will be known inline; also use ValueTask in a polymorphic scenario (virtual, interface) where you can't know the answer.

jjw24 (approved):

Issue #248, opened 2020-12-27
Fody block building


taooceros:
I have encountered an issue that I cannot build Flow due to Fody's issue. That can be solved by updating PropertyChangeFody to the latest version. Should we do that?
![image](https://user-images.githubusercontent.com/45326534/103167368-8f6b4200-4865-11eb-8cfa-39e388139984.png)

jjw24:
how are you build flow? is it via VS 2019?

we kept fody version at it is now due runtime problems encountered after updating the version. for references:
https://github.com/Flow-Launcher/Flow.Launcher/pull/35
https://github.com/Flow-Launcher/Flow.Launcher/pull/52

taooceros:
I use Rider to build it, since I am on another computer at that time.
That's strange, since Rider build it well with version 3+.
Is it possible for us to remove Fody in our project as the PR said?

jjw24:
Yeah of course if it improves flow. I did try removing it and the query result list stop registering/showing up. So I think it'likely we need replace fody instead

taooceros:
Seems that Fody plays a quite important job in Flow, by adding property change to every property for binding that extends the BaseModel, which is almost every model in the Flow. Removing it means that we need to manually add the property changes event raise for every property extends the class, which is huge......

JohnTheGr8:
worth mentioning that Fody also breaks running `dotnet publish` on the solution, which prevents us from fully building the project on appveyor

taooceros:
Check around google and github, if we want to remove Fody, there are only a few choices.
1. Use PostSharp, which is not open-source, but free for community edition. It should do similar jobs as Fody by injecting code for IL after build.
2. Most of the repository to help people implementing inotifypropertychange easier is simply use a backward SetProperty Method with an abstract class. It is possible to choose one or implement our own, but both require significant works, since we have quite significant amount of property that raise the event to inform wpf to update their value.

JohnTheGr8:
Hmmm... after some testing, it looks like the newest version of `PropertyChanged.Fody` resolves the runtime issues we saw in the past (#52). Updating also fixes the issues with `dotnet publish`, so I am looking into getting the self-contained build/release working with CI.

Leave this to me for now, I'll get back with a PR or more details...

PR #260, opened 2021-01-02
Release 1.7.0 | Plugin 1.4.0

jjw24:
Release Notes
=========

## Features
- Added search using acronyms #185 
- Add and remove files or folders in Explorer's Quick Access via context menu, you can now also launch Quick Access files #315
- Added additional functionalities to Flow's IPlugin interface, this means plugins can all utilise: #230 
  - FuzzySearch to search for user's query against strings and text
  - Http.GetAsync, Http.GetStreamAsync and Http.DownloadAsync to download from the web
  - Plugins can also use add & remove ActionKeyword fuctionalities
- Improve Explorer plugin's handling of long running files and folders search, additionally enhanced result list update speed #301 
- Hide Flow's progress bar when search window is hidden #295 
- Auto re-index Programs plugin when re-index has not run for more than 3 days #312 
- New themes Nord, Nord Darker and BlurBlack Darker #304  
- Improved the logic for how search results are rendered and new results are added to the results list #195 
- Added plugin priority setting, you can now set the priority of each plugin's result manually #276 
- Added ability to manually reload the plugin manifest list, use `reload plugin data` #263 
- Updated to avoid the task allocation during image processing #272 
- Improved search window's icon(magnifying glass) rendering and quality #278 
- Improved plugin's asynchronous operations and introduced new async interfaces #267  
- Create and deploy releases from CI #275 
- Upgraded the library used to load settings #253 
- Added Slovak translation for PluginsManager #283 

## Bug fixes
- Fixed issue where the query `c:\>` does not use directory search #309 
- Fixed missing translation files #259 
- Use wildcard to include images in projects #261 
- Fixed error causing PluginsManager unable to install plugins #257 
- Added additional error handling and error logging when updating/downloading plugins #270 #271 
- Fixed several untranslatable settings text #311 
- Fixed Slovak translation for settings, Url and WebSearch plugins text #316 
- Fixed Http Proxy not initilized at startup and not working in some case. #320 
- Fixed Explorer not opening folder path correctly when path contains a comma #321

PR #275, opened 2021-01-05
Create and deploy releases from CI

JohnTheGr8:
This is the continuation of the work done in #35 back in May.

### Changes

The `post_build.ps1` script has been repurposed to be called from `AppVeyor` to create a self-contained version of the project. Squirrel artifacts are deployed to Github as a draft release, which means the only manual work for a release is to edit the draft release with release notes and press `Publish`.

Note that I have removed `post_build.ps1` as a post-build script of the main project because the script calls `dotnet publish` which has to build the main project again, and that would result in a build loop. I have also bumped `PropertyChanged.Fody` to `3.3.1` (which made this entire PR possible, see #248) and added `SatelliteResourceLanguages` to some plugin projects to reduce the output files.

### Objective

- merging into master will create a draft release that we can sanity-check and add release notes before publishing
- publishing the release will create the tag and cause a new build that will update the release assets on github. We could skip this step, but I think it's better to have release assets that match those built from the release tag.

### Todo

I open this PR as a draft with my own Github API credentials so that you guys can see/test that it works as expected and give your feedback. You can see that `AppVeyor` has produced all the [expected artifacts](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher-wl238/builds/37109001/artifacts) and that they have been deployed to a [Github release on my fork](https://github.com/JohnTheGr8/Flow.Launcher/releases/tag/v1.6.0) (which I manually released).

~~Before we can merge this, @jjw24 I will need [a Github token](https://github.com/settings/tokens) with the `public_repo` scope, [encrypted by AppVeyor](https://ci.appveyor.com/tools/encrypt) to replace my own. I will also need to add the conditional to only deploy to Github from the master branch.~~

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,8 +30,6 @@ after_build:
   - ps: .\Scripts\post_build.ps1
 
 artifacts:
-- path: 'Output\Packages\Flow-Launcher-*.zip'
-  name: Zip
 - path: 'Output\Release\Flow.Launcher.Plugin.*.nupkg'
   name: Plugin nupkg
 - path: 'Output\Packages\Flow-Launcher-*.exe'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,6 +26,8 @@ before_build:
 build:
   project: Flow.Launcher.sln
   verbosity: minimal
+after_build:
+  - ps: .\Scripts\post_build.ps1
 
 artifacts:
 - path: 'Output\Packages\Flow-Launcher-*.zip'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,6 +34,12 @@ artifacts:
   name: Zip
 - path: 'Output\Release\Flow.Launcher.Plugin.*.nupkg'
   name: Plugin nupkg
+- path: 'Output\Packages\Flow-Launcher-*.exe'
+  name: Squirrel Installer
+- path: 'Output\Packages\FlowLauncher-*-full.nupkg'
+  name: Squirrel nupkg
+- path: 'Output\Packages\RELEASES'
+  name: Squirrel RELEASES
 
 deploy:
   provider: NuGet
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,9 +42,30 @@ artifacts:
   name: Squirrel RELEASES
 
 deploy:
-  provider: NuGet
-  artifact: /.*\.nupkg/
-  api_key:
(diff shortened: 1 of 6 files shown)


taooceros:
A small question toward that, why not use Github Action?

jjw24:
this should be ready for review yeah? @JohnTheGr8 

Do we need to pin artifact published on master branch?

does this also solve plugin nuget release if no changes made?

appveyor.yml
```diff
@@ -26,17 +26,44 @@ before_build:
...
+
+  - provider: GitHub
+    release: v$(flowVersion)
+    auth_token:
+      secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
+    artifact: Squirrel Installer, Squirrel nupkg, Squirrel RELEASES
+    force_update: true
+    on:
+      branch: master
+      APPVEYOR_REPO_TAG: true
```
jjw24, line 69: is this correct? apparently the tag eval is not going to work if you combine it with branch:
image

JohnTheGr8:
> 
> this should be ready for review yeah? @JohnTheGr8

Well, not exactly. I have yet to finalize the deployment conditions, as you have noticed. I wanted to have PRs to master create and update draft releases... but screw that, let's keep it simple.

I'm pushing the last changes in a moment.

> 
> Do we need to pin artifact published on master branch?

Not sure what you mean here, can you elaborate?

> 
> does this also solve plugin nuget release if no changes made?

This was actually a non-issue. You can see that in any of the tag builds in appveyor, publishing to nuget is simply skipped without failing. [Example here](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/37056791), scroll to the output end.

appveyor.yml
JohnTheGr8: @jjw24 deploy conditions are not finalized, see my other comment

JohnTheGr8:
rebased to resolve merge conflicts, @jjw24 have a look

jjw24:
LGTM 👍 

Can we create a separate issue for this:
I wanted to have PRs to master create and update draft releases

jjw24 (approved):

JohnTheGr8:
> 
> 
> LGTM 👍
> 
> Can we create a separate issue for this:
> 
> > I wanted to have PRs to master create and update draft releases

I tried to do it with appveyor's environment variables with no luck, and I'd rather not waste more time on this... I think just pulling the exe from appveyor artifacts is good enough. So, let's not.

It also makes it possible to update the current stable release when opening a new release PR to master without a version bump (like we had with #260 until recently)

Issue #212, opened 2020-11-18
Text rendering issue on HiDPI screen


mimecry:
**Describe the bug/issue**
On a HiDPI screen, the interface seems to be _somewhat_ blurry and lacking in sharpness.

**To Reproduce**
1. Install Flow on a machine with HiDPI display (i'm using a 14" 2560x1440 screen at 150% scaling)
2. Launch the software

**Screenshots**
The following is a screenshot comparison between Flow 1.4.0 (top, dark blue) and the latest Wox pre-release (1.4.1196) (bottom, teal), both using Ebrima font at regular size.
![image](https://user-images.githubusercontent.com/10742312/99479712-4cc26a00-2989-11eb-9946-46a4466183b5.png)

When zoomed in, you can see that the text on Flow is more pixelated
![image](https://user-images.githubusercontent.com/10742312/99480101-0588a900-298a-11eb-8d51-50eb39c3df27.png)

Edit: Perhaps this issue also applies to other UI elements aside from text. Here is a zoomed in screenshot of the search icon
![image](https://user-images.githubusercontent.com/10742312/99480510-ce66c780-298a-11eb-8dd8-1009a433633b.png)

**Your System**
```
Windows build number: Version 10.0.18363.418
Flow Launcher version: 1.4.0
```

mimecry:
recently redownloaded Flow and it looks great now, thank you for all your work

taooceros:
> recently redownloaded Flow and it looks great now, thank you for all your work

it has been fixed??? Surprise🤣

Garulf:
Great to hear the issue was resolved!

PR #278, opened 2021-01-06
Image quality

### LINKED ISSUE #273: Better magnifying glass icon on query window
Can do with a better magnifying glass icon, current one is too fuzzy/not a hd image

current:
image

taooceros:
fix #273, also link to the image fuzzy for #212


taooceros:
I take a try on both svg and the high quality bitmap scale mode, both displays result similarly. However, both contains a few fuzzy part (not significant and much better than current image). I believe that's due to that we only provide a 48*48 place for displaying the image.

Flow.Launcher/MainWindow.xaml
```diff
@@ -68,8 +69,8 @@
                          Margin="18,0,56,0">
                         <TextBox.Text>
                             <MultiBinding Converter="{StaticResource QuerySuggestionBoxConverter}">
-                                <Binding ElementName="QueryTextBox" Path="Text"/>
-                                <Binding ElementName="ResultListBox" Path="SelectedItem"/>
+                                <Binding ElementName="QueryTextBox" Path="Text" Mode="OneWay"/>
+                                <Binding ElementName="ResultListBox" Path="SelectedItem" Mode="OneWay"/>
```
jjw24, line 73: how come we are adding binding to one way for this change?

Flow.Launcher/MainWindow.xaml
```diff
@@ -92,7 +93,7 @@
                             </ContextMenu>
                         </TextBox.ContextMenu>
                     </TextBox>
-                    <Image Source="{Binding Image, IsAsync=True}" Width="48"  HorizontalAlignment="Right" />
+                    <svgc:SvgControl Source="{Binding Image, Mode=OneWay}" HorizontalAlignment="Right" Width="48" Height="48"/>
```
jjw24, line 96: same how come this is one way now?

Flow.Launcher/MainWindow.xaml
```diff
@@ -81,7 +82,7 @@
                      AllowDrop="True"
                      Visibility="Visible"
                     Background="Transparent"
-                    Margin="18,0,56,0">
+                    Margin="18,1,0,1" HorizontalAlignment="Left" Width="660">
```
jjw24, line 85: what's this change for?

jjw24:
now that we switched to svg, should we archive mainsearch.png into the 'doc' folder only for future reference instead?

Flow.Launcher/MainWindow.xaml
taooceros: Sorry, that was added incidently. The svg seems make by designer break every time after debugging, so I just try to check whether this led to that. I will remove it.

Flow.Launcher/MainWindow.xaml
taooceros: This one is oneway because the search icon won't change.

Flow.Launcher/MainWindow.xaml
taooceros: hmmm that's not intended, I will remove it.

taooceros:
> now that we switched to svg, should we archive mainsearch.png into the 'doc' folder only for future reference instead?

If we do deside to switch to svg, we can archive the png one, since it is only used here.

Flow.Launcher/MainWindow.xaml
taooceros: Found that why I changed here to OneWay. This is the auto suggest factory, so we won't need to let user to change something here to move change to the query text and result list box.

taooceros:
> now that we switched to svg, should we archive mainsearch.png into the 'doc' folder only for future reference instead?

Does svg solve your icon blurry issue? I am not sure because in my computer, the icon looks not that bad. @jjw24 

BTW, seems that the svg control makes an issue displaying in the theme tab in setting window.
image

taooceros:
> > now that we switched to svg, should we archive mainsearch.png into the 'doc' folder only for future reference instead?
> 
> Does svg solve your icon blurry issue? I am not sure because in my computer, the icon looks not that bad. @jjw24
> 
> BTW, seems that the svg control makes an issue displaying in the theme tab in setting window.

solve it....I originally think it is a wrapped mainwindow, but actually they are separately🤣

Flow.Launcher/MainWindow.xaml
jjw24: ok no problems thanks.

Flow.Launcher/MainWindow.xaml
jjw24: hmm i think if were are to implement something like change icon based on result selection or plugin installed indication in other issues, this might be good to remain async and two way, what do you think?

jjw24: > > now that we switched to svg, should we archive mainsearch.png into the 'doc' folder only for future reference instead?
> 
> If we do deside to switch to svg, we can archive the png one, since it is only used here.

haha well, i think we should, image scaling is better from svg.

Flow.Launcher/MainWindow.xaml
jjw24: could we remove this one please

jjw24 (changes requested):
two more changes please per comment before going ahead

Flow.Launcher/MainWindow.xaml
taooceros: sure

Flow.Launcher/MainWindow.xaml
taooceros: > hmm i think if were are to implement something like change icon based on result selection or plugin installed indication from other issues, this might be good to remain async and two way, what do you think?

hmmmm seems you are right, I will change it.

taooceros: Actually, I don't believe we need async. Async may cause a small period of nothing, like the image flickering. Even if we changes the icon, it should be fast because no need to extract thumbnail from files.

jjw24: ok cool no probs

jjw24:
lets keep mainsearch.png around for now for reference

Flow.Launcher/MainWindow.xaml
```diff
@@ -81,7 +82,7 @@
                      AllowDrop="True"
                      Visibility="Visible"
                     Background="Transparent"
-                    Margin="18,0,56,0">
+                    Margin="18,0,56,0" HorizontalAlignment="Left" Width="660">
```
jjw24, line 85: do we need the width to specify width as 660?

jjw24 (approved):
great change, thank you!

taooceros:
Thank you for reviewing!

PR #291, opened 2021-01-14
Add portable release

jjw24:
Add portable release


jjw24:
due to squirrel, update, restart and disable portable will not work

PR #295, opened 2021-01-15
Stop ProgressBar animation when Flow is hidden

taooceros:
#128 

https://github.com/Wox-launcher/Wox/pull/3246
Co-Authored-By: imsh <9258159+imsh@users.noreply.github.com>

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -84,8 +84,15 @@ namespace Flow.Launcher
                             QueryTextBox.SelectAll();
                             _viewModel.LastQuerySelected = true;
                         }
+
+                        ProgressBar.BeginStoryboard(_progressBarStoryboard);
+                    }
+                    else
+                    {
+                        _progressBarStoryboard.Stop();
                     }
                 }
+                
             };
             _settings.PropertyChanged += (o, e) =>
             {
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -1,5 +1,4 @@
-﻿
-using System;
+﻿using System;
 using System.ComponentModel;
 using System.Windows;
 using System.Windows.Input;
@@ -27,6 +26,7 @@ namespace Flow.Launcher
         #region Private Fields
 
         private readonly Storyboard _progressBarStoryboard = new Storyboard();
+        private bool isProgressBarStoryboardPaused;
         private Settings _settings;
         private NotifyIcon _notifyIcon;
         private MainViewModel _viewModel;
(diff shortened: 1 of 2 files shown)


jjw24:
what happens when the query is still loading though? like if its a long running query, i click on some other window and come back a minute later, this change means the bar will stop if it's hidden despite the query is still running right?

taooceros:
hmmm you are right. Let me use the dispatcher to disable and restart animation only when Progressbar visibility change instead of Mainwindow visibility.

jjw24:
i dont think this is working as intended. Visibility doesnt seem to be wired up properly, because regardless whether the mainwindow is hidden or visible, the visibility is always visible.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -84,8 +84,28 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
                     }
+                    else
+                    {
+                    }
+                }
+                else if (e.PropertyName == nameof(MainViewModel.ProgressBarVisibility))
+                {
+                    Dispatcher.Invoke(() =>
+                    {
+                        if (ProgressBar.Visibility == Visibility.Hidden)
```
jjw24, line 97: shouldnt this be MainViewModel.MainWindowVisibility? i thought we are resuming and pausing progress bar animation based on the visibility of mainwindow

taooceros: > what happens when the query is still loading though? like if its a long running query, i click on some other window and come back a minute later, this change means the bar will stop if it's hidden despite the query is still running right?

That's the reason changing this to progressbar visibility

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -84,8 +84,28 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
+                    {
+                        if (_viewModel.ProgressBarVisibility == Visibility.Hidden)
+                        {
+                            _progressBarStoryboard.Pause();
+                        }
+                        else
+                        {
+                            _progressBarStoryboard.Resume();
+                        }
+                    }, System.Windows.Threading.DispatcherPriority.Render);
```
jjw24, line 105: i think the logic should be this instead:
```
Dispatcher.Invoke(() =>
                    {
                        if (_viewModel.ProgressBarVisibility == Visibility.Visible)
                        {
                            if (Visibility == Visibility.Collapsed || Visibility == Visibility.Hidden)
                                _progressBarStoryboard.Pause();

                            if (Visibility == Visibility.Visible)
                                _progressBarStoryboard.Resume();
                        }
                    }, System.Windows.Threading.DispatcherPriority.Render);
```

When the progressbarvisibilty is visible (which means it is running) and main search window is hidden, then pause, otherwise when main search window is visible, resume. 'Visible is the property for main window's visibility'

Can you check?

taooceros: I agree with you. However, the implementation may need a few change, by making the Visibility changed event to the event called by MainWindow.Visibility. I will do that soon, and you can take a look.

taooceros:
Done, please review.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -84,8 +85,33 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
                         }
+
+                        if (_viewModel.ProgressBarVisibility == Visibility.Visible)
+                        {
+                            _progressBarStoryboard.Resume();
+                        }
                     }
+                    else
+                    {
+                        _progressBarStoryboard.Pause();
```
jjw24, line 96: This is going to be called everytime the mainwindow is hidden, how come a check is not added here?

taooceros:
Not sure whether this can save #128. Let's do more testing before close that via this pr.

Flow.Launcher/MainWindow.xaml.cs
jjw24: fixed in https://github.com/Flow-Launcher/Flow.Launcher/pull/295/commits/db63833e93c0b89290fa30a79533b82e28051dfc

taooceros:
> Although not from this change, this code does not set the progress bar when the main window is hidden. To repro, when main window is visible click anywhere else on the screen, the main window will be hidden, but the Visibility property is still visible rather than hidden.
> 
> Wonder if this is an easy fix we can do?
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/d0af7a1b752f8c4b740dccc6c263e8c8a3238d71/Flow.Launcher/MainWindow.xaml.cs#L74-L77

We can use MainViewModel.MainWindowVisibility instead.

jjw24:
> Not sure whether this can fix #128. Let's do more testing before close that via this pr.

oh ok gotcha

jjw24 (approved):

jjw24:
👍 👍

Issue #209, opened 2020-11-15
Program plugin doesn't work


kubalav:
Program plugin doesn't show result in latest version. I tried clean installation.

taooceros:
Could You please upload the log so that it will be easier to debug?

kubalav:
I tried reindex.
[2020-11-16.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/5544499/2020-11-16.txt)

taooceros:
```
2020-11-16 07:38:48.5187|ERROR||------------- BEGIN Flow.Launcher.Plugin.Program exception -------------
2020-11-16 07:38:48.5187|ERROR||
Exception full name: System.Runtime.InteropServices.COMException
Error status: KNOWN
Class name: Win32
Calling method: LnkProgram
Program path: C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Accessories\Windows Media Player.lnk
InnerException number: 1
Exception message: Error HRESULT E_FAIL has been returned from a call to a COM component.
Exception error type: HResult -2147467259
Exception thrown in called method: Void GetDescription(System.Text.StringBuilder, Int32)
Possible interpretation of the error: Error caused likely due to trying to get the description of the program
Possible resolution: Can be ignored and Flow Launcher should still continue, however the program may not be loaded
2020-11-16 07:38:48.5187|ERROR||------------- END Flow.Launcher.Plugin.Program exception -------------
```

Let me take a look, but I think if there is only two program that is unable to load, other programs should be loaded successfully.

kubalav:
It doesn't work at all.

taooceros:
Would you try open Flow with Administer Authorization? One of the log suggests Access denied.
I have just checked the win32 indexing and it works for me.
By the way, is any UWP app can be loaded?

kubalav:
I runs Flow with Admin access. I see UWP app only, not win32.

taooceros:
Let's try clicking this button and see what we get.
![image](https://user-images.githubusercontent.com/45326534/99234062-048f3480-282f-11eb-8779-3f635b799ef4.png)

kubalav:
UWP only from C: drive
![flow_progra_plugin_all_program_list](https://user-images.githubusercontent.com/37414585/99235020-99753c00-27f5-11eb-8a3f-02bbbb804bb1.png)

jjw24:
[...]

PR #300, opened 2021-01-18
fix error message not working + increase delay

jjw24:
fix pm initialisation error message not working due to main window not initialised yet
image

ran into the init error message several times during debug sessions, so increase delay

PR #301, opened 2021-01-19
Add Cancellation token for file system enumeration

### LINKED ISSUE #299: Use cancellation with Explorer's DirectorySearch
The all files and folders search with Explorer's DirectorySearch seems to run quite long, can we add the use of cancellation token to stop when next query comes in.

taooceros:
fix #299 
Also using EnumerationOption to avoid one access denied location breaking the whole searching similar to #209 .

Outcome:
- Improved Explorer's handling of long running results 
- Enhanced result list update speed


jjw24:
Ok so found the bottleneck that causes query to run very long for explorer:

image

when you have a lot of results like below, and it's trying to add new results (called from updateresultview) in after using except to exclude the same results currently display, this operation is very costly:
image

taooceros:
That seems what #195 is trying to do. You can take a look on the same method in that pr.

jjw24:
Ok cool, yeah I haven't looked at the code yet, just being taking it for test drives, will do so very soon

taooceros:
I think enumerate all file including subdirectory for about 200000 results may take about 5s and 200MB memory, which is still costly. This should still be added with an cancellation token.

taooceros:
> Ok so found the bottleneck that causes query to run very long from explorer:
> 
> image
> 
> when you have a lot of results like below, and it's trying to add new results (called from updateresultview) in after using except to exclude the same results currently display, this operation is very costly:
> image
> 
> if you let it just display the new results, blazing fast

Also, due to the lock strategy, by doing huge calculation within NewResutls method will hold the lock, which blocks ongoing results view update.

jjw24:
agree and agree, wonder if there is a better way to do this. also maybe if we want to do it in separate pr so cancellation is not blocked by this other fix

taooceros:
> agree and agree, wonder if there is a better way to do this. also maybe if we want to do it in separate pr so cancellation is not blocked by this other fix

That's already exist in #195, so no worry!

taooceros:
I also fix some code that is used for testing in #195, which isn't a bit deal.

taooceros:
After merging the dev, the recursive search for 180000 result takes about 5s to finish, which is significantly faster than before.

taooceros:
From testing, seems that the large amount of result won't be collected by GC even after a few other queries. By changing the List capacity, it should show a memory lower down but seems no effect....
Not sure whether my testing is wrong.

taooceros:
Unintentionally add a test code in PublicAPIInstance due to #300, should remove it before merging.

jjw24:
> Unintentionally add a test code in PublicAPIInstance due to #300, should remove it before merging.

this one?
```
  Application.Current.Dispatcher.Invoke(() =>
            {
                useMainWindowAsOwner = false; <=============
                var msg = useMainWindowAsOwner ? new Msg { Owner = Application.Current.MainWindow } : new Msg();
                msg.Show(title, subTitle, iconPath);
            });
```

taooceros:
> > Unintentionally add a test code in PublicAPIInstance due to #300, should remove it before merging.
> 
> this one?
> 
> ```
>   Application.Current.Dispatcher.Invoke(() =>
>             {
>                 useMainWindowAsOwner = false; <=============
>                 var msg = useMainWindowAsOwner ? new Msg { Owner = Application.Current.MainWindow } : new Msg();
>                 msg.Show(title, subTitle, iconPath);
>             });
> ```

Yes

jjw24:
getting this error ,

repro:
set explorer hotkey to 'e', type 'e a'. first couple of characters gets this error 

image

jjw24:
open flow's settings get this:
image

Plugins/Flow.Launcher.Plugin.Explorer/Search/DirectoryInfo/DirectoryInfoSearch.cs
```diff
@@ -58,38 +64,38 @@ private List<Result> DirectorySearch(SearchOption searchOption, Query query, str
...
+                        folderList.Add(resultManager.CreateFolderResult(fileSystemInfo.Name, fileSystemInfo.FullName,
+                            fileSystemInfo.FullName, query, true, false));
                     }
                     else
                     {
                         fileList.Add(resultManager.CreateFileResult(fileSystemInfo.FullName, query, true, false));
                     }
+
+                    if (token.IsCancellationRequested)
+                        return null;
```
jjw24, line 80: in your previous changes i have seen you use ThrowIfCancellationRequested() and catch it with an OperationCancelledException. Is one way better than the other? Should we use IsCancellationRequested way to handle it more elegantly?

taooceros: Agree with you. I will change that to throw, which can reduce some code for checking null.

taooceros:
> open flow's settings get this:
> image

Forget to add null check😅, will do now.

taooceros:
> getting this error ,
> repro: set explorer hotkey to 'e', type 'e a'. first couple of characters gets this error

I cannot reproduce the error....Could you please provide the stacktrace for that error?

jjw24:
> > getting this error ,
> > repro: set explorer hotkey to 'e', type 'e a'. first couple of characters gets this error
> 
> I cannot reproduce the error....Could you please provide the stacktrace for that error?

This maybe related to the null check, will check again later

taooceros:
@jjw24 Should we try to fix #308 here or use a new pr to fix it?

jjw24:
Separate pr please

Plugins/Flow.Launcher.Plugin.Explorer/Search/DirectoryInfo/DirectoryInfoSearch.cs
jjw24: Well I thought iscancellationrequested is nicer and instead of returning null you can just return empty results? Rather than having an extra catch.

Is there a reason for returning null rather than empty result list?

taooceros: Just a few difference, not a lot.
We will just throw the error, which will be catched by PluginManager.QueryForPluginAsync, so no worry for doing null check.

taooceros:
> Separate pr please

Sure

Plugins/Flow.Launcher.Plugin.Explorer/Search/DirectoryInfo/DirectoryInfoSearch.cs
jjw24: Yeah but cancellation of previous query is a mechanism we require throughout, so probably not great to be treating it like the program errored out. Maybe change this next time in separate pr, for now just leave it as cancellationrequested

taooceros: Well actually, throw an taskcancelledexcpetion is what Microsoft suggested in cooperative cancellation. This is similar to how we shall handle operationcancelledexception, which is the base of taskcancelledexcpetion.

However, it seems that in Debug mode, throwing a lot of exception may cause stunt……which is not a bit due, but somehow not promising.

jjw24:
> getting this error ,
> 
> repro:
> set explorer hotkey to 'e', type 'e a'. first couple of characters gets this error
> 
> image

yeah ok still getting this error. Set search activation hotkey to 'e' and just type 'e' i got it.

jjw24:
sorted in https://github.com/Flow-Launcher/Flow.Launcher/pull/301/commits/ff5e3695e9e538ceafc997b96adf2a4f7cb0a4f2

taooceros:
> sorted in https://github.com/Flow-Launcher/Flow.Launcher/pull/301/commits/ff5e3695e9e538ceafc997b96adf2a4f7cb0a4f2

Nice! I understand that why I am unable to reproduce, because I have a few quick access.

jjw24:
yeah haha, i should have said remove roaming profile

taooceros:
> yeah haha, i should have said remove roaming profile

Hahaha 🤣

taooceros:
https://github.com/Flow-Launcher/Flow.Launcher/pull/301/commits/9914124d200ef5988c0488f0f65a9e4bc3082d7e remove the extra checking to make the code clear. We don't need to check whether the quick folder list is empty and don't need to check the action keyword as what we have talked about in Window Walker plugin.

taooceros:
@jjw24 Any further change required?

Flow.Launcher/ViewModel/ResultsViewModel.cs
```diff
@@ -139,39 +139,9 @@ public void KeepResultsExcept(PluginMetadata metadata)
...
-            }
-            else
-            {
-                Margin = new Thickness { Top = 0 };
-                Visbility = Visibility.Collapsed;
-            }
+            UpdateResults(newResults);
         }
         /// <summary>
         /// To avoid deadlock, this method should not called from main thread
```
jjw24, line 147: @taooceros do you know if this comment is still relevant?

taooceros: I think yes, because only results from main result set will be added via resultUpdateQueue, the one for context menu and history is still added directly.
However, I am not sure the deadlock meaning here. Maybe just avoiding potential stunt of main ui due to lock machanism.

jjw24: cool, thank you

jjw24 (approved):

PR #309, opened 2021-01-22
Fix unrecognized path

### LINKED ISSUE #308: Unable to recognize first level directory with search wildcard ">" or "*"
When I search with "D:\>flow", the generated query is `SELECT TOP 100 "System.FileName", "System.ItemUrl", "System.ItemType", "System.ItemPathDisplay" FROM "SystemIndex" WHERE (Scope = 'file:D:/' OR CONTAINS(System.FileName,'"D:\"',2052)) AND System.FileName > 'flow' AND scope='file:'`, which doesn't give me the correct result I want.

I don't understand why we use `System.FileName > flow` as condition instead of CONTAINS(System.FileName, "flow"), and the extra `scope='file:'" is redundent.

Also, we should also make the folder result appears earlier similar to the search without index.

taooceros:
fix #308 
The reason is `IsLocationPathString(string)` unable to recongnize `D:\>*` as location string, so that it treat it as a file search.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
@@ -58,7 +58,7 @@ namespace Flow.Launcher.Plugin.Explorer.Search
             // Query is a location path with a full environment variable, eg. %appdata%\somefolder\
             var isEnvironmentVariablePath = querySearch[1..].Contains("%\\");
 
-            if (!FilesFolders.IsLocationPathString(querySearch) && !isEnvironmentVariablePath)
+            if (!querySearch.IsLocationPathString() && !isEnvironmentVariablePath)
             {
                 results.AddRange(await WindowsIndexFilesAndFoldersSearchAsync(query, querySearch, token).ConfigureAwait(false));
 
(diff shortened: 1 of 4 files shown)


taooceros:
Actually, I think we shall not use this method in searchmanager to check whether it is like a path, but better to trim the end part after the last \\ and use directory.exist

taooceros:
image
@jjw24 Would you please tell me why this where restriction is required?

jjw24:
> Actually, I think we shall not use this method in searchmanager to check whether it is like a path, but better to trim the end part after the last \ and use directory.exist

IsLocationPathString is used in scenario where you just do initial string validation and dont need to have the overhead of checking if the directory actually exists.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -70,7 +70,7 @@ internal async Task<List<Result>> SearchAsync(Query query, CancellationToken tok
             if (isEnvironmentVariablePath)
                 locationPath = EnvironmentVariables.TranslateEnvironmentVariablePath(locationPath);
 
-            if (!FilesFolders.LocationExists(FilesFolders.ReturnPreviousDirectoryIfIncompleteString(locationPath)))
+            if (!FilesFolders.ReturnPreviousDirectoryIfIncompleteString(locationPath).IsLocationPathString())
```
jjw24, line 73: Reverting this change because if we dont check whether location exists, it will error when we do directory search:
image

jjw24 (approved):

jjw24 (approved):

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
taooceros: Forget this situation. Thank you!

PR #314, opened 2021-01-25
Optimize manifest downloading behavior

### LINKED ISSUE #313: Plugins Manager check connection message pops up despite manifest loaded
@taooceros getting this message occasionally despite plugins manager's manifest is loaded, any idea why?

https://github.com/Flow-Launcher/Flow.Launcher/blob/5dd12d212b57d2c752c09e8539752d1138be326b/Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs#L41-L52

taooceros:
This branch is intended to fix #313, but haven't done that job.

1. Don't await download manifest when it is unavailable, but only await when calling install method to allow user using uninstall when manifest is not ready.


jjw24:
@taooceros consistently hitting this in vs debug:
image

jjw24:
this does not happen with dev branch

taooceros:
Oh, it was canceled? I will take a look on why this is canceled because it is not supposed to be canceled.

taooceros:
> this does not happen with dev branch

So you mean that this PR fix it?

jjw24:
No as in the dev branch doesn't run into this cancellation

taooceros:
> No as in the dev branch doesn't run into this cancellation

I am not sure about why there exists the cancellation, because the Get method should not include a cancellation.

taooceros:
@jjw24 I still don't understand why there will be cancellation...... Is this replicable? If it is, would you please check whether the manifest has been downloaded successfully? If it has downloaded successfully but canceled, let me make the error message only appears on fault?

jjw24:
I can replicate, will take a look, but I won't be able to do it until later tonight or tomorrow. In the mean time let's roll out the release because I have tested it for a while and haven't run into any issues so pretty comfortable with it. This change can go out with the next release.

What do you think?

taooceros:
> I can replicate, will take a look, but I won't be able to do it until later tonight or tomorrow. In the mean time let's roll out the release because I have tested it for a while and haven't run into any issues so pretty comfortable with it. This change can go out with the next release.
> 
> What do you think?

Yeah, sure. I also quite excited on the new 1.7.0 release, as it brought out the new plugin model.

taooceros:
Or maybe you can just merge the optimization, and let's solve the issue next time.

jjw24:
> Or maybe you can just merge the optimization, and let's solve the issue next time.

I would definitely not be comfortable with this issue if I can replicate it with this PR and not with the dev branch build haha

taooceros:
> > Or maybe you can just merge the optimization, and let's solve the issue next time.
> 
> I would definitely not be comfortable with merging with this issue outstanding if I can replicate it with this PR and not with the dev branch build haha

Haha great. Let's keep it with next release!

jjw24:
I have not experienced issue 313 anymore, do we still want to make this change?

taooceros:
> I have not experienced issue 313 any more, do we still want to make this change? 

Ah yes, I have made a couple changes to make the singleton works better. You can take a look on that.

jjw24:
image

Should we add RanToCompletion  as a pass condition as well?

taooceros:
> image
> 
> Should we add RanToCompletion as a pass condition as well?

No, because this is used to keep one task running, so that we won't have multiple tasks to update the manifest. If the task is ended, we need to start a new one.

jjw24:
oh right i see

jjw24 (approved):

PR #318, opened 2021-01-29
Make most class in Explorer plugin become static class

taooceros:
Make most helper class in Explorer become static class.

I am not sure why by design, they are instance class, and there are multiple duplicate instances. This pr makes them become static class because most of them don't save any information (except for ResultManager, which contains the PluginInitContext to change query), but just be used as a group of methods.


jjw24:
well for now it doesnt keep state, but later on it may such as allowing users to select how many results to display from index search.

What do you mean by there are multiple duplicate instances?

taooceros:
It seems that most class in explorer plugin contains an instance for `ResultManager`. Sometimes, the method call even use new `ResultManager().createFileResult()`....

taooceros:
Even if we may need to check how much results will be listed, most of the class still doesn't have to be instance class. The only class that may store the state maybe ResultManager (store a pointer to pluginAPI), and the queryconstructor (to store how much result to display).

jjw24 (approved):

jjw24 (approved):

PR #319, opened 2021-01-31
Automatic Python installation for Python plugins

jjw24:
If any Python plugins exist upon start up and the PythonDirectory in Settings is empty:

1. Check env "PATH" if Python path exists and set automatically, else
2. Prompt the user if they would like to install or select a Python directory,
3. If chosen to install, use Droplex library to download and install Python 3.9, then set the required variables for the plugin,
4. If chosen to  select the directory, check location exists then set the required variables.
5. Load Python plugins


Zeroto521:
I want to know which python version will be installed. Complete version contains interpreter, idle, some script(includes pip and others). 
So the installed python with interpreter and scripts will be fine

jjw24:
Python 3.9 is installed

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -129,47 +133,104 @@ public static IEnumerable<PluginPair> PythonPlugins(List<PluginMetadata> source,
...
+                            {
+                                MessageBox.Show("Can't find python in given directory");
+                            }
+                        }
+                    }
+                }
+                else
+                {
+                    var installation = Task.Run(async () => { await DroplexPackage.Drop(App.python3_9_1).ConfigureAwait(false); });
+                    installation.Wait();
```
taooceros, line 197: emmm since we are waiting it, why do we need to wrap it within the Task.Run?

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -129,47 +133,104 @@ public static IEnumerable<PluginPair> PythonPlugins(List<PluginMetadata> source,
...
+
+                    var installedPythonDirectory =
+                        Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), @"Programs\Python\Python39");
```
taooceros, line 200: What if user don't set the environment path when installing python?
BTW, isn't python installation require restart computer in order to setup the environmental variable?

pc223:
Just quick test this, when python 3.9 missing, Flow will auto install python 3.9 :thumbsup:

Though my plugin is not usable right now, got 2 problems, related, but maybe outside scope of this PR:
1. When python packages missing (like `flowlauncher`, `python-dotenv`,...), Flow doesn't make a warning, should be at least an alert msgbox.
2. An exception related to RPC/Json stuff

```
2021-02-02 22:35:16.5874|ERROR|JsonRPCPlugin.Query|-------------------------- Begin exception --------------------------
2021-02-02 22:35:16.6100|ERROR|JsonRPCPlugin.Query|Exception when query <ts>
2021-02-02 22:35:16.6272|ERROR|JsonRPCPlugin.Query|Exception full name:
 <System.InvalidOperationException>
2021-02-02 22:35:16.6429|ERROR|JsonRPCPlugin.Query|Exception message:
 <The JSON property name for 'Flow.Launcher.Core.Plugin.JsonRPCQueryResponseModel.Result' collides with another property.>
2021-02-02 22:35:16.6625|ERROR|JsonRPCPlugin.Query|Exception stack trace:
 <   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_SerializerPropertyNameConflict(JsonClassInfo jsonClassInfo, JsonPropertyInfo jsonPropertyInfo)
   at System.Text.Json.JsonClassInfo..ctor(Type type, JsonSerializerOptions options)
   at System.Text.Json.JsonSerializerOptions.GetOrAddClass(Type classType)
   at System.Text.Json.JsonSerializer.ReadCore(Type returnType, JsonSerializerOptions options, Utf8JsonReader& reader)
   at System.Text.Json.JsonSerializer.Deserialize(String json, Type returnType, JsonSerializerOptions options)
   at System.Text.Json.JsonSerializer.Deserialize[TValue](String json, JsonSerializerOptions options)
   at Flow.Launcher.Core.Plugin.JsonRPCPlugin.DeserializedResult(String output) in C:\projects\flow-launcher\Flow.Launcher.Core\Plugin\JsonRPCPlugin.cs:line 66
   at Flow.Launcher.Core.Plugin.JsonRPCPlugin.Query(Query query) in C:\projects\flow-launcher\Flow.Launcher.Core\Plugin\JsonRPCPlugin.cs:line 39>
2021-02-02 22:35:16.6863|ERROR|JsonRPCPlugin.Query|Exception source:
 <System.Text.Json>
2021-02-02 22:35:16.7061|ERROR|JsonRPCPlugin.Query|Exception target site:
 <Void ThrowInvalidOperationException_SerializerPropertyNameConflict(System.Text.Json.JsonClassInfo, System.Text.Json.JsonPropertyInfo)>
2021-02-02 22:35:16.7229|ERROR|JsonRPCPlugin.Query|Exception HResult:
 <-2146233079>
2021-02-02 22:35:16.7402|ERROR|JsonRPCPlugin.Query|-------------------------- End exception --------------------------
```

pc223:
Right now, Flow is using Python 3.9 directly, without virtualenv, this can be pretty annoying for python 3.9 user that use it for anything else other than Flow. 

So, after Flow install Python 3.9, I think Flow should make a new virtualenv,  `python -m venv flow_python_env`, that means Flow can freely install/uninstall any packages it want, no fear of conflict,

and change the python path to something like this in the settings page: `flow_python_env\scripts\python.exe` instead of the current one.

jjw24:
lets tackle this seperately. This PR is to get python installed. We need to figure out how to auto install/uninstall requirement.txt cleanly

Flow.Launcher.Core/Plugin/PluginsLoader.cs
jjw24: @taooceros this should be sychronous, is there another way to do this without converting the calling methods async?

taooceros: You can simply Wait the task do be done. `DroplexPackage.Drop(App.python3_9_1).Wait(timeout)`, while timeout is the time to be waited. However, actually, I don't think it would be great to let user wait downloading python at initialization, since this will block other Flow's functionality.
I think this should be done async, disabling python plugin first if python not installed (also ask whether user has installed python similar to what has done in Everything plugin), and once it has downloaded successfully, enabling them.

jjw24: I think it's fine because the installation has a prompt from Python, so user knows it's installing.

It does also ask the user if they want to select the location beforehand as well

taooceros: oh ok I haven't looked at the newer part of the code.
However, I still think we should wait python installation at initialization, but let it does the job on background.
Is it possible to disable all python plugin and then later enable them after python installed?

jjw24: yeah enabling and disabling is what i dont want to mess with at this stage. If we get any issues from users that it's taking forever to download then we can look to do that, the install is not big and should download and install quickly.

Flow.Launcher.Core/Plugin/PluginsLoader.cs
jjw24: this is the env path for local app data folder, should always be set right

Flow.Launcher.Core/Plugin/PluginsLoader.cs
taooceros: I don't think python includes a great cdn so users in China may encounter significant of time of downloading it. However, I do agree with you that disabling and enabling plugins will bring quite a mess with at this stage.
Maybe don't load python plugin, and force restarting Flow after Python has installed?

Flow.Launcher.Core/Plugin/PluginsLoader.cs
taooceros: Oh see it.

taooceros:
@pc223 the rpc error has been fixed in #324 
You can take a try on the build version of dev branch.
https://ci.appveyor.com/api/buildjobs/k827q6qyfge1a9l0/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.0.exe

pc223:
> @pc223 the rpc error has been fixed in #324
> You can take a try on the build version of dev branch.
> https://ci.appveyor.com/api/buildjobs/k827q6qyfge1a9l0/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.0.exe

Nice! I'll try it out when I got the time 😄

Flow.Launcher.Core/Plugin/PluginsLoader.cs
jjw24: python install file is only 28mbs, can you test it out see if it takes long

taooceros: > python install file is only 28mbs, can you test it out see if it takes long

Will take a try.

taooceros: hmmm seems that people in China cannot directly finish download python because of timeout. I have encountered that the task was cancelled when I try to do this.
image

jjw24: Thank you for testing. How long before it's timeout? 

Is it timing out on connection or during the downloading?

taooceros:
It should be downloading, but I am not sure. I think is is because the downloading speed is really slow, so timeout.




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Jeremy Wu"<notifications@github.com&gt;; 
发送时间: 2021年2月7日(星期天) 晚上7:40
收件人: "Flow-Launcher/Flow.Launcher"<Flow.Launcher@noreply.github.com&gt;; 
抄送: "天煞"<1569418337@qq.com&gt;; "Mention"<mention@noreply.github.com&gt;; 
主题: Re: [Flow-Launcher/Flow.Launcher] Automatic Python installation for Python plugins (#319)





 
@jjw24 commented on this pull request.
 
 
In Flow.Launcher.Core/Plugin/PluginsLoader.cs:
 &gt; +                            if (File.Exists(pythonPath)) +                            { +                                settings.PythonDirectory = pythonDirectory; +                                Constant.PythonPath = pythonPath; +                            } +                            else +                            { +                                MessageBox.Show("Can't find python in given directory"); +                            } +                        } +                    } +                } +                else +                { +                    var installation = Task.Run(async () =&gt; { await DroplexPackage.Drop(App.python3_9_1).ConfigureAwait(false); }); +                    installation.Wait();  
Thank you for testing. How long before it's timeout?
 
Is it timing out on connection or during the downloading?

Zeroto521:
where is the installed path?

Zeroto521:
> Python 3.9 will be installed

This version will install the idle (GUI). The general user doesn't need that thing.

image

And the main version is 3.8 could be better because 3.9 released recently.

I recommend installing [miniconda](https://docs.conda.io/en/latest/miniconda.html)

Miniconda only contains interpreter and scripts.

taooceros:
> > Python 3.9 will be installed
> 
> This version will install the idle (GUI). The general user doesn't need that thing.
> 
> image
> 
> And the main version is 3.8 could be better because 3.9 released recently.
> 
> I recommend installing [miniconda](https://docs.conda.io/en/latest/miniconda.html)
> 
> Miniconda only contains interpreter and scripts.

But isn't miniconda include too much package that generally is not useful for Flow plugin?

Zeroto521:
Official python = idle + script(pip) + interpreter 
Miniconda = script(pip, conda) + interpreter 

If you thought miniconda didn’t have many packages another choice could be anaconda. Anaconda size is x.0GB+. But I’ m not prefer that.

Zeroto521:
Anyway package is the problem of next step not now.

jjw24:
This [build](https://ci.appveyor.com/api/buildjobs/lxoe5ly1lt9u8w05/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.0.exe) should resolve download timeout issues. Thanks @taooceros

jjw24:
This is for 1.8.0 release, do not merge if approved.

taooceros:
Is it any thing we still need to do for 1.7.1?

jjw24:
Hmm not at this stage. Although we don't have anything pending to go out urgently in 1.8.0 as well right, figure we might catch more bugs for 1.7.1

taooceros:
> Hmm not at this stage. Although we don't have anything pending to go out urgently in 1.8.0 as well right, figure we might catch more bugs for 1.7.1

I think we should at least release the fix in #322 since it blocks python plugins from working.

jjw24:
> > Hmm not at this stage. Although we don't have anything pending to go out urgently in 1.8.0 as well right, figure we might catch more bugs for 1.7.1
> 
> I think we should at least release the fix in #322 since it blocks python plugins from working.

Yeah makes sense, let's release 1.7.1, let's still hold off on 1.8.0 PRs for a bit to see if anymore bugs come up. What do you think?

taooceros:
Yeah agree with you. Actually, I think we can release a pre-release before any large change, and see whether some bugs can be catched by that so that we can fix it before official release.

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -129,47 +133,103 @@ public static IEnumerable<PluginPair> PythonPlugins(List<PluginMetadata> source,
...
+                    var pythonPath = Path.Combine(installedPythonDirectory, PythonExecutable);
+                    if (FilesFolders.FileExists(pythonPath))
+                    {
+                        settings.PythonDirectory = installedPythonDirectory;
+                        Constant.PythonPath = pythonPath;
+                    }
+                    else
+                    {
+                        Log.Error("PluginsLoader",
```
taooceros, line 208: What if user choose to install python on other path instead of the default path? I think we should allow selection if that it not found.

jjw24:
> Yeah agree with you. Actually, I think we can release a pre-release before any large change, and see whether some bugs can be catched by that so that we can fix it before official release.

Well, I don't know if people download that though, right now pre-releases needed to be manually downloaded

taooceros:
I think wox allows user to choose downloading prerelease. Isn't flow containing one?

jjw24:
> I think wox allows user to choose downloading prerelease. Isn't flow containing one? 

Nah we haven't built it in. I would prefer something like squirrel's diff release where update is on new changes rather than the while thing. Problem with pre-releases is that it will just sit there forever and too much reliance on it to catch bugs is not effective, nobody wants to use a buggy version.

taooceros:
emm you are right, but we still need more testing before large release, because sometimes it is hard for us to find bugs during our personal usage. Maybe we can ask someone who are willing to try the new feature to take a try on the prerelease, and after it becomes stable, then we release the official release.
I think the squirrel diff should be addressed later, and it is a feature of squirrel, though I don't understand why it doesn't generate the diff file for now. It should be quite useful for Flow because flow is self-contained, using diff update can avoid duplicate downloading the same .net core runtime.

Flow.Launcher.Core/Plugin/PluginsLoader.cs
jjw24: this is to auto set path after auto install. Users can choose the path before hand

taooceros: Well I mean what if user don't choose to install python on the default path?

jjw24: Happy Chinese New Year @taooceros  :)

The installation will just install to the default location, it's a passive instead, no prompt for install location

taooceros: Happy New Year!!! 😃😃
Oh I see. That will be great!

jjw24:
@taooceros This is good to go out for 1.8.0?

taooceros:
> @taooceros This is good to go out for 1.8.0?

Ah yes!

taooceros (approved):

jjw24:
> > Python 3.9 will be installed
> 
> This version will install the idle (GUI). The general user doesn't need that thing.
> 
> image
> 
> And the main version is 3.8 could be better because 3.9 released recently.
> 
> I recommend installing [miniconda](https://docs.conda.io/en/latest/miniconda.html)
> 
> Miniconda only contains interpreter and scripts.

Good pick up, I will need to look into miniconda along with the other suggestions and will change it later if needed.

jjw24:
We can discuss this further in the discord plugin-dev channel. Once decided it's just a matter of changing the droplex package.

PR #320, opened 2021-02-01
Fix Http Proxy

harrynull:
This PR should make connections follow the proxy in the settings correctly.

### DIFF
--- a/Flow.Launcher.Infrastructure/Http/Http.cs
+++ b/Flow.Launcher.Infrastructure/Http/Http.cs
@@ -34,6 +34,7 @@ namespace Flow.Launcher.Infrastructure.Http
 
             client = new HttpClient(socketsHttpHandler, false);
             client.DefaultRequestHeaders.Add("User-Agent", UserAgent);
+            HttpClient.DefaultProxy = WebProxy;
         }
 
         private static HttpProxy proxy;
@@ -45,6 +46,7 @@ namespace Flow.Launcher.Infrastructure.Http
             {
                 proxy = value;
                 proxy.PropertyChanged += UpdateProxy;
+                UpdateProxy(ProxyProperty.Enabled);
             }
         }
 
--- a/Flow.Launcher.Infrastructure/Http/Http.cs
+++ b/Flow.Launcher.Infrastructure/Http/Http.cs
@@ -16,13 +16,7 @@ namespace Flow.Launcher.Infrastructure.Http
     {
         private const string UserAgent = @"Mozilla/5.0 (Trident/7.0; rv:11.0) like Gecko";
 
-        private static HttpClient client;
-
-        private static SocketsHttpHandler socketsHttpHandler = new SocketsHttpHandler()
-        {
-            UseProxy = true,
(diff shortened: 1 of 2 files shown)


Flow.Launcher.Infrastructure/Http/Http.cs
```diff
@@ -34,6 +34,7 @@ static Http()
 
             client = new HttpClient(socketsHttpHandler, false);
             client.DefaultRequestHeaders.Add("User-Agent", UserAgent);
+            HttpClient.DefaultProxy = WebProxy;
```
harrynull, line 37: Not sure why but the proxy setting in `socketsHttpHandler` does not seem to work for me. Maybe you could investigate further but using this `DefaultProxy` would make it work correctly.

I did not verify if it will work if there is no proxy set.

Flow.Launcher.Infrastructure/Http/Http.cs
```diff
@@ -45,6 +46,7 @@ public static HttpProxy Proxy
             {
                 proxy = value;
                 proxy.PropertyChanged += UpdateProxy;
+                UpdateProxy(ProxyProperty.Enabled);
```
harrynull, line 49: Manually trigger an update. Otherwise, `WebProxy` would be `null` until you make a change in the settings.

Flow.Launcher/App.xaml.cs
```diff
@@ -61,6 +61,8 @@ private async void OnStartupAsync(object sender, StartupEventArgs e)
                 _settingsVM = new SettingWindowViewModel(_updater, _portable);
                 _settings = _settingsVM.Settings;
 
+                Http.Proxy = _settings.Proxy;
```
harrynull, line 64: Set http proxy first so that when the plugin manager initializes it will follow the proxy.

Flow.Launcher.Infrastructure/Http/Http.cs
taooceros: Hmmm that's weird, because when I test it with an unusable proxy, making changed to proxy setting do crash due to unavaliable proxy.
Thank you for the fix!
Since we will use DefaultProxy, would you please remove the socketsHttpHandler, which is useless, and make the client initialize with default parameter.

Flow.Launcher.Infrastructure/Http/Http.cs
taooceros: Great fix! Thank you!
Is it possible that the SocketHttpHandler doesn't work is because of this issue?

harrynull: I don't think so. It still does not work with all my other changes.

Flow.Launcher.Infrastructure/Http/Http.cs
harrynull: Done!

taooceros: Thank you!

taooceros (approved):
Great fix! Thank you!

PR #321, opened 2021-02-01
Add "" in FilesFolders.OpenPath to be able to open path with ','

taooceros:


jjw24:
what's an example of this issue?

taooceros:
image
I am unable to open this file with Explorer plugin. By adding the quotation mark, it will work. I believe this is due to the space in the path.

jjw24:
> image
> I am unable to open this file with Explorer plugin. By adding the quotation mark, it will work. I believe this is due to the space in the path.

i tested with space in the path, it still worked.

let me try replicate your path

taooceros:
That's quite weird... Maybe it may caused by the file name instead of path? Because seems that I will be able to open the container folder.

taooceros:
Or maybe the reason is the `,` within the file name?

taooceros:
Yeah test it, the issue is due to the `,` instead of space.

jjw24 (approved):

Issue #323, opened 2021-02-03
Colors plugin installation error (Plugin Manager)


kubalav:
I try to install Colors plugin via Plugin Manager. I use portable version Flow Launcher 1.7.0.
Log:
```
2021-02-03 09:02:58.9054|ERROR|PluginsManager.InstallOrUpdate|-------------------------- Begin exception --------------------------
2021-02-03 09:02:58.9054|ERROR|PluginsManager.InstallOrUpdate|An error occured while downloading plugin
2021-02-03 09:02:58.9054|ERROR|PluginsManager.InstallOrUpdate|Exception full name:
 <System.IO.IOException>
2021-02-03 09:02:58.9054|ERROR|PluginsManager.InstallOrUpdate|Exception message:
 <Source and destination path must have identical roots. Move will not work across volumes.>
2021-02-03 09:02:58.9173|ERROR|PluginsManager.InstallOrUpdate|Exception stack trace:
 <   at System.IO.Directory.Move(String sourceDirName, String destDirName)
   at Flow.Launcher.Plugin.PluginsManager.PluginsManager.Install(UserPlugin plugin, String downloadedFilePath) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.PluginsManager\PluginsManager.cs:line 349
   at Flow.Launcher.Plugin.PluginsManager.PluginsManager.InstallOrUpdate(UserPlugin plugin) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.PluginsManager\PluginsManager.cs:line 136>
2021-02-03 09:02:58.9173|ERROR|PluginsManager.InstallOrUpdate|Exception source:
 <System.IO.FileSystem>
2021-02-03 09:02:58.9173|ERROR|PluginsManager.InstallOrUpdate|Exception target site:
 <Void Move(System.String, System.String)>
2021-02-03 09:02:58.9173|ERROR|PluginsManager.InstallOrUpdate|Exception HResult:
 <-2146232800>
2021-02-03 09:02:58.9173|ERROR|PluginsManager.InstallOrUpdate|-------------------------- End exception --------------------------
```

taooceros:
Is portable version of v1.7.0 working? Because seems that the installed version has some issue generated in the build process, which cause v1.7.0 load assembly fail a lot.

@jjw24 Would you please take a look on this issue, since I am not quite sure how the installation process work?

kubalav:
[...]

PR #325, opened 2021-02-03
Fix move method call only restricted to within the same drive

jjw24:
Fix move method call only restricted to within the same drive

Closing #323


taooceros (approved):

jjw24:
let me get @kubalav to test this first before merging

PR #331, opened 2021-02-08
JSONRPC Async Model

taooceros:
Question:
1. why not use json serialize method in ToString method in JSONRPCModel.
2. whether we want an async version of action?
3. async version for context menu?

### DIFF
--- a/Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
+++ b/Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
@@ -131,26 +131,20 @@ namespace Flow.Launcher.Core.Plugin
             return await ExecuteAsync(start, token);
         }
 
-        protected async Task<string> ExecuteAsync(ProcessStartInfo startInfo, CancellationToken token = default)
+        protected string Execute(ProcessStartInfo startInfo)
         {
             try
             {
                 using var process = Process.Start(startInfo);
-                if (process == null)
-                {
-                    Log.Error("|JsonRPCPlugin.Execute|Can't start new process");
-                    return string.Empty;
-                }
-
+                if (process == null) return string.Empty;
+                
                 using var standardOutput = process.StandardOutput;
-                var result = await standardOutput.ReadToEndAsync();
-                if (token.IsCancellationRequested)
-                    return string.Empty;
+                var result = standardOutput.ReadToEnd();
 
                 if (string.IsNullOrEmpty(result))
                 {
                     using var standardError = process.StandardError;
-                    var error = await standardError.ReadToEndAsync();
+                    var error = standardError.ReadToEnd();
                     if (!string.IsNullOrEmpty(error))
                     {
                         Log.Error($"|JsonRPCPlugin.Execute|{error}");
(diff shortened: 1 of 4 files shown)


taooceros:
Got a question.

```c#
if (result.StartsWith("DEBUG:"))
{
    MessageBox.Show(new Form {TopMost = true}, result.Substring(6));
    return string.Empty;
}
```

Does anyone actually use this to debug? This prevents making it based on Stream, because Stream are impossible to peek.

jjw24:
Does this just show a message box when result starts with debug?

taooceros:
> Does this just show a message box when result starts with debug?

Yes

taooceros:
It will prevent from assuming the returning message is based on JsonRPCQueryResponseModel so that we can't deserialize it based on stream.

jjw24:
Yeah let's get rid of it. Surely there is a better implementation for debugging

jjw24:
Ready for review?

taooceros:
Ah yes!

jjw24:
The to dos in the PR description are no longer relevant right?

taooceros:
> The to dos in the PR description are no longer relevant right?

Yeah they are questions, not quite relate to the change in this branch.

jjw24:
Have we done any testing with this?

jjw24:
> > The to dos in the PR description are no longer relevant right?
> 
> Yeah they are questions, not quite relate to the change in this branch.

Hmmm i havent played enough with non-c# plugins to be able to answer those. General rule of thumb, if it improves Flow and you have tested it to make sure, then lets PR it. Leaving those questions here I will likely forget, unless you will remember :)

taooceros:
> Have we done any testing with this?

I check out that python plugin will work if python works. If It doesn't work, the error reporting part will be triggered (I think the EndofStream works).

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
```diff
@@ -27,23 +27,10 @@ internal abstract class JsonRPCPlugin : IPlugin, IContextMenu
         /// </summary>
         public abstract string SupportedLanguage { get; set; }
 
-        protected abstract string ExecuteQuery(Query query);
+        protected abstract Task<Stream> ExecuteQueryAsync(Query query, CancellationToken token);
```
jjw24, line 30: Is this change backwards compatible?

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
```diff
@@ -130,70 +135,123 @@ private void ExecuteFlowLauncherAPI(string method, object[] parameters)
...
-                Log.Exception($"|JsonRPCPlugin.Execute|Exception for filename <{startInfo.FileName}> with argument <{startInfo.Arguments}>", e);
-                return string.Empty;
+                Log.Exception(
+                    $"|JsonRPCPlugin.ExecuteAsync|Exception for filename <{startInfo.FileName}> with argument <{startInfo.Arguments}>",
+                    e);
+                return Stream.Null;
+            }
+        }
+
+        public async Task<List<Result>> QueryAsync(Query query, CancellationToken token)
```
jjw24, line 237: the change from Query to QueryAsync, will this be backwards compatible?

taooceros:
> Hmmm i havent played enough with non-c# plugins to be able to answer those. General rule of thumb, if it improves Flow and you have tested it to make sure, then lets PR it. Leaving those questions here I will likely forget, unless you will remember :)

Well actually these questions are also relate to c# plugins. What do you think about them?

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
taooceros: The JSONRPCPlugin is in Flow, and what it is doing is that overriding the Query (or QueryAsync) to call the executable (or python) and communicate via console stream. So change here won't affect the compatibility of plugins.

taooceros:
The change in this branch are majorly using the IAsyncPlugin to allow early stop (not completely), and read Stream instead of reading string to reduce the cost of creating string based on Standard output stream.

jjw24:
> > Hmmm i havent played enough with non-c# plugins to be able to answer those. General rule of thumb, if it improves Flow and you have tested it to make sure, then lets PR it. Leaving those questions here I will likely forget, unless you will remember :)
> 
> Well actually these questions are also relate to c# plugins. What do you think about them?

oh yeah you are right. I think why not, we can do it for QueryAsync implementation and test it, not that it's a necessity right now though, but implementing it may allow for future usage and expansion.

jjw24:
Let's create issues for the questions and  can discuss or implement it.

jjw24 (approved):

Issue #86, opened 2020-06-28
Flow Icon is pixelated


Kalaschnik:
![image](https://user-images.githubusercontent.com/9831678/85940681-16267d00-b91e-11ea-9894-22c6d03f295b.png)

I try to figure out why this is happening... maybe it is an issue with the ico containing multiple sizes? Maybe a 16×16 ico might fix this...

This is also happening in the taskbar, compared to its neighbours:
![image](https://user-images.githubusercontent.com/9831678/85940721-6b628e80-b91e-11ea-8b83-d2b837444a05.png)

The tray icon seems fine to me though...

jjw24:
That's interesting pick up. Can you put in a 16 X 16 ico in or and I will test it out.

I haven't notice the pixelation on my machine though.

jjw24:
Closing, please reopen if still an issue.

PR #335, opened 2021-02-12
Use app.ico as icon instead of app.png for setting window

taooceros:
It will help scaling, although seems that our icon is too complex to display well in 16*16 scale.

#86


Flow.Launcher/Flow.Launcher.csproj
```diff
@@ -94,6 +97,12 @@
     <ProjectReference Include="..\Flow.Launcher.Plugin\Flow.Launcher.Plugin.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="Images\app.ico">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+
```
jjw24, line 105: this is not needed right since you are including all .ico extension

jjw24:
Does this actually solve the problem in issue 86?

taooceros:
> Does this actually solve the problem in issue 86? As in will it actually improve the rendering?

The rendering won't be changed, but the ico file contains 16*16 icon so wpf don't need to downscaling the icon.
Well, if the 16*16 icon satisfying, it can be solved, but actually it isn't not that perfect.

Flow.Launcher/Flow.Launcher.csproj
taooceros: oh yes

taooceros:
image
This is the setting panel icon after the pr.

jjw24 (approved):

PR #336, opened 2021-02-12
Remove setting after uninstalling plugin

### LINKED ISSUE #265: PluginsManager does not clean up user settings after plugin removal
**Describe the bug/issue**
PluginsManager does not clean up user settings after plugin removal. Setting which contains custom hotkey is still retained after a plugin is removed.

**To Reproduce**
1. Install any plugin
2. Change the installed plugin hotkey
3. Uninstall the plugin
4. Reinstall and should see the default hotkey for the plugin, but the previously changed hotkey is still there.

This is due to the hotkey setting lives in actual Settings.json file in UserData\Settings folder. 

Have not checked if the plugin setting (in UserData folder\Settings\Plugins) is deleted or not.

taooceros:
fix #265 

TODO: 
create an interface for cleaning up plugin information if they store setting or some data.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
@@ -1,10 +1,9 @@
+using Flow.Launcher.Core.Plugin;
 using Flow.Launcher.Infrastructure;
 using Flow.Launcher.Infrastructure.Http;
 using Flow.Launcher.Infrastructure.Logger;
-using Flow.Launcher.Infrastructure.Storage;
 using Flow.Launcher.Infrastructure.UserSettings;
 using Flow.Launcher.Plugin.PluginsManager.Models;
-using Flow.Launcher.Plugin.PluginsManager.ViewModels;
 using Flow.Launcher.Plugin.SharedCommands;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -400,6 +402,12 @@ internal List<Result> RequestUninstall(string search)
 
         private void Uninstall(PluginMetadata plugin)
         {
+            
+
```
jjw24, line 406: Spaces

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -400,6 +402,12 @@ internal List<Result> RequestUninstall(string search)
...
+            Core.Plugin.PluginManager.Settings.Plugins.Remove(plugin.ID);
+            Core.Plugin.PluginManager.AllPlugins.RemoveAll(p => p.Metadata.ID == plugin.ID);
```
jjw24, line 408: if we can do this, does this mean no restart is required?

jjw24:
This is ready for review?

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
taooceros: Well, we still cannot unload assembly unless we use the library @JohnTheGr8 has mentioned, but we can load assembly dynamically, which means that installation may not require restart.

taooceros:
> This is ready for review?

yes, it is a simple change and it works

jjw24 (approved):

PR #341, opened 2021-02-14
Expand PluginAPI with Log and JsonStorage

### LINKED ISSUE #337: Accessing Flow.Launcher.Infrastructure?
**Question**
Should we allow plugins to access Infrastructure namespace? It seems quite hard to port all utility method to Plugin namespace, and the plugin namespace seems originally supposed to only store models.

taooceros:
close #337


taooceros:
I haven't tested it, but it should be fine to take a look.

jjw24:
👍 👍

jjw24 (approved):

PR #348, opened 2021-02-16
Remove duplicate entry

jjw24:
When result exists from both Quick Access and file search, use unionwith and custom comparer instead of append, so this will return distinct results of duplicates

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
@@ -38,7 +38,7 @@ namespace Flow.Launcher.Plugin.Explorer.Search
             var quickaccessLinks = QuickAccess.AccessLinkListMatched(query, settings.QuickAccessLinks);
 
             if (quickaccessLinks.Count > 0)
-                results.AddRange(quickaccessLinks);
+                results.Union(quickaccessLinks);
 
             var isEnvironmentVariable = EnvironmentVariables.IsEnvironmentVariableSearch(querySearch);
 
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
@@ -54,7 +54,7 @@ namespace Flow.Launcher.Plugin.Explorer.Search
             var quickaccessLinks = QuickAccess.AccessLinkListMatched(query, settings.QuickAccessLinks);
 
             if (quickaccessLinks.Count > 0)
-                results.AddRange(quickaccessLinks);
+                results.UnionWith(quickaccessLinks);
 
             var isEnvironmentVariable = EnvironmentVariables.IsEnvironmentVariableSearch(querySearch);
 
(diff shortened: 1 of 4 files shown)


taooceros:
Union won't work like this. This Union will create a IEnumerable that has been unioned, but not put the results back to the list. It would be better to use HashSet if you want to do the same job.
You may need to use HashSet and Union.

taooceros:
Or actually using an distinct will work, but will create another hashset to do the same job, so it would be better to simply do that with a hashset.

jjw24:
I actually think we should maybe do this so all plugin results will benefit, so plugins dont have to worry about dup results. What do you think

taooceros:
> I actually think we should maybe do this so all plugin results will benefit, so plugins dont have to worry about dup results. What do you think

Just add a distinct for the newResult will work

taooceros:
@jjw24 done, please take a look

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -38,7 +38,7 @@ internal async Task<List<Result>> SearchAsync(Query query, CancellationToken tok
             var quickaccessLinks = QuickAccess.AccessLinkListMatched(query, settings.QuickAccessLinks);
 
             if (quickaccessLinks.Count > 0)
-                results.AddRange(quickaccessLinks);
+                results.Union(quickaccessLinks);
```
jjw24, line 41: Distinct hasnt solved it yet, this change has been reverted. We probably need to write our own comparer

taooceros: Yeah I think you are right. By seeing the code of the equality comparator of Result class, it requires more than just title match. Maybe we still need to do that separately for plugins instead of distinct overall because we seldom will add two results that will be identical (they need to be equal at title, score, and highlight data) to the result list.

jjw24: yeah i agree. Perhaps we can create and add it to IPublicAPI so plugins can utilise the return distinct results themselves

taooceros: Well I don't think it should be included in IPublicAPI, because how they should filter the result should be decided by them. If we allow Equaltycomparater for that, I don't see the different between putting that in IPublicAPI and doing that by distinct or hashset.

taooceros: Let me add a customized add a customized equality comparator to do this job by comparing the subtitle (it is supposed to be the path) to remove duplicate for this one.

taooceros:
Done, please take a look @jjw24 
I haven't tested it, but it supposed to work

jjw24:
should we revert commit 4a3ad08?

taooceros:
> should we revert commit 4a3ad08?

Yeah we can do that. Seems that a simple distinct doesn't work well.

jjw24:
@taooceros good to roll?

taooceros (approved):

PR #358, opened 2021-02-21
add taskcanceledexception to updater exception catch

taooceros:
fix #357
This may because HttpClient throws TaskCanceledException when timeout instead of HttpException, which is previously thrown by the old HttpWebRequest.

PR #359, opened 2021-02-21
Release 1.7.2

jjw24:
Release Notes
=========

## Bug fixes
- Fixed an issue where plugin's action keyword setting gets removed when updating #352
- Fixed an issue where Flow's auto update may crash randomly due to unhandled task cancellation exception #358 
- Fixed duplicate entry when using Quick Access #348
- Fixed an issue with http proxy that caused Flow to crash #361


taooceros (dismissed):

jjw24:
@taooceros figure i release the other fix for plugin update

jjw24:
@taooceros  good to go?

taooceros:
Oh I thought this has been released. Since it hasn't, let's add #361 also, because that will block new users from using.

taooceros (approved):

PR #372, opened 2021-02-28
Change WebSearch default to global

jjw24:
1. Since web searches is performed many times, let's change the default to global action keyword so first time users do not need to go hunting in settings to find out the action keyword to activate it.

2. Update so if a action keyword is matched, display on top of global action keyword ('*') 
before:
image

after: 
image

suggestions also follow the same format.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.WebSearch/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.WebSearch/Main.cs
@@ -25,6 +25,10 @@ namespace Flow.Launcher.Plugin.WebSearch
         internal static string DefaultImagesDirectory;
         internal static string CustomImagesDirectory;
 
+        private readonly int scoreStandard = 50;
+
+        private readonly int scoreSuggestions = 48;
+
         private readonly string SearchSourceGlobalPluginWildCardSign = "*";
 
         public void Save()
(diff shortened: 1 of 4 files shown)


taooceros:
There's an issue for changing one to global keyword, that when you want to use action keyword for other source, this one will appear still. (and search suggestions will be duplicate maybe)
Maybe we still need to disable the global source when query with special keyword.

taooceros:
Maybe we can change the suggestion source to not directly search, but change query to specific suggestion to avoid duplicate suggestion for different source.

jjw24:
Well because it's using global and it's searching whatever you have typed in, therefore results from it will show up, this behaviour is normal, it's not broken. We should probably improve it by if there is an action keyword matched with another source, then show the results from that action keyword above the result from global action keyword, ie results from github search after typing `github flow` is above results from google search

taooceros:
> Well because it's using global and it's searching whatever you have typed in, therefore results from it will show up, this behaviour is normal, it's not broken. We should probably improve it by if there is an action keyword matched with another source, then show the results from that action keyword above the result from global action keyword, ie results from github search after typing `github flow` is above results from google search

Well, the main issue I am concerning about is the suggestion source. Currently that is not supporting multiple suggestion source.  
And the order addressing that should be changed (or else one suggestion source will need to wait till the suggestion source complete to show up).

jjw24:
> > Well because it's using global and it's searching whatever you have typed in, therefore results from it will show up, this behaviour is normal, it's not broken. We should probably improve it by if there is an action keyword matched with another source, then show the results from that action keyword above the result from global action keyword, ie results from github search after typing `github flow` is above results from google search
> 
> Well, the main issue I am concerning about is the suggestion source. Currently that is not supporting multiple suggestion source.
> And the order addressing that should be changed (or else one suggestion source will need to wait till the suggestion source complete to show up).

multi source suggestion seems to be working fine, it's in the foreach loop:
image

taooceros:
> > > Well because it's using global and it's searching whatever you have typed in, therefore results from it will show up, this behaviour is normal, it's not broken. We should probably improve it by if there is an action keyword matched with another source, then show the results from that action keyword above the result from global action keyword, ie results from github search after typing `github flow` is above results from google search
> > 
> > 
> > Well, the main issue I am concerning about is the suggestion source. Currently that is not supporting multiple suggestion source.
> > And the order addressing that should be changed (or else one suggestion source will need to wait till the suggestion source complete to show up).
> 
> multi source suggestion seems to be working fine, it's in the foreach loop:
> image

Oh I thought I haven't changed that on the earlier commit. Never mind, and it seems work great!

taooceros (approved):

PR #374, opened 2021-03-01
move main window width to BaseWindowStyle so it's dynamic

jjw24:
Currently width is fixed because of shadow effect. For some reason when moved to dynamic style shadow effect does not work. This needs further investigation.


taooceros:
I have found an issue that why do shadow effect change the size of text? From my case, the query box font should be `350`, but after opening the shadow effect. It seems that it will become `Regular`? Seems that shadow effect override the setting of font size.
image
image

jjw24:
dynamic width is added in https://github.com/Flow-Launcher/Flow.Launcher/pull/498

Issue #170, opened 2020-10-13
Favourite plugin port requests


jjw24:
Visit [here](https://flow-launcher.github.io/docs/#/port-plugins) on how to port .Net or Python plugins 

Please only request Wox plugins you were using regularly.

Currently ported plugins, use `pm install` to install them:

- WoxDictionary https://github.com/jjw24/WoxDictionary/releases/latest

- Google Search Plus https://github.com/jjw24/Wox.Plugin.GoogleSearch/releases/latest

- SpotifyPremium https://github.com/jjw24/Wox.Plugin.SpotifyPremium/releases/latest

- Runner https://github.com/jjw24/Wox.Plugin.Runner/releases/latest

- Clipboard Manager https://github.com/Xenolphthalein/Flow.Launcher.Plugin.ClipboardHistory

- WindowWalker https://github.com/taooceros/Microsoft.Plugin.WindowWalker

- ipadr https://github.com/taooceros/Flow.Plugin.IPAddress

- UrlEncode https://github.com/taooceros/Flow.Plugin.UrlEncode

- Putty https://github.com/jjw24/Flow.Launcher.Plugin.Putty

- Todoist
https://github.com/DanielBV/Wox.Plugin.Todoist

- isPrime
https://github.com/lvonkacsoh/Flow.Launcher.Plugin.IsPrime

jjw24:
From https://github.com/Flow-Launcher/Flow.Launcher/issues/95

![image](https://user-images.githubusercontent.com/26427004/95836097-c1065e00-0d8a-11eb-870f-c590b5330101.png)

-  Discogs **http://www.wox.one/plugin/90** 
-  SwiftTweet http://www.wox.one/plugin/63
-   Todoist http://www.wox.one/plugin/292
- Timer https://github.com/creativitRy/WoxTimer
- Graphical Timer https://github.com/JaroslawPokropinski/Wox.Plugin.GraphicalTimer

jjw24:
Hey @jeremymeyers, unfortunately the repo Wox.Plugin.Discogs is no longer available. I wont be able to port it.

Xenolphthalein:
Hey i would love to see the Clipboard History Tool from Wox in FlowLauncher.
https://github.com/Wox-launcher/Wox.Plugin.ClipboardManager
That always was my most used tool from Wox / Alfred / Albert.

pc223:
[...]

PR #378, opened 2021-03-04
Added Plugins section with curently compatible plugins

jeremymeyers:
I made a table with all the plugins listed as compatible on https://github.com/Flow-Launcher/Flow.Launcher/issues/170  ... the plugin names would need to be verified in terms of being able to use them to install correctly but since there doesn't seem to be anywhere else in the docs that this lives i thought it might be a good addition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,15 +40,6 @@ Flow Launcher. Dedicated to make your workflow flow more seamlessly. Aimed at be
 
 Windows may complain about security due to code not being signed, this will be completed at a later stage. If you downloaded from this repo, you are good to continue the set up. 
 
-### Integrations
-  - If you use Python plugins:
-    - Install [Python3](https://www.python.org/downloads/), download `.exe` installer.
-    - Add Python to `%PATH%` or set it in flow's settings.
-    - Use `pip` to install `flowlauncher`, open cmd and type `pip install flowlauncher`.
-    - The Python plugin may require additional modules to be installed, please ensure you check by visiting the plugin's website via `pm install` + plugin name, go to context menu and select `Open website`. 
-    - Start to launch your Python plugins.
-  - Flow searches files and contents via Windows Index Search, to use Everything: `pm install everything`.
-
 ### Usage
 - Open flow's search window: <kbd>Alt</kbd>+<kbd>Space</kbd> is the default hotkey.
 - Open context menu: on the selected result, press <kbd>Ctrl</kbd>+<kbd>O</kbd>/<kbd>Shift</kbd>+<kbd>Enter</kbd>.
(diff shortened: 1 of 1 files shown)


JohnTheGr8:
Quick thoughts:

1. listing plugins that have **not** been ported to this project in its Readme will probably confuse people, especially with the text `Here are some actively maintained plugins you can install.`
2. the first two columns can be merged into one to make this less noisy (visually)
3. in my opinion the last column can be omitted

jeremymeyers:
@JohnTheGr8 the "not ported" denotes plugins that were created for Wox that run in Flow without needing to make any changes.  The language could be clearer.

jjw24:
@jeremymeyers I think it would be better to reference this table in our documentation site instead.

What do you guys think if we:
- move Usage section above Integrations
- include it in Integrations section as another bullet point instead.
- A simple one liner for the bullet point along the lines of: `Flow has also integrated plugins from other launchers, visit here for a list of plugins you can use from them`
- Move your Plugins section contents to the docs site: https://github.com/Flow-Launcher/docs as a brand new page called Migrated Plugins (or whatever heading you see fit)

The reason for this is because I would think it's better to keep the readme light and concise, only the necessary information for new users that is needed to get going. Advanced users as well as launcher users can dig into the doco site to find out more what Flow offers additionally.

Let me know what you guys think.

jeremymeyers:
works for me, i just kinda wanted to get it started

jjw24:
You make the changes to the same pr here, and let me know when you have PRed the docs repo

jeremymeyers:
New doc repo is here https://github.com/Flow-Launcher/docs/pull/6 ... will make changes to readme.md in a sec

README.md
```diff
@@ -61,6 +52,19 @@ Windows may complain about security due to code not being signed, this will be c
...
+    - Add Python to `%PATH%` or set it in flow's settings.
+    - Use `pip` to install `flowlauncher`, open cmd and type `pip install flowlauncher`.
+    - The Python plugin may require additional modules to be installed, please ensure you check by visiting the plugin's website via `pm install` + plugin name, go to context menu and select `Open website`. 
+    - Start to launch your Python plugins.
+  - Flow searches files and contents via Windows Index Search, to use Everything: `pm install everything`.
+
+## Plugins
+
+There is a [list of actively maintained plugins](https://github.com/Flow-Launcher/docs/Plugins.md) in the documentation section, some of which are integrated from other launchers.
+
```
jjw24, line 67: instead of a separate section, could this be just a bullet point in Integrations section please

jeremymeyers: it can though i made it a separate thing because I'm not sure an average user would think to look for plugin installation information under an "integrations" section and the extensibility is a core Flow feature. (my focus generally is documentation UX). But happy to do whatever.

jjw24: Good point. Let's keep this section then, once the python auto install feature is rolled out I might merge Integration into the Plugin section as well so we can just use it as a quick guide on using plugins and where to find more info

jjw24 (approved):

PR #387, opened 2021-03-16
Add ConfigureWait and cancellation token & remove unnecessary OperationCanceledException

jjw24:
- Add ConfigureWait and cancellation token for https://github.com/Flow-Launcher/Flow.Launcher/pull/386
- Remove unnecessary catching of OperationCanceledException because it will be caught by PluginManager from each plugin query call

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
@@ -96,11 +96,12 @@ namespace Flow.Launcher.Plugin.Explorer.Search.WindowsIndex
 
             var constructedQuery = constructQuery(searchString);
             return await RemoveResultsInExclusionList(
-                        await ExecuteWindowsIndexSearchAsync(constructedQuery, connectionString, query, token),
-                        exclusionList);
+                        await ExecuteWindowsIndexSearchAsync(constructedQuery, connectionString, query, token).ConfigureAwait(false),
+                        exclusionList,
+                        token).ConfigureAwait(false);
         }
 
-        private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result> results, List<AccessLink> exclusionList)
+        private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result> results, List<AccessLink> exclusionList, CancellationToken token)
         {
             var indexExclusionListCount = exclusionList.Count;
 
@@ -109,27 +110,38 @@ namespace Flow.Launcher.Plugin.Explorer.Search.WindowsIndex
 
             var filteredResults = new List<Result>();
 
-            await Task.Run(() =>
+            try
             {
-                for (var index = 0; index < results.Count; index++)
+                await Task.Run(() =>
                 {
-                    var excludeResult = false;
-
-                    for (var i = 0; i < indexExclusionListCount; i++)
+                    for (var index = 0; index < results.Count; index++)
                     {
-                        if (results[index].SubTitle.StartsWith(exclusionList[i].Path, StringComparison.OrdinalIgnoreCase))
+                        token.ThrowIfCancellationRequested();
+
+                        var excludeResult = false;
+
+                        for (var i = 0; i < indexExclusionListCount; i++)
                         {
-                            excludeResult = true;
-                            break;
+                            token.ThrowIfCancellationRequested();
+
+                            if (results[index].SubTitle.StartsWith(exclusionList[i].Path, StringComparison.OrdinalIgnoreCase))
+                            {
+                                excludeResult = true;
+                                break;
+                            }
                         }
+
+                        if (!excludeResult)
+                            filteredResults.Add(results[index]);
                     }
+                }, token);
 
-                    if (!excludeResult)
-                        filteredResults.Add(results[index]);
-                }
-            });
-
-            return filteredResults;
+                return filteredResults;
+            }
+            catch (OperationCanceledException)
+            {
+                return new List<Result>();
+            }
         }
 
         internal static bool PathIsIndexed(string path)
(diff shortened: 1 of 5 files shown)


Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -109,27 +110,38 @@ private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result
 
             var filteredResults = new List<Result>();
 
-            await Task.Run(() =>
+            try
             {
-                for (var index = 0; index < results.Count; index++)
+                await Task.Run(() =>
```
taooceros, line 115: I would say that we don't need the Task.Run because we don't need to filter result in another thread. It is sync code so using a Task.Run won't create better performance. We just need to add the token.Throw here, and this method don't need to be asynchronous.
The OperationCancelledException is supposed to be catched by PluginManager, so no need  to add another try catch here. I try to do that in WebSearch plugin because I have encoutered debugging latency (the latency only appears when debugger attach, and I haven't found a great way to resolve that), but seems that simply catching the result won't solve that issue. Therefore, no need to do that.

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -96,11 +96,12 @@ internal async static Task<List<Result>> ExecuteWindowsIndexSearchAsync(string i
...
             return await RemoveResultsInExclusionList(
-                        await ExecuteWindowsIndexSearchAsync(constructedQuery, connectionString, query, token),
-                        exclusionList);
+                        await ExecuteWindowsIndexSearchAsync(constructedQuery, connectionString, query, token).ConfigureAwait(false),
+                        exclusionList,
+                        token).ConfigureAwait(false);
         }
 
-        private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result> results, List<AccessLink> exclusionList)
+        private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result> results, List<AccessLink> exclusionList, CancellationToken token)
```
taooceros, line 104:
```suggestion
        private static List<Result> RemoveResultsInExclusionList(List<Result> results, List<AccessLink> exclusionList, CancellationToken token)
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -109,27 +110,38 @@ private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result
 
             var filteredResults = new List<Result>();
 
-            await Task.Run(() =>
+            try
             {
```
taooceros, line 114:
```suggestion
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
taooceros, line 115:
```suggestion
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -109,27 +110,38 @@ private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result
...
-                for (var index = 0; index < results.Count; index++)
+                await Task.Run(() =>
                 {
```
taooceros, line 116:
```suggestion
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -109,27 +110,38 @@ private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result
...
-                    }
 
-                    if (!excludeResult)
-                        filteredResults.Add(results[index]);
-                }
-            });
+                        if (!excludeResult)
+                            filteredResults.Add(results[index]);
+                    }
+                }, token);
```
taooceros, line 137:
```suggestion
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -109,27 +110,38 @@ private async static Task<List<Result>> RemoveResultsInExclusionList(List<Result
...
 
-            return filteredResults;
+                return filteredResults;
+            }
+            catch (OperationCanceledException)
+            {
+                return new List<Result>();
+            }
```
taooceros, line 144:
```suggestion
```

taooceros (changes requested):

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
jjw24: > The OperationCancelledException is supposed to be catched by PluginManager, so no need to add another try catch here.

There is a try catch in IndexSearch class, how come that needs it?

taooceros: I would say no need for that, but since there already contains a try catch for ODBC exception, I add it. We can remove it to keep code cleaner.

jjw24: Sounds good, I will do that. So overall, there is no need to catch token cancellation thrown exception in plugins, because it's handled by the PluginManager, correct?

taooceros: Correct. Exception thrown is by design to avoid result return, so no need to catch them inside any plugin.

jjw24:
@taooceros done, please review

taooceros (approved):
Nice Work!

taooceros (approved):
Nice Work!

Issue #371, opened 2021-02-27
Slow startup


m3Lith:
From the time Windows desktop appears after boot, to the time that the launcher input appears, it takes almost 2 minutes.

I guess that's a bit unusual?

Anything I can do to see what's causing this?

taooceros:
Have you installed any plugin? If the programs index hasn't been cached, it may take some time to do that job before startup. However, it shouldn't be as long as 2 minutes.

jjw24:
> Anything I can do to see what's causing this?

You can restart your comp and once Flow has loaded, upload your logs here, we can take a look if any errors or unusual is going on. Having said this, my flow sometimes is last to start up and sometimes takes a minute or two to load after a reboot.

m3Lith:
Hmm, and you haven't figured out what's the problem in your case?

My logs form last startup: https://pastebin.com/sDCP9NUH

taooceros:
> Hmm, and you haven't figured out what's the problem in your case?
> 
> My logs form last startup: https://pastebin.com/sDCP9NUH
> 
> 

I take a look on the log, and the initialization time of most Plugins are a bit longer, but should only takes a few seconds to be done. Therefore, I think the possible reason for the slow startup is the indexing of Program plugin.

@jjw24 should we stop waiting the program plugin indexing if there are existing Win32 & UWP program that is preloaded? (this happens every three day)

jjw24:
> @jjw24 should we stop waiting the program plugin indexing if there are existing Win32 & UWP program that is preloaded? (this happens every three day)

I think it works better if already preloaded (assuming this means exists in cache data) then do not wait, use cache until indexing is finished

jjw24:
> Hmm, and you haven't figured out what's the problem in your case?

I rarely restart my laptop and and usually Flow finishes starting up after the others so I didn't really register it as an issue in my mind.

jjw24:
@m3Lith could you give this build from PR 381 a go and let me know if it's fixed the slow start up issue:
https://ci.appveyor.com/api/buildjobs/gqqmqu2k9bo2ame3/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.2.exe

m3Lith:
I don't think it did, still takes a while.. Latest log: https://pastebin.com/GUgurrGy

taooceros:
> I don't think it did, still takes a while.. Latest log: https://pastebin.com/GUgurrGy

[...]

PR #389, opened 2021-03-18
Optimize startup

taooceros:
A few enhancement I found during the investigation of #371

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginConfig.cs
+++ b/Flow.Launcher.Core/Plugin/PluginConfig.cs
@@ -20,7 +20,7 @@ namespace Flow.Launcher.Core.Plugin
         public static List<PluginMetadata> Parse(string[] pluginDirectories)
         {
             var allPluginMetadata = new List<PluginMetadata>();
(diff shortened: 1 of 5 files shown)


taooceros:
Oops, let me fix the build fail.

jjw24:
version bumps as well please

jjw24:
@taooceros need to also bump bookmarks plugin version.

Does removing readonly improve anything?

I don't remember, remind me again what's the benefit of using enumerate over get for directories?

jjw24 (changes requested):
See comments

taooceros:
> @taooceros need to also bump bookmarks plugin version.
> 
> Does removing readonly improve anything?
> 
> I don't remember, remind me again what's the benefit of using enumerate over get for directories?

No, but readonly requires initialization in constructor, while we construct instance in sync. I move that to Init() to allow parallel initialization.

jjw24 (approved):

taooceros:
＞ I don't remember, remind me again what's the benefit of using enumerate over get for directories?

I miss this question. It will be faster because we don't need to wait the enumeration finish to create an array and do action earlier.

PR #390, opened 2021-03-23
Make IPlugin inherit from IAsyncPlugin

taooceros:
The currently separated interface create quite significant code duplicate, so I try making the old plugin interface extends from the new plugin interface with the default implementation of interface from c# 8. (This is similar to the first version of the new Async model which directly use the default implementation in IPlugin interface).

It is hard to say which choice is better🤔🤔🤔, but this prevents the weird empty default implementation in legacy Query method.


jjw24:
This is going to force the plugins to implement Async?

taooceros:
> This is going to force the plugins to implement Async?

No, it is the same as the current behavior for IPlugin, which wraps a Task.Run with the Query method.

jjw24 (approved):

Issue #388, opened 2021-03-18
Migrate Direct call to PluginJsonStorage to API.save()


taooceros:

taooceros:
We haven't finish this part, actually 😂.

jjw24:
Oh ok, I thought this is for allowing calling save function via api which is implemented right

taooceros:
> Oh ok, I thought this is for allowing calling save function via api which is implemented right

No, this should move original direct use for JsonStorge for internal plugin to API.save

PR #393, opened 2021-03-27
Make PluginJsonStorage in IPublicAPI become single instance per type

taooceros:
Keep align with the usage of JsonStorage that use an instance holding the data to load and save instead of creating a new one every time. It uses dynamic to fit the generic type, but I also see someone use a generic type with static class to achieve similar result, but that will change quite significantly.
#388


jjw24:
@taooceros what would be a scenario where you would use BackupJsonStorage?

PR #395, opened 2021-03-29
Increase the default width and allow user to resize setting window, and reduce the margin for PluginDirectory

### LINKED ISSUE #375: Format issue for Plugin Directory
image

We should align this with more compatible way.

taooceros:
fix #375 

I increase the default width. The issue is due to the added Priority option, which uses more width than expected. Also, I allow users to resize the window. I don't know why we stop user from doing it before.


jjw24:
Change looks good to me, I am afk and want to load up the change to have a look first before merging.

jjw24:
hmm pluginsmanager and program plugins are still a bit off, could you adjust a little bit more please:

image

image

taooceros:
@jjw24 done

jjw24 (approved):
👍 

Hijacked your PR to do a quick update to translation and text, added a minwidth too

PR #398, opened 2021-04-01
add score bumping with query and do hashcode customized

### LINKED ISSUE #394: Score bump with query
Currently the score bumping is based on the selection of a certain result. However, I would think it would be better to only bump score when not only the result has been matched, but also the query has been match.
This will help users to separate the use of query for different high score results that both can be matched with separated query.

For example, I have a program named WeiXin: Wechat, and another one named Windows Terminal. When I want WeiXin with wx, it will be fine, but when I want Windows Terminal with acronym wt, it will sometimes move Weixin higher than Windows Terminal.

taooceros:
Only bump score when not only the result has been matched, but also the query has been match.
This will help users to separate the use of query for different high score results that both can be matched with separated query.

close #394

### DIFF
--- a/Flow.Launcher/Storage/UserSelectedRecord.cs
+++ b/Flow.Launcher/Storage/UserSelectedRecord.cs
@@ -10,6 +10,9 @@ namespace Flow.Launcher.Storage
 {
     public class UserSelectedRecord
     {
+        private const int HASH_MULTIPLIER = 31;
+        private const int HASH_INITIAL = 23;
+
         [JsonInclude]
         public Dictionary<int, int> records { get; private set; }
 
@@ -20,7 +23,7 @@ namespace Flow.Launcher.Storage
 
         private static int GenerateCustomHashCode(Query query, Result result)
         {
-            int hashcode = 23;
+            int hashcode = HASH_INITIAL;
 
             unchecked
             {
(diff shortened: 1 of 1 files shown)


Flow.Launcher/Storage/UserSelectedRecord.cs
```diff
@@ -8,16 +10,42 @@ namespace Flow.Launcher.Storage
...
+                }
+
+                foreach (var item in result.Title)
+                {
+                    hashcode = hashcode * 31 + item;
+                }
+
+                foreach (var item in result.SubTitle)
+                {
+                    hashcode = hashcode * 31 + item;
```
jjw24, line 39: 31?

taooceros: checkout https://stackoverflow.com/a/5155015
31 will have a better performance without information loss when int overflow

jjw24: Yeah, let's give it a bit of description, set it as a descriptive variable and reference that article so in the future we know where and why the logic is constructed

jjw24:
Could you please change 31 to a variable

taooceros:
> Could you please change 31 to a variable

Sure

jjw24 (approved):

PR #404, opened 2021-04-12
save Shell plugin cmd history

### LINKED ISSUE #403: Shell plugin cmd history does not get saved
Shell plugin cmd history does not get saved. Currently it seems only saved in memory, once flow exists, history is lost.

Fix to save to json file.

jjw24:
- Added saving executed cmds to settings
- Added option to show 5, 10, 20 most used cmds in settings

close #403, #405


taooceros:
is the default value for boolean is false?

jjw24:
Yep, that's why I didn't include default value.

taooceros (approved):

PR #409, opened 2021-04-13
Allow Quick Access Links to be accessed with middle string

taooceros:
Use `Contains` instead of `StartWith` to allows search with middle part of name, also score bump to 100 since it is not a fuzzy, but a full matched.

Or using Fuzzy would be better?

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
@@ -53,11 +53,11 @@ namespace Flow.Launcher.Plugin.Explorer.Search.WindowsIndex
                                                                     dataReaderResults.GetString(0),
                                                                     path,
                                                                     path,
-                                                                    query, true, true));
+                                                                    query, 0, true, true));
                             }
                             else
                             {
-                                fileResults.Add(ResultManager.CreateFileResult(path, query, true, true));
+                                fileResults.Add(ResultManager.CreateFileResult(path, query, 0, true, true));
                             }
                         }
                     }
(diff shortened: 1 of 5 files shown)


jjw24:
Given rest of search in Explorer can't use fuzzy, probably just stick with contains so not too confuse users?

Instead of looping through the matched results again and update score, what about adding an optional score parameter to the create file and create folder methods and assign the score that way?

taooceros:
> Given rest of search in Explorer can't use fuzzy, probably just stick with contains so not too confuse users?

I guess the rest of the search in Explorer still uses fuzzy to calculate the score.

> Instead of looping through the matched results again and update score, what about adding an optional score parameter to the create file and create folder methods and assign the score that way?

Yeah that seems great,.

jjw24:
> I guess the rest of the search in Explorer still uses fuzzy to calculate the score.

Just to calculate score or also uses fuzzy? I'm afk so can't take a look...

taooceros:
> > I guess the rest of the search in Explorer still uses fuzzy to calculate the score.
> 
> Just to calculate score or also uses fuzzy? I'm afk so can't take a look...

calculate the score as I remember previously. Let met take a look

taooceros:
Update, it utilizes fuzzy to create highlighed data......

jjw24:
Ok so actual search is not using fuzzy yeah

jjw24:
Version bump also forgot to mention

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -53,11 +53,11 @@ internal async static Task<List<Result>> ExecuteWindowsIndexSearchAsync(string i
                                                                     dataReaderResults.GetString(0),
                                                                     path,
                                                                     path,
-                                                                    query, true, true));
+                                                                    query, 0, true, true));
                             }
                             else
                             {
-                                fileResults.Add(ResultManager.CreateFileResult(path, query, true, true));
+                                fileResults.Add(ResultManager.CreateFileResult(path, query, 0, true, true));
```
jjw24, line 60: Why are we setting the score to 0 in several of the changes, have they always been 0?

taooceros: Yeah, we don't set the score before, which is default 0.

jjw24 (approved):

taooceros:
Thanks, haven't got time to finish the version bump.

PR #424, opened 2021-05-04
Fix broken link in README

andybergon:
`https://github.com/Flow-Launcher/docs/Plugins.md` gives a 404 to me, I think we need that `blob/<branch>/.


jjw24 (approved):
Good pick up, thank you

PR #427, opened 2021-05-11
Use API call to save instead of direct call

### LINKED ISSUE #388: Migrate Direct call to PluginJsonStorage to API.save()

taooceros:
close #388 

TODO:
- [x] Implement a auto call to Save all plugin's json storage that has registered
- [ ] Implement a better (safer) store to replace Dictionary<Type, object>
- [x] Finish up some settings

### DIFF
--- a/Flow.Launcher.Plugin/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/IPublicAPI.cs
@@ -26,16 +26,16 @@ namespace Flow.Launcher.Plugin
 
         /// <summary>
         /// Restart Flow Launcher
-        /// </summar
+        /// </summary>
         void RestartApp();
 
         /// <summary>
-        /// Saves everything, all of Flow Launcher and plugins' data and settings
+        /// Save everything, all of Flow Launcher and plugins' data and settings
         /// </summary>
         void SaveAppAllSettings();
 
         /// <summary>
-        /// Save Flow's plugins settings
+        /// Save all Flow's plugins settings
         /// </summary>
         void SavePluginSettings();
 
--- a/Flow.Launcher.Plugin/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/IPublicAPI.cs
@@ -186,12 +186,5 @@ namespace Flow.Launcher.Plugin
         /// <typeparam name="T">Type for Serialization</typeparam>
         /// <returns></returns>
         void SaveJsonStorage<T>(T settings) where T : new();
-
-        /// <summary>
-        /// Backup the JsonStorage you loaded from LoadJsonStorage
-        /// </summary>
-        /// <typeparam name="T"></typeparam>
-        /// <param name="settings"></param>
-        void BackupJsonStorage<T>() where T : new();
     }
 }
--- a/Flow.Launcher.Plugin/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/IPublicAPI.cs
@@ -182,6 +182,9 @@ namespace Flow.Launcher.Plugin
         /// <summary>
         /// Save JsonStorage for current plugin. This is the method used to save settings to json in Flow.Launcher
         /// This method will override the original class instance loaded from LoadJsonStorage
+        /// This method allows registering a type with provided instance so that it won't create a new one.
+        /// Only use it when you would like to create a new instance of the type instance and overwrite the original one
+        ///     stored before.
         /// </summary>
         /// <typeparam name="T">Type for Serialization</typeparam>
         /// <returns></returns>
--- a/Flow.Launcher.Plugin/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/IPublicAPI.cs
@@ -35,6 +35,11 @@ namespace Flow.Launcher.Plugin
         void SaveAppAllSettings();
 
         /// <summary>
+        /// Save Flow's plugins settings
+        /// </summary>
+        void SavePluginSettings();
+
+        /// <summary>
         /// Reloads any Plugins that have the 
         /// IReloadable implemented. It refeshes
         /// Plugin's in memory data with new content
--- a/Flow.Launcher.Plugin/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/IPublicAPI.cs
@@ -180,7 +180,7 @@ namespace Flow.Launcher.Plugin
         /// <summary>
         /// Save JsonStorage for current plugin's setting. This is the method used to save settings to json in Flow.Launcher
         /// This method will save the original instance loaded with LoadJsonStorage.
-        /// This API call is for manually Save. Flow will automatically save all setting that has registered. 
+        /// This API call is for manually Save. Flow will automatically save all setting type that has called LoadSettingJsonStorage or SaveSettingJsonStorage previously.
         /// </summary>
         /// <typeparam name="T">Type for Serialization</typeparam>
         /// <returns></returns>
--- a/Flow.Launcher.Plugin/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/IPublicAPI.cs
@@ -165,7 +165,8 @@ namespace Flow.Launcher.Plugin
         void LogException(string className, string message, Exception e, [CallerMemberName] string methodName = "");
 
         /// <summary>
-        /// Load JsonStorage for current plugin. This is the method used to load settings from json in Flow
+        /// Load JsonStorage for current plugin. This is the method used to load settings from json in Flow.
+        /// When the file is not exist, it will create a new instance for the specific type.
         /// </summary>
         /// <typeparam name="T">Type for deserialization</typeparam>
         /// <returns></returns>
--- a/Flow.Launcher.Plugin/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/IPublicAPI.cs
@@ -165,30 +165,20 @@ namespace Flow.Launcher.Plugin
         void LogException(string className, string message, Exception e, [CallerMemberName] string methodName = "");
 
         /// <summary>
-        /// Load JsonStorage for current plugin. This is the method used to load settings from json in Flow.
+        /// Load JsonStorage for current plugin's setting. This is the method used to load settings from json in Flow.
         /// When the file is not exist, it will create a new instance for the specific type.
         /// </summary>
         /// <typeparam name="T">Type for deserialization</typeparam>
         /// <returns></returns>
-        T LoadJsonStorage<T>() where T : new();
+        T LoadSettingJsonStorage<T>() where T : new();
 
         /// <summary>
-        /// Save JsonStorage for current plugin. This is the method used to save settings to json in Flow.Launcher
+        /// Save JsonStorage for current plugin's setting. This is the method used to save settings to json in Flow.Launcher
         /// This method will save the original instance loaded with LoadJsonStorage.
+        /// This API call is for manually Save. Flow will automatically save all setting that has registered. 
         /// </summary>
         /// <typeparam name="T">Type for Serialization</typeparam>
         /// <returns></returns>
-        void SaveJsonStorage<T>() where T : new();
-
-        /// <summary>
-        /// Save JsonStorage for current plugin. This is the method used to save settings to json in Flow.Launcher
-        /// This method will override the original class instance loaded from LoadJsonStorage
-        /// This method allows registering a type with provided instance so that it won't create a new one.
-        /// Only use it when you would like to create a new instance of the type instance and overwrite the original one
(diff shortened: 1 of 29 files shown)


Flow.Launcher/PublicAPIInstance.cs
```diff
@@ -133,15 +133,15 @@ public void OpenSettingDialog()
 
         public void LogException(string className, string message, Exception e, [CallerMemberName] string methodName = "") => Log.Exception(className, message, e, methodName);
 
-        private readonly Dictionary<Type, dynamic> PluginJsonStorages = new Dictionary<Type, dynamic>();
+        private readonly Dictionary<Type, object> PluginJsonStorages = new Dictionary<Type, object>();
```
taooceros, line 136: dynamic seems not working as intended, so I move back to direct cast

jjw24:
Did you want me to review and merge, or you still working on the other two todos?

taooceros:
> Did you want me to review and merge, or you still working on the other two todos?

still working on it. There are a few things that need to be cleaned up.
I am struggling with the second one, but it would be fine to not do it.

taooceros:
I didn't find a better way to replace the Dictionary<Type, object> and reflection for registration. So just review the current one.

jjw24:
Version bumps for plugins please :)

taooceros:
> Version bumps for plugins please :)

Sure

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -113,11 +111,6 @@ public Control CreateSettingPanel()
             return new SettingsControl(_settings);
         }
 
-        public void Save()
-        {
-            _storage.Save();
-        }
-
```
jjw24, line 120: why is this removed? settings still needs to be saved right

jjw24:
Not related to this change but I think we should update the comment to give an example scenario where you will need to override the original class instance, i.e. use the override save method and when you just need to use save the original instance. Since you are in this space maybe you could just give a good explanation in the comment that would be awesome.
https://github.com/Flow-Launcher/Flow.Launcher/blob/b37f9f338aee9c2ecd5a50ae47cecd2fa4dae5cb/Flow.Launcher.Plugin/IPublicAPI.cs#L174-L188

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
taooceros: I add a auto save for all registered jsonstorage, so that we don't need to configure isavable for setting, unless they are going to save something else.

taooceros: https://github.com/Flow-Launcher/Flow.Launcher/blob/b37f9f338aee9c2ecd5a50ae47cecd2fa4dae5cb/Flow.Launcher/PublicAPIInstance.cs#L139-L146

taooceros:
I have updated the comment. In fact, I don't believe we may actually need the one to overwrite it, because the supposed process is to load and use the loaded file (which will be a new when none exist), and save the original one. Should we remove it?

Also, do you think we should modify the name specifically to setting because the file will be stored inside the setting folder.

By the way, would you please give me the access for OAuth from Jetbrains IDE integration? Sometimes I may use rider instead of Visual Studio or vscode. Thank you!

jjw24:
> By the way, would you please give me the access for OAuth from Jetbrains IDE integration? Sometimes I may use rider instead of Visual Studio or vscode. Thank you!

Done

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
jjw24: 1. would this not be a doubling up then, with PluginManager.Save() call?

2. also you need to move this SaveAllJsonStorage() call to here right so when Flow's settings window closed you want to save all plugin settings:
https://github.com/Flow-Launcher/Flow.Launcher/blob/9d4af83fef727a38e22daf6e636351f42e3e1c08/Flow.Launcher/ViewModel/SettingWindowViewModel.cs#L85-L97

jjw24: 3. do we want to remove ISavable?
4. if it's auto saved, might want to document it somewhere right? for those that develop plugins

jjw24:
> I have updated the comment. In fact, I don't believe we may actually need the one to overwrite it, because the supposed process is to load and use the loaded file (which will be a new when none exist), and save the original one. Should we remove it?

i am unsure when will a circumstance be needed to save a new instance to the original, so i would say yes

> Also, do you think we should modify the name specifically to setting because the file will be stored inside the setting folder.

yes

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
taooceros: > would this not be a doubling up then, with PluginManager.Save() call?

I guess that one will call Save from ISavable? So I remove the call to save in exist plugin.

> also you need to move this SaveAllJsonStorage() call to here right so when Flow's settings window closed you want to save all plugin settings:

Yeah you are right.

> if it's auto saved, might want to document it somewhere right? for those that develop plugins

Yeah, I forget to add that into the api description

taooceros:
> > By the way, would you please give me the access for OAuth from Jetbrains IDE integration? Sometimes I may use rider instead of Visual Studio or vscode. Thank you!
> 
> Done

Thank you!

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
taooceros: > do we want to remove ISavable?

I believe not. There are other things that may need to save than setting (e.g. the program cache in program plugin).

taooceros: image

I guess we don't need to add it to viewmodel save. The api save will be called once Flow is disposed.

jjw24: > > do we want to remove ISavable?
> 
> I believe not. There are other things that may need to save than setting (e.g. the program cache in program plugin).

yeah you are right, lets keep it as we do need to save non-settings, but lets document the Save() method in so it's clear, otherwise devs might use it thinking that's how you save settings.

jjw24: > image
> 
> I guess we don't need to add it to viewmodel save. The api save will be called once Flow is disposed.

We need to because it's better to save when user close the settings window, otherwise what of the program crashes, no settings will be saved.

taooceros: Got you

taooceros: > We need to because it's better to save when user close the settings window, otherwise what if the program crashes, no settings will be saved.

Done

Flow.Launcher.Plugin/IPublicAPI.cs
```diff
@@ -34,6 +34,11 @@ public interface IPublicAPI
         /// </summary>
         void SaveAppAllSettings();
 
+        /// <summary>
+        /// Save Flow's plugins settings
+        /// </summary>
+        void SavePluginSettings();
```
jjw24, line 40: what's the difference between this method and SaveSettingJsonStorage()?

Flow.Launcher.Plugin/IPublicAPI.cs
```diff
@@ -165,33 +170,20 @@ public interface IPublicAPI
...
         /// <typeparam name="T">Type for deserialization</typeparam>
         /// <returns></returns>
-        T LoadJsonStorage<T>() where T : new();
+        T LoadSettingJsonStorage<T>() where T : new();
 
         /// <summary>
-        /// Save JsonStorage for current plugin. This is the method used to save settings to json in Flow.Launcher
+        /// Save JsonStorage for current plugin's setting. This is the method used to save settings to json in Flow.Launcher
         /// This method will save the original instance loaded with LoadJsonStorage.
+        /// This API call is for manually Save. Flow will automatically save all setting that has registered.
```
jjw24, line 183: i think it will be better to also include how setting gets registered in the comment so plugin devs do not need to dig into code to find out.

Flow.Launcher.Plugin/IPublicAPI.cs
taooceros: SaveSettingJsonStorage is for a specific Type (or T as generic), but this is for all plugin.

Flow.Launcher.Plugin/IPublicAPI.cs
```diff
@@ -26,11 +26,11 @@ public interface IPublicAPI
 
         /// <summary>
         /// Restart Flow Launcher
-        /// </summary>
+        /// </summar
```
taooceros, line 29: is this typo?

jjw24: yes. i fixed while i was updating the comments.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
jjw24: can you add SaveAllJsonStorage() call to here so when Flow's settings window closed you want to save all plugin settings:
https://github.com/Flow-Launcher/Flow.Launcher/blob/9d4af83fef727a38e22daf6e636351f42e3e1c08/Flow.Launcher/ViewModel/SettingWindowViewModel.cs#L85-L97

taooceros: I think this method is called in SaveAllJsonStorage, so it is impossible to directly call that method here, or else will become infinity recurse.
The call for 
```c#
PluginManager.Save(); 
_storage.Save(); 
```
do save all plugin setting.

jjw24: yeah i see. hmm how do save plugin settings when user closes settings window? PluginManager.Save() only saves plugin settings where the interface ISavable is applied and _storage.Save() saves Flow's own settings.

jjw24: instead of passing the api to the view, which is not even used, we can and should pass it to the viewmodel, this then allows you to call save plugin settings seperately from the view
https://github.com/Flow-Launcher/Flow.Launcher/blob/1e687a213039ba5f0c2fed42c7ef975a2d854141/Flow.Launcher/SettingWindow.xaml.cs#L28

taooceros: > PluginManager.Save() only saves plugin settings where the interface ISavable is applied and _storage.Save() saves Flow's own settings.

I have modified PluginManager.Save() to include saving plugins settings.

taooceros:
```c#
public static void Save()
        {
            foreach (var plugin in AllPlugins)
            {
                var savable = plugin.Plugin as ISavable;
                savable?.Save();
            }
            API.SavePluginSettings();
        }
```

Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
```diff
@@ -9,27 +9,20 @@ namespace Flow.Launcher.Plugin.Explorer.ViewModels
...
-        public Task LoadStorage()
-        {
-            return Task.Run(() => Settings = storage.Load());
-        }
 
         public void Save()
         {
-            storage.Save();
+            Context.API.SaveSettingJsonStorage<Settings>();
         }
```
jjw24, line 26: one more question, why does this have to call the save directly?

taooceros: I guess it is the design of Explorer plugin. I didn't change it. It contains a call for save every time a new item is edited.
It is not ISavable, so I believe it won't be called twice. However, do you think we should remove that?

jjw24: oh right, it's not ISavable, just calls infra's save directly. Nah don't worry about it for this PR, using the API is at least better. Let's get this merged in.

jjw24 (approved):

PR #429, opened 2021-05-14
Add special ctrl for Pluginsmanager and Shell plugin

### LINKED ISSUE #413: Run shell cmds as admin with ctrl
Shell plugin run shell cmds as admin with ctrl + enter/click, without the need to go to context menu

### LINKED ISSUE #421: Ctrl enter/click open Plugin's url
Ctrl enter/click open plugin's url when browsing plugins in PluginsManager plugin

taooceros:
close #413 close #421


taooceros:
By the way, I don't understand the autocomplete feature in the shell plugin. Why do people want to autocomplete a path in shell plugin? It only works when the whole command is a path.....

jjw24:
it's like using cmd or ps window, you can tab to complete the path c:\pro to c:\program files, so you don't have to type out the entire path

taooceros:
> it's like using cmd or ps window, you can tab to complete the path c:\pro to c:\program files, so you don't have to type out the entire path

but it only works for pure path. command like cd D:\ won't be able to autocomplete

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -86,7 +86,7 @@ public List<Result> Query(Query query)
                             IcoPath = Image,
                             Action = c =>
                             {
-                                Execute(Process.Start, PrepareProcessStartInfo(m));
+                                Execute(Process.Start, PrepareProcessStartInfo(m, c.SpecialKeyState.CtrlPressed));
```
jjw24, line 89: you need to do something with this keystate right, passing it in is not going to start the shell as admin right?

taooceros: I guess the CtrlPressed is a bool (when ctrl is pressed, it is true), and this parameter is starting as admin?

jjw24:
> > it's like using cmd or ps window, you can tab to complete the path c:\pro to c:\program files, so you don't have to type out the entire path
> 
> but it only works for pure path. command like `cd D:\` won't be able to autocomplete

yeah so something like `cd d:\prog` then user tabs and becomes `cd d:\program files\` or `cd d:\programdata\` etc

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
jjw24: oohh ok, i didnt look at the method properly. cool

jjw24:
version bumps please

taooceros:
> yeah so something like cd d:\prog then user tabs and becomes cd d:\program files\ or cd d:\programdata\ etc

yeah, I understand the concept, but the current auto complete feature seems uncomplete. I will try to fix it to allow auto complete when the last term is path.

taooceros:
> version bumps please

Sure

jjw24:
> > yeah so something like cd d:\prog then user tabs and becomes cd d:\program files\ or cd d:\programdata\ etc
> 
> yeah, I understand the concept, but the current auto complete feature seems uncomplete. I will try to fix it to allow auto complete when the last term is path.

when did we add autocomplete for shell? it was never there right. You mean the main program's autocomplete? yeah it doesnt really have the actual autocomplete capability, it's a sort of reflection of the current selected result.

jjw24:
you don't really need auto complete for all plugins right, maybe explorer and shell can benefit, maybe just extend it for those plugins only?

taooceros:
> you don't really need auto complete for all plugins right, maybe explorer and shell can benefit, maybe just extend it for those plugins only?

No, I just mean that the autocomplete for shell is not working perfectly. It only works for pure path like `D:\` but not command with path `cd D:\`

jjw24:
i think the merge erased the bump

jjw24 (approved):

PR #440, opened 2021-05-19
stop hiding scroller

### LINKED ISSUE #433: Main setting scrollbar hiding additional settings
Request is to make scrollbar a bit thicker so user notices there's additional settings on the bottom of the General tab.  I consider myself an advanced user was taken aback at how I missed this!

image

taooceros:
resolve #433 

We need votes for this pr
@Flow-Launcher/team

### DIFF
--- a/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
+++ b/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
@@ -63,14 +63,8 @@ namespace Flow.Launcher.ViewModel
 
         public bool AutoHideScrollBar
         {
-            get 
-            {
-                return Settings.AutoHideScrollBar;
-            }
-            set
-            {
-                Settings.AutoHideScrollBar = value;
-            }
+            get => Settings.AutoHideScrollBar;
+            set => Settings.AutoHideScrollBar = value;
         }
 
         // This is only required to set at startup. When portable mode enabled/disabled a restart is always required
@@ -150,11 +144,11 @@ namespace Flow.Launcher.ViewModel
 
         public bool ShouldUsePinyin
         {
-            get 
+            get
             {
-                return Settings.ShouldUsePinyin;            
+                return Settings.ShouldUsePinyin;
             }
-            set 
+            set
             {
                 Settings.ShouldUsePinyin = value;
             }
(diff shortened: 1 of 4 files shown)


Zeroto521:
Is there any screenshot to give a comparison?

taooceros:
**Current**
image

**After**
image

Zeroto521:
I thought that is a style problem.

There have two UI styles, one is win10 UWP style and another is win7 style.

For FlowLauncher the main UI is win7 style, and the scroller is the UWP style. Different UI styles cause that problem.

For now, I do agree with using the **win7** style.

Zeroto521 (approved):

jjw24 (changes requested):
Please add option to settings to autohide or not the scrollbar. This would be a better option.

jjw24:
Also I would consider making it a bit thicker if there is a option in there rather than removing it completely

taooceros:
> Please add option to settings to autohide or not the scrollbar. This would be a better option.

I think the current issue is that new users may not notice that there are some settings at the bottom that they need to scroll down to see. So the default value need to be false?

> Also I would consider making it a bit thicker if there is a option in there rather than removing it completely

Not sure, because this is a wrapped feature.

jjw24:
Just added the option to toggle on off, auto hide, off being the default.

image

@taooceros Let me know if it's good to merge

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
```diff
@@ -61,6 +61,18 @@ public bool AutoUpdates
...
+            get 
+            {
+                return Settings.AutoHideScrollBar;
+            }
+            set
+            {
+                Settings.AutoHideScrollBar = value;
+            }
+        }
+
```
taooceros, line 75: It seems that you directly link the Binding to `Settings.AutoHideScrollBar`, why is this property needed?

jjw24: property is for the settings window

jjw24: because we need to save the state

taooceros: I don't think you need this one. WPF will directly modify Settings.AutoHideScrollBar property, which already save the state.

jjw24: 👍 yes you are right, forgot about that. will make the change

jjw24: @taooceros done, please take another look

jjw24: @taooceros done, please take another look

taooceros:
Let me give it a test

taooceros:
Sorry I was wrong. We still need to backing field because modifying Settings.AutoHideScroller won't trigger the onPropertyChanged.

In this way, the effect will appear once the setting is hit.

jjw24 (approved):

PR #447, opened 2021-05-20
Implement Option to disable program description

### LINKED ISSUE #430: Option to Turn Off App Descriptions in Result List
Please implement this feature as it's badly needed. 

**THE PROBLEM**: In current state, Flow Launcher is pulling the Description of all apps/shortcuts into the results field.  This is really _unnecessary_ as power users tend to name their shortcuts appropriately. It clutters up the results list, makes it more difficult to read & identify the needed app.   I try to avoid comparing Flow Launcher to other apps, but most apps don't append additional text to the end.

image
image

image

**THE SOLUTION**

Give the user the option of turning this off as few may like it, especially newer users.
image

Once turned off, the results will display like this: 
image

taooceros:
resolve #430

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
+++ b/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
@@ -14,6 +14,7 @@
     <system:String x:Key="flowlauncher_plugin_program_indexing">Indexing</system:String>
     <system:String x:Key="flowlauncher_plugin_program_index_start">Index Start Menu</system:String>
     <system:String x:Key="flowlauncher_plugin_program_index_registry">Index Registry</system:String>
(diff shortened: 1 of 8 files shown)


taooceros:
during debug, I found that the cpu usage is being really high. Not sure whether that's expected.

jjw24:
can you do the test on master branch to see if same issue?

jjw24 (approved):

taooceros:
> can you do the test on master branch to see if same issue?

I check it out that the slowness is because I put Explorer plugin back to global plugin. It doesn't happen in Release mode (or debug mode without debugger). I believe that's because of the error handling.

jjw24:
> > can you do the test on master branch to see if same issue?
> 
> I check it out that the slowness is because I put Explorer plugin back to global plugin. It doesn't happen in Release mode (or debug mode without debugger). I believe that's because of the error handling.

Awesome, thank you

PR #449, opened 2021-05-22
Add specific keyword for path explore in explorer plugin

### LINKED ISSUE #292: [Explorer Plugin] Allow Folder search with an extra actionkeyword
Folder search with a special action keyword.
Sometimes we want to open a specific folder, but the result doesn't show it because too much other item that fulfill the requirement has been found.

### LINKED ISSUE #355: Separate actionkeyword of directory and index search
Separate the actionkeyword of the directory and index search.

taooceros:
resolve #355

- Seperated directory and index search, original main search is still intact.

- Allow action keywords (except file content search) to be disabled, reduces query time when not used.

- When an action keyword is not used, disable it automatically

- When user changes a disabled action keyword, enable it automatically

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+﻿[*.cs]
+
+# IDE0011: 添加大括号
+csharp_prefer_braces = when_multiline
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,0 @@
-﻿[*.cs]
-
-# IDE0011: 添加大括号
-csharp_prefer_braces = when_multiline
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
@@ -1,16 +1,7 @@
 ﻿using Flow.Launcher.Plugin.Explorer.ViewModels;
-using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Windows;
-using System.Windows.Controls;
-using System.Windows.Data;
-using System.Windows.Documents;
-using System.Windows.Input;
-using System.Windows.Media;
-using System.Windows.Media.Imaging;
-using System.Windows.Shapes;
 
 namespace Flow.Launcher.Plugin.Explorer.Views
 {
@@ -21,32 +12,43 @@ namespace Flow.Launcher.Plugin.Explorer.Views
     {
         private SettingsViewModel settingsViewModel;
 
-        private ActionKeywordView currentActionKeyword;
-
+        public ActionKeywordView CurrentActionKeyword { get; set; }
 
         private List<ActionKeywordView> actionKeywordListView;
 
-        public ActionKeywordSetting(SettingsViewModel settingsViewModel, List<ActionKeywordView> actionKeywordListView, ActionKeywordView selectedActionKeyword)
-        {
-            InitializeComponent();
+        private Settings settings;
 
+        public Visibility Visible => CurrentActionKeyword.Enabled is not null ? Visibility.Visible : Visibility.Collapsed;
+
+        public ActionKeywordSetting(SettingsViewModel settingsViewModel,
+            List<ActionKeywordView> actionKeywordListView,
+            ActionKeywordView selectedActionKeyword, Settings settings)
+        {
             this.settingsViewModel = settingsViewModel;
 
-            currentActionKeyword = selectedActionKeyword;
+            this.settings = settings;
 
-            txtCurrentActionKeyword.Text = selectedActionKeyword.Keyword;
+            CurrentActionKeyword = selectedActionKeyword;
 
             this.actionKeywordListView = actionKeywordListView;
+            
+            InitializeComponent();
+
         }
 
         private void OnConfirmButtonClick(object sender, RoutedEventArgs e)
         {
-            var newActionKeyword = txtCurrentActionKeyword.Text;
+            var newActionKeyword = TxtCurrentActionKeyword.Text;
 
             if (string.IsNullOrEmpty(newActionKeyword))
                 return;
 
-            if (newActionKeyword == currentActionKeyword.Keyword)
+            // reset to global so it does not take up an action keyword when disabled
+            if (!CurrentActionKeyword.Enabled is not null && newActionKeyword != Query.GlobalPluginWildcardSign)
+                settingsViewModel.UpdateActionKeyword(CurrentActionKeyword.KeywordProperty,
+                    Query.GlobalPluginWildcardSign, CurrentActionKeyword.Keyword);
+
+            if (newActionKeyword == CurrentActionKeyword.Keyword)
             {
                 Close();
 
@@ -54,18 +56,30 @@ namespace Flow.Launcher.Plugin.Explorer.Views
             }
 
             if (settingsViewModel.IsNewActionKeywordGlobal(newActionKeyword)
-                && currentActionKeyword.KeywordProperty == ActionKeywordProperty.FileContentSearchActionKeyword)
+                && CurrentActionKeyword.KeywordProperty == Settings.ActionKeyword.FileContentSearchActionKeyword)
             {
-                MessageBox.Show(settingsViewModel.Context.API.GetTranslation("plugin_explorer_globalActionKeywordInvalid"));
+                MessageBox.Show(
+                    settingsViewModel.Context.API.GetTranslation("plugin_explorer_globalActionKeywordInvalid"));
 
                 return;
             }
 
             if (!settingsViewModel.IsActionKeywordAlreadyAssigned(newActionKeyword))
             {
-                settingsViewModel.UpdateActionKeyword(currentActionKeyword.KeywordProperty, newActionKeyword, currentActionKeyword.Keyword);
+                settingsViewModel.UpdateActionKeyword(CurrentActionKeyword.KeywordProperty, newActionKeyword,
+                    CurrentActionKeyword.Keyword);
 
-                actionKeywordListView.FirstOrDefault(x => x.Description == currentActionKeyword.Description).Keyword = newActionKeyword;
+                actionKeywordListView.FirstOrDefault(x => x.Description == CurrentActionKeyword.Description).Keyword =
+                    newActionKeyword;
+
+                // automatically help users set this to enabled if an action keyword is set and currently disabled
+                if (CurrentActionKeyword.KeywordProperty == Settings.ActionKeyword.IndexOnlySearchActionKeyword
+                    && !settings.EnabledIndexOnlySearchKeyword)
+                    settings.EnabledIndexOnlySearchKeyword = true;
+
+                if (CurrentActionKeyword.KeywordProperty == Settings.ActionKeyword.PathSearchActionKeyword
+                    && !settings.EnabledPathSearchKeyword)
+                    settings.EnabledPathSearchKeyword = true;
 
                 Close();
 
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
@@ -67,11 +67,8 @@ namespace Flow.Launcher.Plugin.Explorer.Views
             }
 
 
-            if (CurrentActionKeyword is
-            {
-                Keyword: Query.GlobalPluginWildcardSign,
-                KeywordProperty: Settings.ActionKeyword.FileContentSearchActionKeyword
-            })
+            if (CurrentActionKeyword.KeywordProperty == Settings.ActionKeyword.FileContentSearchActionKeyword 
+                && ActionKeyword == Query.GlobalPluginWildcardSign)
             {
                 MessageBox.Show(
                     settingsViewModel.Context.API.GetTranslation("plugin_explorer_globalActionKeywordInvalid"));
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
@@ -23,6 +23,7 @@ namespace Flow.Launcher.Plugin.Explorer.Views
 
         private ActionKeywordView currentActionKeyword;
 
+
         private List<ActionKeywordView> actionKeywordListView;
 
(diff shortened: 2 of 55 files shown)


.editorconfig
```diff
@@ -0,0 +1,4 @@
+﻿[*.cs]
+
+# IDE0011: 添加大括号
+csharp_prefer_braces = when_multiline
```
jjw24, line 4: is this needed?

jjw24: @taooceros

taooceros: Sorry I forget to response to this message. I add that because vs is consistently asking me to change var to type name. I don't notice that it will be pushed to the repo. Sorry.

I will be afk for these two days, so I will revise that after tomorrow.

jjw24: All good, I wasn't sure of this option. I might remove it next time I continue to review this pr, if you haven't done do.

taooceros: Thank you

jjw24:
can we keep the action keyword to trigger both index and path search?

jjw24:
i am making some enhancements to allow using the original search action keyword or dedicated path and index action keywords. will push up soon.

jjw24:
this issue should resolve 355 not 292 right?

taooceros:
> this issue should resolve 355 not 292 right?

Ah you are right

taooceros:
> were you ever able to successfully implement allowing duplicate action keywords?

I only allow duplicate * keyword, but it is possible to implement duplicate action keywords without #227

taooceros:
Do we want all actionkeyword in Explorer plugin can be the same? Or only Path and Index search, while full-text search will remain distinct?

jjw24:
Maybe just hold off if you are going to make changes, I have made some new changes but have not pushed up yet. Will do so tomorrow

jjw24:
@taooceros I think it's better to keep the Search Activation action keyword for the scenario where you can use one action keyword to search index & path. We can add additional action keywords this way as well so we get the best of both scenarios.

I made the changes in https://github.com/Flow-Launcher/Flow.Launcher/pull/457, please take a look and let me know what you think, merge if all good.

jjw24:
found an issue with current action keyword binding, to repro- change file content keyword will reveal that the current action keyword is already assigned.

jjw24:
binding causes cancel to be ineffective too, think we can switch OK to Done and remove Cancel

taooceros:
> binding causes cancel to be ineffective too, think we can switch OK to Done and remove Cancel

Yeah if that's OK for you. I think I forget to implement cancel. 😂😖

jjw24:
why not, more modern design i suppose lol, and less having to change the code again

Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
```diff
@@ -21,51 +12,74 @@ public partial class ActionKeywordSetting : Window
...
         {
-            var newActionKeyword = txtCurrentActionKeyword.Text;
+            var newActionKeyword = TxtCurrentActionKeyword.Text;
 
             if (string.IsNullOrEmpty(newActionKeyword))
                 return;
 
-            if (newActionKeyword == currentActionKeyword.Keyword)
+            // reset to global so it does not take up an action keyword when disabled
+            if (!CurrentActionKeyword.Enabled is not null && newActionKeyword != Query.GlobalPluginWildcardSign)
```
jjw24, line 47: dont think this logic is effective, every action keyword except for content search is falling into this clause, hence update action keyword is called here and then down below in the code

jjw24: is this logic just to catch null Enabled as well as that it is false?

jjw24: something like this will work right?
if (CurrentActionKeyword.Enabled is not null && CurrentActionKeyword.Enabled == false && newActionKeyword != Query.GlobalPluginWildcardSign)

taooceros:
> why not, more modern design i suppose lol, and less having to change the code again

Actually, we can still implement the Cancel button, because we still need a temp storage. Let me give it a try.

Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
taooceros: I change it to pattern matching, which I believe is clearer. What do you think?
`if (CurrentActionKeyword is {Enabled: true, Keyword: not Query.GlobalPluginWildcardSign})`

jjw24:
well, if you look what i have done so far, the logic is solid and there is no need for temp storage. Although i just realised there is the 'X' button at the top right corner of the screen which acts as a cancel and screws with the binding. So up to you if you want to implement the Cancel or fix the 'X'.

We will also need to make this backwards compatible since we added the two new action keywords

taooceros:
> well, if you look what i have done so far, the logic is all working and there is no need for temp storage. Although i just realised there is the 'X' button at the top right corner of the window which acts as a cancel and screws with the binding. So up to you if you want to implement the Cancel or fix the 'X'.
> 
> We will also need to make this backwards compatible since we added the two new action keywords

We can use a temp storage (a copy from the original Actionkeyword view) to implement the cancel feature. Or we can just create a temp field with the actionkeyword and enable.

jjw24:
> We can use a temp storage (a copy from the original Actionkeyword view) to implement the cancel feature. Or we can just create a temp field with the actionkeyword and enable.

i think a copy might be nicer? up to you

> We will also need to make this backwards compatible since we added the two new action keywords

this is done, please pull latest

Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
jjw24: I am ok either way, make sure you test it though, the previous condition had the whole thing incorrectly working lol

taooceros:
```c#
if (string.IsNullOrEmpty(CurrentActionKeyword.Keyword))
                return;
```
Are this supposed to do nothing or we shall assume nothing to be global wildcard?

jjw24:
Let's go with nothing = global wildcard

taooceros:
image
I am confused about this logic. Why don't we allow newActionKeyword equal to oldActionKeyword when they are global wildcard?

jjw24:
Sorry not sure what you mean, but this is logic is in place because the actionkeywordregistered call doesn't work for global action keyword

taooceros:
> Sorry not sure what you mean, but this is logic is in place because the actionkeywordregistered call doesn't work for global action keyword

Yeah, but this logic means that we don't want users to assign the same actionkeyword when hitting done.

jjw24:
Yeah we don't want user to assign the same action keyword, only if its global

taooceros:
> Yeah we don't want user to assign the same action keyword, only if its global

No, I mean this logic will prevent user from clicking done when they only want to enable the option.

jjw24:
Basically when new global ==  old global, don't need to update the action keyword, because it's already assigned. Same for non global action keywords

taooceros:
> Basically when new global == old global, don't need to update the action keyword, because it's already assigned. Same for non global action keywords

I got your idea. However, I think this logic should not be put here, because this Assign checking is to check whether there are duplicate actionkeyword, and preventing user from assigning them.

jjw24:
Oh you mean when the action keyword is global, and disabled, user click done it is supposed to enable the action keyword?

taooceros:
> Oh you mean when the action keyword is global, and disabled, user click done it is supposed to enable the action keyword?

yeah something like that. When user want to enable global, it will return false and show users that this is already assigned, because the old one is equal to the new one.

jjw24:
Hold on it should show the message to tell user there is already one assigned in my latest commit, Are you working off that?

jjw24:
Basically it should behave as- do not update the search keyword if it (global/non global) is already assigned.

taooceros:
> Hold on it shouldnt show the message to tell user there is already one assigned in my latest commit, Are you working off that?

Not sure whether I have changed something wrong. Let me take a look on the current commit.

jjw24:
Additionally, if action keyword is disabled, and user changes action keyword, then enable it automatically.

If action keyword is enabled and user disable it, automatically revert to global

taooceros:
> Additionally, if action keyword is disabled, and user changes action keyword, then enable it automatically.
> 
> If action keyword is enabled and user disable it, automatically revert to global

Yes, I see these logic and keep them.

There's a logic issue on the last commit. If the IsActionKeywordAssigned return false, it will still close the window, but not to tell user that they are supposed to change some setting that is not allowed. I believe that why you don't discover the issue.

jjw24:
> > Additionally, if action keyword is disabled, and user changes action keyword, then enable it automatically.
> > If action keyword is enabled and user disable it, automatically revert to global
> 
> Yes, I see these logic and keep them.
> 
> There's a logic issue on the last commit. If the IsActionKeywordAssigned return false, it will still close the window, but not to tell user that they are supposed to change some setting that is not allowed. I believe that why you don't discover the issue.

Yes good pick up lol, I forgot abt that use case.

taooceros:
I finish the editing. Now, ActionKeywordView is linked to setting.actionkeyword, and it will only be trigger when hitting the done button. We can add the Cancel Button back if you want because I use two temp field to store ActionKeyword and Enabled for binding (because I directly linked ActionKeywordView to setting, so using another instance of that won't work for this design).

jjw24:
let me take it for a spin

> We can add the Cancel Button back if you want because I use two temp field to store ActionKeyword and Enabled for binding (because I directly linked ActionKeywordView to setting, so using another instance of that won't work for this design).

I dont really mind not having the cancel button, i think this is more modern design right? the only thing i have think is worth thinking about is it's not consistent with the rest of the plugins, but i would prefer modern design and put in effect to move the default plugins as well as Flow's interface towards that.

jjw24:
Getting this error when i change search action keyword to e (which works and triggers) then switch back to global (then nothing works, no results from any plugin displays):

Exception thrown: 'System.Collections.Generic.KeyNotFoundException' in System.Private.CoreLib.dll
Exception thrown: 'System.Reflection.TargetInvocationException' in System.Private.CoreLib.dll
System.Windows.Data Error: 8 : Cannot save value from target back to source. BindingExpression:Path=QueryText; DataItem='MainViewModel' (HashCode=8662426); target element is 'TextBox' (Name='QueryTextBox'); target property is 'Text' (type 'String') KeyNotFoundException:'System.Collections.Generic.KeyNotFoundException: The given key 'e' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Flow.Launcher.ViewModel.MainViewModel.RemoveOldQueryResults(Query query) in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 564
   at Flow.Launcher.ViewModel.MainViewModel.QueryResults() in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 471
   at Flow.Launcher.ViewModel.MainViewModel.Query() in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 352
   at Flow.Launcher.ViewModel.MainViewModel.set_QueryText(String value) in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 270'
The thread 0x2acc has exited with code 0 (0x0).
The thread 0x54b0 has exited with code 0 (0x0).
The thread 0x35c0 has exited with code 0 (0x0).
The thread 0x4770 has exited with code 0 (0x0).

taooceros:
> > We can add the Cancel Button back if you want because I use two temp field to store ActionKeyword and Enabled for binding (because I directly linked ActionKeywordView to setting, so using another instance of that won't work for this design).
> 
> I dont really mind not having the cancel button, i think this is more modern design right? the only thing i think worth thinking about is it's not consistent with the rest of the plugins, but i would prefer modern design and put in effort to move the default plugins as well as Flow's interface towards that.

Well, I am not sure whether not having a cancel button is a more modern design (in fact, what's a more modern design is hard to be argued).

taooceros:
> Getting this error when i change search action keyword to e (which works and triggers) then switch back to global (then nothing works, no results from any plugin displays):
> 
> Exception thrown: 'System.Collections.Generic.KeyNotFoundException' in System.Private.CoreLib.dll
> Exception thrown: 'System.Reflection.TargetInvocationException' in System.Private.CoreLib.dll
> System.Windows.Data Error: 8 : Cannot save value from target back to source. BindingExpression:Path=QueryText; DataItem='MainViewModel' (HashCode=8662426); target element is 'TextBox' (Name='QueryTextBox'); target property is 'Text' (type 'String') KeyNotFoundException:'System.Collections.Generic.KeyNotFoundException: The given key 'e' was not present in the dictionary.
> at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
> at Flow.Launcher.ViewModel.MainViewModel.RemoveOldQueryResults(Query query) in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 564
> at Flow.Launcher.ViewModel.MainViewModel.QueryResults() in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 471
> at Flow.Launcher.ViewModel.MainViewModel.Query() in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 352
> at Flow.Launcher.ViewModel.MainViewModel.set_QueryText(String value) in C:\Workbench\Git\Flow.Launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 270'
> The thread 0x2acc has exited with code 0 (0x0).
> The thread 0x54b0 has exited with code 0 (0x0).
> The thread 0x35c0 has exited with code 0 (0x0).
> The thread 0x4770 has exited with code 0 (0x0).

That seems some concurrency issue? Let me take a check on that.

taooceros:
image
Do we need this logic? I don't believe we have an increment strategy when updating plugins.

taooceros:
hmmm Let me check for the update logic

taooceros:
I found the issue. When we have a temp query, and change the actionkeyword for a plugin, it will result the dictionary being changed, but this method will access last time actionkeyword to check for removing results. Therefore, it will result the issue. I will fix it soon.

taooceros:
I rethink logic about that. We don't need to make it so complicate. Only if current actionkeyword and lastactionkeyword is null, we need to refresh the view. No need for only removing results for certain plugins because every query results will refresh their plugin results else where (which is not incremental). Removing all will be fine and clear.

jjw24:
From testing, fixed two minor logic issue with file content search and revert to global when disabled, otherwise 💯 , lets get this merged.

jjw24 (approved):

PR #451, opened 2021-05-24
Cache program result

taooceros:
The program plugin will generate same result for same query unless it has been reindex. Therefore, it would be possible to avoid duplicate calculation by creating a cache, since user mostly will use similar query.


jjw24:
Great idea, will take a look later

jjw24:
Have you by any chance compared performance with this change?

taooceros:
> Have you by any chance compared performance with this change?

Well I believe the current program plugin is quite fast? In my computer, the average calculation time is under 20ms (shorter query can be shorter). When cached, the query time will be 0ms.

There will be some memory cost for the cache, but since Result is a quite simple class, the cost is under 50mb? That may require some extra testing.

jjw24:
>Well I believe the current program plugin is quite fast? In my computer, the average calculation time is under 20ms (shorter query can be shorter). When cached, the query time will be 0ms.

Yep very fast thanks to your changes 👍 

>There will be some memory cost for the cache, but since Result is a quite simple class, the cost is under 50mb? That may require some extra testing.

As in ram? If so this we might need to confirm how much larger because I have read some users are not too fond of the large memory footprint currently flow has.

taooceros:
If memory is under higher consideration, we can use WeakReference to cache the result. Or we can use a more specific control over cache like what we have done in ImageCache.

jjw24:
> If memory is under higher consideration, we can use WeakReference to cache the result (which is almost zero cost because we need to create the List of Result even without the cache). Or we can use a more specific control over cache like what we have done in ImageCache.

Yeah the general feeling I get from user comments around this type of applications is that memory usage is quite high, so where possible we would want to minimise it. I was going to take a look at why exactly flow is taking up so much memory and whether it's justifiable as one of next project goals.

Do you think you can add to this change using WeakReference

taooceros:
> > If memory is under higher consideration, we can use WeakReference to cache the result (which is almost zero cost because we need to create the List of Result even without the cache). Or we can use a more specific control over cache like what we have done in ImageCache.
> 
> Yeah the general feeling I get from user comments around this type of applications is that memory usage is quite high, so where possible we would want to minimise it. I was going to take a look at why exactly flow is taking up so much memory and whether it's justifiable as one of next project goals.
> 
> Do you think you can add to this change using WeakReference

Yeah that's the easiest way.

taooceros:
It seems that the GC is really collecting eagerly, so only using a weakreference may not be a great option for caching. It will be collected after one or two GC quickly unless we keep typing the same keyword. Maybe it would be better to keep a stronger caching system.

jjw24 (approved):
👍

jjw24:
would this change bump up the memory usage slightly?

taooceros:
> would this change bump up the memory usage slightly?

I think so, but probably not a lot.

PR #452, opened 2021-05-24
Avoid AppxManifest.xml leak

taooceros:
from PT-Run https://github.com/microsoft/PowerToys/pull/10394

release Com stream Object immediately

Co-Authored-By: Andrey Nekrasov <1828123+yuyoyuppe@users.noreply.github.com>


jjw24:
might be worth doing this for the ShellLinkHelper class as well?
https://github.com/microsoft/PowerToys/blob/f61db8ed3f63c69d1adb514afb006fc692f0b602/src/modules/launcher/Plugins/Microsoft.Plugin.Program/Programs/ShellLinkHelper.cs#L178

in a seperate pr

jjw24 (approved):

PR #455, opened 2021-05-25
Fix loading when no action keywords present for a plugin in user plugin setting

### LINKED ISSUE #432: WebSearch 1.3.2 -  Action keywords collisions with default configurations
**To Reproduce**
1. Use portable version of Flow Launcher (this is optional, but helps with testing on a clean install)
2. Delete all of the default search entries like Wiki or Bing 
3. Close the app completely (optional)
4. Open the app  and try to add Bing back.

Note in this pic, Bing does not exist
image

When I tried to add Bing back (using my URL configuration), it alerted of name collision.  If I used "bing" instead of "b" it allows me to save.
image

Windows 10 64bit LTSC
image

WebSearch 1.3.2. 
Flow Launcher 1.7.2

**Flow Launcher Error Log** 

** ** NOTE** ** I cleared all logs file, reproduced the error, and copied the log file. There's no mention in the log of the action collision.

jjw24:
Web Searches plugin allows users to fully delete all search setting entries, resulting in a scenario where there could be empty action keyword. Current code when flow starts up and there are no action keywords present for the plugin, it switches to populating the action keywords from metadata (plugin.json). 

For Web Searches plugin, this causes a mismatch between the flow's setting plugin action keywords data (has action keywords because repopulated from metadata) and the action keywords in Web Searches plugin's own setting (has no action keywords because all deleted).

close #432 

also added tool tips to make functionality clearer

### DIFF
--- a/Flow.Launcher.Infrastructure/UserSettings/PluginSettings.cs
+++ b/Flow.Launcher.Infrastructure/UserSettings/PluginSettings.cs
@@ -30,6 +30,11 @@ namespace Flow.Launcher.Infrastructure.UserSettings
                         metadata.ActionKeywords = settings.ActionKeywords;
                         metadata.ActionKeyword = settings.ActionKeywords[0];
                     }
+                    else
+                    {
+                        metadata.ActionKeywords = new List<string>();
+                        metadata.ActionKeyword = string.Empty;
+                    }
                     metadata.Disabled = settings.Disabled;
                     metadata.Priority = settings.Priority;
                 }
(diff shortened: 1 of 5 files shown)


Flow.Launcher.Infrastructure/UserSettings/PluginSettings.cs
```diff
@@ -30,6 +30,11 @@ public void UpdatePluginSettings(List<PluginMetadata> metadatas)
                         metadata.ActionKeywords = settings.ActionKeywords;
                         metadata.ActionKeyword = settings.ActionKeywords[0];
                     }
+                    else
+                    {
+                        metadata.ActionKeywords = new List<string>();
+                        metadata.ActionKeyword = string.Empty;
```
taooceros, line 36: won't this causing "" be used as action keyword?

taooceros: Oh I check the code for adding actionkeyword, only actionkeywords will be access.

taooceros (approved):

PR #456, opened 2021-05-27
fix duplicate query searches from same global action keywords

jjw24:
When a plugin has multiple global action keywords, flow loops through the plugin's global (*) and action keywords and fire them off to query. This creates a scenario where a plugin with multiple global action keywords (*), for example 4, running the plugin search 4 times unnecessarily. This also could potentially have a performance cost to it.

This fix excludes adding multiple duplicate global and action keywords even if the plugin has 4 global action keywords, only fire search once.


taooceros:
why not just make GlobalPlugins a hashset?

jjw24:
lgtm, did you want to approve?

taooceros:
Sorry I am reinstalling windows today. Will do it now

taooceros (approved):

PR #457, opened 2021-05-30
update to ExplorerPathActionkeyword branch

jjw24:


taooceros:
Why not use binding instead of a seperated function triggered when checked

jjw24:
> ~Why not use binding instead of a seperated function triggered when checked~

Crossed out?

taooceros:
> > ~Why not use binding instead of a seperated function triggered when checked~
> 
> Crossed out?

Crossed out because the function require additional checking (but still can be done in binding, but is similarly complex). Let me check whether we can simplify this process.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -46,36 +47,39 @@ internal async Task<List<Result>> SearchAsync(Query query, CancellationToken tok
...
-            if (query.ActionKeyword == settings.SearchActionKeyword)
-                return Settings.ActionKeyword.SearchActionKeyword == allowedActionKeyword;
-
-            
-            return (Settings.ActionKeyword.IndexOnlySearchActionKeyword == allowedActionKeyword && settings.EnabledIndexOnlySearchKeyword)
-                || (Settings.ActionKeyword.PathSearchActionKeyword == allowedActionKeyword && settings.EnabledPathSearchKeyword)
-                || settings.SearchActionKeyword == Query.GlobalPluginWildcardSign;
+            return allowedActionKeyword switch
+            {
+                Settings.ActionKeyword.SearchActionKeyword => settings.EnableSearchActionKeyword &&
```
jjw24, line 74: would this `Settings.ActionKeyword.SearchActionKeyword ` work if the action keyword has been updated several times, since you are not referencing the settings instance that contains the updated action keyword?

taooceros: The `Settings.ActionKeyword.SearchActionKeyword` is only an enum, so it is static

jjw24: oh yeah you are right, lol ignore me

taooceros: It is only used for matching which actionkeyword we want to check. `settings.SearchActionKeyword` is the actual reference we are checking.

jjw24: Yeah I actually did several rewrites before pushing this up so forgot which version I had settled on lol

taooceros:
hmmm some logic is weird. Give me some time to fix that.

jjw24:
> Do we want these setting be the same? Or they are not supposed to be the same (the same just to use SearchActionkeyword)

what settings are you referring to?

taooceros:
> > Do we want these setting be the same? Or they are not supposed to be the same (the same just to use SearchActionkeyword)
> 
> what settings are you referring to?


The four action keywords.

jjw24:
same as in using the global action keyword? then yes except for content search which default is `doc:`

taooceros:
> same as in using the global action keyword? then yes except for content search which default is `doc:`

Well in my last commit, I make it to be exclusive. I would consider user should use the SearchActionkeyword when they want index search and path search with one actionkeyword, use PathSearchActionkeyword or IndexOnlyActionkeyword when they only wants a specific location.
Maybe there's no need for allowing duplicate actionkeyword, which will increase the complexity of updating actionkeyword.

jjw24:
> > same as in using the global action keyword? then yes except for content search which default is `doc:`
> 
> Well in my last commit, I make it to be exclusive. I would consider user should use the SearchActionkeyword when they want index search and path search with one actionkeyword, use PathSearchActionkeyword or IndexOnlyActionkeyword when they only wants a specific location.
> Maybe there's no need for allowing duplicate actionkeyword, which will increase the complexity of updating actionkeyword.

Yeah, the changes in this pr is intending to allow search as how it is functioning before, plus those individual action keywords when enabled.

No need to worry about allowing dup action keyword for this, if we going to tackle it better do it in a separate pr.

jjw24:
Only duplicate action keyword that is allowed is global action keyword

taooceros:
> Only duplicate action keyword that is allowed is global action keyword

yes you are right, because that's handled by Flow's plugin system.

jjw24:
if this is good to go, shall we merge it into your pr?

taooceros:
> if this is good to go, shall we merge it into your pr?

Sure, I just waiting for your review.

PR #474, opened 2021-06-15
fix storyboard animation control error

taooceros:
The storyboard is only able to be controlled if using Begin(object, controllable) for beginning.

Maybe this can actually drop cpu usage to 0.

Related to #128

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -88,13 +88,13 @@ namespace Flow.Launcher
 
                         if (_viewModel.ProgressBarVisibility == Visibility.Visible && isProgressBarStoryboardPaused)
                         {
-                            _progressBarStoryboard.Resume();
+                            _progressBarStoryboard.Resume(ProgressBar);
                             isProgressBarStoryboardPaused = false;
                         }
                     }
                     else if (!isProgressBarStoryboardPaused)
                     {
-                        _progressBarStoryboard.Pause();
+                        _progressBarStoryboard.Pause(ProgressBar);
                         isProgressBarStoryboardPaused = true;
                     }
                 }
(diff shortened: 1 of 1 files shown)


taooceros:
In Sysinternal Process explorer, after hiding, the cpu will be lower than 0.01. I believe that's acceptable.
image

jjw24:
Also have you tested it still stops when hidden and starts again when visible as per the original behaviour?

taooceros:
> Also have you tested it still stops when hidden and starts again when visible as per the original behaviour?

Yes, I have tested hide and start a few times, and also triggering progress bar animation by requesting suggestion source. However, let me test more about whether the animation is working properly.

taooceros:
A small issue is that the progress bar animation will start at the position it stop last time. Should we reset the position?

jjw24:
This is the query window re-appearing after hiding and query is still running? Then yeah I think resetting it is fine.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -72,47 +72,57 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
-                        }
-                        else if (_viewModel.MainWindowVisibility == Visibility.Visible && isProgressBarStoryboardPaused)
-                        {
-                            _progressBarStoryboard.Resume();
-                            isProgressBarStoryboardPaused = false;
-                        }
-                    }, System.Windows.Threading.DispatcherPriority.Render);
+                            if (_viewModel.ProgressBarVisibility == Visibility.Hidden && !isProgressBarStoryboardPaused)
+                            {
+                                await Task.Delay(50);
```
jjw24, line 112: why the delay?

taooceros: make it stop after hiding, or else user may feel a bit stunt.

taooceros:
hmmm hold on, after the last commit it seems that flow will still consume some cpu cycle when hiding. Let me take another check.

taooceros:
ok fixed

jjw24 (approved):

PR #475, opened 2021-06-16
Fix plugin documentation link

r-darwish:


jjw24 (approved):
good catch, thank you

PR #476, opened 2021-06-16
Optimize PostBuild Script

taooceros:
1. Use HashTable to check duplicate dll
2. Use dotnet pack instead of nuget pack

- [ ] Allow squirrel create delta

### DIFF
--- a/Flow.Launcher/Flow.Launcher.csproj
+++ b/Flow.Launcher/Flow.Launcher.csproj
@@ -1,5 +1,4 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
     <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
@@ -12,6 +11,8 @@
     <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <SatelliteResourceLanguages>en</SatelliteResourceLanguages>
+    <NuspecFile>../Scripts/flowlauncher.nuspec</NuspecFile>
+    <NuspecProperties>version=$(PackageVersion)</NuspecProperties>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
--- a/Scripts/flowlauncher.nuspec
+++ b/Scripts/flowlauncher.nuspec
@@ -3,7 +3,7 @@
   <metadata>
     <id>FlowLauncher</id>
     <title>Flow Launcher</title>
-    <version>$version$</version>
+    <version>1.0.0</version>
     <authors>Flow-Launcher Team</authors>
     <projectUrl>https://github.com/Flow-Launcher/Flow.Launcher</projectUrl>
     <iconUrl>https://raw.githubusercontent.com/Flow-Launcher/Flow.Launcher/master/Flow.Launcher/Images/app.png</iconUrl>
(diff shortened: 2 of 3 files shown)


taooceros:
@JohnTheGr8 Is it possible to download the latest release nupkg to the CI?

JohnTheGr8:
>     1. Use HashTable to check duplicate dll

Does this actually improve performance? I see the build time for this commit (7c242c8505) is as usual, around 3m30s.

>     2. Use dotnet pack instead of nuget pack

IIRC I used `nuget pack` instead of `dotnet pack` because `dotnet pack` did not behave as expected with multiple project references. Seems like this is no longer the case, though.

> @JohnTheGr8 Is it possible to download the latest release nupkg to the CI?

We can probably do it "manually" using Github's API. However, I think squirrel updates using deltas are not that reliable (at least in older versions of Squirrel), which is why they were not used by Wox in the first place... So do keep that in mind. They will also increase CI build times ;)

taooceros:
> We can probably do it "manually" using Github's API. However, I think squirrel updates using deltas are not that reliable (at least in older versions of Squirrel), which is why they were not used by Wox in the first place... So do keep that in mind. They will also increase CI build times ;)

Yeah I understand that will increase the CI build times, but Flow is self-contained, but wox is not. There will always be a significant amount of same dll in every release, so if delta is possible, it can reduce around 60MB for updating (at the time trimming is not reliable for wpf). We can make it to be release only if that takes too long.

> IIRC I used nuget pack instead of dotnet pack because dotnet pack did not behave as expected with multiple project references. Seems like this is no longer the case, though.

dotnet pack can take nuspec, so it can do the same job if we pass --no-build, but if you see my second commit, it is unable to specify version directly through command line, so csproj need to be changed.

> Does this actually improve performance? I see the build time for this commit (7c242c8) is as usual, around 3m30s.

It reduces 20-30s in my computer, but publish self-contained and squirrel pack takes even longer, and the CI needs time to setup, so the result is not significant.

taooceros:
> However, I think squirrel updates using deltas are not that reliable (at least in older versions of Squirrel)

Do that become more reliable for newer version? Actually, I don't think we need an actual delta update, but just removing the self-contained part when doing update. However, seems that using delta is the easiest way to let squirrel do that.

taooceros:
commit https://github.com/Flow-Launcher/Flow.Launcher/pull/476/commits/f01962f6649a202c85418be90c2f777ee0a1ac6d takes only 2 minutes. Is that caused by using dotnet pack instead of nuget pack?

taooceros:
The last commit related to #479.

jjw24:
> The last commit related to #479.

i am going to pull this commit out and do it as a seperate pr because this pr will need more testing and i want 1.8.0 out soon :)

taooceros:
> > The last commit related to #479.
> 
> i am going to pull this commit out and do it as a seperate pr because this pr will need more testing and i want 1.8.0 out soon :) 

Sure, the last commit need to be changed a bit for CI due to hardcoding issue, and I have only got it run correctly in my computer.

taooceros:
I understand why sometimes script will run extremely slow in my computer. It's because once we have published, there will be much more dll to have to be recursively searched, so it won't affect the CI since it will only do it once.

jjw24:
@taooceros can you merge dev back in this pr please :)

taooceros:
> @taooceros can you merge dev back in this pr please :)

Sure, will do it tonight.

jjw24:
> > @taooceros can you merge dev back in this pr please :)
> 
> Sure, will do it tonight.

All good not in a hurry, whenever

JohnTheGr8:
Maybe close this? Doesn't seem to add any value... 🤔 

I am a bit reluctant about switching from `nuget pack` to `dotnet pack` because I know I had an issue with `dotnet pack` back then but I can't recall exactly what it was...

PR #477, opened 2021-06-17
dev branch: switch to python 3.8.9 embeddable

jjw24:
switch to using python 3.8.9 embeddable

1. prompt user if want to manually set python path or use Python Embeddable 3.8.9
2. if user chooses manually set, dialog box opens for selecting the folder
3. if chooses to let flow install python flow will download and extract the Python Embeddable and set it as the python path for use on python plugins

### DIFF
--- a/Flow.Launcher.Core/Flow.Launcher.Core.csproj
+++ b/Flow.Launcher.Core/Flow.Launcher.Core.csproj
@@ -53,7 +53,7 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Droplex" Version="1.2.0" />
+    <PackageReference Include="Droplex" Version="1.3.0" />
     <PackageReference Include="FSharp.Core" Version="4.7.1" />
     <PackageReference Include="squirrel.windows" Version="1.5.2" />
   </ItemGroup>
--- a/Flow.Launcher.Core/Flow.Launcher.Core.csproj
+++ b/Flow.Launcher.Core/Flow.Launcher.Core.csproj
@@ -53,7 +53,7 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Droplex" Version="1.3.0" />
+    <PackageReference Include="Droplex" Version="1.3.1" />
     <PackageReference Include="FSharp.Core" Version="4.7.1" />
     <PackageReference Include="squirrel.windows" Version="1.5.2" />
   </ItemGroup>
--- a/Flow.Launcher.Core/Plugin/PluginsLoader.cs
+++ b/Flow.Launcher.Core/Plugin/PluginsLoader.cs
@@ -209,7 +209,8 @@ namespace Flow.Launcher.Core.Plugin
                 }
                 else
                 {
-                    DroplexPackage.Drop(App.python3_9_1).Wait();
+                    // Python 3.8.9 is used for Windows 7 compatibility
+                    DroplexPackage.Drop(App.python_3_8_9_embeddable, Path.Combine(DataLocation.DataDirectory(), "Python Embeddable")).Wait();
 
                     var installedPythonDirectory =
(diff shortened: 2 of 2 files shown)


jjw24:
@taooceros @Zeroto521 there is no need to use where.exe or search env PATH to find the python path right? Either set it to the embeddable python executable or ask the users to point it to a python executable themselves.

https://github.com/Flow-Launcher/Flow.Launcher/blob/4909f582254c53a6901dc4bf533c4e66abcf07af/Flow.Launcher.Core/Plugin/PluginsLoader.cs#L124-L147

taooceros:
> @taooceros @Zeroto521 there is no need to use where.exe or search env PATH to find the python path right? Either set it to the embeddable python executable or ask the users to point it to a python executable themselves.
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/4909f582254c53a6901dc4bf533c4e66abcf07af/Flow.Launcher.Core/Plugin/PluginsLoader.cs#L124-L147

We can keep it, unless portable is set. If portable is set, I think it would be better to let user choose whether to use the embeddable python.

jjw24:
This doesn't prompt user at all, just silently assigns if paths has python. I think we remove it so it prompts user to download portable or chooses to  set it themselves. Essentially just two options

taooceros:
> This doesn't prompt user at all, just silently assigns if paths has python. I think we remove it so it prompts user to download portable or chooses to set it themselves. Essentially just two options

ok, got you. Maybe we can have another option to detect python path automatically?

jjw24:
From a quick glance It's just grabbing the python path from PATH, there is no need to do this though, because if the user chooses to set it themselves they will know where to set it, and it may not necessarily be in path, it could be their own python env

jjw24:
> > This doesn't prompt user at all, just silently assigns if paths has python. I think we remove it so it prompts user to download portable or chooses to set it themselves. Essentially just two options
> 
> ok, got you. Maybe we can have another option to detect python path automatically?

Well my perspective, no need to do this, because python installation could be anywhere, no point in guessing where it could be.

Zeroto521:
I need some time to do this, I would keep watching this.

jjw24:
I can get it ready, then we can test it out, if it's more intuitive to use the old code we can always add it back in

jjw24:
https://ci.appveyor.com/api/buildjobs/ncb0jvg5cgy9ogj7/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.2.exe

jjw24:
https://ci.appveyor.com/api/buildjobs/mnnlfrmc5u21egol/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.2.exe

jjw24:
https://ci.appveyor.com/api/buildjobs/50ksa6gaei14mkff/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.2.exe

pc223:
> https://ci.appveyor.com/api/buildjobs/50ksa6gaei14mkff/artifacts/Output%2FPackages%2FFlow-Launcher-v1.7.2.exe

I just quick test this build, working pretty well, I tested with 2 plugins, `Flow.Launcher.Plugin.Currency`, `Flow.Launcher.Plugin.Timestamp` (Windows 10 21H1)

Python is downloaded silently, without a progress bar seems a bit awkward in terms of UX, but no big deal 😄 

Python folder is `Python Embeddable`, personally, I try to avoid spaces in paths as much as possible, this might causing bugs in the future if devs don't aware of it. **Maybe drop the space?**

deefrawley (approved):
Tested, works with my CC plugin

jjw24:
>Python is downloaded silently, without a progress bar seems a bit awkward in terms of UX, but no big deal 😄

yeah true, we can add it later, we dont have a progress bar for anything except for the query window.

> Python folder is Python Embeddable, personally, I try to avoid spaces in paths as much as possible, this might causing bugs in the future if devs don't aware of it. Maybe drop the space?

yeah makes sense, will do this in another pr

PR #470, opened 2021-06-11
Optimize Exception Handling and Log

taooceros:
1. Introduce Demystifier library to make stacktrace more readable. https://github.com/benaadams/Ben.Demystifier
2. Make QueryResult Method async void to allow exception thrown, which will trigger the report window on runtime.
3. Use ExceptionDispatchInfo in Log.Exception to preserve stacktrace
4. Make NLog log in async, and make Log.Exception not implemented synchronously.
5. Use NLog's exception layout renderer to render exception.
6. Use NLog's OutputDebugStringTarget to print to debug window
7. Change layout to `${logger}->${time}|${level}|${message}` (need discussion)    (Exception layout is similar but too more new line is added before and after Exception block)
    - [x] `PluginManager->21:40:16.0845|Debug|"QueryForPluginAsync"|"Cost for System Commands <14ms>"|`
8. Replace all "prefix|UnPrefix" message to call overload that allow a className and methodName
9. Use Conditional("DEBUG") for Debug method in logger.

The Exception block has been simplified to only Exception.ToString, because that will include message, Exception Type, and Stacktrace. I don't believe extra HResult or TargetSite will help us debugging.

TODO:
- [ ] Unify Program Logger

Some Discussion
- [x] Do we want every exception thrown by plugin to report in a report window? (if not, do we want a message to notice user?)

Future (need discussion)
- [ ] Make Logging system not static class but with per instance logger (which allows NLog to catch callsite)

PR #480, opened 2021-06-20
Handle incompatible plugins from major upgrade

jjw24:
@Flow-Launcher/team thinking of how to handle incompatible plugins after a major upgrade. This is brought about by 1.8.0, but we would likely hit this again in the near future. This is for situations where it is not feasible to provide backwards compatibility, i.e. Everything plugin.

When a plugin is incompatible, I think we can do:

1. mark the plugins in the code that will be incompatible, this is for the dev making the major upgrade to do.
2. either provide a new plugin version release that these incompatible plugins have,
3. or disable these plugins
4. if a plugin has a new compatible version release then download, extract the new plugin and make it ready for use

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -79,6 +79,9 @@ namespace Flow.Launcher.Core.Plugin
         /// <param name="settings"></param>
         public static void LoadPlugins(PluginsSettings settings)
         {
+            if (PluginUpgrader.ForceUpgradeRequired(settings))
+                PluginUpgrader.ForceUpgrade(settings);
+
             _metadatas = PluginConfig.Parse(Directories);
             Settings = settings;
             Settings.UpdatePluginSettings(_metadatas);
--- a/Flow.Launcher.Core/Plugin/PluginUpgrader.cs
+++ b/Flow.Launcher.Core/Plugin/PluginUpgrader.cs
@@ -1,0 +1,41 @@
+﻿using Flow.Launcher.Infrastructure.UserSettings;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Flow.Launcher.Core.Plugin
+{
+    public static class PluginUpgrader
+    {
+        /// <summary>
+        /// Indicates these plugins need to have version greater than currently recorded
+        /// </summary>
+        private static readonly Dictionary<string, PluginVersion> PluginsUpgradeRequiredVersion = new()
+        {
+            { "D2D2C23B084D411DB66FE0C79D6C2A6E", new PluginVersion { ForceUpgradeRequiredVersion = "1.4.9" }  }
+        };
+
+        internal static void ForceUpgrade(PluginsSettings settings)
+        {
+            // if new version is added for PluginVersion.NewVersionDownloadUrl=> remove older plugin, download & extract
+            // if PluginVersion.NewVersionDownloadUrl does not contain new version url, disable plugin and prompt to notify user
+        }
+
+        internal static bool ForceUpgradeRequired(PluginsSettings settings)
+        {
+            return 
+                settings.Plugins
+                .Any(t1 => PluginsUpgradeRequiredVersion
+                            .Any(x => x.Key == t1.Key && x.Value.ForceUpgradeRequiredVersion == t1.Value.Version));
+        }
+    }
+
+    internal class PluginVersion
+    {
+        internal string ForceUpgradeRequiredVersion;
+
+        internal string NewVersionDownloadUrl = string.Empty;
+    }
+}


pc223:
To further safe guard againts future bugs :bug: and incompatibility, I think Flow should tell users **where to get help** when errors happen in general :smile:

Like, "BOOM! Error: ExceptionBlaBlaBla: Please go to ... to get help".

Maybe with some contact info like:
- Github issues
- Discord
- Maintainer email?
- Project email? (googlegroups? - Normal users tend to prefer emails more than Github :thinking:)

Maybe some kind of instructions to get logs file/write bug report.

If possible, some kind of auto-capture context data, to help with postmortem debugging. I'm pretty sure C# has some kind of debugging tools like that. (`serilog`?)

taooceros:
> To further safe guard againts future bugs 🐛 and incompatibility, I think Flow should tell users **where to get help** when errors happen in general 😄
> 
> Like, "BOOM! Error: ExceptionBlaBlaBla: Please go to ... to get help".
> 
> Maybe with some contact info like:
> 
> * Github issues
> * Discord
> * Maintainer email?
> * Project email? (googlegroups? - Normal users tend to prefer emails more than Github 🤔)
> 
> Maybe some kind of instructions to get logs file/write bug report.
> 
> If possible, some kind of auto-capture context data, to help with postmortem debugging. I'm pretty sure C# has some kind of debugging tools like that. (`serilog`?)

Flow already have that if a outside bug is triggered. However, the current build will only trigger that if exception other than plugins exception is thrown. #470 is trying to make that perform better.

taooceros:
> Project email? (googlegroups? - Normal users tend to prefer emails more than Github 🤔)

Regarding this, there's an interesting project doing this. https://fire.fundersclub.com/#SetUp

taooceros:
A few thought
1. PluginsManager need to be changed to detect the latest supported plugin version for a specific Flow version
2. Plugin developer will also need to mention supported Flow version (probably in plugin.json?).
3. PluginsManifest need to be changed to include more release version and the supported version information. Probably, users may be allowed to choose which version to download.
4. I don't think these information should be included in Flow like this commit except for the current everything plugin. They should be included in Plugins' repo.

jjw24:
> > Project email? (googlegroups? - Normal users tend to prefer emails more than Github 🤔)
> 
> Regarding this, there's an interesting project doing this. https://fire.fundersclub.com/#SetUp

this is good, we should set up this

jjw24:
> A few thought
> 
> 1. PluginsManager need to be changed to detect the latest supported plugin version for a specific Flow version
> 2. Plugin developer will also need to mention supported Flow version (probably in plugin.json?).
> 3. PluginsManifest need to be changed to include more release version and the supported version information. Probably, users may be allowed to choose which version to download.

is 1-3 to solve downloading latest plugins that old version of flow does not support?

jjw24:
Alternative we can actually add backwards compatibility by re-adding the old methods and classes and updating the new class+ methods to a different name: https://github.com/Flow-Launcher/Flow.Launcher/pull/481

Technicality wise, what do you think? @taooceros

pc223:
> > > Project email? (googlegroups? - Normal users tend to prefer emails more than Github 🤔)
> > 
> > 
> > Regarding this, there's an interesting project doing this. https://fire.fundersclub.com/#SetUp
> 
> this is good, we should set up this

Should tell plugin-dev to provide email too. Maybe put the email field in `plugin.json`? (make sure they aware that this email is going to be public)

I surprise that Github doesn't have built-in support for email 😮 

> Alternative we can actually add backwards compatibility by re-adding the old methods and classes and updating the new class+ methods to a different name: #481
> 
> Technicality wise, what do you think? @taooceros

I prefer this approach, But, have to make plugin-dev knows that they using old API, and Flow will end support for that API in the future, so that plugin-dev knows to upgrade to the new API.

I prefer a Msgbox warning ⚠️ when Flow runs, so that even normal-users aware of the plugin is using an old API, maybe users should consider using alternative plugins. (Kind of pressure plugin-dev to update 😅 )

And a lot of the time, plugin-dev _is_ normal-users too => don't read the new Flow-docs => don't aware of changes of API => Flow have to force them a bit 😅

taooceros:
> Alternative we can actually add backwards compatibility by re-adding the old methods and classes and updating the new class+ methods to a different name: #481
> 
> Technicality wise, what do you think? @taooceros

I think currently no old methods is removed, so no other plugins are currently incompatible (at least those plugin in my computer). Everything plugin is not compatible because of the reference to Old_Infrasturature.dll.

taooceros:
> I surprise that Github doesn't have built-in support for email 😮

I guess they do this because they would like people to at least register an account. Github do allows people to respond issue with email.

> I prefer this approach, But, have to make plugin-dev knows that they using old API, and Flow will end support for that API in the future, so that plugin-dev knows to upgrade to the new API.

> I prefer a Msgbox warning ⚠️ when Flow runs, so that even normal-users aware of the plugin is using an old API, maybe users should consider using alternative plugins. (Kind of pressure plugin-dev to update 😅 )

> And a lot of the time, plugin-dev is normal-users too => don't read the new Flow-docs => don't aware of changes of API => Flow have to force them a bit 😅

Good idea

taooceros:
> is 1-3 to solve downloading latest plugins that old version of flow does not support?

Is for downloading latest plugins that old version of flow does support (it would be more possible that a version of plugin is not supported by older version of flow comparing to a newer version of flow).

jjw24:
This pr will not go ahead. New changes are in https://github.com/Flow-Launcher/Flow.Launcher/pull/481, which adds backwards compatibility via restoring previous class and interface used by Everything plugin.

PR #481, opened 2021-06-21
Add backwards compatibility for Everything plugin v1.4.9

jjw24:
Restored old class and interface for backwards compatibility with Everything plugin v1.4.9 in v1.8.0 Flow release


taooceros:
Why change that to ISettingSavable? I remember in #427, we have moved all saving for JsonSetting in Flow, so ISavable is for other things that needs to be saved.

jjw24:
because the old interface from Everything needs ISavable interface. Do we want to change it from ISettingsSavable to another name to reflect the purpose for other things that needs to be saved?

jjw24:
also because we changed JsonStrorage class name to JsonStorage

taooceros:
Hmmm, but currently, ISavable is not be used as only saving setting. We add a auto json settings save to ISavable, so plugins don't need to implement ISavable when they want to save setting. They only implement ISavable if they want to save other things (such as Program plugin's cache).

jjw24:
i may have made the code a little messy but the intention is ISavable interface because is used by Everything's infra dll, we need to keep it for backwards compatibility. Therefore we cant have another ISavable. We can rename the ISavable to something else.

jjw24:
Also the dll expects ISavable to be in Flow.Launcher.Infrastructure/Storage/ btw

taooceros:
Maybe not that hard. Let ISavable (in infrasturature) implement ISettingSavable shall work. No need for manually check for everything plugin.

taooceros:
@jjw24 I still consider ISavable is better than ISettingsSavable in this case. Since ISavable (Infrastructure) is only be used in everything plugin, keeping both interface as the same name won't be confused, and we can mark ISavable in Infrastructure as Obsolete.

jjw24:
all done @taooceros

taooceros (approved):

PR #484, opened 2021-06-22
[UI] Fix `Settings-window` doesn't show back up when minimized in taskbar by user

pc223:
Bug reproduce steps:
1. Right click on Flow icon in Notification Area, click settings
2. Minimize Settings-window (or Alt+Tab to different window)
3. Right click on Flow icon in Notification Area, click settings
4. Settings-window will not show back up 

I'm not really sure why this solution works tho 😅 

Maybe `.Show()` fails when `window.WindowState == Minimized` (not `Normal`)
Not sure why need .Activate() too

Ref: https://stackoverflow.com/a/59719760/4230390

Please delete unnecessary comments if this is too obvious to C# devs 😅

### DIFF
--- a/Flow.Launcher/Helper/SingletonWindowOpener.cs
+++ b/Flow.Launcher/Helper/SingletonWindowOpener.cs
@@ -11,7 +11,19 @@ namespace Flow.Launcher.Helper
             var window = Application.Current.Windows.OfType<Window>().FirstOrDefault(x => x.GetType() == typeof(T))
                          ?? (T)Activator.CreateInstance(typeof(T), args);
             Application.Current.MainWindow.Hide();
+            
+            
+            // Fix UI bug
+            // If only use `window.Show()`, Settings-window doesn't show when minimized in taskbar 
+            // Not sure why this works tho
+            // Probably because, when `.Show()` failed, `window.WindowState == Minimized` (not `Normal`) 
+            // https://stackoverflow.com/a/59719760/4230390
+            // Not sure why need .Activate() too
+            window.WindowState = WindowState.Normal; 
             window.Show();
+            window.Activate();
+            
+            
             window.Focus();
 
             return (T)window;
--- a/Flow.Launcher/Helper/SingletonWindowOpener.cs
+++ b/Flow.Launcher/Helper/SingletonWindowOpener.cs
@@ -12,7 +12,6 @@ namespace Flow.Launcher.Helper
                          ?? (T)Activator.CreateInstance(typeof(T), args);
             Application.Current.MainWindow.Hide();
             
-            
             // Fix UI bug
             // Add `window.WindowState = WindowState.Normal`
             // If only use `window.Show()`, Settings-window doesn't show when minimized in taskbar 
@@ -21,7 +20,6 @@ namespace Flow.Launcher.Helper
             // https://stackoverflow.com/a/59719760/4230390
             window.WindowState = WindowState.Normal; 
             window.Show();
-            
             
             window.Focus();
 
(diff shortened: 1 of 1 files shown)


taooceros:
https://docs.microsoft.com/en-us/dotnet/api/system.windows.window.activate?view=net-5.0

refer this, it will call a SetForeGround invoke (which is a pinvoke I learnt in WindowWalker plugin), so it will set specific window.

Question, do we need to set the windows' state? or calling Activate will be fine?

pc223:
> Question, do we need to set the windows' state? or calling Activate will be fine?

Lemme quick test that, see if it works without Activate()

pc223:
> > Question, do we need to set the windows' state? or calling Activate will be fine?
> 
> Lemme quick test that, see if it works without Activate()

Yes, seems that without Activate() still works, I think we can remove that line

pc223:
If we don't `window.WindowState = WindowState.Normal`, and only use `Activate()`, it changes color of  the Settings-window in taskbar (activated?) but **not** open Settings-window. The WindowState seems is the key 🤔

taooceros:
> If we don't `window.WindowState = WindowState.Normal`, and only use `Activate()`, then it changes color of the Settings-window in taskbar (activated?) but **not** open Settings-window. The WindowState seems is the key 🤔

Interesting, I finally understand why sometimes windowwalker won't actually show up the window.

taooceros:
Could you change the comment for Activate()?

pc223:
Done!

taooceros:
> Done!

Ha🤔 Does Activate don't need to be included? Could you give it a quick test?
I just mean that it would be better to include more detailed interpretation why Activate() is included though lol.

pc223:
> > Done!
> 
> Ha🤔 Does Activate don't need to be included? Could you give it a quick test?
> I just mean that it would be better to include more detailed interpretation why Activate() is included though lol.

Tested, still works without Activate, I'm to not sure why 😅  I tried remove `.Focus()`, still works 🤔

taooceros:
> Tested, still works without Activate, I'm to not sure why 😅 I tried remove .Focus(), still works 🤔

If setting window is not minimized, but simply hided by other window, it may not work I guess. Though, I am not sure whether Focus can replace the need for Activate.

taooceros:
Could you give that a quick test?

pc223:
> > Tested, still works without Activate, I'm to not sure why 😅 I tried remove .Focus(), still works 🤔
> 
> If setting window is not minimized, but simply hided by other window, it may not work I guess. Though, I am not sure whether Focus can replace the need for Activate.

Hm... 🤔 could you give me like a list of test cases? Like 

> Could you give that a quick test?

Just tested, you're right! Remove `.Focus()` and only use `Show()`, Settings will be hidden by other windows. 

But, still remove `.Focus()`, but add `Activate()`, it works, not hidden by other windows.

Seems that we can use either `Focus()` or `Activate()`.

Any more test case that I can run? 😄

taooceros:
I guess that's enough. Thank you for testing! If Focus work, we can remove Activate().

Would you please remove one extra empty line? I think one is fine for separating the logic.

Then, I will approve, and merge later to let somebody else to review if they want.

Thank you for the fix.

pc223:
> I guess that's enough. Thank you for testing! If Focus work, we can remove Activate().
> 
> Would you please remove one extra empty line? I think one is fine for separating the logic.
> 
> Then, I will approve, and merge later to let somebody else to review if they want.
> 
> Thank you for the fix.

Like this? I tend to go overboard on whitespaces haha

image

taooceros:
Yes, thank you!

taooceros (approved):

PR #486, opened 2021-06-23
Add github dependency bot

taooceros:
Add automatic check for dependency update

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "nuget" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
   - package-ecosystem: "nuget" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
-      interval: "daily"
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,5 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+    ignore:
+      - dependency-name: "squirrel-windows"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,7 @@ updates:
       interval: "weekly"
     ignore:
       - dependency-name: "squirrel-windows"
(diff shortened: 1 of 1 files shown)


jjw24:
Upping the versions will require testing, would it not be better to test while you are choosing to upgrade the version? 

This will create a lot of noise in the pr section

taooceros:
> Upping the versions will require testing, would it not be better to test while you are choosing to upgrade the version?
> 
> This will create a lot of noise in the pr section

We can make that check weekly/monthly, which means it will only create a version update pr once a week/month. In most situation, we don't update version because we are lazy to update hhh🤣.

jjw24:
Isn't this going to run through and create or for every version upgrade? This is going to create a fair bit prs.

Again you have to test to do the upgrade, which I would prefer effort put into enhancements instead and upgrade the required packages if needed

taooceros:
> Isn't this going to run through and create a pr for every version upgrade? This is going to create a fair bit prs.

I don't think so. It shall create one pr once a week for all update if it has detected an update. 

> Again you have to test to do the upgrade, which I would prefer effort put into enhancements instead and upgrade the required packages if needed

Unless there's a breaking change (like squirrel), I don't think updating the a dependency will require a lot of testing.

jjw24:
We have a lot of old packages, it's going to upgrade a lot of them.

Can you check how many updates it will create for please

taooceros:
I will take a try in my fork first.

taooceros:
image
It will behave like this. Each update with an pull request, and it will list the release note and commit difference for each update.

jjw24:
only 5 packages need updating?

taooceros:
> only 5 packages need updating?

No, it is because we can limit the count of pr created one time (which is 5)

jjw24:
Would it be better to make an individual branch, upgrade all the non-breaking upgrades on in this branch and test it out before applying the bot?

taooceros:
> Would it be better before applying the bot to make an individual branch, upgrade all the non-breaking upgrades on this branch, test them out and merge into dev first?

Sounds reasonable

taooceros:
@jjw24 shall we merge the dependabot?

jjw24:
We have a lot of prs in the pipeline atm, will this change add a lot more? If so can we bump all NuGet packages on a seperate branch, test everything ok and merge in before merging this one so we start with a good baseline.

Garulf:
You can limit the maximum PR's dependabot is able to create with: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

taooceros:
> You can limit the maximum PR's dependabot is able to create with: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

the default is 5, which I think is quite reasonable

Garulf:
Seems reasonable to me. Our tests will still run on each PR so in theory we get updated packages for free.

jjw24:
Ok let's merge this after 1.9.5

jjw24 (approved):

PR #495, opened 2021-06-27
Remove top margin of General tab (Settings)

pc223:
Why?
Seems cleaner and similar UX to other apps/website

Before:

After:

### DIFF
--- a/Flow.Launcher/SettingWindow.xaml
+++ b/Flow.Launcher/SettingWindow.xaml
@@ -36,8 +36,8 @@
 
     <TabControl Height="auto" SelectedIndex="0">
         <TabItem Header="{DynamicResource general}">
-            <ScrollViewer ui:ScrollViewerHelper.AutoHideScrollBars="{Binding AutoHideScrollBar, Mode=OneWay}" Margin="60,30,0,0">
-                <StackPanel Orientation="Vertical">
+            <ScrollViewer ui:ScrollViewerHelper.AutoHideScrollBars="{Binding AutoHideScrollBar, Mode=OneWay}" Margin="60,0,0,0">
+                <StackPanel Orientation="Vertical" Margin="0,30,0,0">
                     <ui:ToggleSwitch Margin="10" IsOn="{Binding PortableMode}">
                         <TextBlock Text="{DynamicResource portableMode}" />
                     </ui:ToggleSwitch>


taooceros (approved):

taooceros:
@pc223 's two changes seem to make our setting panel not appears only have one page. Do we still want to hide scroller by default? @Flow-Launcher/Team

pc223:
> @pc223 's two changes seem to make our setting panel not appears only have one page. Do we still want to hide scroller by default? @Flow-Launcher/team

I too prefer to remove the option auto-hide scrollbar in the settings, and set the default to hide scrollbar, I think with some visual trick, we can make users aware of the scrolling space 😄 Provide options for these kinds of thing is seems a bit overkill IMO

deefrawley:
I disagree - I think the scroll bar should be visible by default and large enough to show users there are more options below.

pc223:
> I disagree - I think the scroll bar should be visible by default and large enough to show users there are more options below.

The Plugin tab and Theme tab still hide the scrollbar by default, and Windows 10 settings and Start menu seems to hide scrollbar too. It's pretty common style in Windows 10.

I think users get confuse because 2 margins that cause users to think it's unscrollable space :smile:

deefrawley:
The plugin and theme tabs have the left scroll area which aren't hidden. All I know is I used Wox/Flow for a while before I realised you could scroll in the settings tab (even then it was only when it came up on Discord). We have to have a clear way to show users to scroll for more settings.

pc223:
> The plugin and theme tabs have the left scroll area which aren't hidden. All I know is I used Wox/Flow for a while before I realised you could scroll in the settings tab (even then it was only when it came up on Discord). We have to have a clear way to show users to scroll for more settings.

Yes, before the removing of 2 top and bottom margins, it looks like the settings is unscrollable. But if you remove those margins, it look really clear that this is scrollable space.

Before it looks like this



Now: (looks really clear, this is scrollable space)

deefrawley:
I see what you mean, it certainly is an improvement. Let's implement and see what the feedback is.

jjw24:
> The Plugin tab and Theme tab still hide the scrollbar by default, and Windows 10 settings and Start menu seems to hide scrollbar too. It's pretty common style in Windows 10.

Theme and plugin tabs should mirror the main settings tab in terms hiding or not hiding the scrollbar. By default the scrollbar is shown, this is not the case?

pc223:
> > The Plugin tab and Theme tab still hide the scrollbar by default, and Windows 10 settings and Start menu seems to hide scrollbar too. It's pretty common style in Windows 10.
> 
> Theme and plugin tabs should mirror the main settings tab in terms hiding or not hiding the scrollbar. By default the scrollbar is shown, this is not the case?

Oh yeah, apparently it mirrors the auto-hide scroll bar settings, sorry, I never turn off auto-hide so I never see them in Theme or Plugins haha 😅

PR #498, opened 2021-06-27
auto disable shadow when blur is enabled and make query window width dynamic

### LINKED ISSUE #478: query window when shadow and blur effects enabled, extends outside of the window
currently in dev branch as at commit [2e412a00232998546f987c034adc753c59c36b06](https://github.com/Flow-Launcher/Flow.Launcher/tree/2e412a00232998546f987c034adc753c59c36b06), query window when blur enabled via the theme, extends outside of the window

repro:

1. enable shadow
2. switch to BlurBlack Darker theme
3. restart flow
4. the edges where there should be shadow becomes just blurred:
image

jjw24:
- shadow does not work when blur is enabled via the theme file. This automatically disables shadow when blur is enabled
close #478 

- make query window width dynamic

### DIFF
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -30,10 +30,6 @@ namespace Flow.Launcher.Core.Resource
 
         public bool BlurEnabled { get; set; }
 
-        // due to blur only enabled when main window is loaded for the first time, this is used to track the blur status
-        // when the window is first loaded and prompt user that a restart is required for blur to take effect. 
-        public bool BlurEnabledOnloaded { get; set; }
-
         public Theme()
         {
             _themeDirectories.Add(DirectoryPath);
@@ -98,6 +94,8 @@ namespace Flow.Launcher.Core.Resource
 
                 if (Settings.UseDropShadowEffect && !BlurEnabled)
                     AddDropShadowEffectToCurrentTheme();
+
+                SetBlurForWindow();
             }
             catch (DirectoryNotFoundException e)
             {
@@ -356,6 +354,8 @@ namespace Flow.Launcher.Core.Resource
         private void SetWindowAccent(Window w, AccentState state)
         {
             var windowHelper = new WindowInteropHelper(w);
+            w.Width = 750;
+            windowHelper.EnsureHandle();
             var accent = new AccentPolicy { AccentState = state };
             var accentStructSize = Marshal.SizeOf(accent);
 
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -28,6 +28,8 @@ namespace Flow.Launcher.Core.Resource
         private string DirectoryPath => Path.Combine(Constant.ProgramDirectory, Folder);
         private string UserDirectoryPath => Path.Combine(DataLocation.DataDirectory(), Folder);
 
+        public bool BlurEnabled { get; set; }
+
         public Theme()
         {
             _themeDirectories.Add(DirectoryPath);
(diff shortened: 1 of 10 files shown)


taooceros:
Why not enable Blur when blur theme is selected instead of prompting to restart?

jjw24:
Can't enable it with the current implementation of blur, you can call the set blur function but it comes out with wider blur width, hence the original comment about it being in mainwindow's on_loaded function call

taooceros:
> Can't enable it with the current implementation of blur, you can call the set blur function but it comes out with wider blur width, hence the original comment about it being in mainwindow's on_loaded function call

I know, but the extra boarder will only be enabled when shadow effect is on. I mean if that's not on, we can enable the blur effect so that user won't see a blur theme without blur effect.

jjw24:
cant, blur can only be enabled by mainwindow's onloaded, if you force it else where you get this

image

taooceros:
I take a debug check. It is behaving weird if we set blur after onload because the width of the main window is changed to 1440 instead of expected 750 with unknown reason. If we change it back to 750 (or the dynamic width setting in #374), it seems work correctly.

jjw24:
oh right, good pick up thank you.

taooceros:
```c#
private void SetWindowAccent(Window w, AccentState state)
        {
            var windowHelper = new WindowInteropHelper(w);
            w.Width = 750;
            windowHelper.EnsureHandle();
            var accent = new AccentPolicy { AccentState = state };
            var accentStructSize = Marshal.SizeOf(accent);

            var accentPtr = Marshal.AllocHGlobal(accentStructSize);
            Marshal.StructureToPtr(accent, accentPtr, false);

            var data = new WindowCompositionAttributeData
            {
                Attribute = WindowCompositionAttribute.WCA_ACCENT_POLICY,
                SizeOfData = accentStructSize,
                Data = accentPtr
            };

            SetWindowCompositionAttribute(windowHelper.Handle, ref data);

            Marshal.FreeHGlobal(accentPtr);
        }
```

This will work, because the setting window will hide the main window, so we need to ensure the mainwindow is rendered successfully before assigning it with a shadow effect.

jjw24:
Let's make that width dynamically passed in from theme xmal, should work well since we are making width dynamic in the other pr as well

jjw24:
all done

Flow.Launcher.Core/Resource/Theme.cs
```diff
@@ -320,35 +343,39 @@ private enum WindowCompositionAttribute
...
         public void SetBlurForWindow()
         {
+            if (BlurEnabled)
+            {
+                SetWindowAccent(Application.Current.MainWindow, AccentState.ACCENT_ENABLE_BLURBEHIND);
+            }
+            else
+            {
+                SetWindowAccent(Application.Current.MainWindow, AccentState.ACCENT_DISABLED);
+            }
```
taooceros, line 353: Maybe only change this if blur state is changed?

jjw24: in the original code this state is always set though via on_loaded everytime when flow is started, not sure if we really need to manage this

jjw24: @taooceros did you want to show me what you mean?

taooceros: Oh sorry I forget to reply. I mean only call this method when we switch from Blur to non-blur or non-blur to Blur. Also maybe remove the call of this method in MainWindow.OnLoaded and move that to ChangeTheme

jjw24: > call this method when we switch from Blur to non-blur or non-blur to Blur

there's not really a good way to tell what state blur was previously. is this just so the call is not done multiple times when user cycles through themes? i tested current way there hasnt been any issues.

>  remove the call of this method in MainWindow.OnLoaded and move that to ChangeTheme

this is already removed.

taooceros (approved):

PR #500, opened 2021-06-27
dev branch: remove createdump.exe

### LINKED ISSUE #479: dev branch: 1.8.0 installation creates a .net shortcut
this shortcut is created everytime 1.8.0 is installed:
image

@taooceros any ideas?

jjw24:
close #479

### DIFF
--- a/Scripts/post_build.ps1
+++ b/Scripts/post_build.ps1
@@ -50,10 +50,8 @@ function Delete-Unused ($path, $config) {
 function Remove-CreateDumpExe ($path, $config) {
     $target = "$path\Output\$config"
 
-    $depjson = Get-Content $target\Flow.Launcher.deps.json -raw |ConvertFrom-Json -depth 32
-    $depjson.targets.'.NETCoreApp,Version=v5.0/win-x64'.'runtimepack.Microsoft.NETCore.App.Runtime.win-x64/5.0.6'.native.PSObject.Properties.Remove("createdump.exe")
-    $depjson|ConvertTo-Json -Depth 32|Out-File $target\Flow.Launcher.deps.json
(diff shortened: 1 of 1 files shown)


jjw24:
@taooceros  getting this error in event viewer and flow does not start after installation:

image

taooceros:
We need to check the dotnet version for CI, which may not be 5.0.6. I am not sure whether there's better way of handling the json modification with powershell.

taooceros:
I change it to regex match, which is simpler and more reliable since it won't depend on the version number.

jjw24:
from the last commit, the shortcut is created again. also issue with parsing the json file:

image

taooceros:
> from the last commit, the shortcut is created again. also issue with parsing the json file:
> 
> image

hmm yes I cannot get it run correctly. By the way, where do you get the log? I am not sure how to debug this kind of issues.

jjw24:
I got it from event viewer, that's usually the place windows log when not caught by apps own exception handler.

Did it work when you manually changed it? If so can u let me know the steps please

taooceros:
> I got it from event viewer, that's usually the place windows log when not caught by apps own exception handler.
> 
> Did it work when you manually changed it? If so can u let me know the steps please

Thanks. Let me take a check.

taooceros:
It will work if I manually change the dep.json. Let me take a comparison of the two file.

taooceros:
I think I know the reason..... powershell will output the file as UTF16, but the dep.json is supposed to be UTF8😣

taooceros:
I think I know the reason..... powershell will output the file as UTF16, but the dep.json is supposed to be UTF8😣

jjw24:
just tested installing, is good 👌 thank you

taooceros (approved):

PR #501, opened 2021-06-28
Update README.md

jjw24:
Consolidate sections

Exclude the Q&A issue template now that we are using Discussions


JohnTheGr8:
Sections like "Plugins" and "Documentation" are just one sentence with a link to another page... perhaps it would be nicer to link directly from the menu you've introduced here?

Also, should we mention the github discussions page in "Question/Suggestion" ?

jjw24:
Yeah need to add discussions section. Plugins may get expanded to include a link to how to develop plugins and another to current plugins available. Maybe documentation can be directly linked

deefrawley (approved):

Issue #55, opened 2020-05-18
Change the query loading progress bar


jjw24:
Change the magnifying glass icon to spinning progress bar when query runs longer, and remove existing progress bar line

pc223:
I'm experimenting with a non-animating-progress-bar, like just solid line, no fading, no delay. If plugins searching then a thick blue bar will always show immediately (contrast to waiting for 200ms like normal)

![image](https://user-images.githubusercontent.com/10551242/123676056-a408b100-d86d-11eb-8d64-43af1548a3af.png)

Feel much better, like you can feel exactly what's going on => much less "wrong-enter" (wrongly launch the program because of plugins searching taking too long and they update right before you type Enter)

jjw24:
you can throw a pr up for us to try this, but i like spinning wheels haha 😝

pc223:
Oh I would love to have spinning wheel haha, but this blue bar cost me like 10min, spinner probably takes me at least couple days 😂

This kind of like a experiment, I want to see what is the effect of removing the 200ms delay and make progress bar extremely obvious. The result seems make the UX much better, feel snappier, more reliable 😊

Maybe I'll make a PR for you guys to try it out

PR #531, opened 2021-06-29
Make progress bar bigger, solid, and always visible when have any queries running

pc223:
Why:
For experiment, seems better UX

An experiment for https://github.com/Flow-Launcher/Flow.Launcher/issues/55

> This kind of like a experiment, I want to see what is the effect of removing the 200ms delay and make progress bar extremely obvious. The result seems make the UX much better, feel snappier, more reliable 😊

The difference is more obvious when using plugins that have high latency like Explorer or Currency


jjw24:
will take a look after 1.8.0 :)

PR #532, opened 2021-06-29
Exception log partial opt

taooceros:
Part of the commit in #470

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -157,7 +157,7 @@ namespace Flow.Launcher.Core.Plugin
             }
         }
 
-        public static async Task<List<Result>> QueryForPlugin(PluginPair pair, Query query, CancellationToken token)
+        public static async Task<List<Result>> QueryForPluginAsync(PluginPair pair, Query query, CancellationToken token)
         {
             var results = new List<Result>();
             try
@@ -171,7 +171,7 @@ namespace Flow.Launcher.Core.Plugin
 
                 token.ThrowIfCancellationRequested();
                 if (results == null)
-                    return results;
+                    return null;
                 UpdatePluginMetadata(results, metadata, query);
 
                 metadata.QueryCount += 1;
@@ -183,10 +183,6 @@ namespace Flow.Launcher.Core.Plugin
             {
                 // null will be fine since the results will only be added into queue if the token hasn't been cancelled
                 return null;
-            }
-            catch (Exception e)
-            {
-                Log.Exception($"|PluginManager.QueryForPlugin|Exception for plugin <{pair.Metadata.Name}> when query <{query}>", e);
             }
 
             return results;
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -184,10 +184,6 @@ namespace Flow.Launcher.Core.Plugin
                 // null will be fine since the results will only be added into queue if the token hasn't been cancelled
                 return null;
             }
-            catch (Exception e)
-            {
-                Log.Exception("PluginManager", "Exception", e);
-            }
 
             return results;
         }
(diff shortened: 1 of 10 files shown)


Flow.Launcher/ViewModel/ResultsForUpdate.cs
```diff
@@ -8,15 +8,15 @@ namespace Flow.Launcher.ViewModel
 {
     public struct ResultsForUpdate
     {
-        public List<Result> Results { get; }
+        public IReadOnlyList<Result> Results { get; }
```
taooceros, line 11: It has been changed because I add an emptyResult List singleton for null result

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -63,6 +63,11 @@ internal async static Task<List<Result>> ExecuteWindowsIndexSearchAsync(string i
                     }
                 }
             }
+            catch (OperationCanceledException)
+            {
+                // return empty result when cancelled
+                return results;
+            }
```
taooceros, line 70: Previously OperationCanceledException is thrown as normal exception, causing the slowness of logging exception.

Plugins/Flow.Launcher.Plugin.Program/Logger/ProgramLogger.cs
```diff
@@ -20,36 +20,13 @@ internal static class ProgramLogger
...
-            target.FileName = path.Replace(@"\", "/") + "/${shortdate}.txt";
-#if DEBUG
-            var rule = new LoggingRule("*", LogLevel.Debug, target);
-#else
-            var rule = new LoggingRule("*", LogLevel.Error, target);
-#endif
-            configuration.LoggingRules.Add(rule);
-            LogManager.Configuration = configuration;
-        }
-
```
taooceros, line 43: It is removed because it will override the config setting in Log.cs (which is the same in before so no effect is shown).

Flow.Launcher.Infrastructure/Logger/Log.cs
```diff
@@ -23,23 +28,44 @@ static Log()
...
+                FileName = CurrentLogDirectory.Replace(@"\", "/") + "/${shortdate}.txt",
+                Layout = layout
+            };
+
+            var fileTargetASyncWrapper = new AsyncTargetWrapper(fileTarget);
+
+            var debugTarget = new OutputDebugStringTarget
+            {
+                Layout = layout
+            };
```
taooceros, line 49: The target to output information to debug console (not the console target for command line app).

taooceros (commented):
A few comment about the changes.

Hope we won't experience serious merge conflict in the future XD

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -469,6 +482,12 @@ private void QueryResults()
             ProgressBarVisibility = Visibility.Hidden;
             _isQueryRunning = true;
 
+            // Switch to ThreadPool thread
+            await TaskScheduler.Default;
+
+            if (currentCancellationToken.IsCancellationRequested)
+                return;
```
taooceros, line 489: Add for racing issue reported by @pc223

jjw24:
What's tested so far with this

Have you taken it for a spin to ensure there is no degradation or slowness

taooceros:
> What's tested so far with this
> 
> Have you taken it for a spin to ensure there is no degradation or slowness

1. Log file are logged successfully on #470
2. Everything Plugin works properly in #470 
3. Log message can successfully printed on Debug Ouput

Theoretically, there will only be performance boost due to the use of Async Wrapper in Debug mode (debug mode has quite significant parallel logging). The remove of Synchronize Implementation should make logging Exception from multiple thread much faster than before.

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -171,7 +171,7 @@ public static async Task<List<Result>> QueryForPlugin(PluginPair pair, Query que
 
                 token.ThrowIfCancellationRequested();
                 if (results == null)
-                    return results;
```
jjw24, line 174: just curious why return null when results is already null? does this actually change anything?

jjw24 (approved):
thank you

PR #548, opened 2021-07-04
point pluginsmanifest to new api branch

jjw24:
use the [new branch](https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/53) in PluginsManifest per [discussion](https://github.com/Flow-Launcher/Flow.Launcher/discussions/535)

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
@@ -50,7 +50,7 @@ namespace Flow.Launcher.Plugin.PluginsManager
                     IcoPath = "Images\\request.png",
                     Action = _ =>
                     {
-                        // standard UrlSourceCode format in PluginsManifest's plugins.json file: https://github.com/jjw24/WoxDictionary/tree/master
+                        // standard UrlSourceCode format in PluginsManifest's plugins.json file: https://github.com/jjw24/Flow.Launcher.Plugin.Putty/tree/master
                         var link = pluginManifestInfo.UrlSourceCode.StartsWith("https://github.com") 
                                                                 ? pluginManifestInfo.UrlSourceCode.Replace("/tree/master", "/issues/new/choose") 
                                                                 : pluginManifestInfo.UrlSourceCode;
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/Models/PluginsManifest.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/Models/PluginsManifest.cs
@@ -15,7 +15,7 @@ namespace Flow.Launcher.Plugin.PluginsManager.Models
         {
             try
             {
-                await using var jsonStream = await Http.GetStreamAsync("https://raw.githubusercontent.com/Flow-Launcher/Flow.Launcher.PluginsManifest/main/plugins.json")
+                await using var jsonStream = await Http.GetStreamAsync("https://raw.githubusercontent.com/Flow-Launcher/Flow.Launcher.PluginsManifest/plugin_api_v2/plugins.json")
                                  .ConfigureAwait(false);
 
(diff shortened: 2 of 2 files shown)


jjw24:
@taooceros @JohnTheGr8 as per discussion, good to go?

taooceros:
> @taooceros @JohnTheGr8 as per discussion, good to go? 

Have you checked what @Zeroto521 says?

jjw24:
Just saw it, ok let me think about this

JohnTheGr8:
Has anyone tried to see what happens when Flow updates but a previously installed plugin is no longer compatible? Does new flow simply ignore it (not load it)?

jjw24:
what scenario you are talking about ? i have tested dev branch's build with all plugins we currently have, they are all compatible

taooceros:
> Has anyone tried to see what happens when Flow updates but a previously installed plugin is no longer compatible? Does new flow simply ignore it (not load it)?

I think it will pop out a message saying that a specific plugin is unable to load.

jjw24:
> > Has anyone tried to see what happens when Flow updates but a previously installed plugin is no longer compatible? Does new flow simply ignore it (not load it)?
> 
> I think Flow will pop out a message saying that a specific plugin is unable to load, and disable it.

Yeah, if the plugin fails to load, that message will pop out. If flow is .net core and plugin is .net 5, the same message will pop up

JohnTheGr8:
> Yeah, if the plugin fails to load, that message will pop up. If flow is .net core and plugin is .net 5, the same message will pop up

That's what I meant. Maybe we should be auto-updating these if possible? Or prompting the user to update it (assuming an update is available)

jjw24:
We can't do anything for users on 1.7.2 or earlier, any changes will be 1.8.0 onwards. This scenario is not valid anyway because with the branch system, they won't be able to download unsupported plugins

JohnTheGr8:
> 
> 
> We can't do anything for users on 1.7.2 or earlier, any changes will be 1.8.0 onwards. This scenario is not valid anyway because with the branch system, they won't be able to download unsupported plugins

What I mean is: if a plugin is installed under Flow 1.7.2 that is not compatible with 1.8.0 because of a breaking change, and the user updates their Flow to 1.8.0, the plugin will still be there in the AppData folder.... If this happens and the plugin fails to load, it would be nice to prompt the user to simply update the plugin to a version that is compatible.

taooceros:
After the fix for everything plugin, I don't believe we have introduced some breaking change for plugin systems in 1.8.0. But that's an interesting idea.

jjw24:
> > We can't do anything for users on 1.7.2 or earlier, any changes will be 1.8.0 onwards. This scenario is not valid anyway because with the branch system, they won't be able to download unsupported plugins
> 
> What I mean is: if a plugin is installed under Flow 1.7.2 that is not compatible with 1.8.0 because of a breaking change, and the user updates their Flow to 1.8.0, the plugin will still be there in the AppData folder.... If this happens and the plugin fails to load, it would be nice to prompt the user to simply update the plugin to a version that is compatible.

I don't think this is a necessity for 1.8.0 release, we have tested the current plugins.

This logic is a bit hard to achieve though, if we can figure out the plugin failure to load is due to lower version plugin, we need to check if a newer version is available. What if plugin load failure is because the plugin tried to write data to a protected folder? 

May be a lot faster to just help user read the error logs than to write up logic to determine the type of failure?

jjw24:
@taooceros @JohnTheGr8 can we get this moving along and merged?

taooceros (approved):

JohnTheGr8:
> I don't think this is a necessity for 1.8.0 release, we have tested the current plugins.

Testing all the plugins for every big release cannot be fun, that's all... ;)

But yeah, I think this is good to go.

jjw24:
> > I don't think this is a necessity for 1.8.0 release, we have tested the current plugins.
> 
> Testing all the plugins for every big release cannot be fun, that's all... ;)
> 
> But yeah, I think this is good to go.

Agree, not fun haha. Let's work on your suggestion for the next release after 1.8.0

PR #550, opened 2021-07-04
fix unexpected Task Status

### LINKED ISSUE #549: PluginManifest constantly updating
In debug mode, 

1. plugins manager is constantly updating when i just query pm install windowwalker:

image

This is coming from the task status being always in  cancelled status:

image

2. might also be related, the last update time is never set, even after several updates:

image

@taooceros is this expected?

taooceros:
fix #549


taooceros:
hold on for merging. There's a small logic issue there, which will cause potential multi run of the last updating time updating (though not a big deal).

taooceros:
Done

jjw24:
tested and is good, thank you

jjw24 (approved):

PR #561, opened 2021-07-05
Dispose Disposable or AsyncDisposable plugin on mainwindow close

taooceros:
Dispose Disposable or AsyncDisposable plugin on mainwindow close

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -11,6 +11,7 @@ using Flow.Launcher.Core.Resource;
 using Flow.Launcher.Helper;
 using Flow.Launcher.Infrastructure.UserSettings;
 using Flow.Launcher.ViewModel;
+using Application = System.Windows.Application;
 using Screen = System.Windows.Forms.Screen;
 using ContextMenuStrip = System.Windows.Forms.ContextMenuStrip;
 using DataFormats = System.Windows.DataFormats;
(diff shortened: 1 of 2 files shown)


jjw24:
What's the impact this issue is causing currently?

taooceros:
We currently don't have a strategy to let plugins know Flow is closing. This will allow plugins to do something when Flow is exiting.

jjw24:
Ok cool, but otherwise no issues casuee right?

taooceros:
> Ok cool, but otherwise no issues caused right?

The ClipBoard plugin? It is causing Flow hang while exiting

jjw24:
I remmeber seeing some plugins inherit the IDisposable interface, do we currently have no implementation of it?

taooceros:
> I remmeber seeing some plugins inherit the IDisposable interface, do we currently have no implementation of it?

I don't think so. Probably what you remember is ISavable?

jjw24:
Hmm ok , I'm afk but that's fine.

jjw24:
I saw your issue in https://github.com/Xenolphthalein/Flow.Launcher.Plugin.ClipboardHistory/issues/4, maybe update this issue to let @Xenolphthalein know how to use it to fix the problem.

taooceros:
By the way, I think we may need to update the plugin development tutorial after the release of 1.8.0. Quite a lot new features are added, and some old features are not documented (like IReloadable, IResultUpdated, etc.)

jjw24:
> By the way, I think we really need to update the plugin development tutorial after the release of 1.8.0. Quite a lot new features are added, and some old features are not documented (like IReloadable, IResultUpdated, etc.)

Oh yeah, for sure

jjw24:
Going to be a lot of writing 😂

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -46,10 +47,13 @@ public MainWindow()
             InitializeComponent();
         }
 
-        private void OnClosing(object sender, CancelEventArgs e)
+        private async void OnClosing(object sender, CancelEventArgs e)
```
jjw24, line 50: What happens when user use sys command exit, will it route the call here?

taooceros: sys exit will call MainWindow.Close()

jjw24 (approved):

PR #562, opened 2021-07-05
Json rpc unit test

taooceros:
Unit Test for JsonRPCPlugins

1. Check whether specific return string is parsed correctly
2. Check whether case is actually insensitive

### DIFF
--- a/Flow.Launcher.Core/Flow.Launcher.Core.csproj
+++ b/Flow.Launcher.Core/Flow.Launcher.Core.csproj
@@ -59,10 +59,6 @@
   </ItemGroup>
   
   <ItemGroup>
-    <Folder Include="Properties\" />
-  </ItemGroup>
-  
-  <ItemGroup>
     <ProjectReference Include="..\Flow.Launcher.Infrastructure\Flow.Launcher.Infrastructure.csproj" />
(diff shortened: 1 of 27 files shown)


Flow.Launcher.Core/Flow.Launcher.Core.csproj
```diff
@@ -58,10 +58,6 @@
     <PackageReference Include="squirrel.windows" Version="1.5.2" />
   </ItemGroup>
   
-  <ItemGroup>
-    <Folder Include="Properties\" />
```
jjw24, line 62: is this removal intended? we still specify compile include at line 48 of the core csproj file

jjw24 (changes requested):
very nice addition, we should add more tests in general.

for unit tests, could we follow the GivenWhenThen structure please:
https://martinfowler.com/bliki/GivenWhenThen.html

Have a look at the stringmatcher unit tests as an example

Flow.Launcher.Core/Flow.Launcher.Core.csproj
taooceros: Let me check. I don't know why this is removed.

jjw24:
I removed the unused InternalVisibleTo for Flow.Launcher project. We can add it when it's actually used

jjw24 (approved):

PR #563, opened 2021-07-06
Move IAsyncReloadable.cs, IReloadable.cs, ISavable.cs to IFeatures

taooceros:
Also add a Cancellation Token field to ResultUpdatedEventArgs


jjw24:
How are we planning to use IFeatures?

jjw24:
> How are we planning to use IFeatures?

All good, I just read the code lol

jjw24:
Could we rename Feature.cs to Features.cs please

jjw24:
Do we want to also move the other interfaces in to Features.cs or keep them separate?

jjw24:
> Could we rename Feature.cs to Features.cs please

Before you do this actually, will it break plugins ?

taooceros:
> > Could we rename Feature.cs to Features.cs please
> 
> Before you do this actually, will it break plugins ?

I test the Dictionary plugin, and it works. Not a lot plugin has implemented these interfaces.

taooceros:
> Do we want to also move the other interfaces in to Features.cs or keep them separate?

I am not sure about this, but we should choose one throughoutly. Move all interface to one file or separate all of them.

jjw24:
just been checking this pr, ran into this issue:

image

jjw24:
also ran into this issue when downloaded dictionary plugin and trying to query it, doubt this may be related to this pr but can you see if you can replicate please

image

taooceros:
Will take a look now.

taooceros:
> just been checking this pr, ran into this issue:

image

This issue has been reproduced before. The actual reason is unclear, probably because multithreading.

taooceros:
> also ran into this issue when downloaded dictionary plugin and trying to query it, doubt this may be related to this pr but can you see if you can replicate please
> 
> image

I will redownload dictionary plugin and try.

taooceros:
It is reproducible with current dev branch, but not reproducible with the pre-release build. Interesting

taooceros:
It is reproducible with current dev branch, but not reproducible with the pre-release build. I think the reason is that the release build already maintain a Sqlite.Interop.dll in the runtime, while when we load the plugin, we didn't load the runtime.

taooceros:
@jjw24 It's debug only issue, so won't stop us from checking this branch.

By the way, shall we add publishprofile to dotnet template?

jjw24:
publish profile for dontet template? dont think so, plugins releases zip file, what would be the need for a publish profile?

jjw24:
i renamed Features.cs. Checked that the only external plugins that uses interfaces from there are WindowWalker and Dictionary.

jjw24:
> @jjw24 It's debug only issue, so won't stop us from checking this branch.
> 
> By the way, shall we add publishprofile to dotnet template?

oh you mean the github workflow file? then yes

taooceros:
> > @jjw24 It's debug only issue, so won't stop us from checking this branch.
> > By the way, shall we add publishprofile to dotnet template?
> 
> oh you mean the github workflow file? then yes

No, I mean auto create a publishprofile targeting the Net 5 and with runtime specification win-x64. Or else some native files like the Sqlite.Interop.dll will still reproduce the issue.

taooceros:
Though I think we can include a powershell script for creating release, and add the publish argument there.

taooceros:
> i renamed Features.cs. Checked that the only external plugins that uses interfaces from there are WindowWalker and Dictionary.

And the everything plugin.

jjw24:
> > i renamed Features.cs. Checked that the only external plugins that uses interfaces from there are WindowWalker and Dictionary.
> 
> And the everything plugin.

yep everything plugin tested as well, is good.

jjw24:
> Though I think we can include a powershell script for creating release, and add the publish argument there.

maybe a powershell better? dont think many will use publishprofile

jjw24:
can you see if you can replicate this:

in my ide vs 2019 in debug mode, when i turn on enable suggestions from google, no results are showing anymore. using the tray icon and going into settings to switch it off then restart, results come back again

jjw24:
> can you see if you can replicate this:
> 
> in my ide vs 2019 in debug mode, when i turn on enable suggestions from google, no results are showing anymore. using the tray icon and going into settings to switch it off then restart, results come back again

possibly related to the cancellation token in mainviewmodel

jjw24:
i dont think we should move interfaces into one file, hard to find stuff. Maybe just move them into the Interfaces folder. When we add more interfaces, it's going to be a big file in the future i think

taooceros:
> > can you see if you can replicate this:
> > in my ide vs 2019 in debug mode, when i turn on enable suggestions from google, no results are showing anymore. using the tray icon and going into settings to switch it off then restart, results come back again
> 
> possibly related to adding the cancellation token check in mainviewmodel

will check soon or tomorrow

taooceros:
I found out why you can reproduce the cancellation bug. Setting one of the web search with global wildcard with search source suggestion will reproduce the issue consistently.

jjw24:
let's keep the interfaces in the Interfaces folder :)

jjw24 (approved):

PR #565, opened 2021-07-09
Fix flow's hotkey not working correctly after changing in settings

### LINKED ISSUE #166: Empty last query is not working (neither is select last query)
**Describe the bug/issue**
When I open FlowLauncher with hotkey <kbd>Win</kbd>+<kbd>N</kbd>, the launcher appears but it contains the last query.

This also happened to me in old Wox, as well as sometimes with your fork. But with Flow.Launcher it seems to happen all the time.

**To Reproduce**
1. <kbd>Win</kbd>+<kbd>N</kbd>
2. Type e.g. "g foo"
3. Hit enter
4. Notice Launcher starts a new browser window and disappears.
5. Hit <kbd>Win</kbd>+<kbd>N</kbd>
6. Notice that "g foo" is still there.

**Screenshots**

image

**Your System**
```
Windows build number: Version 10.0.19041.508
Flow Launcher version: 1.2.1 Portable mode
```

jjw24:
**Description**
After changing the hotkey that triggers flow's query window, the query window that get triggered by new hotkey can not be used- unable to move up and down on the result list, unable to move the cursor on the query test box, the query text is not select if you have this option enabled, and query window's position is not in the same place as previous. 

**Changes**
- consolidated duplicate hotkey behaviour code into one class HotkeyMapper
- added focus to custom query hotkey preview so can be used when previewing
- behaviour should be the same for when a hot key/custom query hotkey is changed and when these hotkeys are loaded during startup

**How to reproduce**
Change hotkey. Trigger the query window, then unable to use it with the issues described above.

**Tested with this PR:**
- changed hotkey and made sure new hotkey works afterwards
- changed hotkey and check all 'Last Query Style' options are working afterwards
- changed hotkey and set ignore hotkey in full screen mode
- able to trigger Custom Query Hotkeys and query window position is correct

Close #166

### DIFF
--- a/Flow.Launcher/Helper/HotKeyMapper.cs
+++ b/Flow.Launcher/Helper/HotKeyMapper.cs
@@ -20,7 +20,7 @@ namespace Flow.Launcher.Helper
             settings = mainViewModel._settings;
 
             SetHotkey(settings.Hotkey, OnHotkey);
-            SetCustomPluginHotkey();
+            LoadCustomPluginHotkey();
         }
 
         private static void SetHotkey(string hotkeyStr, EventHandler<HotkeyEventArgs> action)
@@ -91,22 +91,29 @@ namespace Flow.Launcher.Helper
             }
         }
 
-        private static void SetCustomPluginHotkey()
+        internal static void LoadCustomPluginHotkey()
         {
-            if (settings.CustomPluginHotkeys == null) return;
+            if (settings.CustomPluginHotkeys == null)
+                return;
+
             foreach (CustomPluginHotkey hotkey in settings.CustomPluginHotkeys)
             {
-                SetHotkey(hotkey.Hotkey, (s, e) =>
-                {
-                    if (ShouldIgnoreHotkeys())
-                        return;
-
-                    UpdateLastQUeryMode();
-
-                    mainViewModel.MainWindowVisibility = Visibility.Visible;
-                    mainViewModel.ChangeQueryText(hotkey.ActionKeyword);
-                });
+                SetCustomQueryHotkey(hotkey);
             }
+        }
+
+        internal static void SetCustomQueryHotkey(CustomPluginHotkey hotkey)
+        {
+            SetHotkey(hotkey.Hotkey, (s, e) =>
+            {
+                if (ShouldIgnoreHotkeys())
+                    return;
+
+                UpdateLastQUeryMode();
+
+                mainViewModel.MainWindowVisibility = Visibility.Visible;
+                mainViewModel.ChangeQueryText(hotkey.ActionKeyword);
+            });
         }
     }
 }
--- a/Flow.Launcher/Helper/HotKeyMapper.cs
+++ b/Flow.Launcher/Helper/HotKeyMapper.cs
@@ -1,0 +1,110 @@
+﻿using Flow.Launcher.Infrastructure.Hotkey;
+using Flow.Launcher.Infrastructure.UserSettings;
+using System;
+using NHotkey;
+using NHotkey.Wpf;
+using Flow.Launcher.Core.Resource;
+using System.Windows;
+using Flow.Launcher.ViewModel;
+
+namespace Flow.Launcher.Helper
+{
+    internal static class HotKeyMapper
+    {
+        private static Settings settings;
+        private static MainViewModel mainViewModel;
+
+        internal static void Initialize(MainViewModel mainVM)
+        {
+            mainViewModel = mainVM;
+            settings = mainViewModel._settings;
+
+            SetHotkey(settings.Hotkey, OnHotkey);
+            SetCustomPluginHotkey();
+        }
+
+        private static void SetHotkey(string hotkeyStr, EventHandler<HotkeyEventArgs> action)
+        {
+            var hotkey = new HotkeyModel(hotkeyStr);
+            SetHotkey(hotkey, action);
+        }
+
+        internal static void SetHotkey(HotkeyModel hotkey, EventHandler<HotkeyEventArgs> action)
+        {
+            string hotkeyStr = hotkey.ToString();
+            try
+            {
+                HotkeyManager.Current.AddOrReplace(hotkeyStr, hotkey.CharKey, hotkey.ModifierKeys, action);
+            }
+            catch (Exception)
+            {
+                string errorMsg =
+                    string.Format(InternationalizationManager.Instance.GetTranslation("registerHotkeyFailed"),
+                        hotkeyStr);
+                MessageBox.Show(errorMsg);
+            }
+        }
+
+        internal static void RemoveHotkey(string hotkeyStr)
+        {
+            if (!string.IsNullOrEmpty(hotkeyStr))
+            {
+                HotkeyManager.Current.Remove(hotkeyStr);
+            }
+        }
+
+        internal static void OnHotkey(object sender, HotkeyEventArgs e)
+        {
+            if (!ShouldIgnoreHotkeys())
+            {
+                if (settings.LastQueryMode == LastQueryMode.Empty)
+                {
+                    mainViewModel.ChangeQueryText(string.Empty);
+                }
+                else if (settings.LastQueryMode == LastQueryMode.Preserved)
+                {
+                    mainViewModel.LastQuerySelected = true;
+                }
+                else if (settings.LastQueryMode == LastQueryMode.Selected)
+                {
+                    mainViewModel.LastQuerySelected = false;
+                }
+                else
+                {
+                    throw new ArgumentException($"wrong LastQueryMode: <{settings.LastQueryMode}>");
+                }
+
+                mainViewModel.ToggleFlowLauncher();
+                e.Handled = true;
+            }
+        }
+
+        /// <summary>
+        /// Checks if Flow Launcher should ignore any hotkeys
+        /// </summary>
+        /// <returns></returns>
+        private static bool ShouldIgnoreHotkeys()
+        {
+            //double if to omit calling win32 function
+            if (settings.IgnoreHotkeysOnFullscreen)
+                if (WindowsInteropHelper.IsWindowFullscreen())
+                    return true;
+
+            return false;
+        }
+
+        private static void SetCustomPluginHotkey()
+        {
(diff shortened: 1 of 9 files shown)


taooceros:
Could you share which part fixes the bug? It is a bit confuse this commit includes quite a lot refactor.

Flow.Launcher/Helper/HotKeyMapper.cs
```diff
@@ -0,0 +1,110 @@
...
+                    mainViewModel.LastQuerySelected = true;
+                }
+                else if (settings.LastQueryMode == LastQueryMode.Selected)
+                {
+                    mainViewModel.LastQuerySelected = false;
+                }
+                else
+                {
+                    throw new ArgumentException($"wrong LastQueryMode: <{settings.LastQueryMode}>");
+                }
```
taooceros, line 75: This is not needed because the changed of MainViewModel.Visibility will include this as part through events

Flow.Launcher/Helper/HotKeyMapper.cs
```diff
@@ -0,0 +1,110 @@
...
+        private static void SetCustomPluginHotkey()
+        {
+            if (settings.CustomPluginHotkeys == null) return;
+            foreach (CustomPluginHotkey hotkey in settings.CustomPluginHotkeys)
+            {
+                SetHotkey(hotkey.Hotkey, (s, e) =>
+                {
+                    if (ShouldIgnoreHotkeys()) return;
+                    mainViewModel.MainWindowVisibility = Visibility.Visible;
+                    mainViewModel.ChangeQueryText(hotkey.ActionKeyword);
```
taooceros, line 105: Could you also add `mainViewModel.LastQuerySelected = false;` to here because I do have encountered it is selected sometime.

Flow.Launcher/Helper/HotKeyMapper.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// </summary>
+        /// <returns></returns>
+        private static bool ShouldIgnoreHotkeys()
+        {
+            //double if to omit calling win32 function
+            if (settings.IgnoreHotkeysOnFullscreen)
+                if (WindowsInteropHelper.IsWindowFullscreen())
+                    return true;
+
+            return false;
```
taooceros, line 93: Also could you change this to short circuit?

```c#
return settings.IgnoreHotkeysOnFullscreen && WindowsInteropHelper.IsWindowFullscreen();
```

It will omit calling win32 function when the first one is false.

taooceros (changes requested):

jjw24:
Will do those changes tomorrow, haven't got around to this at all this weekend.

Flow.Launcher/Helper/HotKeyMapper.cs
jjw24: hmm not sure what you mean, remove that code will result in different behaviour

taooceros: Oh sorry I misremember the onPropertyChanged invoke in MainWindow🤐. We need to keep this.

taooceros:
By the way, do you think it would be better to change the use of the property `LastQuerySelected`. I think the usage is quite confusing.

Flow.Launcher/Helper/HotKeyMapper.cs
jjw24: it's same thing isnt it, current code wont call the win32 function anyway if settings.IgnoreHotkeysOnFullscreen is false

taooceros: Yeah I know. Just it looks cleaner, and the comment is a bit confusing suggesting that we need the double if to do so.

Flow.Launcher/Helper/HotKeyMapper.cs
```diff
@@ -87,14 +72,37 @@ private static bool ShouldIgnoreHotkeys()
...
             if (settings.CustomPluginHotkeys == null) return;
             foreach (CustomPluginHotkey hotkey in settings.CustomPluginHotkeys)
             {
                 SetHotkey(hotkey.Hotkey, (s, e) =>
                 {
-                    if (ShouldIgnoreHotkeys()) return;
+                    if (ShouldIgnoreHotkeys())
+                        return;
+
+                    UpdateLastQUeryMode();
```
taooceros, line 104: I don't think we need to add this here. The plugin hotkey is supposed to change the query text to a specific keyword. Therefore, it should always only change the text and allow typing. No need for preserve/select/empty last query.

taooceros: Currently it works for most situation except when the query are the same. Probably we can just add one more check to manually deselect the text.

jjw24: I am still refactoring, maybe take a look when i am done

jjw24: > I don't think we need to add this here. The plugin hotkey is supposed to change the query text to a specific keyword. Therefore, it should always only change the text and allow typing. No need for preserve/select/empty last query.

yeah makes sense, let me double check

jjw24: > > I don't think we need to add this here. The plugin hotkey is supposed to change the query text to a specific keyword. Therefore, it should always only change the text and allow typing. No need for preserve/select/empty last query.
> 
> yeah makes sense, let me double check

removed

jjw24:
> By the way, do you think it would be better to change the use of the property `LastQuerySelected`. I think the usage is quite confusing.

how is it confusing? i used it all the time, it makes sure the previous query is highlighted so i can start typing after triggered the window.

jjw24:
@taooceros ready to go, please review 😄

taooceros:
> > By the way, do you think it would be better to change the use of the property `LastQuerySelected`. I think the usage is quite confusing.
> 
> how is it confusing? i use it all the time, it makes sure the previous query is highlighted so i can start typing after triggered the window.

It is confusing because when we don't want to select last query, we need to mark it as `true` to fake a selected. When we want the model to select last time query, we mark it `false`. Maybe a property like `(Should)SelectLastQuery` may be clearer.

Flow.Launcher/Helper/HotKeyMapper.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            try
+            {
+                HotkeyManager.Current.AddOrReplace("HotkeyAvailabilityTest", currentHotkey.CharKey, currentHotkey.ModifierKeys, (sender, e) => { });
+
+                return true;
+            }
+            catch
+            {
+            }
```
taooceros, line 127: We can remove the catch

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -21,7 +21,9 @@ public partial class HotkeyControl : UserControl
         protected virtual void OnHotkeyChanged()
         {
             EventHandler handler = HotkeyChanged;
-            if (handler != null) handler(this, EventArgs.Empty);
+
+            if (handler != null)
+                handler(this, EventArgs.Empty);
```
taooceros, line 26: I like this way `handler?.Invoke(this, EventArgs.Empty)`, but either is fine.

Flow.Launcher/Languages/en.xaml
```diff
@@ -46,7 +46,7 @@
     <system:String x:Key="browserMorePlugins">Find more plugins</system:String>
     <system:String x:Key="enable">Enable</system:String>
     <system:String x:Key="disable">Disable</system:String>
-    <system:String x:Key="actionKeywords">Action keyword:</system:String>
+    <system:String x:Key="actionKeywords">Query</system:String>
```
taooceros, line 49: image
Is this reference to this? I think Query may be a bit confusing here.

taooceros, line 49: image
Is this reference to this? I think Query may be a bit confusing here.

taooceros, line 49: image
Is this reference to this? I think Query may be a bit confusing here.

taooceros (commented):
Do you think we shall change the select query property?

Flow.Launcher/Helper/HotKeyMapper.cs
jjw24: cant remove it, catch is needed to go to `return false` at the bottom, you can just use catch but you duplicate code for no benefit, so may be just leave it?

Flow.Launcher/Languages/en.xaml
jjw24: good catch, i didnt realise that is hooked up to Action keyword field

jjw24:
> > > By the way, do you think it would be better to change the use of the property `LastQuerySelected`. I think the usage is quite confusing.
> > 
> > 
> > how is it confusing? i use it all the time, it makes sure the previous query is highlighted so i can start typing after triggered the window.
> 
> It is confusing because when we don't want to select last query, we need to mark it as `true` to fake a selected. When we want the model to select last time query, we mark it `false`. Maybe a property like `(Should)SelectLastQuery` may be clearer.

oh ok, you mean the code not the actual functionality

jjw24:
@taooceros just done refactor based on my understanding of how the last query modes should work, have a look and let me know if it's correct https://github.com/Flow-Launcher/Flow.Launcher/pull/568

jjw24:
@taooceros i think this pr is ready to go, let's get this in and do the last query mode refactor seperately

taooceros:
> @taooceros i think this pr is ready to go, let's get this in and do the last query mode refactor seperately so we can release 1.8.0 since there is no functionality change for the refactor

Oh sorry, I was not at home these few days so haven't got through the whole pr again.
Sure. Let's keep them after 1.8.0.

taooceros (approved):

jjw24:
> > @taooceros i think this pr is ready to go, let's get this in and do the last query mode refactor seperately so we can release 1.8.0 since there is no functionality change for the refactor
> 
> Oh sorry, I was not at home these few days so haven't got through the whole pr again.
> Sure. Let's keep them after 1.8.0.

All good no stress, I was just doing a fyi

PR #568, opened 2021-07-15
refactor last query mode

jjw24:
simplify the last query mode code

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -84,11 +84,9 @@ namespace Flow.Launcher
                             QueryTextBox.Focus();
                             UpdatePosition();
                             _settings.ActivateTimes++;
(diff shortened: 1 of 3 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -84,11 +84,9 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
                             UpdatePosition();
                             _settings.ActivateTimes++;
-                            if (!_viewModel.LastQuerySelected)
-                            {
-                                QueryTextBox.SelectAll();
-                                _viewModel.LastQuerySelected = true;
-                            }
+
+                            if (_settings.LastQueryMode == LastQueryMode.Selected)
+                                    QueryTextBox.SelectAll();
```
taooceros, line 89: This will make the custom plugin keyword select the query text as well.

PR #570, opened 2021-07-18
fix Custom Plugin Hotkey not deselect text

taooceros:
fix Custom Plugin Hotkey not deselect text when current query text is equal to custom plugin hotkey.

The part I describe in #565.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -125,6 +125,13 @@ namespace Flow.Launcher
 
                         break;
                     }
+                    case nameof(MainViewModel.QueryTextCursorMovedToEnd):
+                        if (_viewModel.QueryTextCursorMovedToEnd)
+                        {
+                            MoveQueryTextToEnd();
+                            _viewModel.QueryTextCursorMovedToEnd = false;
+                        }
+                        break;
                 }
             };
             _settings.PropertyChanged += (o, e) =>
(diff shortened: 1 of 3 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -329,13 +336,9 @@ private void OnKeyDown(object sender, KeyEventArgs e)
...
         }
 
-        private void OnTextChanged(object sender, TextChangedEventArgs e)
+        private void MoveQueryTextToEnd()
         {
-            if (_viewModel.QueryTextCursorMovedToEnd)
-            {
-                QueryTextBox.CaretIndex = QueryTextBox.Text.Length;
-                _viewModel.QueryTextCursorMovedToEnd = false;
-            }
```
taooceros, line 338: We can add this once we want to do something else when text changed, but currently it would be better to call it when the property changed.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -289,8 +289,8 @@ public string QueryText
         /// <param name="queryText"></param>
         public void ChangeQueryText(string queryText)
         {
-            QueryTextCursorMovedToEnd = true;
             QueryText = queryText;
+            QueryTextCursorMovedToEnd = true;
```
taooceros, line 293: change order to avoid unintended position changed before the QueryText Changed

jjw24:
> fix Custom Plugin Hotkey not deselect text when current query text is equal to custom plugin hotkey

Does this happen in dev branch? I remmeber I checked that custom query plugin hotkey doesn't select all text

taooceros:
> > fix Custom Plugin Hotkey not deselect text when current query text is equal to custom plugin hotkey
> 
> Does this happen in dev branch? I remmeber I checked that custom query plugin hotkey doesn't select all text

It should. It didn't select all text, but if the text is already selected, Flow is not able to deselect it.

jjw24:
Oh ok. After this goes in we should be good to release 1.8.0 right?

jjw24 (approved):

PR #571, opened 2021-07-18
[Explorer] Only files and only folders search

pc223:
Should we have options for only files and only folders search? Seems much better UX, more precise 👌 :smile:

This PR serve as demo and starting point, not intended to be merge yet, use `f <file_name>` or `d <folder_name>` to search.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -151,6 +151,22 @@ namespace Flow.Launcher.Core.Plugin
                             NonGlobalPlugins[actionKeyword] = plugin;
                             break;
                     }
+                   
+                    
+                    
+                    
+                    
+                    
+                    if (plugin.Metadata.Name == "Explorer")
+                    {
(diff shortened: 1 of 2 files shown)


taooceros:
What about make `d` or `f` as secondary action keyword?

pc223:
> What about make `d` or `f` as secondary action keyword?

What secondary action keyword looks like? Like `doc`, `pa`, `id`? I'm thinking of something like that :smile:

taooceros:
I am thinking like `e d/f {search}` or `d/f {search}` when actionkeyword is `*`.

pc223:
> I am thinking like `e d/f {search}` or `d/f {search}` when actionkeyword is `*`.

Personally I prefer `f <file_name>` or `d <folder_name>` as main way to use Explorer, seems simpler, less syntax users have to learn.

I use it this way for awhile, seems pretty good, never needed to use `e <...>` anymore, always `d/f <...>`.

jjw24:
Agree with @pc223 , more natural to use as action keywords like the folder path and index search action keywords

Jack251970:
Already available in the latest version

image

PR #572, opened 2021-07-18
[Test.Explorer] Fix hard-coded  LCID (1033 - en-US)

pc223:
This test fails when run on system have different locale than en-US, because of the hard-coded 1033 LCID (en-US).


jjw24:
Is this something that we should be fixing in the code actually?

pc223:
> Is this something that we should be fixing in the code actually?

I don't think we have to fix the code. In the code, the LCID (1033) automatically generated according to the system locale


See this
```
QueryConstructor.CreateQueryHelper()
```

And this

image

pc223:
The test fails because it expects the system locale to be "1033" (en-US) which is not always true :smile:

jjw24 (approved):

PR #579, opened 2021-07-19
add some space to the end of the scrollviewer

taooceros:
When user scroll down to the end, it will have some space around there.


pc223 (approved):
Works like expected on my machine

jjw24 (approved):

PR #595, opened 2021-07-24
fix animation logic

taooceros:
Resolve Animation not start if we hide and display the main window when the operation is still continue


taooceros:
This is a really simple change..... I just add an else after the if. I don't know why it creates so much diff.

jjw24 (approved):

PR #596, opened 2021-07-24
Add portable release

jjw24:
Pack flow for portable release

1. install with silent flag
2. create userdata folder
3. zip to output location
4. upload artifact

This release is also helpful for users who are unable to use the installer

### DIFF
--- a/Scripts/post_build.ps1
+++ b/Scripts/post_build.ps1
@@ -111,7 +111,7 @@ function Publish-Portable ($outputLocation, $version) {
     
     & $outputLocation\Flow-Launcher-v$v.exe --silent | Out-Null
     mkdir "$env:LocalAppData\FlowLauncher\app-$version\UserData"
-    Compress-Archive -Path $env:LocalAppData\FlowLauncher -DestinationPath $outputLocation\Flow-Launcher-v$version-Portable.zip
+    Compress-Archive -Path $env:LocalAppData\FlowLauncher -DestinationPath $outputLocation\Flow-Launcher-Portable.zip
 }
 
 function Main {
--- a/Scripts/post_build.ps1
+++ b/Scripts/post_build.ps1
@@ -107,6 +107,13 @@ function Publish-Self-Contained ($p) {
     dotnet publish -c Release $csproj /p:PublishProfile=$profile
 }
 
+function Publish-Portable ($outputLocation, $version) {
+    
+    & $outputLocation\Flow-Launcher-v$v.exe --silent | Out-Null
+    mkdir "$env:LocalAppData\FlowLauncher\app-$version\UserData"
+    Compress-Archive -Path $env:LocalAppData\FlowLauncher -DestinationPath $outputLocation\Flow-Launcher-v$version-Portable.zip
+}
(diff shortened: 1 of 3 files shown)


Scripts/post_build.ps1
```diff
@@ -107,6 +107,13 @@ function Publish-Self-Contained ($p) {
     dotnet publish -c Release $csproj /p:PublishProfile=$profile
 }
 
+function Publish-Portable ($outputLocation, $version) {
+    
+    & $outputLocation\Flow-Launcher-v$v.exe --silent | Out-Null
+    mkdir "$env:LocalAppData\FlowLauncher\app-$version\UserData"
+    Compress-Archive -Path $env:LocalAppData\FlowLauncher -DestinationPath $outputLocation\Flow-Launcher-Portable.zip
+}
+
```
taooceros, line 116: Why not simply compress the folder that has been published before deleting it?

jjw24: we need to add userdata in there first

taooceros: We may also need to create a default setting.json and mark Portable as true.

jjw24: no, portable is determined simply if UserData folder exists

jjw24: > Why not simply compress the folder that has been published before deleting it?

Also you need to Squirrelfy it first otherwise you dont get the Update.exe, which will cause restart and update functionality to fail.  However Squirrelfy packs everything into an installer so, in the end you just have to install it then convert it to portable.

taooceros: Oh ok, so why not simply create UserData folder and zip it?

jjw24: because that will make the installer portable

pc223: Sorry for interject 😄 but this is a thing that I request long ago on the old repo, is keep the portable release as a zip file, really simple, no magic.

I think portable users like that more than current approach. The current approach is...a bit new, not many users familiar with that workflow.

You see this approach in nearly all portable software, one zip file, no installer, run independently.

https://github.com/goldendict/goldendict/wiki/Early-Access-Builds-for-Windows
https://github.com/NickeManarin/ScreenToGif/releases/tag/2.33
https://github.com/greenshot/greenshot/releases/tag/Greenshot-RELEASE-1.2.10.6

jjw24: lol I know, I remember 😉  that's why this is here

It just wasnt as straight forward because the way we pack flow into an installer

pc223: Got it!, just want to make sure the idea is clear, if it's technical problem then I don't have any more comment haha 😆

taooceros: > because that will make the installer portable

I mean after packing the installer.

jjw24: the directory structure is not setup to use it
image

jjw24:
We good to go with this @taooceros ?

taooceros (approved):

PR #599, opened 2021-07-25
Fix Windows Index Search service not activated

jjw24:
Resolves the issue where the Windows Search service required for Windows Index Search in Explorer plugin is turned off by the user intentionally and currently throwing the error report window on every character typed.

To replicate this issue, you need to stop Windows Search service, disable it so it wont be activated by other services, and then type in the search window to trigger Explorer 's Windows Index Search.

Solution:
- Add exception handling, and then output a warning window to let user know.
- Add option for user to remove the message, as well as an alternative to install Everything plugin.
  Users who intentionally stopped the service usually are inclined to use Everything instead.

image

image

image

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
@@ -1,10 +1,9 @@
-using Flow.Launcher.Infrastructure.Logger;
 using Flow.Launcher.Plugin.Explorer.Search.QuickAccessLinks;
 using Microsoft.Search.Interop;
 using System;
 using System.Collections.Generic;
 using System.Data.OleDb;
-using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
@@ -84,7 +83,8 @@ namespace Flow.Launcher.Plugin.Explorer.Search.WindowsIndex
              return results;
         }
 
-        internal async static Task<List<Result>> WindowsIndexSearchAsync(string searchString, string connectionString,
+        internal async static Task<List<Result>> WindowsIndexSearchAsync(string searchString, 
+                                                                  Func<CSearchQueryHelper> queryHelper,
                                                                   Func<string, string> constructQuery,
                                                                   List<AccessLink> exclusionList,
                                                                   Query query,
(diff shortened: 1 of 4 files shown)


Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -84,7 +86,8 @@ internal static async Task<List<Result>> ExecuteWindowsIndexSearchAsync(string i
              return results;
         }
 
-        internal async static Task<List<Result>> WindowsIndexSearchAsync(string searchString, string connectionString,
+        internal async static Task<List<Result>> WindowsIndexSearchAsync(string searchString, 
+                                                                  Func<CSearchQueryHelper> queryHelper,
```
taooceros, line 90: why not simply accept a CSearchQueryHelper here?

jjw24: Well not much of difference at this stage to pass in the actual object or the function. Just thought to keep it consistent with other parameters, and in the future if we do need to test this function we don't need to mock up the object or change it to a function parameter

taooceros: Hmm even with the delegate seems that we still need to mock the object for testing🧐.

jjw24: Not really, with Explorer's approach you can test each individual function for their behaviour

taooceros: I mean we still need to mock a `CSearchQueryHelper` to test this.

taooceros: It seems that the queryHelper is only used for getting the connectionString, maybe a simple connectionString will be fine here?

jjw24: Creating a connection string calls CSearchManager, that's why I had to change it

taooceros: Sorry i didn't notice the message. 
I think it would be better to keep the dependency cleaner. This method only requires a connection string, so probably creating it outside and pass that in would be better.
I don't think the c search manager belongs to this method, so why not create it outside the method?

Too much delegate may make the code a bit hard to understand.

jjw24: CSearchManager is part of this method call chain, other alternative is to pass in the query helper object, do this instead?

taooceros: Yeah I mean so. I just not quite understand the need to pass a delegate here.

jjw24: done

taooceros (approved):

PR #603, opened 2021-07-26
Add a MemoryStream buffer to ReadStream first

taooceros:
fix #598

PR #554, opened 2021-07-04
Change bulk add condition

taooceros:
From discussion https://github.com/Flow-Launcher/Flow.Launcher/discussions/586

PR #604, opened 2021-07-26
Release 1.8.1

jjw24:
Release Notes
=========

## Portable Zip Download
You can now download portable version directly from the [releases page](https://github.com/Flow-Launcher/Flow.Launcher/releases/latest). This is the same as manually enabling portable mode after the installation.

## Bug fixes
- Fixed an issue where Python plugins could not display more than 49 results #603 
- Fixed Explorer crashing due to Windows Index Search service not activated  #599
- Fixed progress bar animation not rerunning after query window reappears during loading #595 
- Resolved the scenario where settings not saved after Python path is set. Settings now save after flow finished starting up  #594 
- Reduced query window's magnifying glass icon size so it looks more aligned with the rest of the icons and text boxes #590 
- Updated Slovak translations #583 #614
- Added some space to the bottom of the General tab in Settings window #579


jjw24:
@Flow-Launcher/team lets hold off merging feature PRs and just bug fixes for 1.8.1

taooceros:
Could you take a look at the CI build fail? I think it is due to winget update (which is not supposed to run at pr CI)

taooceros:
> `APPVEYOR_REPO_BRANCH` - build branch. For Pull Request commits it is base branch PR is merging into

from appveyor website

jjw24:
yep just saw, working on it

jjw24:
@Flow-Launcher/team let's roll

Zeroto521:
I thought the next step we should keep our eyes on unit test, since v1.7 they're more bugs comes up after new version released.
My work computer still runs old version, only my home computer I would try new version.

taooceros:
> I thought the next step we should keep our eyes on unit test, since v1.7 they're more bugs comes up after new version released.
> My work computer still runs old version, only my home computer I would try new version.
> 

Yeah I agree with you. However, unit tests are not easy to write due to the complex dependency graph in Flow, and writing it is not so fun.
Probably the first step is to list which unit test is still lacking in the unit test project.

jjw24:
> I thought the next step we should keep our eyes on unit test, since v1.7 they're more bugs comes up after new version released.
> 

This will have to be an ongoing thing, we can't hold off releases until enough tests have been added.

> In my experience, some troubles these tests could easily check out.
> 
> Test pipeline should pass before prs are merged. There has some, but not enough.
> 
> The coverage report should be gave one.
> 

We have tests for each pipeline run, bare in mind flow is a desktop application that internally does a lot of different things and connects with a bunch of interfaces, on top of this it runs in an user environment that is quite unique for each users, so there will always be edge cases that we don't catch.

> My work computer still runs old version, only my home computer I would try new version.

We are very quick to patch things up so I don't see much risks in upgrading, we never had a report of complete meltdown where flow was rendered useless.

Overall the general concensus is yeah we need to add more tests to those critical functionalities, at the same time we need to roll out fixes like this one so the bugs we/users find can be quickly stamped out.

@taooceros can you approve this please

taooceros (approved):

pc223:
@taooceros Should we add PR #554 as a temporary fix for the delay?

jjw24:
I was testing pr 554 and got a freeze with nothing logged. we will need to investigate a bit more.

PR #606, opened 2021-07-26
Fix portable readme download link & WinGet pipeline & change Nuget publish on tag only

jjw24:
- Fix portable download link in readme

- Fix WinGet deployment so it occurs only on non-pr master branch builds.

- Change Nuget publish to tag only

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Flow Launcher. Dedicated to make your workflow flow more seamlessly. Aimed at be
 
 ### Installation
 
-| [Windows 7 and up installer](https://github.com/Flow-Launcher/Flow.Launcher/releases/latest) | [Portable](https://github.com/Flow-Launcher/Flow.Launcher/releases/latest/Flow-Launcher-Portable.zip) | `WinGet install "Flow Launcher"` |
+| [Windows 7 and up installer](https://github.com/Flow-Launcher/Flow.Launcher/releases/latest) | [Portable](https://github.com/Flow-Launcher/Flow.Launcher/releases/latest/download/Flow-Launcher-Portable.zip) | `WinGet install "Flow Launcher"` |
 | --------------------------------- | --------------------------------- | --------------------------------- |
 
 Windows may complain about security due to code not being signed, this will be completed at a later stage. If you downloaded from this repo, you are good to continue the set up. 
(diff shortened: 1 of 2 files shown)


taooceros:
Does the winget update work? The check fail on last CI run

JohnTheGr8 (changes requested):
Minor issue: I think it would be better to check `APPVEYOR_REPO_TAG` instead, so that the deployment to WinGet happens from the same CI build as the release deployment to Github

jjw24:
> Does the winget update work? The check fail on last CI run

you mean the ci on the release pr? then yeah update will work because it's just failed to push the update.

jjw24:
> Minor issue: I think it would be better to check `APPVEYOR_REPO_TAG` instead, so that the deployment to WinGet happens from the same CI build as the release deployment to Github

the release happens on push to master or is it triggered by a tag? APPVEYOR_REPO_TAG only equals true when you tag a branch and it builds right?

jjw24:
> Minor issue: I think it would be better to check `APPVEYOR_REPO_TAG` instead, so that the deployment to WinGet happens from the same CI build as the release deployment to Github

Done

On a side note @JohnTheGr8 why are we not publishing to Nuget on tag as well? I see the build history of the 1.8.0 release tried to publish to nuget on tag build too

JohnTheGr8:
> 
> Done

Hmm, I think this:

```ps
if ($env:APPVEYOR_REPO_BRANCH -eq "master" -and $env:APPVEYOR_REPO_TAG)
```

is equivalent to this:

```ps
if ($env:APPVEYOR_REPO_BRANCH -eq "master")
```

because it only checks if `$env:APPVEYOR_REPO_TAG` is set and not its value... I think we need to do it like this:

```ps
if ($env:APPVEYOR_REPO_BRANCH -eq "master" -and $env:APPVEYOR_REPO_TAG -eq "true")
```

> 
> On a side note @JohnTheGr8 why are we not publishing to Nuget on tag as well? I see the build history of the 1.8.0 release tried to publish to nuget on tag build too

I don't think there's a reason, we initially did nuget deployment on master builds and didn't switch to tag builds when we added github deployments later on...

jjw24:
@JohnTheGr8 Fixed

JohnTheGr8 (approved):

PR #609, opened 2021-07-27
Fix PluginsManager context menu crashing

jjw24:
Repro:
Going to any context menu other than pm install and select open website will get the exception pop up

Fix:
- Added handling when the object can not be cast as UserPlugins. This means the selected result was not provided the plugins list from the manifest, as a result we bypass creating menu items for them.
- Added context menu for pm update

Tested:
- pm install and pm update all can use the context menu
- all other context menus will go to the default items with out those options available in install and update

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
@@ -19,6 +19,9 @@ namespace Flow.Launcher.Plugin.PluginsManager
         {
             var pluginManifestInfo = selectedResult.ContextData as UserPlugin;
 
+            if (pluginManifestInfo == null)
+                return new List<Result>();
+
             return new List<Result>
             {
                 new Result
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
@@ -17,9 +17,7 @@ namespace Flow.Launcher.Plugin.PluginsManager
 
         public List<Result> LoadContextMenus(Result selectedResult)
         {
-            var pluginManifestInfo = selectedResult.ContextData as UserPlugin;
-
-            if (pluginManifestInfo == null)
+            if(selectedResult.ContextData is not UserPlugin pluginManifestInfo) 
                 return new List<Result>();
 
             return new List<Result>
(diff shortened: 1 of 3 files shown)


Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
```diff
@@ -19,6 +19,9 @@ public List<Result> LoadContextMenus(Result selectedResult)
         {
             var pluginManifestInfo = selectedResult.ContextData as UserPlugin;
 
+            if (pluginManifestInfo == null)
+                return new List<Result>();
```
taooceros, line 23: Just to let you know C# 9 allows fancy pattern matching like this.🤪
```c#
if(selectedResult.ContextData is not UserPlugin pluginManifestInfo)
    return new List<Result>();
```

taooceros (commented):
Could you also add the contextdata in RequestUninstall

Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
jjw24: Nice!

I'm afk, if u r in front of comp can u make a commit suggestion please

jjw24:
> Could you also add the contextdata in RequestUninstall
> https://github.com/Flow-Launcher/Flow.Launcher/blob/27777956cc1d459719f51be43244b2df00a11e11/Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs#L390-L418

Because we are not creating results from UserPlugins object, it's going to require a whole lot of extra code to match up, bare in mind default plugins are not in the manifest list so it's not in the UserPlugins object. 

Just thought this is not so necessary to implement unless you feel otherwise.

Plugins/Flow.Launcher.Plugin.PluginsManager/ContextMenu.cs
taooceros:
```suggestion
            if(selectedResult.ContextData is not UserPlugin pluginManifestInfo) 
                return new List<Result>();
```

taooceros:
> Because we are not creating results from UserPlugins object, it's going to require a whole lot of extra code to match up, bare in mind default plugins are not in the manifest list so it's not in the UserPlugins object.
> 
> Just thought this is not so necessary to implement unless you feel otherwise.

Yeah probably not that easy to do so. Leave it in the future if needed.

taooceros (approved):

PR #613, opened 2021-07-27
Use 7z instead of powershell native compression

taooceros:
Currently the portable zip is really large, because squirrel will keep a copy of original nupkg. Not sure whether we can remove it (or it will cause some update issue).

### DIFF
--- a/Scripts/post_build.ps1
+++ b/Scripts/post_build.ps1
@@ -111,7 +111,7 @@ function Publish-Portable ($outputLocation, $version) {
     
     & $outputLocation\Flow-Launcher-v$v.exe --silent | Out-Null
     mkdir "$env:LocalAppData\FlowLauncher\app-$version\UserData"
-    Compress-Archive -Path $env:LocalAppData\FlowLauncher -DestinationPath $outputLocation\Flow-Launcher-Portable.zip
+    7z a $outputLocation\Flow-Launcher-Portable.zip $env:LocalAppData\FlowLauncher
 }
 
 function Main {


jjw24:
Doesn't look like much of a size difference though using 7zip

taooceros:
Yes, the main size diff is caused by the duplicate package in packages folder.

pc223:
Wow, the 7z reduce from 127MiB to 21MiB on my machine, we really should consider this

I confirm the duplication of libraries with space sniffer

image

taooceros:
That's because the Release folder is not self contained (framework dependent). Without runtime, regular compression can compress flow to about 20-25 MB.

7z does have advantages, but only 10% when compressing to zip.

pc223:
> 7z does have advantages, but only 10% when compressing to zip.

Yeah, I prefer zip over 7z because built in support

pc223:
Hm...damn, I thought we can reduce to around 20MB, so that means switch to 7z not really change much the size? The build of this PR is similar to other PR. So...we should focus on trying to remove this file right?

image

taooceros:
Yeah that's the case, but not sure whether removing that will cause update issues.

pc223:
> That's because the Release folder is not self contained (framework dependent). Without runtime, regular compression can compress flow to about 20-25 MB.

Can we release the Setup.exe in not self contained style?And if detect syst em not contains required framework (.net 5?) then download for them. Like this: https://github.com/NickeManarin/ScreenToGif

jjw24:
is it worth the extra work for 25mbs? 88mbs is not big of a download and the update is done in the background. Too many installers going to get users confused if we publish a non-self contained, self contained and portable.

pc223:
> is it worth the extra work for 25mbs? 88mbs is not big of a download and the update is done in the background. Too many installers going to get users confused if we publish a non-self contained, self contained and portable.
> 
> Also I dont know how installation of the framework is done, there will be user environments where installation of anything is not allowed.

I'm thinking of replace the current Setup.exe (88MB) with the non-self-contained (25MB). I think it's fairly common practice to have one Installer that download dependencies when needed and have one Portable version that self-contained.

Still, this, personally ,I think should be low priority and not urgent, so no pressure 😄 Just push ideas out there for discussions.

PR #620, opened 2021-07-30
Delay the create query helper call to avoid Windows Search not running

jjw24:
Delay the create query helper call to avoid Windows Search not running, follows on from #599 , missed the query helper call which also requires Windows Search service.


taooceros (approved):
Oh so that's the reason for using delegate huh😂

PR #621, opened 2021-07-30
Add Quick Access action keyword for Explorer

jjw24:
Allows users to trigger and search Quick Access on its own

Tested:
- Changing to global action keyword '*' will show prompt to disallow
- Disabling Quick Access action keyword will return it to global action keyword '*'
- Using Quick Access action keyword can list and search only within the list of saved files/folders
- Using Path Search action keyword can list, search within the list of saved files/folders and as well as its original behaviour of searching paths
- Using other action keywords can not list Quick Access
- Upgrading from Explorer version <1.8.0 to 1.9.0
- Upgrading from Explorer version >1.8.0 to 1.9.0


taooceros:
I have found a bug for explorer plugin....We have added plenty of global action keyword (in 1.8.1, haven't tried this build).
Probably whenever we disable & enable a setting, a new global wildcard will be added.
image

jjw24:
Let's fix this in a seperate PR so we can roll it out as 1.8.3 instead of this feature PR. Are you already looking into this issue? If not i can. Let me know so we dont double up.

jjw24:
i have tried enabling/disabling, changing action keyword and then disabling, i cant replicate. Can you let me know the steps how to repro please.

taooceros:
Sorry I has been a bit busy these few days. I don't know how to replicate this, because I just observe the situation. I will take a check later.
By the way, probably the better to handle actionkeyword here is not assign a global wildcard to every specific option, but just maintain a global wildcard actionkeyword that won't be removed. 

Let me check our logic in change duplicate global wildcard actionkeyword.

taooceros:
I found the way to replicate. Everytime restarting flow, two more "*" will be added to actionkeywords.

taooceros:
Oh it is because the fix for backward compability

jjw24:
> Sorry I has been a bit busy these few days. I don't know how to replicate this, because I just observe the situation. I will take a check later.

All good, no stress :)

Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
```diff
@@ -63,11 +63,19 @@ private void OnDoneButtonClick(object sender, RoutedEventArgs e)
             }
 
 
-            if (CurrentActionKeyword.KeywordProperty == Settings.ActionKeyword.FileContentSearchActionKeyword 
-                && ActionKeyword == Query.GlobalPluginWildcardSign)
+            if (ActionKeyword == Query.GlobalPluginWildcardSign 
+                && (CurrentActionKeyword.KeywordProperty == Settings.ActionKeyword.FileContentSearchActionKeyword 
+                        || CurrentActionKeyword.KeywordProperty == Settings.ActionKeyword.QuickAccessActionKeyword))
```
taooceros, line 68: Why do we need this part of logic? The switch will handle the match

taooceros (changes requested):

Plugins/Flow.Launcher.Plugin.Explorer/Views/ActionKeywordSetting.xaml.cs
jjw24: yeah good pickup

taooceros (approved):

jjw24:
@taooceros I have done a merge of dev branch, no changes added, could you re-approve please

taooceros (approved):

Issue #528, opened 2021-06-28
Remove other search precision levels


taooceros:
### Discussed in https://github.com/Flow-Launcher/Flow.Launcher/discussions/517

<div type='discussions-op-text'>

<sup>Originally posted by **jjw24** April 25, 2020</sup>
Unless I am wrong, other search precision levels- None, Low are not used much by user base. This assumption is that precision level saving is broken and no one has put forward an issue about it in the JJW24 fork.

![image](https://user-images.githubusercontent.com/26427004/80270083-7fa8c600-86f8-11ea-8a2f-d15252195206.png)

Removing others can help simplify logic and focus on fine tuning the search for one level going forward.</div>

skjerns:
Wrong! I use `None`, as otherwise some of my results are not being shown properly, e.g. programms using Umlauts or searching for special chars. E.g. when I want to open Notepad++ I type `++`, but this only shows the app when setting precision to "None" or low

jjw24:
Can you give me some examples of the searches you are not getting with regular

skjerns:
I'm using 1.8.1, I just checked the forum as each version update resets my settings for the precision value.

I can see if I can remember more cases, but for that I need to turn it off and wait a few days

taooceros:
Probably the storage of SearchPrecision Score contains an issue. Will take a look on that soon.

pc223:
Yeah I never use this feature, I always use the default. We should strive for a good default first and options later.

If User fails to find a item with a reasonable query, it should be consider bug.

jjw24:
> ... We should strive for a good default first and options later.
> 
> If User fails to find a item with a reasonable query, it should be consider bug.

agree, absolutely.

skjerns:
I agree!

my problem is, that `++` doesn't ever show up Notepad if setting to `regular`, so I can't really increase the frequency and "train" it. Strangely enough, the c++ compiler shows up! I have the exact same problem on different machines, so it's not just a local problem

pc223:
> my problem is, that `++` doesn't ever show up Notepad if setting to `regular`

Hm, weird, I tried a fresh install of 1.8.1, seems `notepad++` still show in the bottom (with regular on)

jjw24:
> Hm, weird, I tried a fresh install of 1.8.1, seems notepad++ still show in the bottom (with regular on)

[...]

PR #622, opened 2021-07-31
include saving SearchPrecisionScore because it is not public before

taooceros:
Issue discovered in #528


taooceros:
@skjerns Take a try once the build has finished. Previously SearchPrecisionScore was fail to be saved due to json library change.

jjw24 (approved):

PR #615, opened 2021-07-28
Standardise Flow's exe installer + readme update

jjw24:
Standardise Flow's exe installer by removing the version at the end, this will make it easier to link to the latest download e.g. https://github.com/Flow-Launcher/Flow.Launcher/releases/latest/download/Flow-Launcher-Setup.exe

Tested:

1. changed the build version to 1.0.0 for testing
2. downloaded the artifact after build
3. Enabled portable mode
4. moved to different location
5. disabled portable mode
6. update to the latest via update command in query window
7. redid the steps 2-5 and updated flow via settings About tab
8. redid 2-7 for portable zip version

Updated readme to fix c# plugin dev doc and getting started section

PR #626, opened 2021-08-01
Release 1.8.2

jjw24:
Release Notes
=========

## Changing to Flow-Launcher-Setup.exe
We have switched the naming of the setup file from `Flow-Launcher-v<version>` to `Flow-Launcher-Setup.exe`. This will make it easier to link directly to the exe installer.

## Bug fixes

- Fix Explorer's file result subtitle tool tip as file path instead #629  

- No need for the wrapped Lazy instance in image loading #450  

- Adjust JsonRPCPlugin.cs Exception Handling #633

- Fix PluginsManager's InstallOrUpdate focus #628

- Fix actionkeywords inflate when actionkeyword count mismatch in explorer plugin #630

- Standardise Flow's exe installer + readme update #615

- Include saving SearchPrecisionScore because it is not public before #622

- Delay the create query helper call to avoid Windows Search not running #620


pc223:
Quick tested, seems OK on my machine 👍

jjw24:
yeah i have been using this build for a while, it's golden :) let's get this rolling

pc223 (approved):

kubalav (commented):
Tested a while and it works.

PR #628, opened 2021-08-02
Fix PluginsManager's InstallOrUpdate focus

jjw24:
Fix PluginsManager InstallOrUpdate method call's query window focus

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
@@ -120,7 +120,10 @@ namespace Flow.Launcher.Plugin.PluginsManager
                             .ChangeQuery(
                                 $"{Context.CurrentPluginMetadata.ActionKeywords.FirstOrDefault()} {Settings.HotkeyUpdate} {plugin.Name}");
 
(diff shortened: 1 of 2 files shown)


Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -120,7 +120,10 @@ internal async Task InstallOrUpdate(UserPlugin plugin)
                             .ChangeQuery(
                                 $"{Context.CurrentPluginMetadata.ActionKeywords.FirstOrDefault()} {Settings.HotkeyUpdate} {plugin.Name}");
 
-                    Application.Current.MainWindow.Show();
+                    var mainWindow = Application.Current.MainWindow;
+                    mainWindow.Visibility = Visibility.Visible;
+                    mainWindow.Focus();
+
```
taooceros, line 126: Will the mainwindow hide at these time? Haven't taken a try but I think we may not need to change the visibility

jjw24: yeah it will hide because of the message prompt

taooceros (approved):

PR #629, opened 2021-08-03
Fix Explorer's file result subtitle tool tip as file path instead

jjw24:
Fix Explorer's file result subtitle tool tip as file path instead


jjw24:
@pc223 good pick up, thank you. Might be better to use the `path` parameter instead so it's clearer- updated.

taooceros (approved):

PR #653, opened 2021-08-13
Refactor JsonRPC structure

taooceros:
We don't need to provide a virtual method separately for query, context menu, and callback. The virtual method should only provide a way to pass the request model to the executable or something received the jsonrpc request.

Query, ContextMenu, and Callback should be indicated within the JsonRPC model.


jjw24:
No breaking changes here right?

This will go with next release or 1.9.0?

taooceros:
> No breaking changes here right?
> 
Yeah no breaking change. 

> This will go with next release or 1.9.0?

I think so? If this performs well.

jjw24 (approved):

PR #654, opened 2021-08-13
Bump applicable nuget packages to latest versions

taooceros:
#486 
We cannot bump squirrel higher than 1.9.0

### DIFF
--- a/Flow.Launcher.Infrastructure/Flow.Launcher.Infrastructure.csproj
+++ b/Flow.Launcher.Infrastructure/Flow.Launcher.Infrastructure.csproj
@@ -55,7 +55,7 @@
     <PackageReference Include="NLog.Schema" Version="4.7.10" />
     <PackageReference Include="NLog.Web.AspNetCore" Version="4.13.0" />
     <PackageReference Include="System.Drawing.Common" Version="5.0.2" />
-    <PackageReference Include="ToolGood.Words.Pinyin" Version="3.0.2.6" />
+    <PackageReference Include="ToolGood.Words.Pinyin" Version="3.0.1.4" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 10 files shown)


taooceros:
I didn't find some issue currently. Though, our pinyin library has updated its data source, which seems significantly increase the memory usage.

jjw24:
> our pinyin library has updated its data source, which seems significantly increase the memory usage

cpu and memory usage is important, so if the library increases memory, let's not bump it.

Flow.Launcher.Infrastructure/Flow.Launcher.Infrastructure.csproj
```diff
@@ -53,9 +53,9 @@
     <PackageReference Include="Microsoft.VisualStudio.Threading" Version="16.10.56" />
     <PackageReference Include="NLog" Version="4.7.10" />
     <PackageReference Include="NLog.Schema" Version="4.7.10" />
-    <PackageReference Include="NLog.Web.AspNetCore" Version="4.12.0" />
-    <PackageReference Include="System.Drawing.Common" Version="4.7.0" />
-    <PackageReference Include="ToolGood.Words.Pinyin" Version="3.0.1.4" />
+    <PackageReference Include="NLog.Web.AspNetCore" Version="4.13.0" />
+    <PackageReference Include="System.Drawing.Common" Version="5.0.2" />
+    <PackageReference Include="ToolGood.Words.Pinyin" Version="3.0.2.6" />
```
jjw24, line 58: should we not bump this then

taooceros:
The memory inflation will only appear when people use pinyin, but the new version can also increase the performance of pinyin, which probably is acceptable.

jjw24:
> downgrade pinyin library until further discussion

Thanks, i was about to suggest to pull this out as a seperate PR. Did you find more issues from this bump?

jjw24:
havent had the time to look into this, we are getting error with the build:
image

jjw24:
Ok so this error is coming with nuget pack version > 5.8.1. This is part of the wider .net 5 sdk change now requiring you to specify platform version e.g. net5.0-windowsX.Y (X.Y as TPV- TargetPlatformVersion). The target framework net5.0-windows for some of our projects is still valid in the csproj files even though we are not specifying which version of Windows because the .NET SDK provides a default, and NuGet uses that to generate lib/net5.0-windows7.0/your.dll. Since we are not packing via csproj, we wont get the message from there, but we are using command-line to pack so need to append the platform version number to our nuspec.

I am going to separate out this issue so I can isolate the testing for this as I also want to try appending the target platform version to the csproj files.

jjw24 (dismissed):

jjw24 (approved):

PR #657, opened 2021-08-16
Json rpc plugin setting control

### LINKED ISSUE #814: Let setting UI support for other program languages
### Discussed in https://github.com/Flow-Launcher/Flow.Launcher/discussions/514

<div type='discussions-op-text'>

<sup>Originally posted by **Zeroto521** September 21, 2020</sup>
## Question

As title.

More can see [jjw24/Wox#178](https://github.com/jjw24/Wox/issues/178).

## Discussion

I think just like using **JSON-RPC** for supporting other program languages. 

Why not using some **template file** for support setting UI. 

But at the same time, we need a **tool** to compile the file and  **inject in** the Flow main program.

To make it just like a `C#` plugin.

Interaction
</div>

taooceros:
close #814

TODO:
- [ ] Markdown Field
- [ ] File/Folder Selector


taooceros:
It's currently a simple proposal. Only key/value is supported. Bool value use `Checkbox`, and others use `TextBox`.
Numbers are treated as string.

Zeroto521:
[githubs-form-schema](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema) is a good YAML example.

```yaml
---
name: Account Recovery
description: Initiate an account recovery due to lost 2FA or inability to access email address
title: Account recovery request
labels: account-recovery 

body:
- type: markdown
  attributes:
    value: |
      **Request for account recovery.**

- type: input
  attributes:
    label: PyPI Username
    description: >-
      Enter your PyPI `username` wrapped with single backticks
  validations:
    required: true

- type: textarea
  attributes:
    label: Reason for request
    description: |
      State the reason for your request
      - Lost access to email address
      - Lost 2FA authentication app or security token
  validations:
    required: true


- type: checkboxes
  attributes:
    label: Code of Conduct
    description: |
      Read the [PSF Code of Conduct][CoC] first.

      [CoC]: https://github.com/pypa/.github/blob/main/CODE_OF_CONDUCT.md
    options:
    - label: I agree to follow the PSF Code of Conduct
      required: true

- type: checkboxes
  attributes:
    label: Acknowledgement
    description: |
      By their nature account recovery issues are labor intensive. The PyPI team has limited resources available for their completion.

      There may be a delay of a week or more between responses for these issues.
    options:
    - label: I understand that it may take a significant amount of time to process my account recovery request.
      required: true
...
```

image

taooceros:
# Current Proposal

- Configuration File: SettingsTemplate.yaml in the same folder as plugin.json\

Configuration Syntax (almost equal to github one but with small modification

```yaml
- type: textarea
  attributes:
    label: Label to be displayed before the contro
    description: Not yet implemented (probably tooltip)
    placeholder: Not yet implemented 
    name: identifier of the control
    defaultValue: Default Value of the control
- type: dropdown
  attributes:
    label: Version
    description: What version of our software are you running?
    multiple: false
    options:
      - label: 1.0.2 (Default)
      - label: 1.0.3 (Edge)
  validations:
    required: true
- type: checkbox
  attributes:
    label: Code of Conduct
    description: The Code of Conduct helps create a safe space for everyone. We require
      that everyone agrees to it.
    defaultValue: true
```

jjw24:
has this been tested, can you provide a screenshot of an example output please.

taooceros:
> has this been tested, can you provide a screenshot of an example output please.

It is usable. Though some communication detail is still under discussion.

jjw24:
shall we mark it as draft then?

taooceros:
> shall we mark it as draft then?

It depends. The functionality is complete, so plugin is able to create setting via the yaml. The left thing is how we want plugin to change the content in the setting, and how they should access the setting.

jjw24:
> how they should access the setting.

during start up, jsonrpc call to plugin to load the settings and then create the settings in flow's settings window based on the callback.

when user changes the setting in the settings window, do a jsonrpc call to let the plugin know?

>how we want plugin to change the content in the setting

you mean plugin will change the settings contents in the settings window? can you give me an example

taooceros:
Well the current implementation is a template file (in the plugin root directory) will be used to create the settings file and the controls. Plugins may get the place of the setting file, but they currently don't have access to the content changed by user. I am thinking about to pass the whole setting json as part of the query to the plugin, and receive a diff new setting json from plugin when calling back (to modify the setting).
There's no need to let plugin know when setting changes because they don't have state. But it is also possible to save whenever user change the setting, so plugin can directly access that file. Though that way plugin won't be able to modify the setting loaded in flow.

Sometimes we would like to change the setting via query such as we can add path to quickaccesspath in explorer plugin. I think that's reasonable for jsonrpc plugin to do.

Garulf:
After some testing retrieving settings works, the templating works as well. `SettingsChange` would be more effective as an API method like:
```
{"method": "Flow.SettingsChange","parameters": foo}
```

Also we should make sure a malformed template file doesn't crash Flow.

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
```diff
@@ -8,25 +10,37 @@
...
     /// every JsonRPC plugin should has its own plugin instance
     /// </summary>
-    internal abstract class JsonRPCPlugin : IAsyncPlugin, IContextMenu
+    internal abstract class JsonRPCPlugin : IAsyncPlugin, IContextMenu, ISettingProvider, ISavable
     {
         protected PluginInitContext context;
         public const string JsonRPC = "JsonRPC";
 
-        /// <summary>
+        /// <summary=
```
jjw24, line 43:
```suggestion
        /// <summary>
```

Garulf (approved):

PR #664, opened 2021-08-24
Bump Nuget.CommandLine & append csproj target platform

jjw24:
Following on from #654 

Part of the wider .net 5 sdk change now requiring you to specify platform version e.g. net5.0-windowsX.Y (X.Y as TPV- TargetPlatformVersion). The target framework net5.0-windows for some of our projects is still valid in the csproj files even though we are not specifying which version of Windows because the .NET SDK provides a default, and NuGet uses that to generate lib/net5.0-windows7.0/your.dll. Since we are not packing via csproj, we wont get the message from there, but we are using command-line to pack so need to append the platform version number to our nuspec.

**This change includes**
- Nuget.CommandLine to version 5.10.0 from 5.4.0
- Appending the target platform version to the csproj files.

**Tested:**


jjw24:
@taooceros do you remember if there was a reason why we did not append the platform version to all the csproj files?

taooceros:
No a specific reason. I just didn't add them because I am lazy😂

taooceros:
One reason I didn't put all project is to net5.0-windows10.0.19041.0 is to keep plugin doesn't need to use this framework, which includes the WinRT.dll in their reference. Although it won't be loaded in Flow if plugin includes it, it will consume more about 25MB space.

jjw24:
> it will consume more about 25MB space

ok i will need to test this, and maybe find a way to not include it.

jjw24:
completed via https://github.com/Flow-Launcher/Flow.Launcher/pull/1601

PR #674, opened 2021-09-02
Release version bump 1.8.3

jjw24:
Release version bump 1.8.3


jjw24:
@taooceros I think would should get this in and do a 1.8.3 release to sort out the File Content search keyword bug

taooceros:
> @taooceros I think we should get this in and do a 1.8.3 release to sort out the File Content search keyword bug

Yeah agree.

taooceros (approved):

PR #699, opened 2021-09-23
Use CDN for PluginsManifest

taooceros:
https://www.jsdelivr.com/


taooceros:
Probably we can also use CDN for update package download, which will largely improve the experience for Chinese user to get the latest release (though I am not in China for now🤣).

jjw24 (approved):

PR #705, opened 2021-09-27
Fix winget update causing incompatibility with always retrieving latest

jjw24:
This stems from the 1.8.3 release which failed with the latest winget-create version. Simplest fix is just to fix the version we are using since our release is a simple straight forward update.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -78,6 +78,6 @@ on_success:
   - ps: |
         if ($env:APPVEYOR_REPO_BRANCH -eq "master" -and $env:APPVEYOR_REPO_TAG -eq "true")
         {
-            iwr https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
+            iwr https://github.com/microsoft/winget-create/releases/download/v0.2.0.29-preview/wingetcreate.exe -OutFile wingetcreate.exe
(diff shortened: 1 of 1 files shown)


taooceros:
What has the latest winget changed lead to the failure?

jjw24:
image

Zeroto521:
My suggestion is to keep two winget versions.
One is a stable version for now to build, and another is the latest version for the future.
If the latest version building result has any error, that could warn us.

jjw24:
There is no build involved, error comes from submission via their api. Our release is not that complicated as we only update our installer in the winget pkg repo.

I agree there should be a better way to inform of the submission failure because I only picked it up in the last release when I actively checked if the submission went through to winget. Maybe appveyor or GitHub has a way to watch pipeline failures via email,I know GitHub workflow does which is really neat.

jjw24:
@taooceros unless there is alternative, lets get this in.

taooceros:
OK, but I am still wondering why we cannot use the latest winget cli.

jjw24:
@taooceros merge in? if 1.9.0 release still unable to do the update, i can spend time to take another look.

taooceros:
> @taooceros merge in? if 1.9.0 release still unable to do the update, i can spend time to take another look.

ok

taooceros (approved):

PR #713, opened 2021-09-29
Fix Explorer plugin crash on semicolon

jjw24:
Fix Explorer plugin crash on semicolon

Repro:
type `;` with Explorer


jjw24:
yep, they do crash as well. added.

taooceros (approved):

PR #731, opened 2021-10-06
Querybar gadget

onesounds:
Clock in query window test. (sublime theme only)


jjw24:
hey why is this closed, was it working?

onesounds:
> hey why is this closed, was it working?

- This code is old and cannot merge now.
- No one replied to me about this as go or no.
- You talked about the second memory. I don't know exactly how to test this problem and I don't know about its impact. Someone should test it and let you know the results. However, since no one does, this pr remains still.
- Recently, you were checking about the pr whose work was old. It's already a late answer by my standards. It is not ready because the work has passed. It was inconvenient to answer these questions one by one, and I understand that your confirmation is late, so I'm just closing the unprepared pr.

jjw24:
Ok if you have the latest code that is working, feel free to pr again.

Usually I am pretty quick to get back to things in GitHub, but may not always be the case that I immediately respond, if that's the case I can guarantee I will always respond though.

If you need me to look at something urgent, feel free to @ me or directly email. I will also try to check discord too.

For testing the memory usage I have looked for a better way, J just keep an eye on task manager after running the build.

Issue #168, opened 2020-10-12
Inform user when version update is not successful via windows notification


jjw24:
Flow should inform the user when an error occurs or failure to update version manually or automatically via windows notification.

Currently error is written to log.

https://github.com/Flow-Launcher/Flow.Launcher/issues/149

Zeroto521:
How about `window notification`?

jjw24:
> How about `window notification`?

What do you mean? I was thinking of a notification from Flow , the one that pops up briefly.

Zeroto521:
Like this.

![image](https://user-images.githubusercontent.com/25895405/95932561-c30a0480-0dfe-11eb-9c0a-830dcced00f2.png)

It is useful for me. It can group all notifications.

jjw24:
> Like this.
> 
> ![image](https://user-images.githubusercontent.com/25895405/95932561-c30a0480-0dfe-11eb-9c0a-830dcced00f2.png)
> 
> It is useful for me. It can group all notifications.

Is this from the windows 10 sidebar notification area?

Zeroto521:
Yes. You can see it.

![image](https://user-images.githubusercontent.com/25895405/95933135-4bd57000-0e00-11eb-8787-8ea3d02ec42a.png)

Zeroto521:
Yes. You can see it.

![image](https://user-images.githubusercontent.com/25895405/95933135-4bd57000-0e00-11eb-8787-8ea3d02ec42a.png)

Unless you start up the `focus assist`. But notifications are still can show on here.

![image](https://user-images.githubusercontent.com/25895405/95933243-acfd4380-0e00-11eb-8de6-3b3c6dc57d70.png)

jjw24:
Yeah ok. We might have to do this in two prs. I want to roll this out relatively quicker so I will just do the usual pop up notification, the second pr could add this new message toast in for all notifications from Flow, which is a bigger change, bearing in mind also of win 7, 8 support as well.

Zeroto521:
Use window API to do this, notification will have a bubble at the taskbar in win7, 8 as I remember.

PR #748, opened 2021-10-11
Redesign Popup (Msg)

onesounds:
**Before Issue**

BEFORE

- The display and disappearing animation were not natural. and It was in the form of covering the taskbar.
- There was a problem that seemed to overlap in a situation where the two overlapped.
- The design was not a common form.

ezgif

The following was used nuget.
https://github.com/Platonenkov/Notification.Wpf 

- I think It only works on Windows with notification functionality. (over windows 7)
- I tested it. and even in Windows Focus assist mode, a notification pops up.
- The color varies depending on the type, but it was unified to black. (Do we need it?)
- it has 'fontaswome' font. if we don't use, It have to remove.


onesounds:
> #168 우리는 실제로 그것을 Windows API 호출로 변경하고 싶습니다.🤣 나는 그것이 더 좋지만 약간 프레임 워크에 의존한다고 생각합니다.

I thought the same from the beginning. this nuget using windows os notification system.

PR #749, opened 2021-10-12
Fix empty query

onesounds:
for Fix the https://github.com/Flow-Launcher/Flow.Launcher/discussions/735 issue.

**Why it happen.**
- There is no problem with the logic. Just WPF UI thread is not updated first.
- https://social.msdn.microsoft.com/Forums/en-US/e974c849-90ea-4741-9a27-0ae4857bc3c2/wpf-update-ui-in-while-window-is-hidden?forum=wpf
- https://stackoverflow.com/questions/3367430/force-wpf-to-render-a-window-while-hidden

**so...**
1. force update using Dispatcher
https://stackoverflow.com/questions/1644079/change-wpf-controls-from-a-non-main-thread-using-dispatcher-invoke
It is too difficult for me.

2. Using Opacity?
- Visibility hidden or collapsed doesn't update data. Opacity 0 is possible.
- I thought opacity 0 will reserved area. 

**It is my solution**
1. When It close (there are two place, esc command and toggle window) 
2. update query data. (replace query data by setting)
3. window opacity = 0 (it just for hide delay. The user feels that the window closes quickly.)
4. 30ms delay (Time for update data. it's not showing because opacity 0)
5. close (hidden) window.

When I tested it, it worked well. feel so good.

**problem**
- I tested a number of things. I think opacity 0 doesn't reserved area. You can use this as a substitute for closing windows.
- but window visible using in everywhere. I can't change everything.
- If you close the progress bar when you see it, it remains the next time you open it.
- So tricky.

you guys will be able to trimmed more neatly.

### DIFF
--- a/Flow.Launcher/Helper/HotKeyMapper.cs
+++ b/Flow.Launcher/Helper/HotKeyMapper.cs
@@ -21,8 +21,13 @@ namespace Flow.Launcher.Helper
             mainViewModel = mainVM;
             settings = mainViewModel._settings;
 
-            SetHotkey(settings.Hotkey, mainViewModel.OnHotkey);
+            SetHotkey(settings.Hotkey, OnToggleHotkey);
             LoadCustomPluginHotkey();
+        }
+
+        internal static void OnToggleHotkey(object sender, HotkeyEventArgs args)
+        {
+            mainViewModel.ToggleFlowLauncher();
         }
 
         private static void SetHotkey(string hotkeyStr, EventHandler<HotkeyEventArgs> action)
--- a/Flow.Launcher/SettingWindow.xaml.cs
+++ b/Flow.Launcher/SettingWindow.xaml.cs
@@ -127,7 +127,7 @@ namespace Flow.Launcher
             if (HotkeyControl.CurrentHotkeyAvailable)
             {
 
-                HotKeyMapper.SetHotkey(HotkeyControl.CurrentHotkey, mainViewModel.OnHotkey);
+                HotKeyMapper.SetHotkey(HotkeyControl.CurrentHotkey, HotKeyMapper.OnToggleHotkey);
                 HotKeyMapper.RemoveHotkey(settings.Hotkey);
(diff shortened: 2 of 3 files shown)


jjw24:
what do you mean visible/hidden case?

onesounds:
> what do you mean visible/hidden case?

like 'open item' thing. In addition to esc and toggle hotkey, there are situations in which windows are closed.
I think closing should be a function and connected to everything.Flow uses the visibility code everywhere, but I haven't checked it all.

Anyway, looking at Power Toy, It was a similar way to me.
https://github.com/microsoft/PowerToys/commit/3a15276668289e77b264bf89e8b12264de89b108

I'll try copying it, but I don't know if it'll work out. If I success, I'll reopen.

taooceros:
It's similar, just await 100ms before changing to collapse. They use Thread.Sleep(100), which I am not sure whether will cause a bit slow on closing.

onesounds:
It was brought from powertoy to process toggle and updates. I think I fixed it. The opacity trick written above was also used.
Some codes are in the mainviewmodel.cs (when esc and choose item) and the rest are in the hotkeymapper. 
(alt+space toggle uses hotkeymapper) I think I have to transfer hotkeymapper classe them all, but it's too much for me.
I fixed what I saw first. It does not work when the plug-in context menu is selected. I don't know where to fix this.
I leave the rest to you guys.

onesounds:
> Hm why all hotkey mapper method is moved to mainviewmodel? Previously #565 extract those method related to hotkey from mainviewmodel to hotkey mapper. I don't think we should move it back.
> 
> I can handle the rest of the refactor if you think sometimes you encounter trouble in using the method across classes.

I didn't see that issue discussion. I didn't know why this was separated. I just thought the empty query method should be separated from hotkey. I thought it was intuitive to be by the window. 

Anyway, I referred to Power Toy. They probably organized the code better than me. I focused on changing window hiding to a unified method. I just moved what was called together while moving the import empty query. If there seems to be a problem, you can correct it. As you know, I don't know how to structure it.

PR #704, opened 2021-09-27
Modern Themes

onesounds:
- Add Highlight Styling
- Add modern themes and adjust classic themes
- Changing Result item and hotkey layout
- Change Context Menu label "Settings" to language string

Also closes https://github.com/Flow-Launcher/Flow.Launcher/issues/661

PR #754, opened 2021-10-13
Dev branch: fix issue with converters casting to null values

jjw24:
when changing themes, occasionally run into null casts:
image

image

not sure why the ListBoxItem is 'DisconnectedItem', maybe WPF bug?

repro:
go to the themes settings, select and switch themes quickly by holding down the left mouse button and slide across different themes and eventually will run into the exception.

solution:
use pattern matching 

discovered in:
#704


taooceros:
maybe the item is removed?

PR #757, opened 2021-10-16
Preview Feature

onesounds:
image

**WIP**
1. when you press f1, open preview area.
2. you can see full file path and large size thumb.

Working on it. Pr for the test.

### DIFF
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -176,6 +176,7 @@ namespace Flow.Launcher.Core.Resource
                 dict["ItemSubTitleStyle"] is Style resultSubItemStyle &&
                 dict["ItemSubTitleSelectedStyle"] is Style resultSubItemSelectedStyle &&
                 dict["ItemTitleSelectedStyle"] is Style resultItemSelectedStyle &&
+                dict["ItemHotkeyStyle"] is Style resultHotkeyItemStyle &&
                 dict["ItemHotkeySelectedStyle"] is Style resultHotkeyItemSelectedStyle)
             {
                 Setter fontFamily = new Setter(TextBlock.FontFamilyProperty, new FontFamily(Settings.ResultFont));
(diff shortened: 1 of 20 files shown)


onesounds:
image

- preview toggle is F1 key. (I think this is the best shortcut.)
- I can make it hidden when querybox empty, but I think It seems like a space that can be used. (some thing like start menu things) so I left it.
- *lnk path function is related. If hiding lnk path, additional processing is required to show here.
- Original image preview Addition should be considered.
- it should be considered like a plug-in setting window.
- When the size is flexible, when the item is changed while the preview is turned on, a distracting feeling remains. It might be better to get rid of the toggle method. (Instead, it can be changed to alt hold. it close when alt key up.)
- Related style should also be added to the theme.

jjw24:
@onesounds is this ready to go? if so could we resolve conflict please

onesounds:
> @onesounds is this ready to go? if so could we resolve conflict please

Not yet. This pr is a rough version. Modifications such as previews and shortcut keys must be made.

jjw24:
i have marked it as draft

jjw24:
> ...It seems like a space that can be used. (some thing like start menu things) so I left it.

i like this idea, it would be nice if this space offers most used results, but i dont think we can achieve this in this pr because the current model does not persist results and loading results correctly with action context is also something that needs working out. But yeah great idea.

jjw24:
hey @onesounds was there a discussion on to put the preview on top or to the side of the results? what's your opinion on the positioning of the preview window?

onesounds:
> hey @onesounds was there a discussion on to put the preview on top or to the side of the results? what's your opinion on the positioning of the preview window?

There were no other opinions. CitizenDee said He wish he could always turn it on. 

It is recommended that the default position be placed on the right. (like spotlight on mac) However, in that case, various problems such as window size and vertical size arise. Since I can't think of a neat way, I'll think more about this pr. We will work again after the problem of reducing the lnk path path and other design prs are organized.

PR #758, opened 2021-10-17
Change the popup to windows notification

onesounds:
Change the popup to windows notification
https://github.com/Flow-Launcher/Flow.Launcher/pull/748


taooceros:
hm is it possible to not use the Msg class? It is inherited from Window class to become a window. However, if we use toast, no need for that. Maybe change the show code to another place.

PR #763, opened 2021-10-19
Tray Menu Redesign

onesounds:
image

**Problem**
too small.

**Revised history**
- An unselectable header was added for distinction. It's a menu that I use often, so I think it don't need an "About Menu." It interferes with the location selection of other menus. If necessary, it would be nice to include the number of versions. (e.g Flow Launcher 1.9)
- Replace left click with screen on/off and right click with context menu.
- Existing contextmenustrip was used, but the design could not be adjusted, so contextmenu was added. Existing contextmenustrip was maintained because it had the ability to close contextmenu when clicking other areas.
- When the language is changed, the code of the updated part works, but it is not neat. Modifying is always welcome.
- The design is the same as the existing context menu. In the future, it may be necessary to change the color according to the system setting (dark mode, etc.). It will be implementable because there appears to be a code that can fetch the system color. This part will be postponed later.
- Fix empty query PR is included in the code due to the window closing issue. It should be applied after the Fix Empty query is completed.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -162,10 +162,6 @@ namespace Flow.Launcher
         private void UpdateNotifyIconText()
         {
             var menu = contextMenu;
-            //var header = menu.Items[0];
-            //var open = menu.Items[1];            
-            //var setting = menu.Items[2];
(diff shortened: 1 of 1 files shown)


onesounds:
https://github.com/Flow-Launcher/Flow.Launcher/pull/765

PR #767, opened 2021-10-22
Add width in setting and fix max result displayed not reflected in real-time

onesounds:
enhancements:
- ignore width in theme
- min 400, max 900. default 580, tick 10
- Reflected in real time when the window is on. 
- also address the issue where max result displayed by the query window not reflected in real-time

**outstanding issue**
- [x] It's not reflected in the preview window. (I didn't make it. Do we need it?)
- [x] If you turn off shadow mode, there are some themes that are not reflected.
- [x] Some themes work even when shadows are off.
- [x] when I test it,  even though the content is the same theme, (I copied metro server to light) something works and something not works.

close #766


Flow.Launcher.Core/Resource/Theme.cs
```diff
@@ -189,22 +189,11 @@ public ResourceDictionary GetResourceDictionary()
...
-            if (width == null)
-            {
-                windowStyle = dict["BaseWindowStyle"] as Style;
-
-                width = windowStyle?.Setters.OfType<Setter>().Where(x => x.Property.Name == "Width")
-                .Select(x => x.Value).FirstOrDefault();
-            }
-
+            var width = Settings.WindowSize;
+            windowStyle.Setters.Add(new Setter(Window.WidthProperty, width));
```
taooceros, line 195: we don't need that right?

onesounds: maybe?

taooceros: Have you remove all the width setting in the theme?

onesounds: I didn't do it yet. Should I?

taooceros: Isn't this pr does that job? https://github.com/Flow-Launcher/Flow.Launcher/pull/755

onesounds: I did so only for the elements inside the window. The themes currently pushed are fixed in window size.

taooceros: ok nvm I remember the setter will be override by direct setting the width. Therefore, it shouldn't be a problem.

jjw24:
@onesounds @taooceros could we resolve the conflict in this one please

Flow.Launcher.Core/Resource/Theme.cs
```diff
@@ -376,7 +365,6 @@ private void SetWindowAccent(Window w, AccentState state)
             var windowHelper = new WindowInteropHelper(w);
 
             // this determines the width of the main query window
```
jjw24, line 367:
```suggestion
```

jjw24 (changes requested):
looks good. let's get the conflict resolved and merge in

taooceros:
the conflict is a bit hard to resolve. I will do that tonght.

onesounds:
> the conflict is a bit hard to resolve. I will do that tonght.

It's easy for me to do because I did it when I made my version.
I don't know if it's okay to do it like this. anyway I merged with dev and pushed again.

taooceros:
> > the conflict is a bit hard to resolve. I will do that tonght.
> 
> It's easy for me to do because I did it when I made my version. I don't know if it's okay to do it like this. anyway I merged with dev and pushed again.

I think you have done it right?
Let me take a try.

taooceros:
I add auto reformat from plugin https://github.com/Xavalon/XamlStyler

jjw24 (approved):

PR #773, opened 2021-10-27
Guide under result & Hotkey Guide

onesounds:
This PR contains three functions.

1. Small Tip under result list. 

I choose 3 hotkeys. 
- Open Context menu (Shift+Enter) and Right Arrow & Open Containing Folder in result list. (Ctrl+Enter)
- Back to Result in context menu (ESC and Left Arrow)

The display content varies depending on whether the result list/context menu is present.

ctrl+o is left intact because of existing users, 
but I think it is better to exclude it from the guide because it's too small section. This guide on/off by setting>Theme Section . Default is on.

2. Hotkey Popup Guide 
When Press F1 in search box, (The preview is planned to be F2) small popup open. It is located on the right side of the query box.
It automatically close when 1. open selected item 2. close main window 3. focus out. 4. F1 (it's toggle)
Users can turn it on at any time to view the result list along with the shortcut list.

3. Start Message

However, the user does not know that F1 displays a shortcut list. I decided to display it in the query window because I had to show somewhere.

This disappears the moment the query is entered, and is displayed only when the query length is 0 and there is no result window.

**Todo List**
- [x] Change text to string.
- [x] Switch the shortcut list XML to suit the existing code.
- [x] Check the list of shortcuts to display.
- [x] [Bug] Shadow and Round Corner
- [ ] Check the English expression.
- [x] [Bug] When you turn on the shadow, the hot key list window is displayed.
- [x] [Bug] There are cases where the hot key list window is not closed.
- [ ] additional UI & Design in Hotkey Guide Popup (recommend command, open guide document window)
- [x] Disable the selection function the hotkey list
- [x] Rename Methods
- [x] [Bug] Small Tip guide doesn't switch on time
- [x] Don't show the start message in the result window.
- [x] Add Beginner's guide on/off to setting
- [x] Check Hotkey popup position various case
- [x] Add some color in themes (It is necessary to separate color-related information from the theme individually. Working on another pr.)
- [x] Separate template code.


onesounds:
Done. It need a test.

- The start message (press f1...) is always displayed. You can't turn it off with settings. (Is it better to turn it off?) If you enter a query, it disappears.
- A tip message is displayed at the bottom and there are two types. It is switched in two cases: ContextMenu/Result. You can turn it off in settings>Theme. The default setting is on.
- Press f1 to display a list of all available hot keys. If you press it one more time, it turns off.
- There was a problem with the round border processing, so I changed some of the theme xaml and added color information for each guide phrase.
- History, Result, and ContextMenu have been combined into the BorderClip to prevent the code too long.
- Additional styles have been inserted so that the result or contextmenu does not overlap.

onesounds:
This style is pretty, but the code is getting too long....

onesounds:
- Add MenuPanel in Querybox. (from listary) If it is not a result window, display the name of the open list. (Action / History)

https://user-images.githubusercontent.com/6903107/140303942-42d3f263-2edc-4542-a411-25365a370b4f.mp4

WIP : Organizing theme templates.

onesounds:
This pr is too old. There are many changes in other prs. Since it overlaps with the preview function, it should be considered together, and some functions (bug fixes that open results overlapping and display menu panel) should be treated as separate pr.

PR #780, opened 2021-10-29
quick exit when children property is not found

taooceros:
Sometimes the bookmark file may not be constructed property, leading the issue. This patch add a check for that to safely handle that.


taooceros:
@kubalav If you have time you can take a test one it. Though I think after you have added one bookmark, the file structure shouldn't be the issue.

kubalav (approved):

PR #781, opened 2021-10-30
Add OnCultureInfoChanged to IPluginI18n

taooceros:
Add OnCultureInfoChanged to IPublicI18n to allow plugin do action when change language.

It is added with default implementation for interface for backward compatibility (so optional).


taooceros:
With the commit, the windowssetting plugin is truly internationalized.

onesounds:
It was confirmed that the function worked well in Korean language.

PR #784, opened 2021-10-31
Force Exit & Fix HideonStartup (#783)

### LINKED ISSUE #783: "Hide Flow Launcher on startup" doesn't work in prerelease [1.9.0]
If "Hide Flow Launcher on startup" is checked, Flow Launcher is still visible.

onesounds:
Fix #783 

1. There is no problem with 'Hide on startup' and config
2. It related https://github.com/Flow-Launcher/Flow.Launcher/pull/749
3. The Hide() function uses the toggleFlowLauncher function. (Copied from PT)
4. Hide() run when start of the flow in somewhere. Since this function was wearing a toggle, it was a problem that the hidden state was toggled again at the beginning.
5. Power Toy seems to have had something different because it doesn't have Startup Setting.
6. I changed Hide() to just close window.

There was no problem when I tested it, but I don't know if it'll be okay because I didn't understand the code exactly. Please check.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -185,7 +185,7 @@ namespace Flow.Launcher
             var setting = items.Add(InternationalizationManager.Instance.GetTranslation("iconTraySettings"));
             setting.Click += (o, e) => App.API.OpenSettingDialog();
             var exit = items.Add(InternationalizationManager.Instance.GetTranslation("iconTrayExit"));
-            exit.Click += (o, e) => Environment.Exit(0);
+            exit.Click += (o, e) => Close();
 
             _notifyIcon.ContextMenuStrip = menu;
             _notifyIcon.MouseClick += (o, e) =>
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -53,7 +53,7 @@ namespace Flow.Launcher
             _viewModel.Save();
             e.Cancel = true;
             await PluginManager.DisposePluginsAsync();
-            Application.Current.Shutdown();
+            Environment.Exit(0);
         }
 
         private void OnInitialized(object sender, EventArgs e)
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -273,7 +273,7 @@ namespace Flow.Launcher.ViewModel
             ReloadPluginDataCommand = new RelayCommand(_ =>
             {
                 Hide();
-                
+
                 PluginManager
                     .ReloadData()
                     .ContinueWith(_ =>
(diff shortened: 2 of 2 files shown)


taooceros (commented):
Isn't we want to do something here like clear the query box? That code is within the toggleflowlauncher.

onesounds:
> Isn't we want to do something here like clear the query box? That code is within the toggleflowlauncher.

It's been more than 24 hours since I didn't sleep when I was working on that code. 😂
1. Move the 'Emptyquery' from Toggle to the Hide
2. Toggle Using Hide when close window.

How about this?

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -185,7 +185,7 @@ private void InitializeNotifyIcon()
             var setting = items.Add(InternationalizationManager.Instance.GetTranslation("iconTraySettings"));
             setting.Click += (o, e) => App.API.OpenSettingDialog();
             var exit = items.Add(InternationalizationManager.Instance.GetTranslation("iconTrayExit"));
-            exit.Click += (o, e) => Close();
+            exit.Click += (o, e) => Environment.Exit(0);
```
taooceros, line 188: For this one, don't change. Calling Close() will allow saving normal process of saving setting.

onesounds: OK. fixed.

onesounds: When executing 'open maintaining folder' on the file plug-in, the problem in which 'lastemptyquery' does not work is also fixed to this pr.

jjw24: @onesounds should this be Close() not environment.exit acording to @taooceros ?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -704,33 +704,30 @@ public async void ToggleFlowLauncher()
...
-                    default:
-                        throw new ArgumentException($"wrong LastQueryMode: <{_settings.LastQueryMode}>");
-                }
-                MainWindowVisibility = Visibility.Collapsed;
+                Hide();
             }
         }
 
-        public void Hide()
+        public async void Hide()
```
jjw24, line 711: why add async here? @onesounds / @taooceros

taooceros: for handling the task delay for fixing empty query issue.

taooceros (approved):

Flow.Launcher/MainWindow.xaml.cs
onesounds: yes. If I use "environment.ext" here, the process ends immediately, so it don't have time to save settings. If I do as Taoo said, close() will call onclosing(), so it will exit after the save is completed.

This part is not related to this pr topic, just there is a problem that the process continues to remain, so I applied it.

PR #787, opened 2021-11-02
Add search sources Netflix and Youtube Music for WebSearch plugin

jjw24:
Add search sources Netflix and Youtube Music for WebSearch plugin

todo:
backwards compatibility


onesounds:
This work seems more complicated than I thought. I don't mind if this doesn't work. I think the importance is low.

jjw24:
No longer required.

PR #804, opened 2021-11-10
Added trim support to reduce size of installation & spellfix Caculator

XavBryant:


taooceros:
@XavBryant Could you make any change just to trigger the CI?

taooceros:
I check the nuget package for Microsoft.Tools.Trimings, and it hasn't been updated since 2018. Therefore, I think it would be better to use the publish triming. https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained#:~:text=Trimming%20is%20only%20supported%20in.NET%206%2B.%20Components%20that,external%20dependencies%20that%20aren%27t%20visible%20at%20build%20time.

XavBryant:
I have tested that implementation and it does not work. I have got the
publishing tools to work using the nuget package and it does work. You are
using the 'if update then do' which in the world of software... is more of
a hope than a requirement. For instance, as to my knowledge we are not
updating to .net6 yet. I can look at it again, but if it ain't broke...
don't break it.

taooceros:
Could you take a check on the error popped out in the CI build?

XavBryant:
cl build?

jjw24:
it's coming from the bumping the nuget package to 5.11
https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/41525939

Flow.Launcher/Properties/Settings.Designer.cs
```diff
@@ -12,7 +12,7 @@ namespace Flow.Launcher.Properties {
     
     
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "16.4.0.0")]
```
jjw24, line 15: why is this checked in?

jjw24:
i noticed the userdata's folder name has not been corrected:

image

but if we correct this then we need to provide backwards compatibility for users that may have change from the default option in the plugin settings.

jjw24:
@XavBryant i recommend separating the two changes because we may need to test the trim support for a bit assuming this is the change to trim down the release package size

XavBryant:
all I did was describe how to do the trim. id suggest someone else
implement trim as I would have to redownload installation.

ericstj:
`Microsoft.Packaging.Tools.Trimming` was an early experiment we did to demonstrate the possibilities of assembly level trimming.  This predated official support.

You should use the built-in trimming support: https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options

taooceros:
> `Microsoft.Packaging.Tools.Trimming` was an early experiment we did to demonstrate the possibilities of assembly level trimming. This predated official support.
> 
> You should use the built-in trimming support: https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options

Hi, the issue with the built in support of trimming is that it cannot run properly for wpf apps, while the experimental trimming tool seems working.

PR #807, opened 2021-11-11
Remove Unneeded Project Reference for plugins

taooceros:
A complete refactor branch. Currently all plugin is referencing the infrastructure, which creates a lot unnecessary build when modifying.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.BrowserBookmark/Commands/BookmarkLoader.cs
+++ b/Plugins/Flow.Launcher.Plugin.BrowserBookmark/Commands/BookmarkLoader.cs
@@ -1,6 +1,5 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
-using Flow.Launcher.Infrastructure;
 using Flow.Launcher.Plugin.BrowserBookmark.Models;
 using Flow.Launcher.Plugin.SharedModels;
 
@@ -8,13 +7,15 @@ namespace Flow.Launcher.Plugin.BrowserBookmark.Commands
 {
     internal static class BookmarkLoader
     {
+        public static IPublicAPI publicAPI { get; internal set; }
+
         internal static MatchResult MatchProgram(Bookmark bookmark, string queryString)
         {
-            var match = StringMatcher.FuzzySearch(queryString, bookmark.Name);
+            var match = publicAPI.FuzzySearch(queryString, bookmark.Name);
             if (match.IsSearchPrecisionScoreMet())
                 return match;
 
-            return StringMatcher.FuzzySearch(queryString, bookmark.Url);
+            return publicAPI.FuzzySearch(queryString, bookmark.Url);
         }
 
         internal static List<Bookmark> LoadAllBookmarks(Settings setting)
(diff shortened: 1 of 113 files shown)


jjw24:
got this error

image

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Commands/BookmarkLoader.cs
```diff
@@ -1,20 +1,21 @@
...
 using System.Linq;
-using Flow.Launcher.Infrastructure;
 using Flow.Launcher.Plugin.BrowserBookmark.Models;
 using Flow.Launcher.Plugin.SharedModels;
 
 namespace Flow.Launcher.Plugin.BrowserBookmark.Commands
 {
     internal static class BookmarkLoader
     {
+        public static IPublicAPI publicAPI { get; internal set; }
```
jjw24, line 10: why we referencing this interface, should it be via context-> PublicAPIInstance?

Plugins/Flow.Launcher.Plugin.Calculator/LocalizationHelper/LocalizationConverter.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.ComponentModel;
+using System.Globalization;
+using System.Reflection;
+using System.Windows.Data;
+
+namespace Flow.Launcher.Plugin.Caculator.LocalizationHelper
```
jjw24, line 7: how come this and other file is added, dont they already exist?

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Commands/BookmarkLoader.cs
taooceros: That's because I think this file doesn't need the access for context, so just add the api here.

Plugins/Flow.Launcher.Plugin.Calculator/LocalizationHelper/LocalizationConverter.cs
taooceros: I originally want to remove the dependency on flow.plugin.core here, so move that file here. But it's not able to do that easily so revert the change, but the file is kept unexpectedly.

jjw24: ok lets revert this please if it is unexpected

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Commands/BookmarkLoader.cs
```diff
@@ -1,20 +1,21 @@
...
+
         internal static MatchResult MatchProgram(Bookmark bookmark, string queryString)
         {
-            var match = StringMatcher.FuzzySearch(queryString, bookmark.Name);
+            var match = publicAPI.FuzzySearch(queryString, bookmark.Name);
```
jjw24, line 14: are you sure using this here is going to work?

jjw24: because i just tested with similar code, it throw a null reference exception

taooceros: I guess I forget to assign the api🤣. Will check later. These few days are busy...🥲

PR #810, opened 2021-11-12
Add Animation and Sound / Fix Last Executed Position

onesounds:
https://user-images.githubusercontent.com/6903107/141396629-d677c1ed-1af5-4564-bd53-aabc2347ac22.mp4

- [x] Add Slide up animation when open flow
- [x] Fix Remember Last launch position when Relaunch Flow
  - There is a problem that the Y position is mysteriously changed, but it was revised to the code below. (I don't know for sure why it works.) I think Size OfChanged is the reason. It is saved in the settings based on where the program was terminated. I think it would be safer to save it whenever the location changes.
  - **Refs**
  - https://github.com/Wox-launcher/Wox/pull/960/commits/2af751418c77f27775fe39c9a82560df20d7894c
  - https://github.com/Wox-launcher/Wox/issues/823
- [x] On/Off Option
- [x] Sound Effect

**ETC**
- [x] Fix Blink Animation when close setting window
- [x] Change 'Open' tray menu to toggle
- [x] Change Hide/Open Code with out using visibility base
- [x] Reduce Height of setting window <Strike>Change Setting window Height to responsible (80% of Screen Height)</Strike>


onesounds:
If we apply this pr, we should all use this code later. Visibility should not be used directly as possible, but should be toss to animation code.

jjw24:
when started up, the search bar's position is moved to the right, and then readjusted after one deactivate-reactivate

jjw24:
we are adding sound?

onesounds:
> we are adding sound?

yeah, but not default. Since many people ask the same question, I will change the default state to on. It is possible to turn on/off in theme.

Even though you did as above, If you can't hear the sound, there is a problem with the sound file build. I have fixed this problem several times, but sound is not added to the resource folder for some reason.

onesounds:
> when started up, the search bar's position is moved to the right, and then readjusted after one deactivate-reactivate
fixed it.

jjw24:
not yet, it works on first start up, but if you exit then start it up again, it will go to the right.

onesounds:
> not yet, it works on first start up, but if you type `exit` to exit flow then start it up again, it will go to the right.

I fixed it one more time. I tested several cases, but it didn't seem to be a problem.I hope I succeed this time.🙏

jjw24:
No problem, will test it later

jjw24:
> > not yet, it works on first start up, but if you type `exit` to exit flow then start it up again, it will go to the right.
> 
> I fixed it one more time. I tested several cases, but it didn't seem to be a problem.I hope I succeed this time.🙏

fixed, thank you

jjw24:
when trigger show query window after opening Settings window,  the query window flickers once.

repro:
1. open Settings window
2. trigger flow hotkey to show query window
3. observe it flickers once

it also does flicker once after closing settings window, after which it will be back to normal.

jjw24:
flickering is from the animation, turning it off doesnt have the same issue.

onesounds:
> when trigger show query window after opening Settings window, the query window flickers once.
> 
> repro:
> 
> 1. open Settings window
> 2. trigger flow hotkey to show query window
> 3. observe it flickers once
> 
> it also does flicker once after closing settings window, after which it will be back to normal.

There were two problems.
1. When opening the setting window, the main window was closed with the old Hide(). I can close the new Hide(), but the code was removed because it closes when it is deactive anyway.

2. When dragging or closing the setting window, the main window's inactive immediately worked. I added delay in deactive code. because I needed time to initialize the animation.

Flow.Launcher/ResultListBox.xaml
```diff
@@ -30,55 +37,102 @@
...
+                            BorderBrush="Transparent"
+                            BorderThickness="0">
+                            <Image
+                                x:Name="ImageIcon"
+                                Width="32"
+                                Height="32"
+                                Margin="0,0,0,0"
+                                HorizontalAlignment="Center"
+                                Source="{Binding Image}"
+                                Stretch="Uniform"
```
onesounds, line 90: It's not related to this pr, but I added it as a minor correction. Corrects the problem of being cut when the icon size is large (cloud icon of the e.g plug-in manager). When I fixed this part before, the icon was small, so there seemed to be no problem.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -267,14 +307,18 @@ private void OnPreviewDragOver(object sender, DragEventArgs e)
...
+            await Task.Delay(100);
             App.API.OpenSettingDialog();
         }
 
 
-        private void OnDeactivated(object sender, EventArgs e)
+        private async void OnDeactivated(object sender, EventArgs e)
         {
+            _viewModel.Save();
+            await Task.Delay(100);
```
jjw24, line 321: > When dragging or closing the setting window, the main window's inactive immediately worked. I added delay in deactive code. because I needed time to initialize the animation.

You are referring to this delay right?

jjw24:
```suggestion
            // need time to initialize the main query window animation
            if (_settings.UseAnimation)
                await Task.Delay(100);
```

jjw24: we can add a condition since this is only a requirement for animation.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -292,19 +339,37 @@ private void UpdatePosition()
             {
                 Left = WindowLeft();
                 Top = WindowTop();
+
```
jjw24, line 342: remove extra spacing

```suggestion
```

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -292,19 +339,37 @@ private void UpdatePosition()
...
             }
         }
 
+
```
jjw24, line 346: remove extra spacing
```suggestion
```

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -292,19 +339,37 @@ private void UpdatePosition()
...
+            UpdatePosition();
+            if (_settings.HideOnStartup)
+            {
+                _viewModel.Hide();
+            }
+            else
+            {
+                _viewModel.Show();
+            }
+
```
jjw24, line 369: remove extra spacing
```suggestion
```

Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -44,6 +44,7 @@ private void OnLoaded(object sender, RoutedEventArgs e)
             HwndSource hwndSource = PresentationSource.FromVisual(this) as HwndSource;
             HwndTarget hwndTarget = hwndSource.CompositionTarget;
             hwndTarget.RenderMode = RenderMode.SoftwareOnly;
+
```
jjw24, line 47: remove extra spacing
```suggestion
```

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -708,36 +711,56 @@ private void SetOpenResultModifiers()
...
+                    await Task.Delay(100); //Time for change to opacity
                     break;
                 case LastQueryMode.Preserved:
+                    MainWindowOpacity = 0;
+                    await Task.Delay(100);
                     LastQuerySelected = true;
                     break;
                 case LastQueryMode.Selected:
+                    MainWindowOpacity = 0;
+                    await Task.Delay(100);
```
jjw24, line 755: are these Task.Delay(100) related to animation too?

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
```diff
@@ -82,6 +82,9 @@ public CustomExplorerViewModel CustomExplorer
             }
         };
 
+        public bool UseAnimation { get; set; } = true;
+        public bool UseSound { get; set; } = true;
+
```
jjw24, line 87: remove extra spacing
```suggestion
```

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -708,36 +711,56 @@ private void SetOpenResultModifiers()
...
                     LastQuerySelected = false;
                     break;
                 default:
                     throw new ArgumentException($"wrong LastQueryMode: <{_settings.LastQueryMode}>");
             }
+            WinToggleStatus = false;
             MainWindowVisibility = Visibility.Collapsed;
+
```
jjw24, line 763: add new space
```suggestion

            WinToggleStatus = false;
            MainWindowVisibility = Visibility.Collapsed;
```

Flow.Launcher/Flow.Launcher.csproj
```diff
@@ -77,6 +77,7 @@
       <SubType>Designer</SubType>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <None Remove="Resources\open.wav" />
```
jjw24, line 80: this can be removed
```suggestion
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -83,6 +83,10 @@
     <system:String x:Key="theme_load_failure_parse_error">Fail to load theme {0}, fallback to default theme</system:String>
     <system:String x:Key="ThemeFolder">Theme Folder</system:String>
     <system:String x:Key="OpenThemeFolder">Open Theme Folder</system:String>
+    <system:String x:Key="SoundEffect">Sound Effect</system:String>
+    <system:String x:Key="SoundEffectTip">Play a small sound when window open</system:String>
```
jjw24, line 87:
```suggestion
    <system:String x:Key="SoundEffectTip">Play a small sound when the search window opens</system:String>
```

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -235,6 +237,44 @@ private void InitProgressbarAnimation()
...
+                Storyboard.SetTargetProperty(da2, new PropertyPath(Window.TopProperty));
+                sb.Children.Add(da);
+                sb.Children.Add(da2);
+                sb.Completed += (_, _) => _animating = false;
+                _settings.WindowLeft = Left;
+                _settings.WindowTop = Top;
+                sb.Begin(FlowMainWindow);
+            }
+        }
+
```
jjw24, line 276: remove spacing
```suggestion
```

Flow.Launcher/ViewModel/MainViewModel.cs
onesounds: yep

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -267,14 +306,20 @@ private void OnPreviewDragOver(object sender, DragEventArgs e)
...
+            _viewModel.Hide();
+            await Task.Delay(100);
             App.API.OpenSettingDialog();
         }
 
 
-        private void OnDeactivated(object sender, EventArgs e)
+        private async void OnDeactivated(object sender, EventArgs e)
         {
+            _viewModel.Save();
```
jjw24, line 319: why are we calling save here everytime search window is not in focus/deactivated? that's too excessive. This save call does writes to disk.

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: hmm, i am finding these task delays are causing Hide to be called twice when toggle the search window very quickly three times.
repro:
1. turn off animation
2. hide search window
3. quickly press the toggle hotkey 3 times
4. the expected behaviour should be show->hide->show, but it's coming up as show->hide->show->hide

you can use Debug.WriteLine("Hide") to write out to output.

Flow.Launcher/MainWindow.xaml.cs
onesounds: This is a mistake. It was put in the middle of the work for testing. I thought I deleted it.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -720,45 +698,42 @@ public void ToggleFlowLauncher()
...
         public async void Hide()
         {
-            MainWindowOpacity = 0;
             switch (_settings.LastQueryMode)
             {
                 case LastQueryMode.Empty:
                     ChangeQueryText(string.Empty);
                     MainWindowOpacity = 0; // Trick for no delay
-                    await Task.Delay(100); //Time for change to opacity
+                    await Task.Delay(50); //Time for change to opacity
```
onesounds, line 717: As you know, in this case, there should be a delay regardless of animation because the query should be erased.

Flow.Launcher/ViewModel/MainViewModel.cs
onesounds: done. If the animation setting is off, there is no delay.

onesounds:
If you toggle continuously very quickly with the animation setting on, there are cases where the frame is still wrong. (Blinking momentarily when the flow is turned on in this state.) It needs to be very fast, I think it's an unusual situation, and I think it's okay to leave it as it is.

Overall, since the delay time was reduced, the response thought to be a little slow would have decreased.

jjw24:
looks like the extra hide is gone when switching on off three times quickly.

the old problem where it flickers when Settings window is opened or closed is back.

onesounds:
I repeated the correction several times. However, a problem still occurred when the animation was turning on. (This is because the initialization time does not work when the animation setting is turned off.) After working on it a few times, I just handled the window show/hiding to be applied unconditionally only by the opacity, and deleted all delays and visible/collapsed/hidden.

In this case, there is only one problem expected. When alt tab is performed, an empty window exists. This can be modified to another code. First of all, we are working on it, and it is necessary to review whether this method will be okay.

onesounds:
> looks like the extra hide is gone when switching three times quickly.
> 
> the old problem where it flickers when Settings window is opened or closed is back.

This problem occurs when the window is opened in the "animation off state" and changed back to the "animation on state". This is because when the window is turned on in the animation off state, there is no delay, so it remains in the opacity 1 state. In other words, problems arise because it is not initialized when closing the window.

Therefore, the window hiding during ondeactivation was modified to add a delay regardless of animation setting.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -267,29 +267,42 @@ public void WindowAnimator()
...
                 Storyboard.SetTargetProperty(da, new PropertyPath(Window.OpacityProperty));
                 Storyboard.SetTargetProperty(da2, new PropertyPath(Window.TopProperty));
+                Storyboard.SetTargetProperty(da3, new PropertyPath(TopProperty));
                 sb.Children.Add(da);
                 sb.Children.Add(da2);
+                iconsb.Children.Add(da3);
                 sb.Completed += (_, _) => _animating = false;
                 _settings.WindowLeft = Left;
                 _settings.WindowTop = Top;
+                iconsb.Begin(SearchIcon);
```
jjw24, line 305: is this change itended?

onesounds: yeah, for Icon animation.

onesounds:
Is there anything else you want me to fix?
There is still an issue that flashes when setting is changed in a specific situation, but I don't think there is a problem with actual use. I think it's a problem that you can just go. This part is difficult to fix.

jjw24:
i am still doing some testing with it and making minor updates, will push up later.

jjw24:
what is WinToggleStatus used for exactly?

onesounds:
> what is WinToggleStatus used for exactly? Seems to function the same as Visibility check

Logically, there's no problem. but there was a blinking problem. I added a variable for toggle separately because I saw an advice not to use visibility as a trigger when using this kind of animation. and I felt that the data was modulated because there were codes everywhere that changed visibility.

I tested it once in the middle of this PR, and if all the visibility code is changed to opacity = 0 or 1, it will be updated even if it is not visible on the screen, so it will work even after removing all the delays, and logically, it will all work properly if I use trigger base visibility. In this method, a window exists in alt+tab, and consideration was canceled due to the problem of receiving input even if it was not visible on the screen. If we solve the alt tab and focus problem, I think this method is also good.

In principle, if the UI is updated even in the hidden/visible state, the animation can be initialized, the delay can be eliminated, and all blinking problems disappear. It's the same issue as lastemptyquery. These codes were created because this method was not available. To solve a number of problems existing in this code, a method of updating the UI even if a window is hidden on the screen must be implemented first. Things such as background worker, dispatcher, and invoke are needed.

Flow.Launcher/MainWindow.xaml.cs
jjw24: The search icon seems to just jump up a little as if to reposition itself to the right height. Is this the same case for you?

jjw24 (approved):

jjw24:
I just did some refinements based on extensive testing, mainly around delay and show/hide calls. In some extreme scenarios as you mentioned as well, there are still flickering, but those edge cases aside I think this now is good.

jjw24:
Some points as a result of testing this change I want to record below for future reference incase needed:

- MainWindowOpacity is used to stop from mainwindow slow flickering when used with animation

- Task.Delay(100) is used to stop from mainwindow fast flickering when used with animation

- Application.Current.MainWindow.Hide(); is removed from SingletonWindowOpener class because it caused flickering when Setting window is opened

- Toggling mainwindow to show immediately after opening Setting window when animation is off triggers an extra OnDeactivated call at the end, causing the mainwindow to be hidden. This could be happening due to Settings window loading at the same time. The Application.Current.MainWindow.Hide() was used to combat this so with it the mainwindow will actually show as intended. However using this caused flickering with animation when Settings window is opened, so it's removed.

- toggling very quickly on off mainwindow, with animation on, the Hide() is called sometimes instead of Show(), so the hide->show->hide->show sequence is not exactly honoured. toggling slower will not have this issue, neither does doing it without animation. This is likely due to the Task.Delay(100).

PR #817, opened 2021-11-15
Dark mode

onesounds:
image

I didn't want to work, but there will be steady requests in the long run, and if I don't do it now, what I have to continue working on will accumulate. So I'm working on it now.

- [x]  Add Custom Title Bar
- [x]  Add Dark Mode Setting Window
- [x]  Add OS detection Theme
- [x]  OS Dark mode detection 
- [x]  Add custom dark mode setting
- [x]  Remove Hardcoded Color in setting window
- [x]  Remove Hardcoded Color in custom control template
- [x]  Fix Window State issue (maximize/restore)
- [x]  Redesign About Panel
- [x]  Adjust TabHeader Design
- [x]  Apply all popup window
- [x]  Fix Window Deactive Style
- [x]  Adjust Plugins setting panel

Optional
- [ ]  Add High Contrast Setting Window Theme


onesounds:
I'm planning to rework.

onesounds:
There is only one item in the theme that uses Solid Color Brush.

**Ref**
https://stackoverflow.com/questions/45740435/how-to-update-solidcolorbrush-color-with-a-dynamicresource

In this case, there is a problem that does not apply when switching System Theme. In the case of System theme, the theme must be re-applied. Switching to style solves the problem, but it was very difficult to use style in templates. System Theme treated Light and Dark as using common colors.

Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -297,5 +311,60 @@ private void OnExternalPluginInstallClick(object sender, RoutedEventArgs e)
...
+                ModernWpf.ThemeManager.Current.ApplicationTheme = ModernWpf.ApplicationTheme.Dark;
+            }
+        }
+
+        /* Custom TitleBar */
+
+        private void OnMinimizeButtonClick(object sender, RoutedEventArgs e)
+        {
+            this.WindowState = WindowState.Minimized;
+        }
```
taooceros, line 332: why we need to add this😶‍🌫️

Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -297,5 +311,60 @@ private void OnExternalPluginInstallClick(object sender, RoutedEventArgs e)
...
+            else
+            {
+                this.WindowState = WindowState.Maximized;
+            }
+        }
+
+        private void OnCloseButtonClick(object sender, RoutedEventArgs e)
+        {
+            this.Close();
+        }
```
taooceros, line 349: Same, also close button possibly better to not close? Let me check later

Flow.Launcher/SettingWindow.xaml.cs
onesounds: because titlebar cannot change the color. So I had to remove default titlebar and made custom titlebar. In this case, there is no button,(maximize, close)  so it must be made. Other pop-up windows were applied like this.

onesounds:
I corrected all the parts that were problematic. If nothing else is found, we can merge it.

jjw24 (approved):

PR #823, opened 2021-11-18
Detect URL and install using Plugin Manager

Garulf:
image

`pm install <URL>` will allow a user to install a plugin directly from an URL bypassing the plugin store.

Should be useful for debugging and sharing a plugin that may not be ready or appropriate for the plugin store.

Option only appears when a formatted URL is detected.

User will be prompted if the downloaded plugin zip contains a plugin that already exists and is of same or lesser version 

Added an option to warn user when installing from a source not in github and not by an author from the plugin store.

Also added logic to load only unique latest version plugins


Garulf:
Strange... does Flow not restart when in debug mode?

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -302,6 +302,38 @@ internal List<Result> Search(IEnumerable<Result> results, string searchName)
...
         }
 
+        internal List<Result> InstallFromWeb(string url)
+        {
+            var fileName = url.Split("/").Last();
+            var filePath = Path.Combine(DataLocation.PluginsDirectory, fileName);
+            var plugin = new UserPlugin
+            {
+                ID = "",
+                Name = fileName.Split(".").First(),
```
taooceros, line 312: I think it would be better to simply trim the last part (.zip). Also, it is possible that the link doesn't contain file name. Though, I think this won't be a large issue since this feature is mainly for debugging.

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -372,8 +408,8 @@ private void Install(UserPlugin plugin, string downloadedFilePath)
                 MessageBox.Show(Context.API.GetTranslation("plugin_pluginsmanager_install_errormetadatafile"));
                 return;
             }
-
-            string newPluginPath = Path.Combine(DataLocation.PluginsDirectory, $"{plugin.Name}-{plugin.Version}");
+            var directory = String.IsNullOrEmpty(plugin.Version) ? $"{plugin.Name}" : $"{plugin.Name}-{plugin.Version}";
```
taooceros, line 411: this may create some issue. I don't think flow can handle the plugin with same name. @jjw24 any potential solution for this?

jjw24: to solve the problem of downloading plugin of the same name you can append the zip name with a random guid, but will need to test how flow loads it up and how to uninstall duplicate plugins

jjw24:
this will potentially allow for downloading duplicate plugins right?

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
Garulf: Could I just run the `Uninstall` method with the filename before the `Install` method?

jjw24: You still can run uninstall after install because, in the end it will require a restart. the uninstall places a needdelete.txt file on the plugin to uninstall, so when restarted the to be deleted plugin will be deleted, and your new downloaded plugin will be loaded.

only thing is if you going to do this, might need to prompt the user to let them know, because they might want to try the newer plugin but still wants to keep their existing.

taooceros:
> this will potentially allow for downloading duplicate plugins right?

not really. ID check will update old to new

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
Garulf: So in it's current state if one tries to re-download the same zip flow shows there was an error installing. You can uninstall normally.

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -104,7 +104,7 @@ internal List<Result> GetDefaultHotKeys()
 
         internal async Task InstallOrUpdate(UserPlugin plugin)
         {
-            if (PluginExists(plugin.ID))
+            if (PluginExists(plugin.ID) || Directory.Exists(Path.Combine(DataLocation.PluginsDirectory, plugin.Name)))
```
Garulf, line 107: Blocks user from installing the same plugin manually without uninstalling first.

jjw24:
@Garulf pushing up a couple changes, brace for impact...

Garulf:
Very cool! We should probably have the Uninstall method uninstall all manual installed plugins at once.

jjw24:
I added that the user will be prompted if the downloaded plugin zip contains a plugin that already exists and is of same or lesser version, so we dont have to deal with duplicate plugins from this install method.

Also added a check to warn user if the source is not from github and the author is not found from the currently installed plugin list

jjw24:
@Garulf @taooceros  all done, please test and review.

Garulf:
> @Garulf @taooceros all done, please test and review.

Incredibly comprehensive! This seems to handle all the niche situations that can happen with the old system. 👌 

The windows notifications might be a little overwhelming?

jjw24:
Hm, which notification should we remove?

Garulf:
> this will potentially allow for downloading duplicate plugins right?

I think success/fail should be sufficient.

Flow.Launcher.Core/Plugin/PluginConfig.cs
```diff
@@ -45,8 +45,56 @@ public static List<PluginMetadata> Parse(string[] pluginDirectories)
...
+                                    string.Format("Duplicate plugin name: {0}, id: {1}, version: {2} " +
+                                                    "not loaded due to version not the highest of the duplicates",
+                                                    x.Name, x.ID, x.Version),
+                                                    "GetUniqueLatestPluginMetadata"));
+
+            return uniqueList;
+        }
+
+        internal static (List<PluginMetadata>, List<PluginMetadata>) GetUniqueLatestPluginMetadata(List<PluginMetadata> allPluginMetadata)
+        {
```
jjw24, line 63: hey @taooceros can you see a better way of doing this loop?

jjw24:
@Garulf try now

Garulf:
> @Garulf try now

Much better! Windows toast has a slight delay so it got a bit confusing.

We should probably skip the install dialog if unknown source dialog is shown.

jjw24:
> We should probably skip the install dialog if unknown source dialog is shown.

nah because that unknown source warning can be turned off

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -555,8 +555,8 @@ private bool SameOrLesserPluginVersionExists(string metadataPath)
             var newMetadata = JsonSerializer.Deserialize<PluginMetadata>(File.ReadAllText(metadataPath));
             return Context.API.GetAllPlugins()
                                 .Any(x => x.Metadata.ID == newMetadata.ID 
-                                    && (x.Metadata.Version == newMetadata.Version) 
-                                        || newMetadata.Version.CompareTo(x.Metadata.Version) < 0);
+                                    && (x.Metadata.Version == newMetadata.Version 
+                                        || newMetadata.Version.CompareTo(x.Metadata.Version) < 0));
```
jjw24, line 559: good pick up @Garulf , i remmember i was thinking of needing to put them into brackets, not sure why i then forgot lol.

jjw24:
if it's good to go @taooceros @Garulf let me know i will merge

Flow.Launcher.Core/Plugin/PluginConfig.cs
taooceros: ok I will take a try

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
Garulf: No worries, glad we caught it.

Garulf:
As far as my testing goes everything works with no bugs 🤞

Flow.Launcher.Core/Plugin/PluginConfig.cs
taooceros: Hm I think it is good enough. The plugin count will be relatively small, so no need to preoptimize it until we reach a bottleneck for that.

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -302,6 +310,62 @@ internal List<Result> Search(IEnumerable<Result> results, string searchName)
...
+        }
+
+        private bool InstallSourceKnown(string url)
+        {
+            var author = url.Split('/')[3];
+            var acceptedSource = "https://github.com";
+            var contructedUrlPart = string.Format("{0}/{1}/", acceptedSource, author);
+
+            return url.StartsWith(acceptedSource) && Context.API.GetAllPlugins().Any(x => x.Metadata.Website.StartsWith(contructedUrlPart));
+        }
```
taooceros, line 367: github is known source🧐🤔
Probably it will be better to treat all url download as unknown source

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -311,7 +375,11 @@ internal async ValueTask<List<Result>> RequestInstallOrUpdate(string searchName,
 
             token.ThrowIfCancellationRequested();
 
-            var searchNameWithoutKeyword = searchName.Replace(Settings.HotKeyInstall, string.Empty).Trim();
+            var searchNameWithoutKeyword = searchName.Replace(Settings.HotKeyInstall, string.Empty, StringComparison.OrdinalIgnoreCase).Trim();
+
+            if (Uri.IsWellFormedUriString(searchNameWithoutKeyword, UriKind.Absolute)
```
taooceros, line 380: what if the url doesn't end with zip?

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
taooceros: A simply <=0 will be fine

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
jjw24: then no option to download, we support zip currently

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
jjw24: github is only a known source if the author of the plugin is found in the pluginsmanifest, otherwise it's still treated as unknown source

jjw24 (approved):

PR #824, opened 2021-11-19
GameMode

onesounds:
https://user-images.githubusercontent.com/6903107/142577704-384ffac7-376d-4b52-8e36-bd75ec0afad5.mp4

In my case, I tend to turn off all tray programs that are likely to affect game play.
Since games are often played without using the entire screen, this setting is necessary.

- Add "Game Mode" in Tray Menu (Toggle)
- The flow-on/custom query hotkey shortcut does not work when the game mode is on.
- it will change Tray Icon to red when turn on Game Mode

**ETC**
- "Pause Hotkey" is better name?
- It would be nice if we could pause the some part (indexing?) that can affect the speed.


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -44,6 +44,7 @@ public class MainViewModel : BaseModel, ISavable
         private readonly History _history;
         private readonly UserSelectedRecord _userSelectedRecord;
         private readonly TopMostRecord _topMostRecord;
+        private MainWindow _mainWindow;
```
jjw24, line 47: i dont think this is used right, it's private and not referenced in this vm
```suggestion
```

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -304,6 +305,8 @@ private void InitializeKeyCommands()
         public ResultsViewModel ContextMenu { get; private set; }
         public ResultsViewModel History { get; private set; }
 
+        public bool gameModeStatus = false;
```
jjw24, line 308: tip, bools are initialised as false, so you dont need to set it to false.
we can use a getter and setter here actually
```suggestion
        public bool GameModeStatus { get; set; }
```

jjw24: public variables use pascal case where as private variables are lower camel case.

Flow.Launcher/Properties/Resources.resx
```diff
@@ -112,13 +112,16 @@
     <value>2.0</value>
   </resheader>
   <resheader name="reader">
-    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
jjw24, line 115: these version changes are intentional?

Flow.Launcher/Flow.Launcher.csproj
```diff
@@ -104,6 +104,9 @@
     <ProjectReference Include="..\Flow.Launcher.Plugin\Flow.Launcher.Plugin.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Resource Include="Images\gamemode.ico" />
```
jjw24, line 108: we dont need to do this separately because up top it's already including .ico files :)

jjw24:
i will make the recommendations via direct commits if it's ok.

Flow.Launcher/Properties/Resources.resx
onesounds: no. visual studio did it.

onesounds:
> i will make the recommendations via direct commits if it's ok.

yeah, please.

jjw24:
not related to this pr but i noticed ignore hotkey when in full screen mode is not working, neither in dev. i will fix that up later

jjw24 (approved):

PR #834, opened 2021-11-27
[dev] Global hotkey refactor

### LINKED ISSUE #788: Win+R with Shell plugin doesn't work well (with windows walker)
**Describe the bug/issue**
- Window close when press win+r after 1.83 ver (dev)

**To Reproduce**
1. Open Flow launcher 
2. Press Win+R
3. Check that the window is hidden.

Occurred in Dev version. Normal operation at 1.83. One of the modifications appears to have an effect.

taooceros:
fix #788


jjw24:
@taooceros fair bit of method signature changes. can you give me a quick overview of why they are changed please

taooceros:
> @taooceros fair bit of method signature changes. can you give me a quick overview of why they are changed please

Only the global hotkey event in publicapi is obsolete, and two new api is added to handle the same work.

taooceros:
Unsafe is used because
1. It's reasonable to have high performance here because every key hit will trigger the running of the function once.
2. I would like to try the new function pointer feature🤔😂
3. I am not sure whether some issues may exist because of unsafe, so we should take a small set of testing here. However, it is unlikely we need to unload assembly for infrastructure, so it should be fine.

onesounds (dismissed):

Garulf (approved):

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -337,12 +337,9 @@ bool API_GlobalKeyboardEvent(int keyevent, int vkcode, SpecialKeyState state)
 
         private void OnWinRPressed()
         {
-            context.API.ChangeQuery($"{context.CurrentPluginMetadata.ActionKeywords[0]}{Plugin.Query.TermSeparator}");
-
             // show the main window and set focus to the query box
-            Window mainWindow = Application.Current.MainWindow;
-            mainWindow.Show();
-            mainWindow.Focus();
```
jjw24, line 345: does this still need focus applied?

Flow.Launcher.Infrastructure/Hotkey/GlobalHotkey.cs
```diff
@@ -91,17 +78,17 @@ private IntPtr LowLevelKeyboardProc(int nCode, UIntPtr wParam, IntPtr lParam)
             {
                 return InterceptKeys.CallNextHookEx(hookId, nCode, wParam, lParam);
             }
-            return (IntPtr)1;
+            return (IntPtr)(-1);
```
jjw24, line 81: what's this change?

jjw24:
what's been tested with this change so far?

also need to bump Shell plugin

onesounds:
> what's been tested with this change so far?
> 
> also need to bump Shell plugin

It is a pr to solve the problem that when win+r is pressed, the flow closes and the run command (win+r by os) operates. I understand that the code has become complicated because there is a part in the middle of the work that does not work properly related to animation.

jjw24:
well i dont exactly understand this change, so if @taooceros is sure this is good to go and not going to break anything then happy for one of you to approve and merge this.

taooceros:
> well i dont exactly understand this change, so if @taooceros is sure this is good to go and not going to break anything then happy for one of you to approve and merge this.

The main change is because the multicast delegate (event) fails to return all return values so that the win+r may not work with the presence of the windows walker. The main change of the pull request change the event to a list of Func to retrieve the return value correctly.
Also, because c# 9 added the function pointer, so I think it's cool to use that in the hook for better performance because the hook is fired in every single keyboard click.

jjw24 (changes requested):
somehow the artifacts are smaller, doesnt seem normal.

image

Garulf:
- Build looks good
- <kbd>Win</kbd>+<kbd>R,/kbd> fixed
- Fix for Tab complete
- Working with Window Walker
- No crashes

Good to go?

jjw24 (approved):
unsure of code changes but tested working fine, will merge since this addresses an issue in current dev

PR #835, opened 2021-11-27
Hide Apps Path & Adjust program plugin setting panel

onesounds:
https://github.com/Flow-Launcher/Flow.Launcher/discussions/761
https://github.com/Flow-Launcher/Flow.Launcher/discussions/734

This pr contains the following.
- Add Hide Apps Path setting (UWP / *.lnk). Default is "True"

- Change string "Enable Program Description" to "Search in Program Description"
- Change "Search in Program Description" default setting to "False" from "true"
  - If necessary, it is useful, but not necessary for general use, and it is recommended not to activate it by default because there is a problem that the name of the item is prolonged.
- Adjust Popup Window Design and color 
- Dark mode pr is merged.

Additional work is needed, but it is difficult for me to do.
- The Hide Apps Path setting does not change in real time. When changing the setting, it must be changed in real time.
- The list view style is not applied. When applied, it is displayed as follows. There is a need for a way to come out properly when applied. Since the existing state has a problem in dark mode, it is necessary to apply the style.

Broken Style (Default)
image

Applied Style
image


Plugins/Flow.Launcher.Plugin.Program/Settings.cs
```diff
@@ -13,7 +13,8 @@ public class Settings
 
         public bool EnableStartMenuSource { get; set; } = true;
 
-        public bool EnableDescription { get; set; } = true;
+        public bool EnableDescription { get; set; } = false;
+        public bool EnableHideLnkPath { get; set; } = true;
```
jjw24, line 17: feel like this variable should be `EnableHideAppPath` because you are hiding not only Lnk but also could be the file path to the app executable right?

onesounds: yeah. I'll chage it.

jjw24 (approved):

PR #838, opened 2021-11-29
[dev] Change Icon back to field to maintain backward compatibility

taooceros:
It will break backward compatibility for plugin use Icon field (although I think not a lot use it).


taooceros:
@jjw24 if you have time, just merge it. It's super simple.

jjw24 (approved):

PR #840, opened 2021-11-29
Switch off the default launch on system start up

jjw24:
Switch off the default launch on system start up, user can set it if they wish to launch with system start up

### DIFF
--- a/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
+++ b/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
@@ -133,7 +133,7 @@ namespace Flow.Launcher.Infrastructure.UserSettings
         public bool DontPromptUpdateMsg { get; set; }
         public bool EnableUpdateLog { get; set; }
 
-        public bool StartFlowLauncherOnSystemStartup { get; set; } = true;
+        public bool StartFlowLauncherOnSystemStartup { get; set; } = false;
         public bool HideOnStartup { get; set; }
         bool _hideNotifyIcon { get; set; }
         public bool HideNotifyIcon


onesounds (approved):
👼

Garulf:
So we want Flow _not_ to auto start by default? I think most users would expect it to.

jjw24:
i think usually programs do not auto start by default and i personally wouldnt also want any program to auto start unless i tell it too. This is also the case especially if it is portable. Auto start means editing the registry, so not everyone is fond of having a program do that without explicit permission.

onesounds:
> 



> I think usually programs do not auto start by default and i personally wouldnt also want any program to auto start unless i tell it to. This is also the case especially if it is portable. Auto start means editing the registry, so not everyone is fond of having a program do that without explicit permission.



> I think usually programs do not auto start by default and i personally wouldnt also want any program to auto start unless i tell it to. This is also the case especially if it is portable. Auto start means editing the registry, so not everyone is fond of having a program do that without explicit permission.

How about put the setting in 'welcome screen'? it will showing when first launch. surely, default is uncheck.

Garulf:
> I think usually programs do not auto start by default and i personally wouldnt also want any program to auto start unless i tell it to. This is also the case especially if it is portable. Auto start means editing the registry, so not everyone is fond of having a program do that without explicit permission.

Sorry deleted my comment due to some confusion... Yeah I guess that makes sense. 

I think @onesounds welcome screen idea will be the best "fix" for giving the User the option.

jjw24:
Putting in welcome screen is great idea @onesounds

onesounds:
Okay. I think I can work quickly because I used to make something before. Let's release 1.9 within this week and include its features.

PR #843, opened 2021-11-29
Change App missing icon

onesounds:
Old
image

New
app_missing_img

- Change Design
- Resize to 48x48

It is an icon that you often see when you use it, but the file size is larger than other icons and it looks ugly in recent designs, so I replace it. 

p.s It's not the source from online.


jjw24:
hmm the new design looks a bit too plain ?

onesounds:
> hmm the new design looks a bit too plain ?

old design is not good. I put in the window(small 4 square) and tested it. but plain is better.

jjw24:
When you say not good, r you talking aesthetic wise? 

The new look looks a bit 'too old school broken', I think put something in there to lighten it up feels a bit more eye pleasing, doesn't have to be four windows, just something easy and light-hearted on the eyes was what the old design was going for. Like these days with browser response errors.

onesounds:
> When you say not good, r you talking aesthetic wise?
> 
> The new look looks a bit 'too old school broken', I think put something in there to lighten it up feels a bit more eye pleasing, doesn't have to be four windows, just something easy and light-hearted on the eyes was what the old design was going for. Like these days with browser response errors.

Yes, I'm talking from a design point of view. This icon has been my concern since the beginning of flow use. I would have designed it like your words if it was seen simply in the absence of an icon, but this icon is often seen when scrolling. (Maybe it's because icon didn't load.) You're right if flow don't show the icon at all when it's not loaded, and only show if icon really doesn't exist.

It is difficult to say that the error icon is also a good icon. Personally, icons that have come down since the days of wox are not bad, but they are far from modern design and give a feeling that the software is old.  If it was the same theme as before, it would be a problem-free icon, but it is not an appropriate icon in a modernized design change as it is now.

onesounds:
Anyway, there seems to be a disagreement because this icon is used interchangeably in two situations.

1. icon does not exist (you)
2. temporary icon that before the icon is loaded. (me)

Can we make completely distinguish between the temporary icon and the case where the icon does not exist? If we can, I can create a few more icons corresponding to your words.

jjw24:
Ok, thanks for explaining, yep makes sense. If you can, that would be great.

onesounds:
> Ok, thanks for explaining, yep makes sense. If you can, that would be great.

yep, but not now. I'll do first welcome window.

taooceros:
> temporary icon that before the icon is loaded. (me)

yes that's under expected. It's added in #195

onesounds:
how about change all the icons? Let's use icon8.com's fluenticon.

image
image
image
image
image
image
image

jjw24:
> how about change all the icons? (from icon8.com's fluenticon)

yeah sounds good

jjw24 (dismissed):

jjw24 (dismissed):

jjw24 (approved):

PR #847, opened 2021-11-30
Provide Flow Launcher program data in Environmental vars

Garulf:
Providing this data can help a plugin determine which version of Flow Launcher it's running on as well as help find settings.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PythonPlugin.cs
+++ b/Flow.Launcher.Core/Plugin/PythonPlugin.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 using System.IO;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


jjw24:
Would it be better to expose this via api(IPublicAPI)?

Garulf:
> Would it be better to expose this via api(IPublicAPI)?

I don't understand, could you please explain?

jjw24:
Like what you did for copy to clipboard method

Garulf:
> Like what you did for copy to clipboard method

These values should help a plugin determine what environment it's running in and act accordingly.

For example if my plugin uses some newer API calls I might check for Flow's version like so:
```
import os
import sys

if os.getenv("FLOW_VERSION") != "x.x.x":
    sys.exit(1)
```
I'm unaware of any method to retrieve output from IPublicAPI and and act on it in Python.

jjw24 (approved):

PR #848, opened 2021-11-30
Add Clipboard copy to jsonRPC API

Garulf:
Simply adds a method for jsonRPC to call and place test into the system's clipboard.

```
print(json.dumps({"method": "flow.CopyToClipboard","parameters":["text"]}))
```

### DIFF
--- a/Flow.Launcher/PublicAPIInstance.cs
+++ b/Flow.Launcher/PublicAPIInstance.cs
@@ -114,6 +114,11 @@ namespace Flow.Launcher
             var startInfo = ShellCommand.SetProcessStartInfo(filename, arguments: args, createNoWindow: true);
             ShellCommand.Execute(startInfo);
         }
+        
+        public void CopyToClipboard(string text)
+        {
+            Clipboard.SetText(text);
+        }
 
         public void StartLoadingBar() => _mainVM.ProgressBarVisibility = Visibility.Visible;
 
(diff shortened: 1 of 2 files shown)


Flow.Launcher/PublicAPIInstance.cs
```diff
@@ -114,6 +114,11 @@ public void ShellRun(string cmd, string filename = "cmd.exe")
             var startInfo = ShellCommand.SetProcessStartInfo(filename, arguments: args, createNoWindow: true);
             ShellCommand.Execute(startInfo);
         }
+        
+        public void CopyToClipboard(string text)
+        {
+            Clipboard.SetText(text);
```
jjw24, line 120: Pretty sure this method has been causing all kinds of grief back in the days, you want this instead:
https://github.com/Flow-Launcher/Flow.Launcher/blob/365f7e73c8ea8064eeedeced432a9ada8311d9d3/Plugins/Flow.Launcher.Plugin.Calculator/Main.cs#L83

jjw24 (changes requested):

Flow.Launcher/PublicAPIInstance.cs
Garulf: "Clipboard.SetDataObject() : This method attempts to set the data ten times in 100-millisecond intervals, and throws an ExternalException if all attempts are unsuccessful."

Interesting...

jjw24 (approved):

PR #849, opened 2021-12-01
Welcome window

onesounds:
https://user-images.githubusercontent.com/6903107/144161299-ddc2d923-1a25-4f7c-bff1-fffeeb21a388.mp4

https://github.com/Flow-Launcher/Flow.Launcher/pull/840#issuecomment-982227813
https://github.com/Flow-Launcher/Flow.Launcher/discussions/770

**Plan**
- Show Simple Guide and Setting when First Launch.
- In ancient times, people called it 'Wizard'.

**Todo**
- [x] Add Setting Control 
- [x] Check First Launch 
- [x] String Text
- [x] Add Testing Button in About panel
- [x] Language expression correction.
- [x] Animation
- [x] Scrollbar (but I'm not use)
- [x] Pagination Number


onesounds:
Almost done. @taooceros The direction of the animation changed during the work. Check it please. 

Currently, FirstLaunch is applied only when the flow is terminated. First of all, I don't think there will be a problem, but it would be nice to have a code that safely stores settings inside CheckFirstLaunch().

Flow.Launcher/Languages/en.xaml
```diff
@@ -231,4 +232,40 @@
     <system:String x:Key="update_flowlauncher_update_files">Update files</system:String>
     <system:String x:Key="update_flowlauncher_update_upadte_description">Update description</system:String>
 
+    <!--  Welcome Window  -->
+    <system:String x:Key="Skip">Skip</system:String>
+    <system:String x:Key="Welcome_Page1_Title">Welcome to Flow Launcher</system:String>
+    <system:String x:Key="Welcome_Page1_Text01">Hello, It's your first time to run a Flow Launcher!</system:String>
```
deefrawley, line 238: Hello, this is the first time you are running Flow Launcher!

Flow.Launcher/Languages/en.xaml
```diff
@@ -231,4 +232,40 @@
...
+    <system:String x:Key="Welcome_Page1_Text02">Before starting, proceed with a simple setting. Surely, You can skip. Please choose the language you use.</system:String>
```
deefrawley, line 239: Before starting, this wizard will assist in setting up Flow Launcher. You can skip this if you wish. Please choose a language

Flow.Launcher/Languages/en.xaml
```diff
@@ -231,4 +232,40 @@
...
+    <system:String x:Key="Welcome_Page2_Title">Search and run everything in everywhere</system:String>
```
deefrawley, line 240: Search and run all files and applications on your PC

Flow.Launcher/Languages/en.xaml
```diff
@@ -231,4 +232,40 @@
...
+    <system:String x:Key="Welcome_Page2_Text01">Search everything from applications, files, bookmarks, YouTube, Twitter and more. All from the comfort of your keyboard without ever touching the mouse.</system:String>
+    <system:String x:Key="Welcome_Page2_Text02">Flow Launcher starts with the hotkey below, start with the hotkey below. To change, select input and press the wanted hotkey in keyboard.</system:String>
+    <system:String x:Key="Welcome_Page3_Title">Hotkeys</system:String>
+    <system:String x:Key="Welcome_Page4_Title">Action Keyword and Commands</system:String>
+    <system:String x:Key="Welcome_Page4_Text01">Search the web or run various functions. Certain functions start with the expression action keyword, and if necessary, they can be used without action keywords. Try the queries below in Flow Launcher.</system:String>
```
deefrawley, line 245: Search the web, launch applications or run various functions through Flow Launcher plugins. Certain functions start with an action keyword, and if necessary, they can be used without action keywords. Try the queries below in Flow Launcher.

jjw24:
maybe on first run we should hide the query bar or have the welcome screen in front of it, looks a bit weird like this:
image

jjw24:
is there a way to stop listening for hotkey change once it's set? because realizing once user has set the hotkey in the welcome screen, they might one to try it out, but currently doing so will cause the hotkey to change again:

image

jjw24:
with recommending the `VS` tip, i think not every user will have Visual Studio installed and it actually brings up a whole bunch of settings, probably not a good idea to recommend this.

image

jjw24:
maybe `gc` is a better recommendation?

image

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -232,6 +233,20 @@ private void InitializeNotifyIcon()
             };
         }
 
+        private void CheckFirstLaunch()
+        {
+            if (_settings.FirstLaunch)
+            {
+                _settings.FirstLaunch = false;
+                _viewModel.Save();
```
jjw24, line 241: i dont think this is the right save call:
image

it's not saving the setting, so if you terminate after flow is launched, it will display the welcome screen again on next launch. is this intended?

onesounds:
> maybe on first run we should hide the query bar or have the welcome screen in front of it

I thought the default setting was hiding. Initial execution, HideOnStartup was changed to True.

> maybe sn for Sticky Note is a better recommendation?
true. I changed it.

>is there a way to stop listening for hotkey change once it's set? because realizing once user has set the hotkey in the welcome screen, they might one to try it out, but currently doing so will cause the hotkey to change again:

I add keyboard.clearfocus() when set hotkey. but when close searchbox, focus will back to hotkeycontrol. In this case, unintended input may occur again. I don't know how to take the focus away. Additionally, if click somewhere else, will lose focus.

Flow.Launcher/MainWindow.xaml.cs
onesounds: I thought it should be saved in that location.  I knew the problem that it could not be saved this timing . but I didn't know how to call save(). so I commented it above.

>Currently, FirstLaunch is applied only when the flow is terminated. First of all, I don't think there will be a problem, but it would be nice to have a code that safely stores settings inside CheckFirstLaunch().

I wanted anyone to work on it, but I thought it was okay because no one responded. 

Anyway, taoo checked and added the save code.

jjw24 (approved):

PR #850, opened 2021-12-01
Global TAB complete

Garulf:
Pressing <kbd>Tab</kbd> will replace the current Query with the selected result's `Title`.

https://user-images.githubusercontent.com/535299/144630236-62f19803-49a0-4fa2-b0da-217d28b71024.mp4

# Todo:
- [x] ~~Cycle to next result if query matches result~~ (Feels weird and unnatural and doesn't work very well)
- [x] Allow plugins to change inserted text
- [x] <kbd>Shift</kbd>+<kbd>Tab</kbd> to insert result subtitle

### DIFF
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -31,14 +31,7 @@ namespace Flow.Launcher.Plugin
         /// </summary>
         public string ActionKeywordAssigned { get; set; }
 
-        public string InsertText
-        {
-            get { return string.IsNullOrEmpty(_insertText) ? Title : _insertText; }
-            set
-            {
-                _insertText = value;
-            }
-        }
+        public string SuggestionText { get; set; } = string.Empty;
 
         public string IcoPath
         {
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -13,6 +13,8 @@ namespace Flow.Launcher.Plugin
         
         private string _icoPath;
 
+        private string _insertText;
+
         /// <summary>
         /// Provides the title of the result. This is always required.
         /// </summary>
@@ -28,6 +30,15 @@ namespace Flow.Launcher.Plugin
         /// If result is triggered by global keyword: *, this should be empty.
         /// </summary>
         public string ActionKeywordAssigned { get; set; }
+
+        public string InsertText
+        {
+            get { return string.IsNullOrEmpty(_insertText) ? Title : _insertText; }
+            set
+            {
+                _insertText = value;
+            }
+        }
 
         public string IcoPath
         {
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -29,7 +29,12 @@ namespace Flow.Launcher.Plugin
         /// </summary>
         public string ActionKeywordAssigned { get; set; }
 
-        public string SuggestionText { get; set; } = string.Empty;
+        /// <summary>
+        /// This holds the text which can be provided by plugin to help Flow autocomplete text
+        /// for user on the plugin result. If autocomplete action for example is tab, pressing tab will have 
+        /// the default constructed autocomplete text (result's Title), or the text provided here if not empty.
+        /// </summary>
+        public string AutoCompleteText { get; set; }
 
         public string IcoPath
         {
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -13,8 +13,6 @@ namespace Flow.Launcher.Plugin
         
         private string _icoPath;
 
-        private string _insertText;
-
         /// <summary>
         /// Provides the title of the result. This is always required.
         /// </summary>
--- a/Flow.Launcher/Converters/QuerySuggestionBoxConverter.cs
+++ b/Flow.Launcher/Converters/QuerySuggestionBoxConverter.cs
@@ -43,13 +43,12 @@ namespace Flow.Launcher.Converters
                 if (!selectedResultPossibleSuggestion.StartsWith(queryText, StringComparison.CurrentCultureIgnoreCase))
                     return string.Empty;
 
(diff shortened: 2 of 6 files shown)


onesounds:
- I think alt+right arrow is better. Rather than using the tab as an automatic completion, it seems more appropriate to give it to a one-handed manipulator. It is used more often.
- The cursor must move to the end.

Garulf:
> * I think alt+right arrow is better. Rather than using the tab as an automatic completion, it seems more appropriate to give it to a one-handed manipulator. It is used more often.
> * The cursor must move to the end when pressed key.

Maybe we can add setting to allow the user to change hotkeys? I feel <kbd>Tab</kbd> is most natural as default however.

onesounds:
> > * I think alt+right arrow is better. Rather than using the tab as an automatic completion, it seems more appropriate to give it to a one-handed manipulator. It is used more often.
> > * The cursor must move to the end when pressed key.
> 
> Maybe we can add setting to allow the user to change hotkeys? I feel Tab is most natural as default however.

yeah, I think we should. However, it seems a little strange to provide only this function with a key change. For now, it would be good to proceed with tab, and work on that part in another pr later.

Garulf:
> * The cursor must move to the end when pressed key.

Text cursor should be moving when <kbd>Tab</kbd> is used. Is it not working on your end?

Garulf:
Replacing Query with `QuerySuggestionBox.Text`, using `result.Title` if its unavailable for some reason.

Above all else we use the `result.SuggestionText` field

Flow.Launcher.Plugin/Result.cs
```diff
@@ -29,6 +31,8 @@ public class Result
         /// </summary>
         public string ActionKeywordAssigned { get; set; }
 
+        public string SuggestionText { get; set; } = string.Empty;
```
jjw24, line 34: do we want to change this to 'AutoCompleteText' to be more specific? I didnt get what it is for on first glance until i read the code. 

This is good btw, it will allow plugins to do their own autocomplete, as @onesounds have indicated, if we change how Explorer display Title and Subtitle result, this will allow Explorer to put its own autocompelete text

Flow.Launcher.Plugin/Result.cs
```diff
@@ -13,6 +13,8 @@ public class Result
         
         private string _icoPath;
 
+        private string _insertText;
+
```
jjw24, line 17: this is not needed right? it's a private string and not used in this class.
```suggestion
```

Flow.Launcher.Plugin/Result.cs
jjw24, line 34: No need to add a default here since you are already checking for null and string empty below.
```suggestion
        public string SuggestionText { get; set; }
```

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -228,6 +228,24 @@ private void InitializeKeyCommands()
                 }
             });
 
+            InsertSuggestion = new RelayCommand(_ =>
+            {
+                var results = SelectedResults;
+
+                var result = results.SelectedItem?.Result;
```
jjw24, line 235: can just combine these two
```suggestion
                var result = SelectedResults.SelectedItem?.Result;
```

Flow.Launcher/Converters/QuerySuggestionBoxConverter.cs
```diff
@@ -44,6 +44,12 @@ public object Convert(object[] values, Type targetType, object parameter, Cultur
                     return string.Empty;
 
                 // When user typed lower case and result title is uppercase, we still want to display suggestion
+
+                string _suggestion = queryText + selectedResultPossibleSuggestion.Substring(queryText.Length);
+                if (String.IsNullOrEmpty(selectedResult.SuggestionText))
+                {
+                    selectedItem.Result.SuggestionText = _suggestion;
+                }
                 return queryText + selectedResultPossibleSuggestion.Substring(queryText.Length);
```
jjw24, line 53:
```suggestion
                // construct autocomplete with suggestion
                string _suggestion = queryText + selectedResultPossibleSuggestion.Substring(queryText.Length);
                if (String.IsNullOrEmpty(selectedResult.SuggestionText))
                {
                    selectedItem.Result.SuggestionText = _suggestion;
                }
                
                // When user typed lower case and result title is uppercase, we still want to display suggestion
                return queryText + selectedResultPossibleSuggestion.Substring(queryText.Length);
```

Flow.Launcher/Converters/QuerySuggestionBoxConverter.cs
```diff
@@ -44,6 +44,12 @@ public object Convert(object[] values, Type targetType, object parameter, Cultur
                     return string.Empty;
 
                 // When user typed lower case and result title is uppercase, we still want to display suggestion
+
+                string _suggestion = queryText + selectedResultPossibleSuggestion.Substring(queryText.Length);
+                if (String.IsNullOrEmpty(selectedResult.SuggestionText))
+                {
+                    selectedItem.Result.SuggestionText = _suggestion;
```
jjw24, line 51: @taooceros will this modifying result cause results comparison issue afterwards?

jjw24 (changes requested):
lgtm, just some small tweaks

Flow.Launcher/Converters/QuerySuggestionBoxConverter.cs
jjw24: Instead of using Result to hold the new constructed suggestion string, use [ResultViewModel](https://github.com/Flow-Launcher/Flow.Launcher/blob/2ba6ce9e874f80db739da756f79e8b6c3b759ed0/Flow.Launcher/ViewModel/ResultViewModel.cs#L145) instead. As a general we dont want to modify an object after it is constructed, otherwise it's hard to reason what state/information the object has.

jjw24 (approved):
@Garulf ready to go, just mad some minor tweaks, please test it, and if it's good then merge it in.

Garulf:
Looks good to me. 👍

PR #851, opened 2021-12-01
Global copy to clipboard

Garulf:
Invoking <kbd>CTRL</kbd>+<kbd>C</kbd> will copy the selected results title to clipboard.

https://user-images.githubusercontent.com/535299/144389689-635316b5-9803-4803-abe4-c766ba343f47.mp4

- [x] Don't copy result if Query text selected
- [x] Add logic to copy *actual* file object to clipboard

Selected Query text is copied otherwise result title is added to the clipboard.

Edited:
Invoking <kbd>CTRL</kbd>+<kbd>C</kbd> on a selected result will:
- if the result is directory/file path, copy the actual file/folder
- if the result is not directory/file path then the subtitle is copied
- if the query text is highlighted, then copy the highlighted query text
- if not above, then the selected result's subtitle is copied, but plugin can pass through the text to be copied to override this default behaviour.
- notifications for file/folder/text copy


Garulf:
I think it's best we let the plugin delegate what's copied to clipboard. Default behavior is to copy `result.Title`.

We can supply a new field for plugins to populate with what they think is most relevant for the clipboard.

Should we delay this feature for this functionality?

onesounds:
> I think it's best we let the plugin delegate what's copied to clipboard. Default behavior is to copy `result.Title`.
> 
> We can supply a new field for plugins to populate with what they think is most relevant for the clipboard.
> 
> Should we delay this feature for this functionality?

Yes. I don't think it's actually very useful that the default value is Title. I think most basic plugins and users will use ctrl+c as either file copy or path copy. Shortcut keys should not be inserted in a state that does not operate with this function. It is good to work for the purpose used in the plug-in, but this function must be preceded first. This is because it causes confusion if it changes in the future.

Anyway, this is my opinion, and I don't know what others think. We need to discuss more. I hate answering tightly, but three items you pr take up the important hotkey. (It's in the middle of New York) The functions you choose to fix are also very important in terms of long-term UX, and I think it's better to plan ahead because if the shortcut changes, there will be confusion.

On the one hand, if we don't tell the shortcut anyway, no one will use it, and I think it's okay to put it in because it's necessary for plug-in development for now.  I don't think flow launcher will be downloaded a lot.😂

jjw24:
Agree with @onesounds to use ctrl + c  as a file copy and ctrl+shift+c as a file path copy.

I think this maybe best implemented at plugin level, where plugin decide when ctrl + c is pressed, what exactly to copy- title, subtitle, title+subtitle, or the actual file. This reduces complications because otherwise flow will have to decide in what scenario to copy what.

@Garulf you got it to copy title ATM, what scenario will require in copying title ATM?

Garulf:
https://user-images.githubusercontent.com/535299/151084312-c0d66e8e-de85-4c8f-97bd-2013911b5723.mp4

jjw24:
@Garulf I moved the actual copy logic into MainViewModel and added notification for file/folder/text copied. Just a FYI, general rule of thumb, avoid application/business logic in your Views, i.e. MainWindow.xaml.cs, this will make it easier when later wanting to change the view and follows the MVVM pattern.

jjw24 (approved):

PR #852, opened 2021-12-01
Up Arrow key cycles through Query history

Garulf:
Similar to how arrow keys work in most shells this PR allows the user to press <kbd>Up</kbd> to cycle through their last used queries.

# Todo:
- [x] Cycle through entire history if user hasn't moved selection
- [x] Fix not being able to move selection in History view
- [ ]  ~~Add custom hotkey support~~

https://user-images.githubusercontent.com/535299/144291288-73ffa191-5deb-4f42-9675-90b748a82e32.mp4


onesounds:
- The same problem exists in the context menu.
- I think there should be a shortcut to erasing it.  (ctrl+u?)

jjw24:
i cant seem to cycle past the first history that showed, basically if i keep pressing up after the first history it just cycle through the results from it.

also need to check and not trigger when in context menu

is this pr still draft?

Garulf:
> i cant seem to cycle past the first history that showed, basically if i keep pressing up after the first history it just cycle through the results from it.
> 
> also need to check and not trigger when in context menu
> 
> is this pr still draft?

Yeah it's still just a proof-of-concept...

Garulf:
https://user-images.githubusercontent.com/535299/145690824-1d2a8f17-d8ee-4463-a792-55eff6351b41.mp4

Working

jjw24:
i am getting this error. repro:
1. only one history result present
2. up arrow to insert the query
3. alt + down arrow to forward search

image

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -29,16 +29,19 @@ public class MainViewModel : BaseModel, ISavable
         #region Private Fields
 
         private const string DefaultOpenResultModifiers = "Alt";
+        private const string DefaultCycleHistoryModifiers = "Alt";
 
         private bool _isQueryRunning;
         private Query _lastQuery;
         private string _queryTextBeforeLeaveResults;
 
+
```
jjw24, line 38:
```suggestion
```

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -775,7 +833,7 @@ public async void Hide()
                 default:
                     throw new ArgumentException($"wrong LastQueryMode: <{_settings.LastQueryMode}>");
             }
-
+            _lastHistoryIndex = 1;
```
jjw24, line 836: why is this needed here and why 1?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -29,16 +29,19 @@ public class MainViewModel : BaseModel, ISavable
...
         private Query _lastQuery;
         private string _queryTextBeforeLeaveResults;
 
+
         private readonly FlowLauncherJsonStorage<History> _historyItemsStorage;
         private readonly FlowLauncherJsonStorage<UserSelectedRecord> _userSelectedRecordStorage;
         private readonly FlowLauncherJsonStorage<TopMostRecord> _topMostRecordStorage;
         internal readonly Settings _settings;
         private readonly History _history;
+        private int _lastHistoryIndex = 1;
```
jjw24, line 44: this doesnt need to be underscore, code base may have some private variables with underscores but new code should use just lowercase and we will gradually rename these underscores. the standard naming practice for private variables is no longer underscore.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -189,6 +193,51 @@ private void InitializeKeyCommands()
...
+            InsertLastQuery = new RelayCommand(_ => {
+                var results = SelectedResults;
+
+                if (_history.Items.Count > 0 
+                    && _queryText == String.Empty
+                    && !HistorySelected() 
+                    && !ContextMenuSelected())
+                {
+                    ChangeQueryText(_history.Items[_history.Items.Count - 1].Query.ToString());
+                    _lastHistoryIndex = 2;
```
jjw24, line 232: why 2? this is the cause for the above error in the comment. try not to use plain numbers and strings please

jjw24 (changes requested):
hey just a couple changes needed please :)

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -409,8 +458,12 @@ private ResultsViewModel SelectedResults
         public ICommand ReloadPluginDataCommand { get; set; }
         public ICommand ClearQueryCommand { get; private set; }
         public ICommand AutocompleteQueryCommand { get; set; }
+        public ICommand ReverseHistory { get; set; }
+        public ICommand ForwardHistory { get; set; }
+        public ICommand InsertLastQuery { get; set; }
```
jjw24, line 463: these command variable naming should follow the others- e.g. ReverseHistoryCommand

jjw24:
so i am finding it a bit weird when i press up arrow on an empty query and only the one history result (the last one shows) then the subsequent up arrow presses will just cycle through the result. what are your throughts, up arrow only is able to cycle up one result, do we want to do this where the alt + up/down arrows work like a treat.

Garulf:
> so i am finding it a bit weird when i press up arrow on an empty query and only the one history result (the last one shows) then the subsequent up arrow presses will just cycle through the results. what are your throughts, up arrow only is able to cycle up one result, do we want to do this where the alt + up/down arrows work like a treat.

This intended behavior as the previous way got in the way of default behavior. Sorry I haven't documented yet as I was still testing.

Flow.Launcher/ViewModel/MainViewModel.cs
Garulf: Oh this is set so if the user hits <kbd>Up</kbd> on an empty Query and decides to use <kbd>Alt</kbd>+<kbd>Up</kbd> they don't repeat inserting the first item.

Flow.Launcher/ViewModel/MainViewModel.cs
Garulf: Resets the index when Flow is deactivated/hidden

Flow.Launcher/ViewModel/MainViewModel.cs
Garulf: I should add logic to make sure we dont set this to an index that doesn't exist

Garulf:
Currently <kbd>Alt</kbd>+<kbd>Up</kbd> & <kbd>Alt</kbd>+<kbd>Down</kbd> Only trigger if defined in `MainWindow.xaml` and will not trigger from `MainWindow.xaml.cs`

Flow.Launcher/ViewModel/MainViewModel.cs
Garulf: Fixed: https://github.com/Flow-Launcher/Flow.Launcher/pull/852/commits/c6d5525dec46631db6cf587573d6ecc55c8b9419

Garulf:
> i am getting this error. repro:
> 
> 1. only one history result present
> 2. up arrow to insert the query
> 3. alt + down arrow to forward search
> 
> image

Fixed: https://github.com/Flow-Launcher/Flow.Launcher/pull/852/commits/c6d5525dec46631db6cf587573d6ecc55c8b9419

jjw24:
@Garulf this is ready for review again? (Merge Dev if so)

Garulf:
Could not find a proper method to set <kbd>Alt</kbd> & <kbd>Ctrl</kbd> modifiers. I could gut it and hard code to <kbd>Alt</kbd> or keep as is?

jjw24:
The modifier hotkey can set Alt, does that part of the code offer any help?

image

Garulf:
> The modifier hotkey can set Alt, does that part of the code offer any help?
> 
> image

The problem is the current implementation requires the hotkey be coded in two separate files. Doesn't feel like a great fix.

Garulf:
Removed from GUI for now

taooceros:
close in favor of #2696

PR #854, opened 2021-12-02
Hide suggestions if query matches

Garulf:
There's no need to display a suggestion when the query is the same as the suggestion. This helps alleviate issues on very long text.
Old:
OLD
New:
NEW


onesounds:
works well.

taooceros:
Close this in favor of #860

PR #859, opened 2021-12-03
Open selected result in quicklook

Garulf:
<kbd>F1</kbd> Opens [Quicklook](https://github.com/QL-Win/QuickLook) on your selection.

~~Right now it's just a proof of concept.... Logic and features will be added soon™~~
# Todo:
- [x] Change preview when selection changes
- [x] Logic to determine what to open
- [x] Add field for Plugin's to use
...and more

# Preview:
https://user-images.githubusercontent.com/535299/144653778-e0188ee1-ed1b-4647-92de-5341fced6919.mp4


jjw24 (commented):
I am getting freezing with any query I type. This seems to be the error:
System.Windows.Data Error: 23 : Cannot convert '<null>' from type '<null>' to type 'System.Windows.Media.ImageSource' for 'en-US' culture with default conversions; consider using Converter property of Binding. NotSupportedException:'System.NotSupportedException: ImageSourceConverter cannot convert from (null).
   at System.ComponentModel.TypeConverter.GetConvertFromException(Object value)
   at System.Windows.Media.ImageSourceConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   at MS.Internal.Data.DefaultValueConverter.ConvertHelper(Object o, Type destinationType, DependencyObject targetElement, CultureInfo culture, Boolean isForward)'

Garulf:
Does it say what line is triggering this exception?

jjw24:
Nah sorry, that's why I didn't dig into it, I will leave it with you since I want to continue onto other prs. You can replicate the issue right?

Garulf:
> Nah sorry, that's why I didn't dig into it, I will leave it with you since I want to continue onto other prs. You can replicate the issue right?

I never got this error. Is this from the CI build?

jjw24:
it's coming from local run in visual studio

jjw24:
Tested with the CI build, it's happening there as well. Simply type 'a' or jpg causes flow to hang indefinitely.

Garulf:
> Tested with the CI build, it's happening there as well. Simply type 'a' or jpg causes flow to hang indefinitely.

Found the issue!

If QuickLook is unavailable and Flow tries to open a pipe and gets stuck. Adding a timeout helps but isn't the full solution.

jjw24:
When would quick look not be available? Is it when the file is not an image?

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -512,6 +512,10 @@ private void OnKeyDown(object sender, KeyEventArgs e)
                         e.Handled = true;
                     }
                     break;
+                case Key.F1:
+                    _viewModel.OpenQuickLook.Execute(null);
+                    e.Handled = true;
+                    break;
```
taooceros, line 518: I think we have registered another `F1` keystork so you may want to replace that with OpenQuickLook instead of setting that here.

Garulf: Done

jjw24:
so this needs quicklook installed before can be used? do we want to use Droplex to handle it?

Garulf:
> so this needs quicklook installed before can be used? do we want to use Droplex to handle it?

I'm unsure. I originally designed this so it would just work no matter if it was installed or not. 

@Flow-Launcher/team any opinions?

jjw24:
I wonder if there is a way we can set QL to stay on top, hide and set window size via the code? This will help make QL usage feel more integrated with flow.

jjw24:
TODO: need to check if window is open to avoid wasting CPU time.

jjw24:
You can set QL to stay on top on each launch by setting QuickLook.config's attribute TopMost to true

Garulf:
> TODO: need to check if window is open to avoid wasting CPU time.

Quicklook author seems to think detecting the window is more cpu time then just sending the pipe and failing. I haven't tested this but it seems logical.

jjw24:
I got the impression from the linked issue the author meant better to check window exists so to avoid wasting CPU, where did you see sending the pipe and failing is better? If failing is better than yeah that would be good.

Garulf:
> I got the impression from the linked issue the author meant better to check window exists so to avoid wasting CPU, where did you see sending the pipe and failing is better? If failing is better than yeah that would be good.

You’re correct. I must have misread his intentions.

Garulf:
Closing in favor of #2082

Issue #883, opened 2021-12-09
Wizzard window can run multiple times


kubalav:
Wizzard button can run Wizzard multiple times. It should be can run only once.

PR #888, opened 2021-12-10
Small Design Fix in Setting Window

onesounds:
1. https://github.com/Flow-Launcher/Flow.Launcher/issues/883 Change Wizard window to modal when open in about only
2. Change Label Width Size 
image

3. Change Checkbox Position
image

4. Add Tab Key description in page 3 in wizard window

### DIFF
--- a/Flow.Launcher/CustomQueryHotkeySetting.xaml
+++ b/Flow.Launcher/CustomQueryHotkeySetting.xaml
@@ -4,7 +4,7 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:flowlauncher="clr-namespace:Flow.Launcher"
     Title="{DynamicResource customeQueryHotkeyTitle}"
-    Width="500"
+    Width="530"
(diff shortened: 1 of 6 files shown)


taooceros:
What about make the width auto?

onesounds:
> What about make the width auto?

it's not easy. "auto" is made wrong grid so I should use grid layout (very annoying work!) and 
change window width to responsive. and it has some issue with custom title bar. (it need fixed width)
Anyway, I change it to responsive.

taooceros:
> > What about make the width auto?
> 
> it's not easy. "auto" is made wrong grid so I should use grid layout (very annoying work!) and 
> change window width to responsive. and it has some issue with custom title bar. (it need fixed width)
> Anyway, I change it to responsive. 

LOL ok🤣.

onesounds:
Remove Priority "Minimum number = 0"
-> negative number is possible.

onesounds:
Fix Impossible set negative number in priority

jjw24 (approved):

Issue #68, opened 2020-06-18
Discussion: Removal of result highlighting


jjw24:
Just wanting to get some thoughts on whether we should remove result highlighting. 

It was introduced as a way of combating the messy search results back then in Wox, but over the many iterations this has been improved to the point I dont think it's a necessity any more, and results now is clear enough. Any missed results would be more or less edge cases which we should just take as opportunity to further improve search experience.

Thoughts? @Flow-Launcher/team + anyone else

jjw24:
Also want to add- removal of search precision option in Settings tab. Regular is a good standard and 'Low' + 'None' is not utilised much.

Anyone out there still use the other two options?

JohnTheGr8:
> 
> 
> Just wanting to get some thoughts on whether we should remove result highlighting.
> 
> It was introduced as a way of combating the messy search results back then in Wox, but over the many iterations this has been improved to the point I dont think it's a necessity any more, and results now is clear enough. Any missed results would be more or less edge cases which we should just take as opportunity to further improve search experience.
> 
> Thoughts? @Flow-Launcher/team + anyone else

Does highlighting cause any problems other than the icon flickering issue (which is probably fixable)? Whatever performance/complexity overhead it adds is probably negligible.

I agree with what you said, it's not really a necessity any more... the core plugins could do without highlighting. But, at the same time, I think it's a nice to have, and that there could be potential use-cases for it from 3rd-party plugins that we haven't considered.

I'd vote to keep it.

thanhmssl10:
I think the whole point of highlighting is for helping user to understand **why** Wox return those results. Maybe we can expand from that idea, let highlighting is a feature for debugging or advanced user mode. And add more features (even simple as log more detail info) to help user understand the search algorithm (why this item have lower point than that result etc ...)

For my personal usage, I have to do this to debug the "grading algorithm" and observe the performance of plugins :D

![image](https://user-images.githubusercontent.com/10551242/85016431-e6e07500-b193-11ea-8482-853434282e32.png)

Kalaschnik:
Just to be clear we are talking about this:
![image](https://user-images.githubusercontent.com/9831678/85023161-a9202400-b174-11ea-901d-680ab354e4cd.png)

jjw24:
> Just to be clear we are talking about this:
> ![image](https://user-images.githubusercontent.com/9831678/85023161-a9202400-b174-11ea-901d-680ab354e4cd.png)

[...]

PR #890, opened 2021-12-10
Remove Subtitle Highlight

onesounds:
https://github.com/Flow-Launcher/Flow.Launcher/issues/68

1. I think the highlight of the path part is the made bad UI/UX design and functionally unnecessary. if highlight is blue, it spread blue character whole window. it made ugly design. This also destroys the highlight visibility of the title area. 
2. For the same reason, I think this function does not require on/off setting and should be erased.
3. The highlight of the title area must exist. (It is a feedback on whether the query worked normally.) I don't erase this part.

I first pr considered on/off setting, but I changed my mind. It's right to just erase this.

Tested:
- Everything plugin

TODO:
REMOVE FROM DOC https://github.com/Flow-Launcher/docs/blob/main/API-Reference/Flow.Launcher.Plugin/result.md

### DIFF
--- a/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
+++ b/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
@@ -40,7 +40,6 @@ namespace Flow.Launcher.Infrastructure.UserSettings
         public string ResultFontStretch { get; set; }
         public bool UseGlyphIcons { get; set; } = true;
         public bool UseAnimation { get; set; } = true;
-        public bool SubtitleHighlight { get; set; } = false;
         public bool UseSound { get; set; } = true;
         public bool FirstLaunch { get; set; } = true;
 
--- a/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
+++ b/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
@@ -40,6 +40,7 @@ namespace Flow.Launcher.Infrastructure.UserSettings
         public string ResultFontStretch { get; set; }
         public bool UseGlyphIcons { get; set; } = true;
         public bool UseAnimation { get; set; } = true;
+        public bool SubtitleHighlight { get; set; } = false;
         public bool UseSound { get; set; } = true;
         public bool FirstLaunch { get; set; } = true;
 
--- a/Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
+++ b/Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
@@ -14,10 +14,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>2.1.0</Version>
-    <PackageVersion>2.1.0</PackageVersion>
-    <AssemblyVersion>2.1.0</AssemblyVersion>
-    <FileVersion>2.1.0</FileVersion>
+    <Version>2.1.1</Version>
+    <PackageVersion>2.1.1</PackageVersion>
+    <AssemblyVersion>2.1.1</AssemblyVersion>
+    <FileVersion>2.1.1</FileVersion>
     <PackageId>Flow.Launcher.Plugin</PackageId>
     <Authors>Flow-Launcher</Authors>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Windows.Media;
@@ -77,9 +77,7 @@ namespace Flow.Launcher.Plugin
         /// </summary>
         public IList<int> TitleHighlightData { get; set; }
 
-        /// <summary>
-        /// A list of indexes for the characters to be highlighted in SubTitle
-        /// </summary>
+        [Obsolete("Deprecated as of Flow Launcher v1.9.1. Subtitle highlighting is no longer offered")]
         public IList<int> SubTitleHighlightData { get; set; }
 
         /// <summary>
@@ -110,8 +108,7 @@ namespace Flow.Launcher.Plugin
             var equality = string.Equals(r?.Title, Title) &&
                            string.Equals(r?.SubTitle, SubTitle) &&
                            string.Equals(r?.IcoPath, IcoPath) &&
-                           TitleHighlightData == r.TitleHighlightData &&
-                           SubTitleHighlightData == r.SubTitleHighlightData;
+                           TitleHighlightData == r.TitleHighlightData;
 
             return equality;
         }
--- a/Flow.Launcher/Converters/HighlightTextConverter.cs
+++ b/Flow.Launcher/Converters/HighlightTextConverter.cs
@@ -8,15 +8,11 @@ using System.Windows;
 using System.Windows.Data;
 using System.Windows.Media;
 using System.Windows.Documents;
-using Flow.Launcher.Infrastructure.UserSettings;
-using Flow.Launcher.Core.Resource;
 
 namespace Flow.Launcher.Converters
 {
     public class HighlightTextConverter : IMultiValueConverter
     {
-        public Settings Settings { get;}
-        Settings settings = new Settings();
         public object Convert(object[] value, Type targetType, object parameter, CultureInfo cultureInfo)
         {
             var text = value[0] as string;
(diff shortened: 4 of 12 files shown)


jjw24:
which plugin has highlighting on subtitle?

also we need to remove the code too as this is only removal of the view

onesounds:
> which plugin has highlighting on subtitle?

I looked around and there is no plug-in that uses this part specifically.

> also we need to remove the code too as this is only removal of the view

This part operates in the highlight converter.  Compared to the query, the code that puts colors in is used in duplicate in two places: title. In other words, only the subtitle is not separately worked.  I think all you have to do is erase the view code.

jjw24:
> I looked around and there is no plug-in that uses this part specifically.

WindowSettings plugin uses subtitle search + highlight (although highlighting is off)

>This part operates in the highlight converter. Compared to the query, the code that puts colors in is used in duplicate in two places - title and subtitle. In other words, only the subtitle is not separately worked. I think we only need to erase the view code.

@onesounds @taooceros well that's the thing, do we want to completely rip out the subtitle highlighting or just 'disable' it and keep the code behind intact?

onesounds:
> * WindowSettings plugin uses subtitle search + highlight (although highlighting is a bit off)
> * Everything plugin uses subtitle highlighting
> * Sys plugin uses subtitle search + highlight

I didn't think it was a function to "use" highlights. Highlight is just to indicate the matched letters. I only considered the possibility of some different kind of function in this part. For example, in the case of a home assistant plug-in, the highlights operate differently depending on on on/off. Of course, this plug-in doesn't work like that.

> Searching WindowSettings subtitle might be useful if you want to find all settings related to e.g. Privacy

The same applies to other plug-ins. Although not in name, there may be cases where it matches in path.

This function is a function that tells you which part matches the user's search intention when searching.  However, in general, it often appears overlapping in the path area, which damages the visual design. Also, erasing this function does not damage the actual path search results, This is ultimately the same as Title Highlight, but considering the user's general search intention, it is correct that highlights exist only in Title. If we intended to search for Subtitle, we wouldn't question the search results even if we didn't show the highlights Also, even if it was not intended to be so, there is no reason to question the current search results. This is because the search accuracy is high. This function has a reason to exist if there are really often unintended search results, but in my opinion, it does not exist.

> @onesounds @taooceros well that's the thing, do we want to completely rip out the subtitle highlighting or just 'disable' it and keep the code behind intact?

Your intention is understood by me as 'delete unused code'. However, this highlight code is just a code that checks and highlights whether the designated text block matches the query. How can we take out the subtitle highlight? We should use Title Highlight. Is it possible to separate only the subtle highlight?

jjw24:
> Your intention is understood by me as 'delete unused code'. However, this highlight code is just a code that checks and highlights whether the designated text block matches the query. How can we take out the subtitle highlight? We should use Title Highlight. Is it possible to separate only the subtle highlight?

yeah that's what i mean, take out the code that offers the subtitle highlighting (they are also found in the api so we need to deprecate the method as well).

> This function is a function that tells you which part matches the user's search intention when searching. However, in general, it often appears overlapping in the path area, which damages the visual design. Also, erasing this function does not damage the actual path search results, This is ultimately the same as Title Highlight, but considering the user's general search intention, it is correct that highlights exist only in Title. If we intended to search for Subtitle, we wouldn't question the search results even if we didn't show the highlights Also, even if it was not intended to be so, there is no reason to question the current search results. This is because the search accuracy is high. This function has a reason to exist if there are really often unintended search results, but in my opinion, it does not exist.

sorry for the confusion, i meant, subtitle search will be kept, but remove subtitle highlighting and yes i agree, search accuracy is high enough you wont need to search the subtitle most of the time.

Just so we are on the same page, when user search something like this, it should be fine that 'computer' in the subtitle is no longer highlighted:

image

onesounds:
> yeah that's what i mean, take out the code that offers the subtitle highlighting (they are also found in the api so we need to deprecate the method as well).

OK. I didn't know there was one in API. If so, it would be better to remove the code from the relevant plug-in, and if so, it is not simple. It's not urgent, so let's proceed slowly.

jjw24:
hmm this kind of searches wont have any highlighting after removal:

image

taooceros:
> hmm this kind of searches wont have any highlighting after removal:
> 
> image

I think it is ok? the title is enough to tell what the result will do. The subtitle highlight is like something that user won't notice.

JohnTheGr8:
Is there no UI fix to this issue? Maybe a more subtle highlight in the subtitle (say, bold instead of blue) ? Removing this entirely seems like a drastic measure...

> 
> I think it is ok? the title is enough to tell what the result will do. 

The point is to also see why a result matched the typed query...

>
> The subtitle highlight is like something that user won't notice.

Disagree, but I might be biased since I added this a long, long time ago (5+ years, time flies...)

onesounds:
Hi, I know you are an old man in this repo. Nice to meet you. 

> Is there no UI fix to this issue? Maybe a more subtle highlight in the subtitle (say, bold instead of blue) ? Removing this entirely seems like a drastic measure...

There is no setting. I think it is unnecessary.

>The point is to also see why a result matched the typed query...

Why should the user know which part was matched for typed query? If this is really necessary, we'd rather remove the highlights in the all area, and just show the score.

- One of the important rules of modern design is to subtract unnecessary things as much as possible. We shouldn't try to show all the information. 


- In general, the user searches based on the title area.
- If the search result is correct, the user does not need to know what was matched.
- Adding highlights to the subtitle area reduces the highlighting effect of the highlight. In my standard, this is the same as applying Bold to the entire document because the contents of the document as a whole are important. It's not aesthetically good either. I think the highlight of the subtle area reduces the readability of the search results

- For example,  https://github.com/Flow-Launcher/Flow.Launcher/pull/835 in the case of an app, I did not display the path at all. In this case, the user may not see an unnecessary path, and from the user's point of view, the user may feel that the search result is an App.

- In this case, as in the example of jjw, a search result with no highlights may be displayed. However, this is a rare case in general use experience. and indicating that itself a 'search result based on the subtitle', and the user can feel it.

- Other programs don't use highlights in the subtitle area. This is not our feature, our disadvantage.

image

image

image

image

image

image

image
⬆️ For your information, this program is the ugliest program and is similar to our 1.8 version.

JohnTheGr8:
> I know you are an old man😋 
> 

I know that, too 😢 

> 
> Why should the user know which part was matched for typed query? If this is really necessary, we'd rather remove the highlights in the all area, and just show the score.

I mention this as an example, because I remember it as the main reason people requested this feature back in the day.

> 
>     * In general, the user searches based on the title area.
>     * If the search result is correct, the user does not need to know what was matched.

You assume that plugins are searching things, which is not always the case... The calculator plugin is not searching things, the Spotify plugin is not always searching things, and so on... There has to be some creative/practical use of highlighting the subtitle other than to show a search match... Why completely kill that possibility?

> 
>     * One of the important rules of modern design is to subtract unnecessary things as much as possible. We shouldn't try to show all the information.

This project is supposed to target "power users", and modern design is very much not afaik 😛 

> [...]

All I'm arguing for here is: why not just go with the original plan? add a toggle, make it off by default...?

onesounds:
>I mention this as an example, because I remember it as the main reason people requested this feature back in the day.

At that time, I don't know why users requested it. However, it may have been necessary in that the results were not appropriate. However, the search results are excellent now. I don't think users have come up with the right solution.

> You assume that plugins are searching things, which is not always the case... The calculator plugin is not searching things, the Spotify plugin is not always searching things, and so on... There has to be some creative/practical use of highlighting the subtitle other than to show a search match... Why completely kill that possibility?

Currently, there is no plug-in using such a function. There is a possibility, but if it is a plug-in with such a possibility, I think it is better to use it more in other factors. In places such as icons and title areas. Giving a degree of freedom in this area undermines design consistency.

> This project is supposed to target "power users", and modern design is very much not afaik 😛
I agree with the saying that it is a project for power users. However, I don't think this part is necessary even for a power user. And modern power users value design more. 😂

>All I'm arguing for here is: why not just go with the original plan? add a toggle, make it off by default...?

I saw the code, and it has such a difficult structure to apply. We need to make a new highlight converter. I concluded that the value for the cost was low. If you make it, it doesn't matter if this function goes into setting. how about that?

jjw24:
> This project is supposed to target "power users", and modern design is very much not afaik

I would say design is still very much important whilst this project targets power users, it would also be a win if we could get beginner to intermediate users on board too.

I always was under the impression that the highlighting was only implemented to help users see how a query is matched. @JohnTheGr8 i suppose the functionality could have further usage, we can always reimplemented specifically for that potential future usage, which will be better designed than to re-use the existing structure.

Worst case we can just bring back the code again from the commit history.

jjw24 (approved):

PR #893, opened 2021-12-11
Windows setting result slowness

taooceros:


Flow.Launcher/ViewModel/ResultsViewModel.cs
```diff
@@ -290,7 +290,7 @@ public void Update(List<ResultViewModel> newItems, CancellationToken token = def
                 if (Count == 0 && newItems.Count == 0 || _token.IsCancellationRequested)
                     return;
 
-                if (editTime < 10 || newItems.Count < 30)
```
jjw24, line 293: editTime is no longer used, remove it?

Plugins/Flow.Launcher.Plugin.WindowsSettings/Helper/ResultHelper.cs
```diff
@@ -54,7 +54,6 @@ internal static class ResultHelper
                     if (areaMatch.IsSearchPrecisionScoreMet())
                     {
                         var settingResult = NewSettingResult(areaMatch.Score + midScore, entry.Type);
-                        settingResult.SubTitleHighlightData = areaMatch.MatchData.Select(x => x + 6).ToList();
```
jjw24, line 57: why removed?

jjw24:
which change actually resolves the slowness?

has this been tested to confirm?

Plugins/Flow.Launcher.Plugin.WindowsSettings/Helper/ResultHelper.cs
taooceros: because translation will make that behave very weird.

Flow.Launcher/ViewModel/ResultsViewModel.cs
taooceros: This change may resolve the issue. The issue looks like to be the time consuming when adding elements separately.

taooceros: > editTime is no longer used, remove it?

yeah remove it

taooceros:
It's similar to the issue of @pc223 and @kubalav indicates before. Only it is easier for people to notice because windowssettings plugin brings a lot of results.

jjw24:
so it should resolve https://github.com/Flow-Launcher/Flow.Launcher/discussions/586 ?

taooceros:
> so it should resolve #586 ?

I think so? but have you encountered some freeze when debugging first query?

jjw24:
i just tested in visual studio with debug mode. i havent encountered any complete freezing, once where it froze for a second or two but then became ok, i think maybe because it's in debug mode? i havent experienced this with a release build.

However i am not finding a difference between this and the dev branch, using visual studio debug mode.

Flow.Launcher/ViewModel/ResultsViewModel.cs
jjw24: why you went from newItems.Count < 30 to newItems.Count < 100? What's the reason for changing this number

jjw24:
This does not seem to address the problem. Marking it draft for now unless it has changes that can be beneficial.

jjw24:
@taooceros ok to delete branch?

PR #895, opened 2021-12-11
Hotfix toggle switch & Fix Strings

onesounds:
For languages with long on/off label, Fix the problem that the toggle switch is not visible properly.

image

- Change Toggle Switch Width Size to Responsive


onesounds:
#901 Additional modification work will be carried out in this pr.

onesounds:
image

onesounds:
It was confirmed that the Context menu did not change depending on the language. This was also changed so that it could be changed to a string.

jjw24 (approved):

PR #897, opened 2021-12-12
Hide Search icon (Magnifying glass)

Garulf:
Simply hides the "Search Icon". 

Should we add option to GUI? or leave as an advanced feature?

image


jjw24:
Leave it as an 'advanced feature's where it needs to be done via xaml. Sorry I replied late.

Garulf:
My PR keeps the setting out of GUI and is available to `settings.json` editing?

jjw24:
The magnifying glass together with the search window identify as the aesthetic of flow, preferably the option is not so easily available but is for theming purposes.

PR #908, opened 2021-12-16
fix calling mainwindow animation after loaded

### LINKED ISSUE #902: Flow Launcher got an error
Flow Launcher version: 1.9.0
OS Version: Microsoft Windows NT 10.0.19043.0
IntPtr Length: 8
x64: True

Python Path: 
Date: 12/14/2021 10:44:46
Exception:
System.NullReferenceException: Object reference not set to an instance of an object.
   at Flow.Launcher.ViewModel.MainViewModel.Show() in C:\projects\flow-launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 749
   at Flow.Launcher.ViewModel.MainViewModel.ToggleFlowLauncher() in C:\projects\flow-launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 727
   at Flow.Launcher.Helper.HotKeyMapper.OnToggleHotkey(Object sender, HotkeyEventArgs args) in C:\projects\flow-launcher\Flow.Launcher\Helper\HotKeyMapper.cs:line 29
   at NHotkey.HotkeyManagerBase.HandleHotkeyMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled, Hotkey& hotkey)
   at NHotkey.Wpf.HotkeyManager.HandleMessage(IntPtr hwnd, Int32 msg, IntPtr wparam, IntPtr lparam, Boolean& handled)
   at System.Windows.Interop.HwndSource.PublicHooksFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndSubclass.DispatcherCallbackOperation(Object o)

taooceros:
fix #902
It's a fix that prevent the error. It looks like it's because mainwindow takes too long to be created and assigned, which results in the issue. We need to investigate that in the future.


jjw24:
@taooceros were you able to replicate this issue?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -746,7 +746,7 @@ public void Show()
             MainWindowVisibilityStatus = true;
             
             if(_settings.UseAnimation)
-                ((MainWindow)Application.Current.MainWindow).WindowAnimator();
+                ((MainWindow)Application.Current.MainWindow)?.WindowAnimator();
```
jjw24, line 749: so if the mainwindow is null, there wont be animation but will still make the window visible or will nothing show at all?

taooceros: If it is null, something unwanted happens 🤣

taooceros:
> @taooceros were you able to replicate this issue?

Sometimes

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: i dont think this is a necessary, when null it will not animate and just do the original show mainwindow, and MainWindow should never be null.
 ```suggestion
                ((MainWindow)Application.Current.MainWindow).WindowAnimator();
```

jjw24:
there is an IsLoaded method attached to the mainwindow class if you want to actually check that the window is loaded.

jjw24 (changes requested):
see comments

Flow.Launcher/ViewModel/MainViewModel.cs
taooceros: ok xd. Just let it throw so we know something weird happen.

taooceros:
> there is an IsLoaded method attached to the mainwindow class if you want to actually check that the window is loaded.

I will take al look

taooceros:
This way should solve the issue. The animation is called inside MainWindow, so no more potential of null reference.

jjw24 (approved):

PR #911, opened 2021-12-17
Plugin activation icon

Garulf:
See @onesounds comment below: https://github.com/Flow-Launcher/Flow.Launcher/pull/911#issuecomment-996713960

### DIFF
--- a/Flow.Launcher/MainWindow.xaml
+++ b/Flow.Launcher/MainWindow.xaml
@@ -197,7 +197,8 @@
                             Margin="0"
                             Data="{DynamicResource SearchIconImg}"
                             Stretch="Fill"
-                            Style="{DynamicResource SearchIconStyle}" />
+                            Style="{DynamicResource SearchIconStyle}"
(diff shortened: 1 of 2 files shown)


onesounds:
This PR for displaying what activated plugin.

It only showing when activated action keyword (not display general query and *) in magnify icon location. It may be a little complicated in design, but it is functionally meaningful in that it gives feedback on whether the keyword works. Also, I don't think it's harmful to the design because it's not shown in the general results.

Garulf:
Possible to have a very fast fade and slide from right animation?

onesounds:
> Possible to have a very fast fade and slide from right animation?

I tried it, but it didn't go well. I can do it to do it, but  It will be very complicated. 
I think it's in good condition now.

Garulf:
Okay, didn't know how much complexity it would add.

jjw24:
hold for 1.10

jjw24 (approved):

PR #917, opened 2021-12-19
Display hotkey in tray context menu

onesounds:
The trigger key is very important, but it is currently notified on the Wizard screen, and in the settings, you need to go to the hotkey tab to know the set key. Add display the trigger key so that you can know immediately when you open the tray menu.


jjw24 (approved):

jjw24:
hold off for 1.10

PR #926, opened 2021-12-22
Insert file path from file on clipboard

Garulf:
This PR allows Flow to retrieve the path of a file that has been copied from the file browser.

https://user-images.githubusercontent.com/535299/147052557-f0a0633d-c329-4076-9065-1fdb56f21ed2.mp4


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -62,6 +67,30 @@ private async void OnClosing(object sender, CancelEventArgs e)
             Environment.Exit(0);
         }
 
+        private void onPreviewCanExecute(object sender, CanExecuteRoutedEventArgs e)
+        {
+            // Allow paste command to accept non-text items
```
jjw24, line 72: what does this mean sorry, pasting non-text into the query window?

jjw24:
lets move the logic into MainViewModel

jjw24:
is this ready?

jjw24:
Just thinking, if you already have the file open in explorer, what usage scenario would you need to paste the file path into flow?

Flow.Launcher/MainWindow.xaml.cs
Garulf: Yeah, normally TextBox elements ignore non-text data. With this we intercept all data input and capture it.

jjw24: oh so this is just required so you can retrieve path from the file object in clipboard?

Garulf:
> Just thinking, if you already have the file open in explorer, what usage scenario would you need to paste the file path into flow?

It can help if you have a file but want to interact with it with Flow... kinda in reverse to what explorer plugin does.

Garulf:
> lets move the logic into MainViewModel

I'll be honest I'm not sure how much can be moved into `MainViewModel`. I'll take a look.

jjw24:
> > Just thinking, if you already have the file open in explorer, what usage scenario would you need to paste the file path into flow?
> 
> It can help if you have a file but want to interact with it with Flow... kinda in reverse to what explorer plugin does.

yeah but everything you can do in flow you can already do in windows explorer?

Garulf:
For now, could be handy if I have a plugin that crops images or uploads them to a page.

jjw24:
Sorry can you help me understand what those workflow would look like? I am still not sure how this could benefit when you already have windows explorer open already.

Garulf:
> Sorry can you help me understand what those workflow would look like? I am still not sure how this could benefit when you already have windows explorer open already.

On second thought it's probably a bad example. Coming from MacOS it just always felt odd Wox or Flow don't know how to handle this.

jjw24:
That's ok, could you give me an example of using this in flow, like how do you crop image then upload with this feature? i am lacking context because I don't use MacOS.

PR #930, opened 2021-12-23
Custom Query Shortcut + built-in variables

taooceros:
1. Expandable Shortcut
2. Built-in variables- {clipboard} for accessing latest clipboard entry


taooceros:
Do we want to reference to variable? So probably a later shortcut can reference back to an earlier shortcut (then we need a move up and move down button).

taooceros:
It will be interesting if we introduce variable like {{clipboard}}

jjw24 (changes requested):
- The title and subtitle of the query shortcut config window need to be changed.
- Please add preview button so it's consistent with custom query hotkey
- Change the name 'Shortcut' to 'Custom Query Shortcut' so it's consistent.

image

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
```diff
@@ -205,6 +204,10 @@ public bool HideNotifyIcon
 
         // This needs to be loaded last by staying at the bottom
         public PluginsSettings PluginSettings { get; set; } = new PluginsSettings();
+        internal ObservableCollection<ShortCutModel> ShortCuts { get; set; } = new()
+        {
+            ("spp", "sp play")
```
jjw24, line 209: this is not neccessary, please remove

Flow.Launcher/Flow.Launcher.csproj
```diff
@@ -12,6 +12,7 @@
     <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <SatelliteResourceLanguages>en</SatelliteResourceLanguages>
+    <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
```
jjw24, line 15: dont think we are referencing other executables, is this needed?

jjw24 (commented):
Shortcut is not being saved.

jjw24:
is this how you would use env variables?

image

jjw24:
>some environmental variables. e.g. clipboard

How do you use this with clipboard?

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
```diff
@@ -220,4 +223,44 @@ public enum ColorSchemes
         Light,
         Dark
     }
+
+    public struct ShortCutModel
```
jjw24, line 227: can we have the naming consistent with custom hotkey please i.e. CustomQueryShortcut/CustomShortcut -> CustomQueryShortcutModel/CustomShortcutModel.

taooceros: Sure, just feel like it is a bit redundant.

taooceros:
> Shortcut is not being saved.

Oh forget to add jsoninclude i guess. Thanks for checking.

Flow.Launcher/Flow.Launcher.csproj
taooceros: I don't know Why, but my visual studio is preventing me to build without this (the latest preview). It is possible that it is Ms issue.

taooceros:
> >some environmental variables. e.g. clipboard
> 
> 
> 
> How do you use this with clipboard?

I manually set {clipboard} to be the content in clipboard.

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
jjw24: Well naming it shortcut is too broad, since they both relate to setting custom queries, i think they fall into the same category and better to maintain consistency in the codebase.

jjw24:
> > > some environmental variables. e.g. clipboard
> > 
> > 
> > How do you use this with clipboard?
> 
> I manually set {clipboard} to be the content in clipboard

Can you show me a screenshot how it's setup?

taooceros:
> > > > some environmental variables. e.g. clipboard
> > > 
> > > 
> > > How do you use this with clipboard?
> > 
> > 
> > I manually set {clipboard} to be the content in clipboard
> 
> Can you show me a screenshot how it's setup?

https://github.com/Flow-Launcher/Flow.Launcher/blob/0c75dfe46fc9a0116f6c69f9aec0397298df0a96/Flow.Launcher/ViewModel/MainViewModel.cs#L679

just write `{clipboard}` in the expanded query.

jjw24:
>some environmental variables. e.g. clipboard

I thought you added this user side that allows expanding of env variables?

taooceros:
> > some environmental variables. e.g. clipboard
> 
> I thought you added this user side that allows expanding of env variables?

oh that might be interesting as well. We can add that feature

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
jjw24: @taooceros

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
jjw24: @taooceros

jjw24:
> Shortcut is not being saved.

@taooceros

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
taooceros: Oh sure i forget sry

VictoriousRaptor:
Can Plugins create shortcuts? If so, shortcuts shouldn't be limited to pure string replacement. They should have the ability to call a function like `string foo()`.

taooceros:
> Can Plugins create shortcuts? If so, shortcuts shouldn't be limited to pure string replacement. They should have the ability to call a function like `string foo()`.

I didn't get the idea? you mean let plugin provide variable?

> 
> And what if user wants to search the shortcut word rather than triggering the shortcut? Like if I want to search `ab` but there's a shortcut `ab->123456`. What if I want to search a file or whatever else called "ab"? Maybe this is a inappropriate shortcut but we should consider this case.

I guess that's user choice? We can always require prefix `$` for shortcut though, but I think it is convenient to do a full string replacement if user only type shortcut.

VictoriousRaptor:
> I didn't get the idea? you mean let plugin provide variable?

Yes.

taooceros:
> > I didn't get the idea? you mean let plugin provide variable?
> 
> Yes.

Ah that's cool. Though I feel like it is not super easy? We may need extra api for that. Mind leave it in the future?

VictoriousRaptor:
I tried to to finish this PR. One last thing to decide, is how to let users know there are builtin shortcuts like `{clipboard}`? Just write them down in documents or put them in the shortcut list?

jjw24:
> I tried to to finish this PR. One last thing to decide, is how to let users know there are builtin shortcuts like `{clipboard}`? Just write them down in documents or put them in the shortcut list?

I thought they are currently displayed like this?
image

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
```diff
@@ -177,6 +177,13 @@ public string QuerySearchPrecisionString
 
         public ObservableCollection<CustomPluginHotkey> CustomPluginHotkeys { get; set; } = new ObservableCollection<CustomPluginHotkey>();
 
+        public ObservableCollection<CustomShortcutModel> CustomShortcuts { get; set; } = new ObservableCollection<CustomShortcutModel>();
+
+        [JsonIgnore]
+        public ObservableCollection<BuiltinShortcutModel> BuiltinShortcuts { get; set; } = new ObservableCollection<BuiltinShortcutModel>() { 
+            new BuiltinShortcutModel("{clipboard}", "Get text from clipboard.", Clipboard.GetText)  // TODO: translation?
```
VictoriousRaptor, line 184: How to translate this string?

taooceros: Try `Internationalization.Instance.GetTranslation(string key)`? I am not sure whether it is available when Settings is being constructed, but probably should?

VictoriousRaptor: It's not. `Internationalization.Instance` initializes after settings because it needs to know selected language.

jjw24:
What plans do we have for adding built in variables apart from clipboard?

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
jjw24: You cannot use translation in this class and infrastructure space in general because it is instantiated late and has a dependency on settings class.

jjw24 (changes requested):

VictoriousRaptor:
> What plans do we have for adding built in variables apart from clipboard?

Not yet.

jjw24:
Explorer plugin doesn't play well with `{`, getting exception when I type `{cli`, need to fix that in explorer.

jjw24:
> > What plans do we have for adding built in variables apart from clipboard?
> 
> Not yet.

So what else can we do with this functionality? Clipboard access can be done better with the Clipboard plugin I feel.

Flow.Launcher/CustomShortcutSetting.xaml.cs
```diff
@@ -0,0 +1,85 @@
+﻿using Flow.Launcher.Core.Resource;
+using Flow.Launcher.Infrastructure.UserSettings;
+using System;
+using System.Linq;
+using System.Windows;
+using System.Windows.Input;
+
+namespace Flow.Launcher
+{
+    public partial class CustomShortcutSetting : Window
```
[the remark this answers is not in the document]
VictoriousRaptor: Should it be in a new VM or in the SettingsWindowVM?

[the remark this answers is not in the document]
VictoriousRaptor: Should it be in a new VM or in the SettingsWindowVM?

VictoriousRaptor:
> Explorer plugin doesn't play well with `{`, getting exception when I type `{cli`, need to fix that in Explorer.

Maybe it is related to the vscode workspace plugin (default action word `{`). I remember I had similar issues when testing along with the plugin.



> > > What plans do we have for adding built in variables apart from clipboard?
> > 
> > 
> > Not yet.
> 
> So what else can we do with this functionality? Clipboard access can be done better with the Clipboard plugin I feel.

This design is to provide functionality more than simple static string replacment. I don't have any good idea about what else we can do with it yet.

Flow.Launcher/CustomShortcutSetting.xaml.cs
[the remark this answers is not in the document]
jjw24: I think we can chuck them in SettingsWindowVM.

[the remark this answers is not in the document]
jjw24: and place inside a region.

jjw24:
>Maybe it is related to the vscode workspace plugin (default action word {). I remember I had similar issues when testing along with the plugin.

Nah, actually reproducible with Dev branch.

jjw24:
>This design is to provide functionality more than simple static string replacment. I don't have any good idea about what else we can do with it yet.

Feels like half-baked at the moment because users wouldn't be able to add new ones in and clipboard access can be done with the clipboard plugin. What about pulling it out to a separate pr?

VictoriousRaptor:
> > This design is to provide functionality more than simple static string replacment. I don't have any good idea about what else we can do with it yet.
> 
> Feels like half-baked at the moment because users wouldn't be able to add new ones in and clipboard access can be done with the clipboard plugin. What about pulling it out to a separate pr?

Maybe we should discuss it later, when other problems of this pr are fixed.

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
VictoriousRaptor: Translation can be done when calling `get` method.

jjw24: Are we placing this in the get method in this pr?

jjw24: Maybe instead of the actual string, put the key in, and in the get call do the translation of the key.

VictoriousRaptor: > Are we placing this in the get method in this pr?

If we decide to include built-in shortcuts in this pr I will implement this translation.

PR #933, opened 2021-12-24
Use Global Hook for HotkeyControl

taooceros:
Use Global Hook when HotkeyControl get focus to allow stop win key activate start menu.


Jack251970:
@jjw24 Since you have supported Win hotkey in #3262, can we close it?

jjw24:
https://github.com/Flow-Launcher/Flow.Launcher/pull/3157

PR #935, opened 2021-12-27
Revise ttf/otf code

taooceros:
WPF requires font name after the font source. This patch adds support for that.


jjw24:
has this part been working though? what's this improvement addressing?

taooceros:
> has this part been working though? what's this improvement addressing?

well it's a bug. I previously doesn't understand how the font works, so I assume putting the ttf file to the fontfamily will work (when I test it, the text display but with a fallback font which I don't notice). In fact, it needs to indicate the font name after the ttf file address.

Flow.Launcher/ViewModel/ResultViewModel.cs
```diff
@@ -41,6 +62,8 @@ public ResultViewModel(Result result, Settings settings)
             Settings = settings;
         }
 
+        
+
```
jjw24, line 66:
```suggestion
```

PR #937, opened 2021-12-28
JsonRPCPlugin Version 2

taooceros:


jjw24:
why is this merging into the dotnet 6 branch instead? is this ready?

taooceros:
> why is this merging into the dotnet 6 branch instead? is this ready?

Not really, but it works.

jjw24:
should we change the base to Dev branch instead of dotnet 6?

taooceros:
> should we change the base to Dev branch instead of dotnet 6?

well why not merge dotnet 6 before this one XD? I don't see a specific reason to not include dotnet 6 in 1.10

jjw24:
Have you tested .net 5/core (if we still have any) with the Flow.Launcher.Plugin build?

taooceros:
> Have you tested .net 5/core (if we still have any) plugins with the Flow.Launcher.Plugin build?

Yes, they will be separate.

jjw24:
See #1353

PR #938, opened 2021-12-29
Update XML Docs

taooceros:


Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
```diff
@@ -42,6 +42,7 @@
     <WarningLevel>4</WarningLevel>
     <PlatformTarget>AnyCPU</PlatformTarget>
     <Prefer32Bit>false</Prefer32Bit>
+    <DocumentationFile>..\Output\Debug\Flow.Launcher.Plugin.xml</DocumentationFile>
```
JohnTheGr8, line 45: This option is redundant, isn't it? You have only added it to the Debug group, currently, but the CI-built dll file has all the comments...

Flow.Launcher.Plugin/Result.cs
```diff
@@ -36,6 +36,9 @@ public class Result
         /// </summary>
         public string AutoCompleteText { get; set; }
 
+        /// <summary>
+        /// Image Displayed on the result
```
JohnTheGr8, line 40: I think it would help to point out what kind of values are expected here.. I assume we only support local file paths (do we support URLs here yet?)

Also, let's make it clear that this is ignored if `Glyph` is set...

Flow.Launcher.Plugin/Result.cs
```diff
@@ -60,16 +63,23 @@ public string IcoPath
...
-        /// Information for Glyph Icon
+        /// Information for Glyph Icon (Prioritized than IcoPath/Icon if user enable Glyph Icons)
         /// </summary>
-        public GlyphInfo Glyph { get; init; } 
+        public GlyphInfo Glyph { get; init; }
 
 
         /// <summary>
-        /// return true to hide flowlauncher after select result
+        /// Delegate activated when result is been selected
```
JohnTheGr8, line 72:
```suggestion
        /// Delegate activated when the result has been selected
```

Flow.Launcher.Plugin/Result.cs
```diff
@@ -116,22 +127,25 @@ public override bool Equals(object obj)
...
+        /// <inheritdoc />
         public override string ToString()
         {
             return Title + SubTitle;
         }
 
-        public Result() { }
-
         /// <summary>
         /// Additional data associate with this result
```
JohnTheGr8, line 145:
```suggestion
        /// Additional data associated with this result
```

Flow.Launcher.Plugin/Result.cs
```diff
@@ -83,9 +93,9 @@ public string IcoPath
         public IList<int> SubTitleHighlightData { get; set; }
 
         /// <summary>
-        /// Only results that originQuery match with current query will be displayed in the panel
+        /// Query information associated with the result
         /// </summary>
-        internal Query OriginQuery { get; set; }
+        public Query OriginQuery { get; set; }
```
JohnTheGr8, line 98: Any practical reason for making this public?

JohnTheGr8 (commented):
my last nitpicking session of 2021 😂

Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
taooceros: I add it for generating the api docs (which I hasn't published, and directly reference this project). It doesn't need to present in release.

Flow.Launcher.Plugin/Result.cs
taooceros: We stop using this to filter result, so no reason to mark it as internal. Either we can remove it to avoid some duplicate assignments or just mark it as public so plugin user may access it.

taooceros:
> my last nitpicking session of 2021 😂

Long time no see😉

Flow.Launcher.Plugin/Result.cs
```diff
@@ -60,16 +65,23 @@ public string IcoPath
...
-        /// Information for Glyph Icon
+        /// Information for Glyph Icon (Prioritized than IcoPath/Icon if user enable Glyph Icons)
         /// </summary>
-        public GlyphInfo Glyph { get; init; } 
+        public GlyphInfo Glyph { get; init; }
 
 
         /// <summary>
-        /// return true to hide flowlauncher after select result
+        /// Delegate activated when the result has been selected
```
jjw24, line 74:
```suggestion
        /// Delegate. An action in the form of a method call to take when the result has been selected
```

Flow.Launcher.Plugin/Result.cs
JohnTheGr8: if we can remove it, I vote that we do.. seems like it can prove confusing...

jjw24: lets revert back to internal and remove later

JohnTheGr8: doing this now

JohnTheGr8 (approved):

jjw24 (approved):

PR #952, opened 2022-01-03
Fix Toggle Switch Knob (White Circle) Positioning

onesounds:
**Before** 
- The knob position of the toggle switch is wrong.
- This is a problem caused by the use of RTL for right alignment.

image

- When the window size increases, the positions of the Priority button and the toggle switch operate incorrectly.

image

**After**
- Toggle Switch with LTR

image

- Fix Toggle Switch Circle Positioning
- Change "Open Theme Folder" Button Width to Responsive
- Adjust Priority Button Position
- Even if the window size changes, it is aligned to the right.

image

**Test / Validate**
- When the toggle switch is turned on, the inner round circle (aka Knob) must be on the right.
- For languages with long toggle switch on/off labels, the area is automatically adjusted and all must be displayed. 
- The dot mark in Slovak language should be displayed on the right.
- When adjusting the size of the setting window, the Priority button and label should be displayed well.

### DIFF
--- a/Flow.Launcher/Resources/CustomControlTemplate.xaml
+++ b/Flow.Launcher/Resources/CustomControlTemplate.xaml
@@ -1469,6 +1469,7 @@
             <Setter.Value>
                 <ControlTemplate TargetType="ui:ToggleSwitch">
                     <Border
+                        Width="Auto"
                         Background="{TemplateBinding Background}"
                         BorderBrush="{TemplateBinding BorderBrush}"
                         BorderThickness="{TemplateBinding BorderThickness}"
@@ -1498,7 +1499,7 @@
                                 Visibility="Collapsed" />
                             <Grid
                                 Grid.Row="1"
-                                MinWidth="{DynamicResource ToggleSwitchThemeMinWidth}"
+                                MinWidth="10"
                                 HorizontalAlignment="Right"
                                 VerticalAlignment="Top">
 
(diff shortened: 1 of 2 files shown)


taooceros:
remove of this line change the direction of the toggle button?
`<Setter Property="FlowDirection" Value="RightToLeft" />`

onesounds:
> remove of this line change the direction of the toggle button? `<Setter Property="FlowDirection" Value="RightToLeft" />`

Theoretically, Yes. but this should not be used.  it is a temporary measure used in monolingual programs.

taooceros:
> > remove of this line change the direction of the toggle button? `<Setter Property="FlowDirection" Value="RightToLeft" />`
> 
> Theoretically, Yes. but this should not be used. it is a temporary measure used in monolingual programs.

So the change in this pull request is basically using a bunch of right horizontal alignment to replace the flowdirection?

onesounds:
1. The toggle switch used in the past is only aligned on the left and is not aligned on the right. The label is also inserted only on the right and cannot be changed.
2. I had to move the label to the left and change it to be aligned to the right.
3. I checked on Google that changing Flowdirection changes this, and I used some of it.
4. However, RTL is officially a function for RTL language. So Kubalav and Citizendee found the problem.
- The dot is displayed on the left side of the label.
- The knob of the toggle switch is reversed.  In general, the right side is in the On state.

image

5. So I removed the part using RTL and rearranged the toggle switch control completely. That's this PR.

jjw24 (approved):

Issue #950, opened 2022-01-03
Windows + R bringing up Flow Launcher even when not set as the hotkey


STaRDoGG:
**Describe the bug/issue**

I've got my trigger hotkey set to ctrl + space .... when I press Winkey + r the FL search bar pops up. This prevents me from using the standard Windows Run hotkey.

**To Reproduce**
Press Winkey + R ...

**Screenshots**

![image](https://user-images.githubusercontent.com/1524526/147930083-51b5e706-f658-4f02-a5a0-3b95aec36800.png)

**Your System**
```
Windows build number: (run "ver" at a command prompt)
Microsoft Windows [Version 10.0.19043.1415]

Flow Launcher version: (Settings => About) 
1.9.1

```
**Flow Launcher Error Log**

[2022-01-03.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/7801886/2022-01-03.txt)

onesounds:
![image](https://user-images.githubusercontent.com/6903107/147930962-a580b7b0-3624-4f07-af89-403047f59649.png)

- Turn off "Replace Win+R" in Shell Plugin.

STaRDoGG:
@onesounds aha! ty! Didn't pour through all the settings hard enough I guess.

onesounds:
I've experienced the same confusion as you at first. First of all, I don't think it's a big problem because it indicates that the ">" mark is intended to appear. I'll think about how to add a simple guide.

And that plug-in is a plug-in that allows you to use shell commands. Try to get used to it.
- Turn On " Do not close command prompt after command excution"
- and try this. "> start shell:RecycleBinFolder" or "> ipconfig"

STaRDoGG:
@onesounds Maybe FL can detect if the Shell plugin is enabled/installed/Replace Win + R is enabled, and if so, add a text link under the normal hotkey mentioning it, and clicking the link navigates to that plugin's setting.  That would make it more intuitive for people like me, lol.

Another think I just noticed that I'm not sure is intended or not; the `>` was automatically being added to the search bar when the bar was opened, and selected (highlighted), with nothing after it, presumably to make it a bit easier to run commands. But in order to run a command you'd have to first unhighlight it either manually with a mouse click, or a right arrow, and then start typing. And if you hit the delete/backspace while the `>` is selected when the bar comes up, it gets deleted, and doesn't get auto-added the next time the bar comes up. Sort of inconsistent behavior imho.

[...]

Issue #391, opened 2021-03-24
Stop awaiting plugin initialization before showing up the main window


taooceros:
A thought from #371, what about let's stop awaiting the plugin initialization, but instead marking the task (so we won't query the secific plugin until it has been initialized), to improve the startup time (also avoidng potential 3rd party plugin that contains long start up time).

When plugin is not initialized, if it is a global plugin, we stop querying them without showing sign, but if it is a non-global plugin, we provide one result that is showing that it is still loading.

@Flow-Launcher/team What do you think?

jjw24:
When you say global plugin, you mean plugin with global action keyword?

taooceros:
> When you say global plugin, you mean plugin with global action keyword?

Yes

jjw24:
I would say this feels more like a hack than a feature, because plugins should not take too longer to load (I consider more than 2 minutes as long) and if that is the case then should be addressed by the plugin developer. By just letting plugin to finish loading after a long wait, it is also not a good user experience. If the user wants to find something and the plugin is not ready then how will the user know for plugins that use global action keyword. When the query box appears, it should be ready to go, not a plugin maybe ready and user may have to wait.

I would rather have plugins that take too long to load be logged, made known to the user and then addressed by developer than to just let it be.

Great idea, but I don't think this solves the root of the problem.

taooceros:
Yeah you are right, this is more like a hack, but I have a thought that user may not need all plugin feature at the time they start up flow. What if we only wait for the global Plugins, so that user won't be confused about whether one of them are not initialized? 

This can also simplify some development burden to consider whether I need to reduce the Init time (e.g. In the Pluginsmanager plugin) that they just need to await it.

jjw24:
> user may not need all plugin feature at the time they start up flow.  What if we only wait for the global Plugins, so that user won't be confused about whether one of them are not initialized?

This is going to be hard to implement because how do we know what user actually needs at start up, you can also switch plugins from global to non-global.

PluginsManager start up slow is a network thing, not really something we can solve with Flow itself.

jjw24:
Any init slowness that is non-network, then can be addressed by flow or plugin's developers.

taooceros:
> This is going to be hard to implement because how do we know what user actually needs at start up, you can also switch plugins from global to non-global.

[...]

PR #953, opened 2022-01-03
stop trigger global hotkey when shell plugin is not enabled

taooceros:
temporary fix for #950 
We should stop init disabled plugin when startup. corresponding to idea of #391


jjw24:
i thought disabled plugins are not loaded at all?

jjw24:
any chance we can stop init disabled plugin when startup? this means they wont even show up in Settings window right?

taooceros:
> any chance we can stop init disabled plugin when startup? this means they wont even show up in Settings window right?

Possible. Currently the Init method does two thing, passing context to the plugin, and allow it to do some initialization thing. If the plugin needs the initialization for setting panel, they may fail to create it without the initialization.

jjw24:
Just checked out the rest of the code, i see your point, will merge this in for now.

jjw24 (approved):

PR #963, opened 2022-01-07
Use Microsoft Package to handle toast notification for registry 

### LINKED ISSUE #964: Notification not registered in the system
### Discussed in https://github.com/Flow-Launcher/Flow.Launcher/discussions/946

<div type='discussions-op-text'>

<sup>Originally posted by **stax76** January  1, 2022</sup>
I disabled Flow Launcher notifications system-wide in Windows 11, I think I did this with a UI setting, and now I'm not able to re-enable notifications, I'm not getting a notification when I press F5 using v1.9.1. I've searched the Windows 11 settings UI and the registry but could not find a way to be re-enabled notifications. In the Windows settings under 'System > Notifications' Flow Launcher is not listed, I tried reinstalling, but it did not help.</div>

taooceros:
#946
close #964

### DIFF
--- a/Flow.Launcher/Notification.cs
+++ b/Flow.Launcher/Notification.cs
@@ -1,4 +1,5 @@
 ﻿using Flow.Launcher.Infrastructure;
+using Microsoft.Toolkit.Uwp.Notifications;
 using System;
 using System.IO;
 using Windows.Data.Xml.Dom;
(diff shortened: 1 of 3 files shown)


Flow.Launcher/Notification.cs
```diff
@@ -8,10 +9,17 @@ namespace Flow.Launcher
 {
     internal static class Notification
     {
+        internal static bool legacy = Environment.OSVersion.Version.Build < 19041;
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Interoperability", "CA1416:Validate platform compatibility", Justification = "<Pending>")]
+        internal static void Uninstall()
+        {
+            if (!legacy)
+                ToastNotificationManagerCompat.Uninstall();
```
jjw24, line 17: What does this Uninstall do?

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -317,7 +317,7 @@ public void Init(PluginInitContext context)
 
         bool API_GlobalKeyboardEvent(int keyevent, int vkcode, SpecialKeyState state)
         {
-            if (_settings.ReplaceWinR)
+            if (!context.CurrentPluginMetadata.Disabled && _settings.ReplaceWinR)
```
jjw24, line 320: Is it easy to full disable the plugin without doing it here? Obviously taking into consideration that disabled plugin still needs to be shown in the Settings window

Flow.Launcher/Notification.cs
taooceros: https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp#step-4-handling-uninstallation

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
taooceros: Not easy due to the current model. We cannot make sure the plugin won't do anything or register anything under its constructor or Init method, so we can only assume that.

jjw24 (approved):

PR #968, opened 2022-01-09
Support Application Uri in OpenUrl API call

taooceros:
Fallback to direct process open when uri is not started with http/https


JohnTheGr8:
Maybe add a different overload that accepts a `Uri` for such cases? Having the `inPrivate` parameter there when this might not run in browser is awkward.

taooceros:
> Maybe add a different overload that accepts a `Uri` for such cases? Having the `inPrivate` parameter there when this might not run in browser is awkward.

Hm if that's the case, I would doubt whether we should include a new public api because it is simply a helper function.

JohnTheGr8:
> 
> Hm if that's the case, I would doubt whether we should include a new public api because it is simply a helper function.

nothing about this suggests that you can pass an application Uri and it will be started as a process:

https://github.com/Flow-Launcher/Flow.Launcher/blob/b64dd27d5310d196adfe81cb18c10e76ed8bde8f/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs#L230-L233

jjw24 (changes requested):
please update

jjw24:
> > Hm if that's the case, I would doubt whether we should include a new public api because it is simply a helper function.
> 
> nothing about this suggests that you can pass an application Uri and it will be started as a process:
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/b64dd27d5310d196adfe81cb18c10e76ed8bde8f/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs#L230-L233

resolved.

jjw24 (approved):

jjw24:
@taooceros i dont think the logic prior to my change is correct. Can you give me an App URI example you were testing with.

jjw24:
@taooceros sorry, my mistake. This is now fixed in PR1002

PR #969, opened 2022-01-10
Move Tasklink Name into Resource (except altName) & Glyph Icon Support

taooceros:
It is creating too much warning in the log file because of translation. No need for the warning because it is simply the resources doesn't exist.


jjw24:
Should we perhaps change it to `if debug` to log when only debugging? Would there be a case where we need to determine why a resource is not loaded?

Also plugin version bump please

taooceros:
> Should we perhaps change it to `if debug` to log when only debugging? Would there be a case where we need to determine why a resource is not loaded?
> 
> Also plugin version bump please

Generally it is simply because of the translation doesn't exist, which should be totally fine. We cannot distinguish this from other reason because it is simply return null.

taooceros:
wait a moment. If it's translation issue, shouldn't be so much resources can't be found. I will check again.

taooceros:
huh those strings are for control panel (probably for task link added by @pc223 ) I remember he adds the translation for that previously. We can add that first.

jjw24:
> huh those strings are for control panel (probably for task link added by @pc223 ) I remember he adds the translation for that previously. We can add that first.

I thought we included his changes already

jjw24:
> I remember he adds the translation for that previously. We can add that first.

I wasnt too invovled with the PR/changes, do you know where we can find the translations he added previously?

taooceros:
> > I remember he adds the translation for that previously. We can add that first.
> 
> I wasnt too invovled with the PR/changes, do you know where we can find the translations he added previously?

https://github.com/pc223/MSWindowsControlPanelTaskLinks
he has added part of that here

taooceros:
I will fix the resources by adding a english one first

jjw24 (approved):

PR #682, opened 2021-09-06
Refactor PluginsManager.

taooceros:
1. Stop showing pre-built plugin with pm uninstall
2. show updatable plugin with pm install
3. refactor code
resolves #264

PR #972, opened 2022-01-11
Improvements to PluginManager plugin

JohnTheGr8:
The main change introduced here is the use of conditional http requests when fetching the plugin manifest json. This allows us to always check for changes to the manifest repo when running one of `pm install`, `pm update`, with very little performance overhead (for me it is about 50ms). It also allows us to simplify the relevant code a bit.

Also some quick refactoring in `PluginsManager` and a couple quality-of-life enhancements:

1. typing `pm inst` and hitting tab would autocomplete to `pm install` without a space at the end (same for the other base commands)
2. fix incorrect link (for some plugins) when selecting to open an issue from the context menu

PS: I missed #682 until I started writing this PR text... my apologies. I see similarities, perhaps we can pick some of the missing changes into this PR?

### DIFF
--- a/Flow.Launcher.Core/ExternalPlugins/PluginsManifest.cs
+++ b/Flow.Launcher.Core/ExternalPlugins/PluginsManifest.cs
@@ -1,7 +1,8 @@
-﻿using Flow.Launcher.Infrastructure.Http;
 using Flow.Launcher.Infrastructure.Logger;
 using System;
 using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


taooceros:
Should we expand the helper class instead of creating a new httpclient for plugins manager only? Also this may not follow the proxy setting.

Flow.Launcher.Core/ExternalPlugins/PluginsManifest.cs
```diff
@@ -10,43 +11,45 @@ namespace Flow.Launcher.Core.ExternalPlugins
 {
     public static class PluginsManifest
     {
-        static PluginsManifest()
-        {
-            UpdateTask = UpdateManifestAsync();
-        }
-
-        public static List<UserPlugin> UserPlugins { get; private set; } = new List<UserPlugin>();
+        private const string manifestFileUrl = "https://raw.githubusercontent.com/Flow-Launcher/Flow.Launcher.PluginsManifest/plugin_api_v2/plugins.json";
```
taooceros, line 14: By the way, could you change this to jsdelivr link to help Chinese users? Not sure whether they support etag as well.

JohnTheGr8:
> Should we expand the helper class instead of creating a new httpclient for plugins manager only? Also this may not follow the proxy setting.

Yup, you're right, I wanted to keep it simple but I forgot about proxy settings... Will add a helper method inside `Http` that accepts a `HttpRequestMessage`

Flow.Launcher.Core/ExternalPlugins/PluginsManifest.cs
JohnTheGr8: done, it's significantly faster as well, github takes ~50ms to return a 304 for me, jsdelivr takes ~20ms

JohnTheGr8:
There's one issue I just noticed: launching Flow and opening `Settings > Plugin Store` (without first using `pm install/update`) will show no plugins...

Possible workarounds:

1. actually fetch the plugin data in `PluginsManager`'s `InitAsync` (which will add a few ms to the init time)
2. call `PluginsManifest.UpdateManifestAsync()` when the user opens settings (or something similar)

thoughts?

taooceros:
I would consider that we just call download manifest in init? We can decide not to await that so it won't change init time.

JohnTheGr8:
> I would consider that we just call download manifest in init? We can decide not to await that so it won't change init time.

Alright, was hoping for this since I am not very familiar with WPF 😛 It's pretty fast with the CDN link, anyways...

taooceros:
> > I would consider that we just call download manifest in init? We can decide not to await that so it won't change init time.
> 
> Alright, was hoping for this since I am not very familiar with WPF 😛 It's pretty fast with the CDN link, anyways...

lol it's not related to WPF. Flow waits for plugin init before showing up.

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -523,36 +494,6 @@ private void Uninstall(PluginMetadata plugin, bool removedSetting = true)
             using var _ = File.CreateText(Path.Combine(plugin.PluginDirectory, "NeedDelete.txt"));
         }
 
-        private List<Result> AutoCompleteReturnAllResults(string search, string hotkey, string title, string subtitle)
```
taooceros, line 526: All of this is changed to `GetDefaultHotkey.Where...` right

JohnTheGr8: after some head-scratching, the only use for `AutoCompleteReturnAllResults` that I could find was that it allowed queries like `pm install install install` 😂 

I think it can safely go....

Flow.Launcher.Infrastructure/Http/Http.cs
```diff
@@ -153,5 +153,13 @@ public static async Task<Stream> GetStreamAsync([NotNull] string url, Cancellati
             var response = await client.GetAsync(url, HttpCompletionOption.ResponseHeadersRead, token);
             return await response.Content.ReadAsStreamAsync();
         }
+
+        /// <summary>
+        /// Asynchrously send an HTTP request.
+        /// </summary>
+        public static async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken token = default)
+        {
+            return await client.SendAsync(request, token);
```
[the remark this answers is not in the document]
JohnTheGr8: Added, good call...

taooceros (approved):

jjw24 (approved):

PR #1002, opened 2022-01-31
[Dev] Fix application URI logic when opening via WebSearch plugin

jjw24:
Fix incorrect logic added in https://github.com/Flow-Launcher/Flow.Launcher/pull/968 where application URIs opened from WebSearch plugin are not be triggered

### DIFF
--- a/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
@@ -228,13 +228,26 @@ namespace Flow.Launcher.Plugin
         public void OpenDirectory(string DirectoryPath, string FileName = null);
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


JohnTheGr8 (commented):
This seems to have been done in a bit of a rush 😆 Can we refactor to make it look a bit nicer?

I think we can:

1. move the `Process.Start` call from `OpenAppUri` to the private `OpenUri` and have everything in one place, and call `OpenUri` from `OpenAppUri` (similar to `OpenUrl`)
2. use an `else` statement in `OpenUri` and remove those early returns

Also, as mentioned before, I think it's a good idea to add an overload for `OpenAppUri` (maybe also for `OpenUrl`) that accepts a `Uri` instead of a `string`. I think plugins that use this method might already have a Uri before calling it, so why create a new one in `OpenUri`...

jjw24:
done, please review @JohnTheGr8 @taooceros

JohnTheGr8 (approved):

PR #1013, opened 2022-02-08
Allow plugins to capture hotkey actions on query window results

jjw24:
**Goal:**
Allow plugins to capture hotkey presses on query window results, such as Ctrl+P, F1, Space, Ctrl+Backspace, Up or Down. 

This information is then passed onto each plugin, which will allow them to do some action based on the hotkey and information supplied.

This is a proof of concept PR, using F1 key press as an example to demonstrate when F1 is pressed on a result, the information of the selected result is translated to a path which is then passed onto each plugin. Only if the plugin inherits IPathSelected, will it be able to capture and use the path and hotkey information to do something further.
This example uses Explorer plugin, and works with JsonRPC plugins (i.e. Python, TS, JS and etc).

**What success looks like:**
1. Allow plugin to utilise QuickLook from #859 and previews should work on Up, Down or F1 hotkeys on the selected result.  
2. Allow Explorer or Everything plugin to utilise Ctrl+C on a selected result that is a file/folder and have it copied on to clipboard. 
3. JsonRPC plugins to also utilise the new feature.

### DIFF
--- a/Flow.Launcher.Core/Plugin/ExecutablePlugin.cs
+++ b/Flow.Launcher.Core/Plugin/ExecutablePlugin.cs
@@ -21,12 +21,15 @@ namespace Flow.Launcher.Core.Plugin
                 RedirectStandardOutput = true,
                 RedirectStandardError = true
             };
-            
-            // required initialisation for below request calls 
+
+            // required initialisation for below request calls
             _startInfo.ArgumentList.Add(string.Empty);
         }
 
-        protected override Task<Stream> RequestAsync(JsonRPCRequestModel request, CancellationToken token = default)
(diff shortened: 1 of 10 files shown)


jjw24:
@taooceros please have a look at this concept when you can. This will allow any plugin to use hotkeys not just the default ones. Feeds into the Quicklook work.

jjw24:
On second thought, not feasible to map hotkeys ongoing. Better to allow plugin support via functionality, i.e. for external preview support with a QuickLook plugin design https://github.com/Flow-Launcher/Flow.Launcher/pull/2082/commits/666211dd4cecae901f0a5509447fc07e2a72feaa. 

Closing.

PR #1018, opened 2022-02-10
Quickswitch

taooceros:


taooceros:
Some threading issues seem appearing. Not sure the detailed reason.

Flow.Launcher/QuickSwitch/NativeHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+        public const uint WINEVENT_OUTOFCONTEXT = 0;
+        public const uint EVENT_SYSTEM_FOREGROUND = 3;
+
+        [DllImport("user32.dll")]
+        public static extern unsafe IntPtr SetWinEventHook(uint eventMin, uint eventMax, IntPtr hmodWinEventProc,
+            delegate* unmanaged<IntPtr, uint, IntPtr, int, int, uint, uint, void> lpfnWinEventProc,
+            uint idProcess, uint idThread, uint dwFlags);
+
+        [DllImport("user32")]
+        public static extern IntPtr GetForegroundWindow();
```
stefnotch, line 20: Quick design questions: 
- Would it ever make sense for a plugin to also query which *other* explorer windows are open? Stuff like having two explorer windows open side by side...?
- And, *how* should the case be handled where the user clearly has one explorer window open, but it's not the topmost window? Like having the explorer window opened, but being focused on the task manager?

taooceros: > Would it ever make sense for a plugin to also query which other explorer windows are open? Stuff like having two explorer windows open side by side...?

Then only sync the path that is moved foreground. I didn't design this feature, but just try to mimic the design.

Here's the original design. https://www.listary.com/quick-switch

> And, how should the case be handled where the user clearly has one explorer window open, but it's not the topmost window? Like having the explorer window opened, but being focused on the task manager?

It may be useful to sync explorer path that is not at the foreground. But I guess the case you describe is somewhat different. The original design gives user control about what they want to sync or when they want to sync the explorer path. Therefore, they need to put the explorer up to front to notify which path they want to sync for.

taooceros:
> Okay, my main questions would be
> 
>     * What's the intended use-case for this? As in, when does a Flow plugin need to navigate in the actual file explorer, instead of opening a new one with the correct path.

I think the major use case is to sync the path of an opened explorer to a open file dialog to select a file easily.

> Would opening a new file explorer window with the desired path, and closing the old one work? Or does that not handle certain cases? (e.g. [the file browser/chooser](https://i.stack.imgur.com/gx6IT.png) )

Sry I don't get the idea.

Flow.Launcher/QuickSwitch/NativeHelper.cs
stefnotch: Oh, I understand! Then all of this makes perfect sense.

stefnotch:
Okay, I understand what this is used for now. I'd have to dig a lot deeper into what the IUIAutomation can do to be able to improve this.

I think the rule of thumb is to avoid sending keyboard events, and instead always use an API if one exists. Keyboard events can be delayed and whatnot.

taooceros:
> Okay, I understand what this is used for now. I'd have to dig a lot deeper into what the IUIAutomation can do to be able to improve this.
> 
> I think the rule of thumb is to avoid sending keyboard events, and instead always use an API if one exists. Keyboard events can be delayed and whatnot.

Yeah that's what I would like to see. It is possible to use PInvoke directly without IUIAutomation though, so it will be cool if you are familiar with that as well.

Another thing is the original listary seems implement this feature without changing the textbox and sending an enter signal, so I wonder whether you may have some clues about that.

stefnotch:
I tried searching for what I could, but that's apparently quite tricky to hook into. So I don't really have a better solution at the moment.

taooceros:
> I tried searching for what I could, but that's apparently quite tricky to hook into. So I don't really have a better solution at the moment.

okay thanks🤣

stefnotch:
There might be a alternate design:

So the file manager has the "quick access" sidebar. Flow could add its own entry there, and that entry always redirects to the currently open folder. An additional advantage might be that it's easier to discover this, compared to a keyboard shortcut.

image

(Note: I have no idea how hard that would be to efficiently pull that off.)

taooceros:
So you mean to add a entry that redirect to the most recent opened explorer path?🤔Interesting

stefnotch:
Yep, spot-on.

taooceros:
> Yep, spot-on.

If that's the case, we may be able to create a plugin for it.

taooceros:
Do you have any docs for that?

stefnotch:
@taooceros I haven't looked into this all that much (just a few cursory google searches

## Programmatic access

Apparently there's a way of programmatically adding folders to the quick access area.

https://stackoverflow.com/questions/30051634/is-it-possible-programmatically-add-folders-to-the-windows-10-quick-access-panel


## Special Links folder
https://blogs.msmvps.com/kenlin/2017/06/14/537/

Steps:
1. Enable a special, built-in folder by setting a value in the system registry. Anything in this folder will land in the "quick access".
2. Put a *shortcut* in that folder. (A `.lnk` shortcut)
3. And then always update the shortcut's path to point at the currently open file explorer.

## Symbolic links or Hardlink

I bet there's some trickery that could be done with those


## Extra harddrive

We could add an in-memory harddrive, mount it and provide a single shortcut in there.

mcthesw:
Could this be done? I really love this feature.

stefnotch:
Yet another option would be to add a "switch to" context menu entry

Sort of like how 7zip has a dynamic context menu, except that we'd populate it with the titles of other explorer windows.
image

stefnotch:
Apparently Windows 11 can add files to quick access. [That might let us pin a program to quick access](https://developerpublish.com/how-to-pin-or-unpin-files-via-quick-access-in-windows-11/)

rockingmax:
Really hope we can get the quick switch function :( the Ctrl+G in Listary is so useful

onesounds:
@Jack251970 It works fine with "Save As," but when using "Open," issues arise (such as the program crashing or an error popup appearing). The problem is not with the workflow but occurs within the program itself.

onesounds:
The dialog swapping works well. But I think we should improve the search feature that shows up at the bottom. Right now, it’s just a display-only feature. In Listary, when you type a path or filename at the bottom of the dialog, it directly inputs that into the dialog. (In other words, typing a specific path or file and hitting enter inputs it into the dialog.) If we’re not going to support that behavior just yet, I think the “show” feature should be turned off and either hidden or disabled in the UI, otherwise we shouldn't include it in the release.

Jack251970:
> The dialog swapping works well. But I think we should improve the search feature that shows up at the bottom. Right now, it’s just a display-only feature. In Listary, when you type a path or filename at the bottom of the dialog, it directly inputs that into the dialog. (In other words, typing a specific path or file and hitting enter inputs it into the dialog.) If we’re not going to support that behavior just yet, I think the “show” feature should be turned off and either hidden or disabled in the UI, otherwise we shouldn't include it in the release.

You can use right click to open file or directory. And Explorer plugin already supports quick switch. I am planning release this feature as beta in this version if you think it is good to go.

Flow.Launcher.Infrastructure/QuickSwitch/QuickSwitch.cs
```diff
@@ -0,0 +1,545 @@
...
+        // The class name of a dialog window
+        private const string DialogWindowClassName = "#32770";
+
+        private static readonly Settings _settings = Ioc.Default.GetRequiredService<Settings>();
+
+        private static readonly object _lastExplorerViewLock = new();
+
+        private static readonly object _dialogWindowHandleLock = new();
+
+        private static IWebBrowser2 _lastExplorerView = null;
```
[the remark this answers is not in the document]
Jack251970: Good

Flow.Launcher.Infrastructure/QuickSwitch/QuickSwitch.cs
```diff
@@ -0,0 +1,548 @@
...
+                // Navigate to path
+                if (_settings.AutoQuickSwitch)
+                {
+                    // Check if we have already switched for this dialog
+                    bool alreadySwitched;
+                    lock (_dialogWindowHandleLock)
+                    {
+                        alreadySwitched = _autoSwitchedDialogs.Contains(hwnd);
+                    }
+
```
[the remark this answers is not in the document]
Jack251970: Good!

Flow.Launcher.Infrastructure/QuickSwitch/QuickSwitch.cs
```diff
@@ -0,0 +1,548 @@
...
+                _destroyChangeHook = null;
+            }
+
+            // Release ComObjects
+            if (_lastExplorerView != null)
+            {
+                Marshal.ReleaseComObject(_lastExplorerView);
+                _lastExplorerView = null;
+            }
+        }
```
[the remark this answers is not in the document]
Jack251970: Good!

onesounds (dismissed):

jjw24:
@Jack251970 I haven't looked at the code but for Files Explorer feature, can we separate it into another PR? Try to keep PR focused on one feature (ideally smaller so it's easier for review but I know that's not always applicable)

Jack251970:
> @Jack251970 I haven't looked at the code but for Files Explorer feature, can we separate it into another PR? Try to keep PR focused on one feature (ideally smaller chunks of a large PR so it's easier for review but I know that's not always applicable)

Sorry we cannot do that. Files support is from Quick Switch feature. I just refactor original `FileExplorerHelper.cs` to incorporate changes from Quick Switch.

Jack251970:
Well, I can remove all changes in `FileExplorerHelper.cs` and follow on with another PR after this is merged.

Jack251970:
> @Jack251970 I haven't looked at the code but for Files Explorer feature, can we separate it into another PR? Try to keep PR focused on one feature (ideally smaller chunks of a large PR so it's easier for review but I know that's not always applicable)

Well, removed all changes for Files explorer support.

jjw24:
@Jack251970 is this ready for review?

Jack251970:
@jjw24 It is ready for review now. I have done a through test.

Jack251970:
@jjw24 I have notified the author if DirQuickJump plugin: https://github.com/idkidknow/Flow.Launcher.Plugin.DirQuickJump/issues/9

onesounds:
image
- The right-click works fine, but the left-click doesn't seem to be working.
-  Does this only work with clicks? I was hoping to operate it using the Enter key.

- It seems there should be a note somewhere indicating that this only works in File Explorer.
- When multiple tabs are open in File Explorer, it doesn't operate based on the most recently active tab. It would be great if it could function according to the most recent tab.
- Auto mode is not working at all. If improving it is difficult, it might be better to remove this feature. 

- The behavior of Alt+G is much more stable now compared to when I tested it previously.

Jack251970:
@onesounds 

> * The right-click works fine, but the left-click doesn't seem to be working.

Please double check this. It works fine on me.

https://github.com/user-attachments/assets/d27fb357-2ef5-4100-b2c9-f364c50537c5

> * Does this only work with clicks? I was hoping to operate it using the Enter key.

Using Enter key will cause some strange behaviours since Flow window & file dialog window will both get the enter system event.

> * It seems there should be a note somewhere indicating that this only works in File Explorer.

We can add support for third-party explorers (like Files) and third-party file dialogs (like WinRAR file dialog). And it is better to notify users that.

> * When multiple tabs are open in File Explorer, it doesn't operate based on the most recently active tab. It would be great if it could function according to the most recent tab.

I cannot find any api functions to do this. So I am afraid I cannot improve that.

> * Auto mode is not working at all. If improving it is difficult, it might be better to remove this feature.

I would like to mark this as experimental feature since it can work under many cases.

> * The behavior of Alt+G is much more stable now compared to when I tested it previously.

Good.

Jack251970:
To let users know supported explorers and dialogs, I think we can show them in Flow Launcher documents.

onesounds:
- I think the Enter key issue must be addressed one way or another. If not, it would be better to just keep the Alt+G feature and either remove or hide the settings shown at the bottom. Relying solely on mouse clicks feels like too much of a workaround, and ultimately I believe users will expect Enter key functionality. I think the dialog state should be treated as a kind of "mode", and when this mode is active, only the Explorer plugin should function. If version 1 has neither Enter nor mouse click support, and version 2 supports the Enter key, then version 1.5 isn’t worth offering. You should either go with version 1 or version 2 — not both. I understand that implementation is difficult, but this kind of workaround doesn’t feel good. (If you tested this with 100 people, I’m confident they’d all agree that "Enter key support is necessary.")

- If additional file manager support is possible, either merge it in with that support included or at least add some form of notice. I use the docs too, but let’s be real — most people don’t read them. As a result, you’ll see a bunch of issues being reported and you’ll end up responding with “Please check the docs.” But really, just adding a one-line infobar like “Currently only Explorer is supported” would solve it. From your perspective, it seems like you don’t want to include this message because “(Eventually) other file managers might be supported,” but from a QA standpoint, this is a simple binary: either support the feature or inform the user.

- As for auto mode — unless it works under certain conditions I’m not aware of, it didn’t work at all for me. I’ll do more testing and get back to you.

onesounds:
I have a question. For save/open dialog windows that don't work properly, is it possible to support them if the user provides the dialog window's ID or something similar? From what I understood before, it seemed like that's how the structure worked. If that's the case, would it be possible to create a dataset or allow users to add or modify the information themselves?

Jack251970:
> I have a question. For save/open dialog windows that don't work properly, is it possible to support them if the user provides the dialog window's ID or something similar? From what I understood before, it seemed like that's how the structure worked. If that's the case, would it be possible to create a dataset or allow users to add or modify the information themselves?

I think it is impossible. If you check listary interface, you can find all things should be customized.

Jack251970:
@onesounds Added information bar as follows:

image

Jack251970:
@onesounds I checked and found if you selected "Left click", enter key should also work. So I changed the localization like this:

image

onesounds:
https://github.com/user-attachments/assets/967e8964-d15b-4e1b-adb5-675a154c4e8b

- Enter and Left click doesn't work
- Issue of attempting to overwrite when the "Save As" dialog is opened in automatic mode.

jjw24:
@Jack251970 are all issues raised by onesounds resolved?

Please resolve conflict.

Jack251970:
@onesounds Could you please give me your settings file which can reproduce left click and enter issue?

`%AppData%\FlowLauncher\Settings\Settings.json` &
`%AppData%\FlowLauncher\Settings\Plugins\Flow.Launcher.Plugin.Explorer\Settings.json`

VictoriousRaptor:
Seems like not working well with multi tab explorer in windows 11. It goes for the first tab rather than the active tab.

Jack251970:
> Seems like not working well with multi tab explorer in windows 11. It goes for the first tab rather than the active tab.

I noticed this problem before but I cannot find the correct API to get the active tab path. Feel free to change the codes in `Flow.Launcher.Infrastructure.QuickSwitch.Models.WindowsExplorer` in `WindowsExplorer.cs`

VictoriousRaptor:
> > Seems like not working well with multi tab explorer in windows 11. It goes for the first tab rather than the active tab.
> 
> I noticed this problem before but I cannot find the correct API to get the active tab path. Feel free to change the codes in `Flow.Launcher.Infrastructure.QuickSwitch.Models.WindowsExplorer` in `WindowsExplorer.cs`

https://github.com/w4po/ExplorerTabUtility/blob/fad736353210d36595c6e6caa77ac7ad9129b3ee/ExplorerTabUtility/Helpers/Helper.cs#L336

https://github.com/w4po/ExplorerTabUtility/blob/fad736353210d36595c6e6caa77ac7ad9129b3ee/ExplorerTabUtility/Hooks/ExplorerWatcher.cs#L736

This might help.

Jack251970:
> > > Seems like not working well with multi tab explorer in windows 11. It goes for the first tab rather than the active tab.
> > 
> > 
> > I noticed this problem before but I cannot find the correct API to get the active tab path. Feel free to change the codes in `Flow.Launcher.Infrastructure.QuickSwitch.Models.WindowsExplorer` in `WindowsExplorer.cs`
> 
> https://github.com/w4po/ExplorerTabUtility/blob/fad736353210d36595c6e6caa77ac7ad9129b3ee/ExplorerTabUtility/Helpers/Helper.cs#L336
> 
> https://github.com/w4po/ExplorerTabUtility/blob/fad736353210d36595c6e6caa77ac7ad9129b3ee/ExplorerTabUtility/Hooks/ExplorerWatcher.cs#L736
> 
> This might help.

It helps, and now active explorer path feature is implemented

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
...
+    <system:String x:Key="showQuickSwitchWindow">Show Quick Switch Window</system:String>
+    <system:String x:Key="showQuickSwitchWindowToolTip">Show quick switch search window when file dialogs are opened to navigate its path (Only work for file open dialog).</system:String>
+    <system:String x:Key="quickSwitchWindowPosition">Quick Switch Window Position</system:String>
+    <system:String x:Key="quickSwitchWindowPositionToolTip">Select position for quick switch window</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionUnderDialog">Fixed under dialogs. Displayed after dialogs are created and until it is closed</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionFollowDefault">Floating as search window. Displayed when activated like search window</system:String>
+    <system:String x:Key="quickSwitchResultBehaviour">Quick Switch Result Navigation Behaviour</system:String>
+    <system:String x:Key="quickSwitchResultBehaviourToolTip">Behaviour to navigate file dialogs to paths of the results</system:String>
```
```suggestion
    <system:String x:Key="quickSwitchResultBehaviourToolTip">Behaviour to navigate file dialog to the result path</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
...
+    <system:String x:Key="showBadgesGlobalOnlyToolTip">Show badges for global query results only</system:String>
+    <system:String x:Key="quickSwitchHotkey">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchHotkeyToolTip">Enter shortcut to quickly navigate the path of a file dialog to the path of the current file manager.</system:String>
+    <system:String x:Key="quickSwitch">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchToolTip">Quickly navigate to the path of the current file manager when a file dialog is opened.</system:String>
+    <system:String x:Key="autoQuickSwitch">Quick Switch Automatically</system:String>
+    <system:String x:Key="autoQuickSwitchToolTip">Automatically navigate to the path of the current file manager when a file dialog is opened. (Experimental)</system:String>
+    <system:String x:Key="showQuickSwitchWindow">Show Quick Switch Window</system:String>
+    <system:String x:Key="showQuickSwitchWindowToolTip">Show quick switch search window when file dialogs are opened to navigate its path (Only work for file open dialog).</system:String>
```
jjw24: I have been able to use it for file save dialogue as well, can you also confirm?

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
...
+    <system:String x:Key="quickSwitchWindowPositionToolTip">Select position for quick switch window</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionUnderDialog">Fixed under dialogs. Displayed after dialogs are created and until it is closed</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionFollowDefault">Floating as search window. Displayed when activated like search window</system:String>
+    <system:String x:Key="quickSwitchResultBehaviour">Quick Switch Result Navigation Behaviour</system:String>
+    <system:String x:Key="quickSwitchResultBehaviourToolTip">Behaviour to navigate file dialogs to paths of the results</system:String>
+    <system:String x:Key="QuickSwitchResultBehaviourLeftClick">Left click or Enter key</system:String>
+    <system:String x:Key="QuickSwitchResultBehaviourRightClick">Right click</system:String>
+    <system:String x:Key="quickSwitchFileResultBehaviour">Quick Switch File Navigation Behaviour</system:String>
+    <system:String x:Key="quickSwitchFileResultBehaviourToolTip">Behaviour to navigate file dialogs when paths of the results are files</system:String>
```
```suggestion
    <system:String x:Key="quickSwitchFileResultBehaviourToolTip">Behaviour to navigate file dialog when the results are file paths</system:String>
```

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -285,8 +288,16 @@ public void RegisterResultsUpdatedEvent()
 
                     var token = e.Token == default ? _updateToken : e.Token;
 
-                    // make a clone to avoid possible issue that plugin will also change the list and items when updating view model
-                    var resultsCopy = DeepCloneResults(e.Results, token);
+                    IReadOnlyList<Result> resultsCopy;
+                    if (e.Results == null)
```
jjw24, line 292: Why do we need this? When will it be null?

VictoriousRaptor:
Issue:
1. Search window is fixed to the previous location even after the dialog is closed, and can't use other

Suggestion:
Show the folder we're about to switch to in the first result, like Listaryi.

Jack251970:
> Issue:
> 
> 1. Search window is fixed to the previous location even after the dialog is closed, and can't use other

Should be fixed in https://github.com/Flow-Launcher/Flow.Launcher/pull/1018/commits/8af41b6014b1723ee9922b0df077b4ccb114b9e7 and https://github.com/Flow-Launcher/Flow.Launcher/pull/1018/commits/e50fd8049d13e8ddb64ecdf7ff274651ff2c88ca. Please help me confirm that.

> Suggestion: Show the folder we're about to switch to in the first result, like Listaryi.

Do not plan to introduce new features in this PR. And we can create a new PR in future.

VictoriousRaptor:
While press the hotkey it takes about half a second to switch to the target folder. Is it possible to get faster?

VictoriousRaptor:
> > Issue:
> > 
> > 1. Search window is fixed to the previous location even after the dialog is closed, and can't use other
> 
> Should be fixed in [8af41b6](https://github.com/Flow-Launcher/Flow.Launcher/commit/8af41b6014b1723ee9922b0df077b4ccb114b9e7) and [e50fd80](https://github.com/Flow-Launcher/Flow.Launcher/commit/e50fd8049d13e8ddb64ecdf7ff274651ff2c88ca). Please help me confirm that.
> 
> > Suggestion: Show the folder we're about to switch to in the first result, like Listaryi.
> 
> Do not plan to introduce new features in this PR. And we can create a new PR in future.

Looks like it's fixed.

Jack251970:
> While press the hotkey it takes about half a second to switch to the target folder. Is it possible to get faster?

I tested in Notepad file dialog and the navigation is very fast. And I wonder if this problem could be related to your Windows system or the file dialog application.

And I think it is little hard to optimize😢

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
...
+    <system:String x:Key="autoQuickSwitch">Quick Switch Automatically</system:String>
+    <system:String x:Key="autoQuickSwitchToolTip">Automatically navigate to the path of the current file manager when a file dialog is opened. (Experimental)</system:String>
+    <system:String x:Key="showQuickSwitchWindow">Show Quick Switch Window</system:String>
+    <system:String x:Key="showQuickSwitchWindowToolTip">Show quick switch search window when file dialogs are opened to navigate its path (Only work for file open dialog).</system:String>
+    <system:String x:Key="quickSwitchWindowPosition">Quick Switch Window Position</system:String>
+    <system:String x:Key="quickSwitchWindowPositionToolTip">Select position for quick switch window</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionUnderDialog">Fixed under dialogs. Displayed after dialogs are created and until it is closed</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionFollowDefault">Floating as search window. Displayed when activated like search window</system:String>
```
```suggestion
    <system:String x:Key="QuickSwitchWindowPositionFollowDefault">Default search window position. Displayed when triggered by search window hotkey</system:String>
```

Flow.Launcher/Languages/en.xaml
jjw24:
```suggestion
    <system:String x:Key="showQuickSwitchWindowToolTip">Display quick switch search window when the open/save dialog is shown to quickly navigate to file/folder locations.</system:String>
```

jjw24: Checked save dialogue works fir word, visual studio, notepad and notepad++. Please confirm.

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
     <system:String x:Key="showBadges">Show Result Badges</system:String>
     <system:String x:Key="showBadgesToolTip">For supported plugins, badges are displayed to help distinguish them more easily.</system:String>
     <system:String x:Key="showBadgesGlobalOnly">Show Result Badges for Global Query Only</system:String>
+    <system:String x:Key="showBadgesGlobalOnlyToolTip">Show badges for global query results only</system:String>
+    <system:String x:Key="quickSwitchHotkey">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchHotkeyToolTip">Enter shortcut to quickly navigate the path of a file dialog to the path of the current file manager.</system:String>
+    <system:String x:Key="quickSwitch">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchToolTip">Quickly navigate to the path of the current file manager when a file dialog is opened.</system:String>
```
```suggestion
    <system:String x:Key="quickSwitchToolTip">When Open/Save As dialog opens, quickly navigate to the current path of the file manager.</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
...
+    <system:String x:Key="showQuickSwitchWindow">Show Quick Switch Window</system:String>
+    <system:String x:Key="showQuickSwitchWindowToolTip">Show quick switch search window when file dialogs are opened to navigate its path (Only work for file open dialog).</system:String>
+    <system:String x:Key="quickSwitchWindowPosition">Quick Switch Window Position</system:String>
+    <system:String x:Key="quickSwitchWindowPositionToolTip">Select position for quick switch window</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionUnderDialog">Fixed under dialogs. Displayed after dialogs are created and until it is closed</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionFollowDefault">Floating as search window. Displayed when activated like search window</system:String>
+    <system:String x:Key="quickSwitchResultBehaviour">Quick Switch Result Navigation Behaviour</system:String>
+    <system:String x:Key="quickSwitchResultBehaviourToolTip">Behaviour to navigate file dialog to the result path</system:String>
```
```suggestion
    <system:String x:Key="quickSwitchResultBehaviourToolTip">Behaviour to navigate Open/Save As dialog to the selected result path</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
...
+    <system:String x:Key="QuickSwitchResultBehaviourLeftClick">Left click or Enter key</system:String>
+    <system:String x:Key="QuickSwitchResultBehaviourRightClick">Right click</system:String>
+    <system:String x:Key="quickSwitchFileResultBehaviour">Quick Switch File Navigation Behaviour</system:String>
+    <system:String x:Key="quickSwitchFileResultBehaviourToolTip">Behaviour to navigate file dialog when the results are file paths</system:String>
```
```suggestion
    <system:String x:Key="quickSwitchFileResultBehaviourToolTip">Behaviour to navigate Open/Save As dialog when the result is a file path</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
...
+    <system:String x:Key="quickSwitch">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchToolTip">Quickly navigate to the path of the current file manager when a file dialog is opened.</system:String>
+    <system:String x:Key="autoQuickSwitch">Quick Switch Automatically</system:String>
+    <system:String x:Key="autoQuickSwitchToolTip">Automatically navigate to the path of the current file manager when a file dialog is opened. (Experimental)</system:String>
+    <system:String x:Key="showQuickSwitchWindow">Show Quick Switch Window</system:String>
+    <system:String x:Key="showQuickSwitchWindowToolTip">Show quick switch search window when file dialogs are opened to navigate its path (Only work for file open dialog).</system:String>
+    <system:String x:Key="quickSwitchWindowPosition">Quick Switch Window Position</system:String>
+    <system:String x:Key="quickSwitchWindowPositionToolTip">Select position for quick switch window</system:String>
+    <system:String x:Key="QuickSwitchWindowPositionUnderDialog">Fixed under dialogs. Displayed after dialogs are created and until it is closed</system:String>
```
```suggestion
    <system:String x:Key="QuickSwitchWindowPositionUnderDialog">Fixed under the Open/Save As dialog. Displayed on dialog open and stays until dialogue is closed</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
     <system:String x:Key="showBadges">Show Result Badges</system:String>
     <system:String x:Key="showBadgesToolTip">For supported plugins, badges are displayed to help distinguish them more easily.</system:String>
     <system:String x:Key="showBadgesGlobalOnly">Show Result Badges for Global Query Only</system:String>
+    <system:String x:Key="showBadgesGlobalOnlyToolTip">Show badges for global query results only</system:String>
+    <system:String x:Key="quickSwitchHotkey">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchHotkeyToolTip">Enter shortcut to quickly navigate the path of a file dialog to the path of the current file manager.</system:String>
+    <system:String x:Key="quickSwitch">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchToolTip">Quickly navigate to the path of the current file manager when a file dialog is opened.</system:String>
+    <system:String x:Key="autoQuickSwitch">Quick Switch Automatically</system:String>
+    <system:String x:Key="autoQuickSwitchToolTip">Automatically navigate to the path of the current file manager when a file dialog is opened. (Experimental)</system:String>
```
```suggestion
    <system:String x:Key="autoQuickSwitchToolTip">When Open/Save As dialog is displayed, automatically navigate to the path of the current file manager. (Experimental)</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -366,6 +366,28 @@
     <system:String x:Key="showBadges">Show Result Badges</system:String>
     <system:String x:Key="showBadgesToolTip">For supported plugins, badges are displayed to help distinguish them more easily.</system:String>
     <system:String x:Key="showBadgesGlobalOnly">Show Result Badges for Global Query Only</system:String>
+    <system:String x:Key="showBadgesGlobalOnlyToolTip">Show badges for global query results only</system:String>
+    <system:String x:Key="quickSwitchHotkey">Quick Switch</system:String>
+    <system:String x:Key="quickSwitchHotkeyToolTip">Enter shortcut to quickly navigate the path of a file dialog to the path of the current file manager.</system:String>
```
```suggestion
    <system:String x:Key="quickSwitchHotkeyToolTip">Enter shortcut to quickly navigate the Open/Save As dialogue to the path of the current file manager.</system:String>
```

Jack251970:
@coderabbitai full review

jjw24:
@Jack251970 ready to merge?

Jack251970:
> @Jack251970 ready to merge?

Yeah, all reviews are resolved.

Jack251970:
All thins are good to go

jjw24 (approved):
💯

PR #1036, opened 2022-02-25
Fix Shared Assembly Issue

taooceros:
This fixes the issue where multiple plugins using shared assembly like SQLite, where the ones loaded later will not be able to access the shared assembly.

Previous I check all the loaded assembly before allowing the pluginassemblyloader to load the assembly. However, this doesn't seem like a good practice, since this will prevent a assembly context to load the same assembly if another assembly context has already loaded it (which are isolated).

I have tested it for the winrt.dll situation as well. The default assembly loader will load the winrt.dll before we trigger a deferred load so all assembly should be good with that. However, I am curious whether we should match with Name instead of FullName.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginAssemblyLoader.cs
+++ b/Flow.Launcher.Core/Plugin/PluginAssemblyLoader.cs
@@ -11,57 +11,37 @@ namespace Flow.Launcher.Core.Plugin
 {
     internal class PluginAssemblyLoader : AssemblyLoadContext
     {
-        private readonly AssemblyDependencyResolver dependencyResolver;
+        private readonly AssemblyDependencyResolver _dependencyResolver;
 
-        private readonly AssemblyName assemblyName;
-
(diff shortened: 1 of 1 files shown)


jjw24:
let's keep this for 1.10.0 so it can go through daily usage tests longer.

taooceros:
This one will break Plugins with shared assembly (for example sqlite). So somehow user may encounter weird issue when installing multiple plugin.

jjw24 (approved):
no need underscores for privates :)

jjw24:
> This one will break Plugins with shared assembly (for example sqlite). So somehow user may encounter weird issue when installing multiple plugin.

Oh ok, in this case let's roll out for 1.9.3 then.

jjw24:
@taooceros could this potentially solve #1048 and #1046 ?

taooceros:
> @taooceros could this potentially solve #1048 and #1046 ?

I don't think so. They are not related to assembly loading.

PR #1074, opened 2022-03-10
caret position fix to fix #1025 and to fix #1070

### LINKED ISSUE #1025: Shell plugin "WIN+R" hotkey cursor default position error
use WIN+R hotkey launch input box, default curosr position is at the front. 
If you want to run the command, you need to move the cursor to after the ">" character
image

### LINKED ISSUE #1070: Custom Query Hotkey feature not usable because of incorrect caret position
**Describe the bug/issue**

Maybe I'm the only user wanting to use this feature. The Custom Query Hotkey feature is not usable because the caret position is at the beginning of the query at index 0. At the end of the query should be a space and after that space is the supposed caret position.

**To Reproduce**

1. Assign 'b' as 'Action Keyword' for the 'Browser Bookmarks' plugin.
2. Create a Custom Query Hotkey with 'b ' as query.
3. Press the hotkey, caret is at pos 0 instead of pos 2.

**Screenshots**

Screenshot (18)

Screenshot (19)

**Your System**
```
(It's Windows 11)
Microsoft Windows [Version 10.0.22000.493]
Flow Launcher version: 1.9.3
```

stax76:
fix #1025 
fix #1070

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -346,6 +346,7 @@ namespace Flow.Launcher.ViewModel
             {
                 // re-query is done in QueryText's setter method
                 QueryText = queryText;
+                QueryTextCursorMovedToEnd = false;
             }
             else if (reQuery)
             {
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -346,6 +346,8 @@ namespace Flow.Launcher.ViewModel
             {
                 // re-query is done in QueryText's setter method
                 QueryText = queryText;
+                // set to false so the subsequent set true triggers
+                // PropertyChanged and MoveQueryTextToEnd is called
                 QueryTextCursorMovedToEnd = false;
             }
             else if (reQuery)


taooceros:
Could you briefly explain why this works?

stax76:
It triggers PropertyChanged, to run code here:

https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher/MainWindow.xaml.cs#L160

taooceros:
Probably a better solution is to not use the inotifypropertychanged, but change that to an event?

stax76:
It's a lot of complex code I don't understand well, I can test it if you want to try something different.

taooceros:
> It's a lot of complex code I don't understand well, I can test it if you want to try something different.

Thank you! I feel this should work so I will merge it first.

stax76:
Comment is added.

taooceros (approved):

gissehel:
What I'll write concerns #1078, but I suppose #1025 and #1070 may be in the same case.

Using the reproductible use case described in #1078, I tried to see if this PR was closing the issue or not, and to my surprise, it didn't.

**Context**: 
- branche `dev` with just PR #1074 applied on top
- Plugin https://github.com/gissehel/BarLauncher-Volume/releases/tag/v1.0.0 installed
- I type "vol +"
- Expected result: "vol change +" with cursor at the end
- Actual result: "vol change +" with cursor at the start

**In debug mode**:
When the code:
image

is reached, the value of `QueryText` is correctly updated. The value of `QueryTextCursorMovedToEnd` was already set to `false`.
Then it's set to `false` and to `true`

Then the `PropertyChanged` code is reached:
image

It calls `MoveQueryTextToEnd`:
image

which correctly set the caret at the end... But...
The value of `QueryTextBox` is still `vol +` (and the caret is already at the end).

What was previously changed was `QueryText` not `QueryTextBox`.

And then later the `QueryTextBox` content goes to `vol change +`, but it's too late, the "caret move" code has already been executed before.

I haven't yet studied the relation between `QueryTextBox` and `QueryText`, but I suppose it's the source of the problem, and probably #1025 and #1070 have the same problem.

gissehel:
# Follow up of the analysis

- The binding of the `QueryText` from the VM to `QueryTextBox` seems OK.
image
- Various calls of `PropertyChanged` are in the excepted order (first `QueryText`, then `QueryTextCursorMovedToEnd` - twice as excepted by the code to set it back to `false` inside the `PropertyChanged` hook, with some other ones unrelated to this problems)
image
- Inside the tread UI, the code dispatched at line **541** is played before the automatic binding of `QueryText`.
image

And here stop where I'm useful : I don't know how to synchronise the automatic binding and an explicit `Dispatcher.Invoke`. I would have expected the automatic binding to use the same `Dispatcher`, and I would have expected that `Dispatcher` to guaranty the order of execution in the target thread based on the order of invocation in the source thread.

gissehel:
I was about to abandon, but I think I've understood now...

After searching if the `Dispatcher` was really running calls in the same order as the invocations, the answer is... Yes almost...

From [that page on docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcher.begininvoke?redirectedfrom=MSDN&view=windowsdesktop-6.0#System_Windows_Threading_Dispatcher_BeginInvoke_System_Windows_Threading_DispatcherPriority_System_Delegate_System_Object_) :
> If multiple [BeginInvoke](https://docs.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcher.begininvoke?view=windowsdesktop-6.0) calls are made at the same [DispatcherPriority](https://docs.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcherpriority?view=windowsdesktop-6.0), they will be executed in the order the calls were made.

Ok, so I suppose Dispatch are not done with the same priority...

From [that other page on docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcherpriority?view=windowsdesktop-6.0) (except sorted in priority order, because MS doc seems to live in a world where sorting priority alphabetically is more meaningful than by priority):

<html>
<body>


Send | 10 | The enumeration value is 10. Operations are processed before other asynchronous operations. This is the highest priority.
-- | -- | --
Normal | 9 | The enumeration value is 9. Operations are processed at normal priority. This is the typical application priority.
DataBind | 8 | The enumeration value is 8. Operations are processed at the same priority as data binding.
Render | 7 | The enumeration value is 7. Operations processed at the same priority as rendering.
Loaded | 6 | The enumeration value is 6. Operations are processed when layout and render has finished but just before items at input priority are serviced. Specifically this is used when raising the Loaded event.
Input | 5 | The enumeration value is 5. Operations are processed at the same priority as input.
Background | 4 | The enumeration value is 4. Operations are processed after all other non-idle operations are completed.
ContextIdle | 3 | The enumeration value is 3. Operations are processed after background operations have completed.
ApplicationIdle | 2 | The enumeration value is 2. Operations are processed when the application is idle.
SystemIdle | 1 | The enumeration value is 1. Operations are processed when the system is idle.
Inactive | 0 | The enumeration value is 0. Operations are not processed.
Invalid | -1 | The enumeration value is -1. This is an invalid priority.


</body>
</html>

I try to Invoke the "caret change" with priority `DataBind` (cause it look like the priority a data binding should use). It didn't work. I tried to use `Render`, it didn't work neither. I tried `Background` because "it should be low enough". It still didn't work.
With `ApplicationIdle` and `ContextIdle` it seems to works every time ! Still, because we're dealing with a synchronisation bug, I can't be sure that it's not due to my hardware or me being lucky.

So in the end, replacing:

```
        private void MoveQueryTextToEnd()
        {
            Dispatcher.Invoke(() => QueryTextBox.CaretIndex = QueryTextBox.Text.Length);
        }
```

by:

```
        private void MoveQueryTextToEnd()
        {
            Dispatcher.Invoke(() => QueryTextBox.CaretIndex = QueryTextBox.Text.Length, System.Windows.Threading.DispatcherPriority.ContextIdle);
        }
```

"appears" to solve the problem. I'm just bothered by the fact that I can't find the priority used for binding the QueryText into QueryTextBox.

stax76:
With your modification, mine is still needed here, so apparently it was independent issues.

stax76:
There is a minor cosmetic issue left, the caret is first and shortly at 0 and after that jumps to the end.

Please ignore if it cannot be reproduced or is too difficult to fix.

taooceros:
> There is a minor cosmetic issue left, the caret is first and shortly at 0 and after that jumps to the end.
> 
> Please ignore if it cannot be reproduced or is too difficult to fix.

Yes I have experienced that as well. Especially when we set a lower priority for the caret change.
I don't think it is easy to fix under the current situation.

taooceros:
close in favor of #1080

Issue #1064, opened 2022-03-04
"Explorer" plugin can't search some folders


minawaheed9697:
Hi there, 

Thanks for this great app. It really changed my life.

I have an issue with "Explorer" plugin.
When I search folders in C:\, it doesn't find them. 
In indexing options, there are always excludes in search that I can't get rid of them. 
Can those exclude be the cause of this issue?? 

Window 10 [Version 10.0.19044.1526]

<img width="368" alt="04_03_2022 07_52_10_674" src="https://user-images.githubusercontent.com/73688266/156710114-5cc0746f-b1a7-4a86-afe8-9ddf20ee91a0.png">

deefrawley:
Hi. Are the things you are searching for located in the excluded directories?
To change what is and isn't excluded click the Modify button and next to the displayed drive/folder checkbox, just to the right, you should see a greater than sign `>`. Click that and it expands the sub directories and you can change as needed (it's not easy to see unless you are looking for iti)

minawaheed9697:
I know how to chick folders to be searched, but I can't find those exclude folders. 
The folders I want to search are located in AppData folder. I think AppData is a hidden folder. 
Can Explorer plugin search in those hidden files??

minawaheed9697:
I installed "Everything" plugin and it's so good in search, but it doesn't have a great feature in "Explorer". Explorer can show the files and folders inside the searched folder, but Everything open the folder directly. 
If there is a way to make Everything make the same thing, it'll be perfect.

deefrawley:
Just to confirm:

![2022-03-04 17_49_57-Window](https://user-images.githubusercontent.com/18594055/156714131-e0b72c7c-6f7d-4db9-954a-64bdcadd7fb5.png)

I do have Windows Explorer set to show hidden files and directories and I don't know if this affects what you see in the Indexing options.
I'm not sure what the `Data` and `CSC` folders are from your original screenshot. They don't seem to be standard Windows folders but I could be wrong.

minawaheed9697:
I also made Windows Explorer to show hidden files, but nothing happened. It still can't search those folders.

deefrawley:
Yeah I can confirm - I enabled the AppData folder, did a full reindex, restarted Flow and can't search for anything in the AppData folder. @jjw24 @taooceros is this expected behaviour?

Garulf:
Odd, I don't have any issues?

[...]

PR #1076, opened 2022-03-13
Fix the use of index in path search

jjw24:
- By default path search should not use index, this solves the issue where if a path e.g. C:\blah\  is not included users should still able to pull it up with the path search.
- Allow option to enable/disable the use of index in path search via settings.

Related issue #1064

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Settings.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Settings.cs
@@ -14,7 +14,7 @@ namespace Flow.Launcher.Plugin.Explorer
         // as at v1.7.0 this is to maintain backwards compatibility, need to be removed afterwards.
         public List<AccessLink> QuickFolderAccessLinks { get; set; } = new List<AccessLink>();
 
-        public bool UseWindowsIndexForDirectorySearch { get; set; } = true;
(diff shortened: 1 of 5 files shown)


Plugins/Flow.Launcher.Plugin.Explorer/Settings.cs
```diff
@@ -14,7 +14,7 @@ public class Settings
         // as at v1.7.0 this is to maintain backwards compatibility, need to be removed afterwards.
         public List<AccessLink> QuickFolderAccessLinks { get; set; } = new List<AccessLink>();
 
-        public bool UseWindowsIndexForDirectorySearch { get; set; } = true;
+        public bool UseWindowsIndexForDirectorySearch { get; set; } = false;
```
taooceros, line 17: Have we already checked this condition?

jjw24: Sorry what do you mean check?

jjw24: Oh yeah it's used, when I wrote the plugin I had already considered this would be something we need to expose to the ui, so it's already in use and I didn't have to add any for this change.

taooceros (approved):

PR #1077, opened 2022-03-13
Fix incorrect button visibility in Explorer's expander control

jjw24:
Fix the issue where button visibility is still visible when expander is collapsed in Explorer.

Repro issue:
In Explorer plugin's settings, expand then collapse any of the three option sections, observe the button (add/edit/delete) still there even after the option has been collapsed/closed.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Views/ExplorerSettings.xaml.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Views/ExplorerSettings.xaml.cs
@@ -60,9 +60,7 @@ namespace Flow.Launcher.Plugin.Explorer.Views
 
             lbxExcludedPaths.Items.SortDescriptions.Add(new SortDescription("Path", ListSortDirection.Ascending));
 
-            btnDelete.Visibility = Visibility.Hidden;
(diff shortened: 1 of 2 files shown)


Plugins/Flow.Launcher.Plugin.Explorer/Views/ExplorerSettings.xaml.cs
```diff
@@ -123,8 +121,8 @@ private void expActionKeywords_Click(object sender, RoutedEventArgs e)
 
         private void expActionKeywords_Collapsed(object sender, RoutedEventArgs e)
         {
-            if (!expActionKeywords.IsExpanded)
```
taooceros, line 126: why is this condition removed?

taooceros, line 126: why is this condition removed?

jjw24: Yeah because this method is hooked up to the ui which triggers when collapsed, so this condition is obsolete.

PR #1080, opened 2022-03-13
Caret position fix : Include PR #1074

### LINKED ISSUE #1025: Shell plugin "WIN+R" hotkey cursor default position error

### LINKED ISSUE #1070: Custom Query Hotkey feature not usable because of incorrect caret position

### LINKED ISSUE #1078: The cursor position is at the start of the field instead of the end of the field after a call to Context.API.ChangeQuery("...") from inside a plugin
**Describe the bug/issue**
I'm developing a plugin (which target both Flow Launcher and Wox using an abstraction layer for both).

At a certain point in parsing the command, I decide to rewrite the current query from "vol +" to "vol change +".
To do that I use PluginInitContext.API.ChangeQuery("vol change +")

(Note: I happen to use the same API for both Wox and Flow Launcher)

**Expected result**:
- The new query is "vol change +" and the cursor is at the end of the query so any new char typed is inserted at the end of the string

**Actual result**:
- The new query is "vol change +" but the cursor is at the start of the query so any new char typed is inserted at the start of the string

**Notes**:

- Under Wox, the cursor is at the end of the query as excepteded.
- While the context is different, the problem may be related to #1070 and/or #1025 
- This context may help in solving the problem of the above issues
- I haven't yet looked at Flow Launcher source code about this problem

**Your System**:
```
Microsoft Windows [version 10.0.19043.1586]
Flow Launcher version: 1.9.3
```

**Flow Launcher Error Log**:
No log is generated during this sequence.

gissehel:
I include the PR #1074 because they both tried to address the same problem. I don't know if there are cases where #1074 applies or not, but it's more robust code, thus better.

Reason for this PR are explained in various comments of #1074

Should close #1025 close #1070 and close #1078

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -538,7 +538,9 @@ namespace Flow.Launcher
 
         private void MoveQueryTextToEnd()
         {
-            Dispatcher.Invoke(() => QueryTextBox.CaretIndex = QueryTextBox.Text.Length);
+            // QueryTextBox seems to be update with a DispatcherPriority as low as ContextIdle.
+            // To ensure QueryTextBox is up to date with QueryText from the View, we need to Dispatch with such a priority
+            Dispatcher.Invoke(() => QueryTextBox.CaretIndex = QueryTextBox.Text.Length, System.Windows.Threading.DispatcherPriority.ContextIdle);
         }
 
         public void InitializeColorScheme()
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -346,6 +346,7 @@ namespace Flow.Launcher.ViewModel
             {
                 // re-query is done in QueryText's setter method
                 QueryText = queryText;
+                QueryTextCursorMovedToEnd = false;
             }
             else if (reQuery)
             {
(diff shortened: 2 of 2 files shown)


taooceros:
Thank you! It's a very thoughtful debugging process, and the explanation is very valuable!
I will take a test now.

taooceros:
I encounter a slightly delay when using low priority dispatching option, but I think it's not significant in release version.

jjw24:
These changes are all tested and ready to roll?

taooceros:
> These changes are all tested and ready to roll?

I think it is ok to roll. There's nothing bugs here, but only a small tradeoff to reduce the priority. Probably in the future we can adjust the event better.

jjw24:
Make a new issue for this?

taooceros:
> Make a new issue for this?

It's not really a serious bug and I have no clues whether it can be addressed.

jjw24:
Yeah ok, I haven't started testing the new changes in Dev yet but will keep an eye out.

PR #1088, opened 2022-03-19
Add Spanish and Spanish Latin America to language selection

jjw24:
Add Spanish and Spanish Latin America to language selection


taooceros:
So previously we don't have Spanish?

jjw24:
Nop, we didn't have Spanish.

taooceros (approved):

Issue #646, opened 2021-08-11
Add open with for files and folders in Explorer plugin


jjw24:
Add 'open with' for files and folders in Explorer plugin, this will help e.g. open a folder with vscode, or a file with notepad++

pc223:
Cool idea!

Issue #526, opened 2021-06-28
Explorer plugin- Open directory location  in cmd, bash or powershell via context menu


taooceros:
### Discussed in https://github.com/Flow-Launcher/Flow.Launcher/discussions/509

<div type='discussions-op-text'>

<sup>Originally posted by **jjw24** November 18, 2020</sup>
**Is your feature request related to a problem? Please describe.**

Allow Explorer plugin to open via context menu the directory location in cmd, bash or powershell

**Describe the solution you'd like**

In the Explorer plugin, when in a directory, go to context menu and have the options to open the directory location in cmd, bash or ps.</div>

PR #1097, opened 2022-03-25
Merge Explorer & Everything Plugins

taooceros:
#1502

1. Shared Interface for Windows Index and Everything Index
2. Settings Merge (Part 1)
3. Include Everything dll

TODO:
- [x] finish Path traversal unify
- [x] implement Engine selection
- [x] #526 
- [x] #646 
- [x] implement Everything auto download
- [x] Context Menu for windows (ported from everything)
- [x] Everything Translation


taooceros:
Thinking about enumeration design. Probably it would be better to only use index search when doing recursive search for enumerating a directory, and do a direct search when only traverse the first level. @jjw24 what do you think

jjw24:
I think directory recursive search with Windows Search could be restricted by user's indexing options/excluded directories, so allowing an option to use DirectoryInfo search method is preferred per current design in this area.

What do you mean by direct search when only traverse the first leve?

taooceros:
> I think directory recursive search with Windows Search could be restricted by user's indexing options/excluded directories, so allowing an option to use DirectoryInfo search method is preferred per current design in this area.
> 
> What do you mean by direct search when only traverse the first leve?

oh I mean we may always use the DirectoryInfo Search for the non-recursive traversal.

taooceros:
@onesounds mind creating a engine selection ui for this? I am not sure whether it is a good practice to use interface for purpose selection, but probably we can adjust that easier.

- IndexSearchEngine
  - WindowsIndex
  - Everything
- ContentSearchEngine
  - WindowsIndex
  - Everything
- PathEnumerator
  - Direct
  - WindowsIndex
  - Everything

taooceros:
@Flow-Launcher/team I have almost finished most features and the UI Refactor. There're some small features in everything plugin that I haven't ported, but it will be fairly easy to do so. Anyone want to give this pr a check?

taooceros:
@jjw24 feel free to review. The remaining part is only about everything's translation, and I will try to fix it through crowdin recently.

jjw24:
Could you fix the build failure please :)

jjw24:
I haven't had a chance to look at the code yet but just thought of something- have you added handling of situation where user already has Everything service/plugin installed?

taooceros:
> I haven't had a chance to look at the code yet but just thought of something- have you added handling of situation where user already has Everything service/plugin installed?

not for everything plugin. For everything service, I just copied the code from everything plugin, so as long as it is handled there, we are fine.

taooceros:
I will fix the conflict tmr.

jjw24:
@taooceros could you please fix the CI

jjw24:
Couple of things:
1. Path search seems to be always using index search, 'Use Index Search For Path Search' unticked and selecting 'Direct Enumeration' do nothing.
2. Please complete Everything's translation, the keys should at least exist.
3. Selecting Everything should prompt user if they want to download and install Everything, when 'Yes' flow should automatically download and install it.
4. Clicking on the plugin's settings:
image

5. What is 'Launch Hidden' and I remember it's part of Everything plugin? Should move to the Everything setting tab if the case.
6. Using 'Index Search Excluded Path' is not working, could be same issue as point 1.

jjw24:
@taooceros please update description on what you have tested.

onesounds:
Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/1495

taooceros:
> Couple of things:
> 
>     1. Path search seems to be always using index search, 'Use Index Search For Path Search' unticked and selecting 'Direct Enumeration' do nothing.

It works, but the path equality check have some issue due to previous change to make subtitle only display parent path, which makes everything duplicate except the first one. fixed. 


>     2. Please complete Everything's translation, the keys should at least exist.

Fixed


>     3. Selecting Everything should prompt user if they want to download and install Everything, when 'Yes' flow should automatically download and install it.
>

I think we can do it when user actually search (when the result display everything not working).

>     4. Clicking on the plugin's settings:
>        image
> 

I cannot reproduce

>     5. What is 'Launch Hidden' and I remember it's part of Everything plugin? Should move to the Everything setting tab if the case.

This is a weird setting and I think we have made the decision to remove it.

>     6. Using 'Index Search Excluded Path' is not working, could be same issue as point 1.

Will handle that soon.

jjw24:
Translations are done?

Plugins/Flow.Launcher.Plugin.Explorer/Languages/en.xaml
```diff
@@ -17,13 +17,21 @@
...
+    <system:String x:Key="plugin_explorer_everything_sort_option">Sort Option:</system:String>
+    <system:String x:Key="plugin_explorer_everything_installed_path">Everything Path:</system:String>
+    <system:String x:Key="plugin_explorer_launch_hidden">Launch Hidden</system:String>
+    <system:String x:Key="plugin_explorer_editor_path">Editor Path</system:String>
+    <system:String x:Key="plugin_explorer_shell_path">Shell Path</system:String>
     <system:String x:Key="plugin_explorer_indexsearchexcludedpaths_header">Index Search Excluded Paths</system:String>
+    <system:String x:Key="plugin_explorer_use_location_as_working_dir">Use search result's location as executable working directory</system:String>
     <system:String x:Key="plugin_explorer_usewindowsindexfordirectorysearch">Use Index Search For Path Search</system:String>
     <system:String x:Key="plugin_explorer_usewindowsindexfordirectorysearch_tooltip">Turning this on will return indexed directories/files faster, but if a directory/file is not indexed it will not show up. If a directory/file has been added to Index Search Excluded Path then it will still show up even if this option is on</system:String>
```
jjw24, line 36: This should change to `If directory is indexed, turning this on will return indexed directories/files faster, but if a directory/file is not indexed it will still use path enumeration so it can still be shown.`
Because UseWindowsIndexForDirectorySearch call checks and returns the outcome of whether the path is indexed.

jjw24: hmm on second thought this option seems confusing now because feels like a 'maybe use index search' but also depends also on Paths Enumeration Engine.

Plugins/Flow.Launcher.Plugin.Explorer/Settings.cs
```diff
@@ -38,8 +56,94 @@ public class Settings
...
+        };
+
+        public enum PathEnumerationEngineOption
+        {
+            [Description("plugin_explorer_engine_windows_index")]
+            WindowsIndex,
+            [Description("plugin_explorer_engine_everything")]
+            Everything,
+            [Description("plugin_explorer_path_enumeration_engine_none")]
+            Direct
```
jjw24, line 108: Maybe call this DirectEnumeration so it's consistent with UI

Plugins/Flow.Launcher.Plugin.Explorer/Settings.cs
```diff
@@ -1,22 +1,40 @@
...
     public class Settings
     {
         public int MaxResult { get; set; } = 100;
 
-        public List<AccessLink> QuickAccessLinks { get; set; } = new List<AccessLink>();
+        public ObservableCollection<AccessLink> QuickAccessLinks { get; set; } = new();
 
         // as at v1.7.0 this is to maintain backwards compatibility, need to be removed afterwards.
-        public List<AccessLink> QuickFolderAccessLinks { get; set; } = new List<AccessLink>();
+        public ObservableCollection<AccessLink> QuickFolderAccessLinks { get; set; } = new();
```
jjw24, line 23: Remove this and the other call, no longer needed.

jjw24 (commented):
>3. Selecting Everything should prompt user if they want to download and install Everything, when 'Yes' flow should automatically download and install it.

This is still not fixed. Using Everything engine for index search/content search without everything installed just errors out:
image

jjw24:
>5. What is 'Launch Hidden' and I remember it's part of Everything plugin? Should move to the Everything setting tab if the case.
This is a weird setting and I think we have made the decision to remove it.

Please remove it.

jjw24:
Shouldn't these settings be moved to Everything settings tab?

image

taooceros:
will check that later today

jjw24:
Could this be Everything is installed but not running?

taooceros:
> Could this be Everything is installed but not running?

I do haven't uninstalled everything, but I believe both should throw IPCError? Not so sure why it may be different. Will check tongiht

jjw24:
Changing action keywords do not take effect on first try. Repro:
Remove the settings.json file in user data directory. Start up flow, go to Explorer and change Path action keyword. Trigger the keyword and it does not happen. Only after changing the action keyword again then it will trigger.

jjw24:
Please address:
1. When Everything is selected as Index Search Engine, press and hold any key will cause below error:
image

2. Bring this message back when Windows Index is disabled:
image

3. Hide the search window when the prompt to install Everything is shown:
image

4. Please add icon for this result:
image

jjw24:
> Please address:
> 
> 1. When Everything is selected as Index Search Engine, press and hold any key will cause below error:
>    image
> 2. Bring this message back when Windows Index is disabled:
>    image
> 3. Hide the search window when the prompt to install Everything is shown:
>    image
> 4. Please add icon for this result:
>    image

@taooceros I am fixing 2,3,4

taooceros:
the concurrency bug is ....very weird but interesting (I know what's causing it, and try to find a good solution).

jjw24 (approved):

PR #1104, opened 2022-03-29
Add Stale issue workflow

Garulf:
With this workflow Issues and PRs that have not received any updates will be marked with a `stale` label. 

Any Issues will be closed after 5 days if they are not updated after being marked. 
PRs will never be closed.
Any PR that has a milestone will be exempt from the stale label.

### DIFF
--- a/.github/ISSUE_TEMPLATE/workflows/stale.yml
+++ b/.github/ISSUE_TEMPLATE/workflows/stale.yml
@@ -1,0 +1,25 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v4
+        with:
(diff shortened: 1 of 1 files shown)


deefrawley:
For issues that are waiting for a release to fix, where the release is taking longer than thirty days, is the idea to add an update comment or remove the label to keep it open and waiting?

Garulf:
That's how it would work I guess.

jjw24 (changes requested):
We would have issues that are waiting to be fixed, yet to be fixed or  lower priority, would it be better to implement closing based on a label i.e. 'waiting for response'?

Also don't close PRs regardless of how inactive they are, because the team might not get around to them or they need further implementation.

Garulf:
> We would have issues that are need investigation, waiting to be fixed, yet to be fixed or acknowledged but lower priority, would it be better to implement closing based on a label i.e. 'waiting for response'? So we would mark stale the ones that have been waiting from user response > 5 days and close after 15 days. All other issues still open will mean they are of the four mentioned status.
> 
That could work, but I feel if we haven't responded in the current cut off period (30 + 5 days) it's probably not being responded to because it is just a stale issue. 

Also marking an issue as stale isn't necessarily a bad thing. It could help draw attention to issues that may have fallen through the cracks. 
> 
> Also don't close PRs regardless of how inactive they are, because the team might not get around to them or they need further implementation.

Current settings only mark PRs with a "Stale" label unless they're included in a milestone. Never are they closed. Hopefully this can give a kick start to some PRs that have been sitting and help indicate what's been sitting around for an extended period and may need some attention.

jjw24:
Yeah ok. I personally keep open only issues that need work on and strive to maintain a low count, but we can give this a try too.

jjw24 (approved):

PR #1108, opened 2022-03-31
Respect Decimal Separator for query not just result

taooceros:


jjw24:
Is the intention to not calculate 2.10/2 when the decimal calculator is `,` ? If so, it's not working.

jjw24:
Just pushed an update, should work as intended now.

jjw24 (approved):

taooceros:
nice, haven't noticed I reverse the logic.

PR #1121, opened 2022-04-10
Pass query without trimming whitespace

Garulf:
Trailing whitespace can be very helpful for argument detection.

This PR removes any whitespace trimming from Query before being passed to the Plugin.


taooceros:
I would say that we still need to trim start since it would be weird to use space as one of the action keyword?

Garulf:
> I would say that we still need to trim start since it would be weird to use space as one of the action keyword?

No worries the keyword is captured by `string possibleActionKeyword = terms[0];` and is unaffected by this change.

jjw24 (approved):

PR #1123, opened 2022-04-12
Implement Async Result Execution

taooceros:
Result Execution can take quite long (Especially for JsonRPCPlugin), which cannot handle threadpool easily (requiring subprocess to prevent long execution time).

### DIFF
--- a/Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
+++ b/Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
@@ -115,7 +115,7 @@ namespace Flow.Launcher.Core.Plugin
 
             foreach (var result in queryResponseModel.Result)
             {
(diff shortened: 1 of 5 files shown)


taooceros:
@Flow-Launcher/team Is ActionAsync named better or AsyncAction named better🤔?

jjw24:
AsyncAction

jjw24 (approved):

taooceros:
@jjw24 I couldn't reproduce the issue you said earlier, and I take a look on the code, only possible issue is that `RequestAsync` will throw an error when response is empty, but I remember I have suppressed it earlier. Could you share a more detailed reproduction process?

jjw24:
Able to replicate with BrowserHistory plugin, as soon as you open one of its results.

PR #1205, opened 2022-05-15
Fix shell cmd command with quote and space

### LINKED ISSUE #1085: The Shell plugin cannot parse file directory that contains blanks
**Describe the bug/issue**
The plugin `Shell` cannot handle the file directory that contains blanks.

**To Reproduce**
1. type `> mklink "root\..path1" "root\..path2"`, where path1 (path2) is the file directory containing blanks.
2. 
3. ...

**Screenshots**

**Your System**
```
Windows build number: 10.0.19044.1586
Flow Launcher version: 1.93
```

taooceros:
fix #1085


Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -195,16 +193,18 @@ private ProcessStartInfo PrepareProcessStartInfo(string command, bool runAsAdmin
...
             {
                 case Shell.Cmd:
                     {
                         info.FileName = "cmd.exe";
-                        info.ArgumentList.Add(_settings.LeaveShellOpen ? "/k" : "/c");
-                        info.ArgumentList.Add(command);
+                        info.Arguments = $"{(_settings.LeaveShellOpen ? "/k" : "/c")} {command}";
+                        
+                        // ArgumentList may break original shell command separation with quote.
+                        // info.ArgumentList.Add(_settings.LeaveShellOpen ? "/k" : "/c");
```
jjw24, line 206: These commented out code is not removed?

jjw24:
Any chance we can add tests to this?

taooceros:
> Any chance we can add tests to this?

I have limited idea about how to write test for this.

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
taooceros: yes we can remove it. Probably it would be better to change all argumentlist to plain arguments, because we want to allow user to have the most control over the argument they are writing.

jjw24:
This is still not working, you can test by running `mkdir c:\new test` with Shell plugin, should create a folder called 'new test' in c:\

taooceros:
I haven't given another round of test, but I think you need to type `mkdir "C:\new test"` instead?

jjw24:
yes you are right 🤦 too early in the day for me right now lol

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
jjw24: yes agree.

jjw24 (approved):
👍

PR #1275, opened 2022-07-16
Insert active explorer path 

stefnotch:
I refactored the code from https://github.com/Flow-Launcher/Flow.Launcher/pull/904

Please test this code, as I haven't really done it.


Flow.Launcher/Helper/FileExplorerHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <summary>
+        /// Gets the path of the file explorer that is currently in the foreground
+        /// </summary>
+        public static string GetActiveExplorerPath()
+        {
+            var explorerWindow = GetActiveExplorer();
+            string locationUrl = explorerWindow.LocationURL;
+            if (!string.IsNullOrEmpty(locationUrl))
+            {
+                return new Uri(locationUrl).LocalPath;
```
stefnotch, line 20: This ended up being the most reliable way of getting the explorer path. I think edge cases are folders like the desktop, recycle bin, documents folder, etc.

Flow.Launcher/Helper/FileExplorerHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+            var shellWindows = new SHDocVw.ShellWindows();
+
+            // loop through all windows
+            foreach (var window in shellWindows)
+            {
+                if (window is SHDocVw.InternetExplorer explorerWindow && new IntPtr(explorerWindow.HWND) == handle)
+                {
+                    // we have found the desired window, now let's make sure that it is indeed a file explorer
+                    // we don't want the Internet Explorer or the classic control panel
+                    if (explorerWindow.Document is not Shell32.IShellFolderViewDual2)
```
stefnotch, line 46: This check could be removed, since the check below catches everything.

(The classical control panel is a pretty nasty edge case, try it out)

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -749,6 +751,11 @@ public void Show()
                 ((MainWindow)Application.Current.MainWindow).WindowAnimator();
             
             MainWindowOpacity = 1;
+            if (_explorerPath != null)
```
stefnotch, line 754: This check has been simplified, so that's nice

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -734,6 +733,9 @@ public void ToggleFlowLauncher()
 
         public void Show()
         {
+            string _explorerPath = FileExplorerHelper.GetActiveExplorerPath();
```
stefnotch, line 736: I think this might be a useful function to expose to all plugins.

Flow.Launcher/Helper/FileExplorerHelper.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Text;
+using System.Runtime.InteropServices;
+using System.IO;
+
+namespace Flow.Launcher.Helper
+{
+    public class FileExplorerHelper
```
stefnotch, line 8: The other function that *might* be interesting would be getting all open file explorer paths (sorted by depth/zindex). (See also https://github.com/jjw24/Wox.Plugin.Runner/pull/27 )
Thoughts?

Flow.Launcher/Flow.Launcher.csproj
```diff
@@ -77,6 +77,24 @@
...
+    </COMReference>
+    <COMReference Include="Shell32">
+      <VersionMinor>0</VersionMinor>
+      <VersionMajor>1</VersionMajor>
+      <Guid>50a7e9b0-70ef-11d1-b75a-00a0c90564fe</Guid>
+      <Lcid>0</Lcid>
+      <WrapperTool>tlbimp</WrapperTool>
+      <Isolated>false</Isolated>
+      <EmbedInteropTypes>true</EmbedInteropTypes>
+    </COMReference>
```
taooceros, line 97: What do you think about the COMREference? Is it better to use dynamic typing as what you did in the runner plugin? Somehow the dotnet cli doesn't support this feature, and the previous workaround is to copy the reference dll and directly reference that dll.

stefnotch: I think if the COMReference works without any issues, then that's pretty nice. But I honestly have no clue.

taooceros: I would say since you have shared a way to prevent usage of comreference, let's use that. Previously I workaround this by copying out the dll and reference that directly.

taooceros: Though I personally think it may be better to have strongly type?

taooceros: https://github.com/dotnet/msbuild/issues/3986

stefnotch: Apparently one can define the COM import interfaces themselves. Like this app does https://github.com/emoacht/Monitorian/blob/c16345d1aa679ca463a14b4c1ef0917a524fe31a/Source/Monitorian.Core/Models/Monitor/LightSensor.cs , or like how the .NET framework does it https://referencesource.microsoft.com/#PresentationFramework/src/Framework/MS/Internal/AppModel/ShellProvider.cs,3ec518c827af4fd7

taooceros: yeah for sure, but I think that's pretty messy. We do have some example about that.

Flow.Launcher/ViewModel/MainViewModel.cs
taooceros: It seems like this is a utility function? Maybe we can put the helper in Flow.Launcher.Plugin?

Flow.Launcher/Helper/FileExplorerHelper.cs
taooceros: I agree, maybe we should return the paths sorted by depth/zindex (since we will need to get all of them). The overhead can be ignored.

Flow.Launcher/ViewModel/MainViewModel.cs
stefnotch: @taooceros Yes, that sounds useful

Garulf:
Are we able to fix the CI build issue?

taooceros:
> Are we able to fix the CI build issue?

It's because of merge conflict. Just resolve it.

Garulf:
> > Are we able to fix the CI build issue?
> 
> 
> 
> It's because of merge conflict. Just resolve it.

Ah! Didn't see that. That's what I get for glancing.

taooceros:
@stefnotch Do you have time to remove the comreference and move everything to either dynamic or a template class? Then we shall be able to move to the final stage.

stefnotch:
@taooceros Okay, I've changed it to use dynamic. However, I didn't test the code at all, since I don't have the .NET 5 framework on this machine. (Only got .NET 6)

taooceros:
> @taooceros Okay, I've changed it to use dynamic. However, I didn't test the code at all, since I don't have the .NET 5 framework on this machine. (Only got .NET 6)

I think flow has been upgraded to .Net 6. The original pr is pretty old, so do you mind taking sometimes to merge the dev branch? Thanks

stefnotch:
@taooceros Lovely, that's super nice to hear. Okay, I merged it in and tested it out. Seems to work rather nicely, the only thing I'd tweak is: The entire text should be selected when it automatically inserts the file explorer path. (aka it should be easy to get rid of the inserted file explorer path)

taooceros:
> @taooceros Lovely, that's super nice to hear. Okay, I merged it in and tested it out. Seems to work rather nicely, the only thing I'd tweak is: 

It does take almost a whole year for .Net 6 branch to be merged🤣. That's pretty nice. Thanks

> The entire text should be selected when it automatically inserts the file explorer path. (aka it should be easy to get rid of the inserted file explorer path)

I am not sure whether it is easy to be done. I will reach back soon about that once I finish the current stuff.

taooceros:
One thing I am a bit worried by this feature is that it only applies to explorer plugin (although it is a built in plugin). Therefore, do you think it would be better if we put the code into plugin project instead of the main project?

Regarding text selection, there will be much more that needs to be done. Maybe we should expose an api for plugins to select a part of the query base on index.
Another idea is to use a special actionkeyword to query only in the specific path. Thereby we don't actually need to handle the long preceding path (and also don't need to select it). However, that will lose us the possibility to traverse around it (but search around it may be enough?). If this way makes more sense, maybe we need a way to quickly switch to normal path search under the same path.

I think we can put the helper class in Flow.Launcher.Plugin project, since it maybe quite useful but somewhat non-trivial to implement.

jjw24:
>Therefore, do you think it would be better if we put the code into plugin project instead of the main project?

Thought the idea was to expose this functionality via the API? If not then yes move it to plugin project so all plugins can take advantage of.

taooceros:
> > Therefore, do you think it would be better if we put the code into plugin project instead of the main project?
> 
> Thought the idea was to expose this functionality via the API? If not then yes move it to plugin project so all plugins can take advantage of.
> 
> EDIT: functionality as in grabbing the active window, is this what you are asking?

I don't think we need to expose as api. It's nothing related to flow.

jjw24:
But isn't this functionality potentially can be used by several plugins e.g. Explorer & Runner plugins? Even though placing in the plugin project can achieve the same, but wouldn't putting it in API will mean it will be documented and potentially allow it to be used (assuming) by other non-C# language plugins?

taooceros:
> But isn't this functionality potentially can be used by several plugins e.g. Explorer & Runner plugins? Even though placing in the plugin project can achieve the same, but wouldn't putting it in API will mean it will be documented and potentially allow it to be used (assuming) by other non-C# language plugins?

hmmm make sense? Though I think there's a really long path to allow non-c# plugin to use feature like this. I just feel like the plugin api may be better to include api that is configurable in flow or depends on flow. This looks like simply a utility function.

jjw24:
Yeah, at this stage, up to you guys, ultimately placing it in the plugin project is a good start.

taooceros:
Yeah let's put it into the plugin project first. We can add it to the api later.

taooceros:
I have a concern at the current stage. We actually allow customize the action keyword for exploring a particular directory. That means it is possible that the inserted directory won't do anything (which is partially why I think it would be better to move the code to explorer plugin, where it has control over its own action keyword).

taooceros:
@stefnotch maybe let's delay moving to Flow.Launcher.Plugin after we finish #1097. It contains too much change to explorer plugin that I didn't want to handle the merge conflict.

taooceros:
for example this page
image

stefnotch:
@taooceros Sounds fine, we can always do more in future pull requests.

taooceros:
@stefnotch Let's make this a built-in command😉(Sorry for such a long delay).

stefnotch:
@taooceros That sounds like a great idea! Would you like to take care of it, since it might take me a while to get around to doing so?

taooceros:
> @taooceros That sounds like a great idea! Would you like to take care of it, since it might take me a while to get around to doing so?

yeah sure!

taooceros:
Hmmm I guess we need to adjust the logic a bit, since the shortcut will be activated after flow showing up, so the foreground window will become flow. What do you think about getting the explorer window with the highest z-index? @stefnotch

taooceros:
@stefnotch mind give it a test?

Assign something like this or use the custom shortcut image

taooceros:
Great! I will merge it!

taooceros (approved):

Issue #637, opened 2021-08-08
Theme query and result item fonts working?


jjw24:
The fonts option for query and result item list do not apply immediately, requires a restart or a switch to and back from another theme 

### Discussed in https://github.com/Flow-Launcher/Flow.Launcher/discussions/625

<div type='discussions-op-text'>

<sup>Originally posted by **jjw24** August  2, 2021</sup>
is it me or does changing this options not working fully? the preview doesnt seem to refresh and seems like font changes works after a restart, but not all font options are working.

![image](https://user-images.githubusercontent.com/26427004/127785534-31ca0515-3809-47fa-8883-e1b2bc351296.png)
</div>

jjw24:
hey @taooceros , is this resolved by https://github.com/Flow-Launcher/Flow.Launcher/pull/722?

jjw24:
I can give it a crack later, if not then let's just close it

PR #1283, opened 2022-07-19
Force reload theme resources when changing fonts etc in settings 

nachmore:
Fixes #625 & #637 

This will mean that changing some settings will be a bit slower than before, especially when enabling blur which will reload resources twice but this is negligible and only hits in settings.

### DIFF
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -85,7 +85,7 @@ namespace Flow.Launcher.Core.Resource
 
                 Settings.Theme = theme;
 
-                // #625, #637 - reload all resources even if the theme itself hasn't changed in order to pickup changes
+                // reload all resources even if the theme itself hasn't changed in order to pickup changes
                 // to things like fonts
                 UpdateResourceDictionary(GetResourceDictionary());
 
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -85,10 +85,13 @@ namespace Flow.Launcher.Core.Resource
 
                 Settings.Theme = theme;
 
+                // #625, #637 - reload all resources even if the theme itself hasn't changed in order to pickup changes
+                // to things like fonts
+                UpdateResourceDictionary(GetResourceDictionary());
+
                 //always allow re-loading default theme, in case of failure of switching to a new theme from default theme
                 if (_oldTheme != theme || theme == defaultTheme)
                 {
-                    UpdateResourceDictionary(GetResourceDictionary());
                     _oldTheme = Path.GetFileNameWithoutExtension(_oldResource.Source.AbsolutePath);
                 }
 


jjw24 (changes requested):
Hey there, thank you for taking the time to look into this issue and making a fix. The current code is not written with enough flexibility to allow directly calling the required changes to fonts without going through the whole ChangeTheme call, which is confusing and probably why you needed to move UpdateResourceDictionary method out of the if condition.

I think it's better design is to move the query font update code in this UpdateResourceDictionary method into its own method so you can call it directly from the view model:
- https://github.com/Flow-Launcher/Flow.Launcher/blob/8942e5567a888068aaf7eabde29aa909387155ca/Flow.Launcher.Core/Resource/Theme.cs#L149-L160
- https://github.com/Flow-Launcher/Flow.Launcher/blob/8942e5567a888068aaf7eabde29aa909387155ca/Flow.Launcher.Core/Resource/Theme.cs#L168-L172

Also another method for updating the result fonts:
https://github.com/Flow-Launcher/Flow.Launcher/blob/8942e5567a888068aaf7eabde29aa909387155ca/Flow.Launcher.Core/Resource/Theme.cs#L175-L191

So then you can call them directly from the view model here:
- https://github.com/Flow-Launcher/Flow.Launcher/blob/8d71b433def76162f6d6aeb95c27864441f52b28/Flow.Launcher/ViewModel/SettingWindowViewModel.cs#L426
- https://github.com/Flow-Launcher/Flow.Launcher/blob/8d71b433def76162f6d6aeb95c27864441f52b28/Flow.Launcher/ViewModel/SettingWindowViewModel.cs#L450
- https://github.com/Flow-Launcher/Flow.Launcher/blob/8d71b433def76162f6d6aeb95c27864441f52b28/Flow.Launcher/ViewModel/SettingWindowViewModel.cs#L471
- https://github.com/Flow-Launcher/Flow.Launcher/blob/8d71b433def76162f6d6aeb95c27864441f52b28/Flow.Launcher/ViewModel/SettingWindowViewModel.cs#L495

Let me know if you can make these changes, happy to help/make them also :)

nachmore:
Thanks for the detailed response! I looked around for ways to do what you're suggesting, and avoided it because of how invasive it was :)

I'll see if I can spend some time this weekend refactoring some of the code as you suggest and then resubmit!

jjw24:
Fantastic. You can just push the new changes to this PR :)

nachmore:
I spent a bit of time looking at this today and it's not clear that a refactor actually achieves much. [`GetResourceDictionary`](https://github.com/Flow-Launcher/Flow.Launcher/blob/8942e5567a888068aaf7eabde29aa909387155ca/Flow.Launcher.Core/Resource/Theme.cs#L145) is mostly about font updating (other than a small bit updating `CaretBrush` and `Foreground`). 

Additionally, to change this you either need to edit the live resource dictionaries or reapply them anyway using [`UpdateResourceDictionary`](https://github.com/Flow-Launcher/Flow.Launcher/blob/8942e5567a888068aaf7eabde29aa909387155ca/Flow.Launcher.Core/Resource/Theme.cs#L125) which does a very coarse `Remove` and `Add`.

Reloading the full monte on font change is relatively cheap, since it only happens in Settings, and is safe as it just reloads everything.

A different option could be to call the resource dictionary updates directly (i.e. `UpdateResourceDictionary(GetResourceDictionary());` instead of going through [`ChangeTheme`](https://github.com/Flow-Launcher/Flow.Launcher/blob/8942e5567a888068aaf7eabde29aa909387155ca/Flow.Launcher.Core/Resource/Theme.cs#L76)) but that's not a particularly large change and you lose the safety of having a single (mostly) path to call when theme resources change.

taooceros:
why is this one closed? I think you are right that it is quite hard to refactor?

nachmore:
I was futzing with branches and github decided to be helpful and close the PR. I still want to fix this one... 

If there is consensus on the original fix (or a different one) then I will bring it back as a new PR.

taooceros:
> I was futzing with branches and github decided to be helpful and close the PR. I still want to fix this one...
> 
> If there is consensus on the original fix (or a different one) then I will bring it back as a new PR.

Sure, the original code sounds good for me though (as a fix). I think what @jjw24 would like to do is to rewrite the whole theme loading system, which may take a bit more time.
Currently we can merge the small fix first. I don't think it will affect performance in a large scale.

jjw24:
Yeah, I will do a small rewrite and push up soon.

nachmore:
> Yeah, I will do a small rewrite and push up soon.

Let me know if you want me to re-PR the original fix, or a different fix if the overall refactor is going to take a while 😀

jjw24:
nah should be fine, let's keep in this PR, not a big refactor and you were right, it's not going to bring much efficiency but at least the code will be more flexible and readable. Will find some time today or tomorrow.

Flow.Launcher.Core/Resource/Theme.cs
```diff
@@ -85,10 +85,13 @@ public bool ChangeTheme(string theme)
 
                 Settings.Theme = theme;
 
+                // #625, #637 - reload all resources even if the theme itself hasn't changed in order to pickup changes
```
jjw24, line 88:
```suggestion
                // reload all resources even if the theme itself hasn't changed in order to pickup changes
```

jjw24:
I don't think I will have time to do a proper rewrite, will merge in and tackle it later when needed.

jjw24 (approved):

PR #1289, opened 2022-07-23
[Dev] add check query length when calling backspace command action

jjw24:
When browsing directories and using ctrl backspace to go back up one directory, flow will throw error if ctrl backspace is pressed when there is no query text.

Currently error exists in Dev branch.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -517,6 +517,7 @@ namespace Flow.Launcher
                     if (specialKeyState.CtrlPressed)
                     {
                         if (_viewModel.SelectedIsFromQueryResults()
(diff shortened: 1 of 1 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -517,6 +517,7 @@ private void OnKeyDown(object sender, KeyEventArgs e)
                     if (specialKeyState.CtrlPressed)
                     {
                         if (_viewModel.SelectedIsFromQueryResults()
+                            && QueryTextBox.Text.Length > 0
```
taooceros, line 520: I think we can use a $?$ operator lower to resolve this as well. But this shall work as well.

taooceros (approved):

PR #1296, opened 2022-07-25
Fix theme fonts not changing separately for query box and results

### LINKED ISSUE #1284: Font change for query and result
**Describe the bug/issue**
Changing the font for query will also change the font for result and vice versa

**To Reproduce**
1. Go to settings
2. Go to themes
3. Change the font for query and the font for result also changes to the same
4. Or change the font for result and font for query will change to the same too.

**Screenshots**
image

Found the issue with dev build, have not tried release build.

nachmore:
Fixes #1284: `CollectionViewSource` will synchronize the current item across all users of the same view source unless they explicitly set `IsSynchronizedWithCurrentItem`.


jjw24 (approved):

jjw24:
Thanks for the fix :)

PR #1311, opened 2022-08-06
Proposal for Http Plugin.

st0le:
Adding a new type of JsonRPCPlugin that retrieves results from an API endpoint that can talk the same protocol. 

Not completely done. Want to configure the Http methods, etc. but wondering if this will be accepted.

### DIFF
--- a/Flow.Launcher.Core/Plugin/HttpPlugin.cs
+++ b/Flow.Launcher.Core/Plugin/HttpPlugin.cs
@@ -1,0 +1,37 @@
+﻿using System.IO;
+using System.Net.Http;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Flow.Launcher.Plugin;
+
+namespace Flow.Launcher.Core.Plugin
+{
+    internal class HttpPlugin : JsonRPCPlugin
+    {
+        private readonly HttpClient client;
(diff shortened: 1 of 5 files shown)


taooceros:
Would you please share a sample plugin? I couldn't see how exactly it will be.

Flow.Launcher.Core/Plugin/HttpPlugin.cs
```diff
@@ -0,0 +1,37 @@
...
+        public HttpPlugin(string url)
+        {
+            this.client = new HttpClient();
+            this.url = url;
+        }
+
+        protected override string Request(JsonRPCRequestModel rpcRequest, CancellationToken token = default)
+        {
+            HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Post, url);
+            requestMessage.Content = new StringContent(rpcRequest.ToString(), Encoding.UTF8);
```
taooceros, line 26: I have trouble mostly with this part. It only directly use the rpcRequest as payload, which is mostly flow's style. Therefore, the use case can only be a customized http server.
I would like to see a http plugin not only provide the endpoints address, but also translate flow's query to the specific payload (a parser), and translate the response back to flow's standard response.

PR #1314, opened 2022-08-08
Fix (some) build warnings

nachmore:
Primarily fixing a bunch of:

1. CS1572
2.    CS1573
3.    CS0168
4.    CS8073
5.    CA2200
1.    VSTHRD110
1.    VSTHRD200
1.    VSTHRD105
1.    SYSLIB0013
1.    CS8524

Not a comprehensive fix, but removes a large number of warnings.


nachmore:
Before: [400 warnings](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/44257746/messages)
After: [161 warnings](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/44400262/messages)

.editorconfig
```diff
@@ -0,0 +1,146 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
```
taooceros, line 1: Is this file added by intention?

Flow.Launcher.Core/Configuration/Portable.cs
```diff
@@ -127,7 +127,7 @@ public void CreateUninstallerEntry()
 
             using (var portabilityUpdater = NewUpdateManager())
             {
-                portabilityUpdater.CreateUninstallerRegistryEntry();
+                _ = portabilityUpdater.CreateUninstallerRegistryEntry();
```
taooceros, line 130: I think we should make CreateUninstallerEntry either `async Task` or `async void`. I am curious what this method is doing.

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -98,7 +98,7 @@ public static IEnumerable<PluginPair> DotNetPlugins(List<PluginMetadata> source)
                                    + (erroredPlugins.Count > 1 ? "plugins have " : "plugin has ")
                                    + "errored and cannot be loaded:";
 
-                Task.Run(() =>
+                _ = Task.Run(() =>
```
taooceros, line 101: I don't think it makes sense to use a Task.Run here. Would you please replace it with the notification class?

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -103,7 +103,7 @@ public async Task SetHotkey(HotkeyModel keyModel, bool triggerValidate = true)
 
         public void SetHotkey(string keyStr, bool triggerValidate = true)
         {
-            SetHotkey(new HotkeyModel(keyStr), triggerValidate);
+            _ = SetHotkey(new HotkeyModel(keyStr), triggerValidate);
```
taooceros, line 106: I don't think it makes sense to have different return value for these two methods. Could you make it also either async void or async Task.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -93,7 +93,8 @@ private bool ActionKeywordMatch(Query query, Settings.ActionKeyword allowedActio
                 Settings.ActionKeyword.IndexSearchActionKeyword => Settings.IndexSearchKeywordEnabled &&
                                                                    keyword == Settings.IndexSearchActionKeyword,
                 Settings.ActionKeyword.QuickAccessActionKeyword => Settings.QuickAccessKeywordEnabled &&
-                                                                        keyword == Settings.QuickAccessActionKeyword
+                                                                        keyword == Settings.QuickAccessActionKeyword,
+                _ => false
```
taooceros, line 97: I think we should throw an exception since it doesn't make sense here.

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -378,7 +378,7 @@ public List<Result> LoadContextMenus(Result selectedResult)
                     Title = context.API.GetTranslation("flowlauncher_plugin_cmd_run_as_different_user"),
                     Action = c =>
                     {
-                        Task.Run(() => Execute(ShellCommand.RunAsDifferentUser, PrepareProcessStartInfo(selectedResult.Title)));
+                        _ = Task.Run(() => Execute(ShellCommand.RunAsDifferentUser, PrepareProcessStartInfo(selectedResult.Title)));
```
taooceros, line 381: Let's move this to `AsyncAction` since we have added that.

Plugins/Flow.Launcher.Plugin.WebSearch/SuggestionSources/Baidu.cs
```diff
@@ -23,7 +23,7 @@ public override async Task<List<string>> SuggestionsAsync(string query, Cancella
             try
             {
                 const string api = "http://suggestion.baidu.com/su?json=1&wd=";
-                result = await Http.GetAsync(api + Uri.EscapeUriString(query), token).ConfigureAwait(false);
+                result = await Http.GetAsync(api + Uri.EscapeDataString(query), token).ConfigureAwait(false);
```
taooceros, line 26: what's the difference between these two?

taooceros (changes requested):
It's a non-trivial refactor. Thank you so much for taking time to do it!

.editorconfig
nachmore: yes - it adds:

1. forces tab sizes for the project to 4, since VS sometimes can't auto detect it (and it seems to be the standard across the project)
2. it adds explicit ignores for `dotnet_diagnostic.CA1416.severity = silent`

VS adds all the other boilerplate junk :P I can try to cut out all the other defaults if that's preferable. Note that it shouldn't change anything significant other than the editing experience.

Plugins/Flow.Launcher.Plugin.WebSearch/SuggestionSources/Baidu.cs
nachmore: `EscapeUriString` is deprecated: https://docs.microsoft.com/en-us/dotnet/api/system.uri.escapeuristring?view=net-6.0

Flow.Launcher.Core/Configuration/Portable.cs
nachmore: `async void` needs to be [avoided](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void), in fact I want to clean up some of those in the next step if I get a chance.

For this one, it just so happens that Squirrel's function is async, but we're not bothering to call it asynchronously. In fact, nothing in [`IPortable.cs`](https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher.Core/Configuration/IPortable.cs) seems to be `async`.

I'm happy to create a follow up task for this, but changing the signature of the interface seems like a more invasive task than just removing the warning from the existing flow.

.editorconfig
taooceros: Yeah it makes sense to have a uniform editor preference for everybody working on Flow. @jjw24 what do you think about this?

Flow.Launcher.Core/Plugin/PluginsLoader.cs
nachmore: Completely agree - but [`Notification`](https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher/Notification.cs) is not accessible outside of `Flow.Launcher`. I can open a follow up task to refactor `Notification` into `Flow.Launcher.Infrastructure` as part of a different PR, if that works?

Flow.Launcher.Core/Configuration/Portable.cs
taooceros: if we didn't await the task, any exception thrown won't be caught. I wonder whether that may create some issue (previously I have been deeply hurt by this suppression).

.editorconfig
jjw24: Yes absolutely, I think this is a great idea and something we should have implemented a long time ago. Thank you @nachmore

taooceros:
@nachmore One of the build test has failed. Would you please take a look on what's causing that?

nachmore:
I'm looking at this now - it's due to a change in behaviour handling `null` in `json`. So much for "just replace X with Y" 😜

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
```diff
@@ -69,7 +69,7 @@ public List<Result> LoadContextMenus(Result selectedResult)
         private static readonly JsonSerializerOptions options = new()
         {
             PropertyNameCaseInsensitive = true,
-            IgnoreNullValues = true,
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
```
taooceros, line 72: Basically because of this?

nachmore: exactly, but the change (`IgnoreNullValues` is deprecated) should have been the same. Looking into the alternative, but might just give up and stay with the deprecated property.

taooceros: What's the difference between this two property? If they are supposed to replace each other, there shouldn't be an issue with that.

taooceros: Let's remove this property. I see no reason we want to ignore null/default property.

taooceros: If I remember correctly, the original usage is to ignore null value when deserialize. Do we have a similar api for that?

taooceros: I take a look on the test case, and it is testing whether different naming case will create same result, which should be responsible for `PropertyNameCaseInsensitive` but not `IgnoreNullValue`. I am quite curious what's causing that.

taooceros: My test result is surprising. It will not be `null` only when we have the `IgnoreNullValue` property set to true. I have limited clues about it.

nachmore: The new property only kicks in during serialization, which means setting an explicit null, whereas the previous value ignores the null and seems to auto-set the default value (which in this case is an empty List for [`Result`](https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher.Core/Plugin/JsonPRCModel.cs#L44)).

This is fixable, though not sure it gets us anything.

taooceros: The weird thing is only the second result is null. I don't understand what's causing that.

nachmore:
Looking at build mismatch...

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
taooceros: I try to take a debug but there seems like it cannot deserialize the second result at all.

nachmore: I think it's because the second result matches case exactly. It just seems... broken, which is why I propose not fixing this for now.

nachmore:
I'm going to pause the warnings elimination work as this PR is large enough already. So when folks have time, can you please review these changes, I'll address comments and then continue in a separate PR. Thanks!

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
taooceros: That seems reasonable. Though I am still stuck at the reason of the null here....Maybe it's somewhat because of how I handle the stream or what? Let's investigate that in the future.

Flow.Launcher.Core/Plugin/PluginsLoader.cs
taooceros: I agree. Let's move to `Flow.Launcher.Infrastructure` to make it usable across the solution.

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
```diff
@@ -82,12 +88,15 @@ public List<Result> LoadContextMenus(Result selectedResult)
...
 
-        private Dictionary<string, FrameworkElement> _settingControls = new();
+        private readonly Dictionary<string, FrameworkElement> _settingControls = new();
 
         private async Task<List<Result>> DeserializedResultAsync(Stream output)
         {
             if (output == Stream.Null) return null;
 
+            var temp = System.Text.Encoding.UTF8.GetString(((MemoryStream)output).ToArray());
+            System.Diagnostics.Debug.WriteLine(temp);
```
taooceros, line 98: Is this accidentally added for debugging?

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -46,8 +45,10 @@ public static IEnumerable<PluginPair> DotNetPlugins(List<PluginMetadata> source)
                         var assembly = assemblyLoader.LoadAssemblyAndDependencies();
                         var type = assemblyLoader.FromAssemblyGetTypeOfInterface(assembly,
                             typeof(IAsyncPlugin));
-
+#pragma warning disable IDE0019 // "Use Pattern Matching" - disabled because the pattern is useful for RELEASE
                         var plugin = Activator.CreateInstance(type) as IAsyncPlugin;
```
taooceros, line 49: why not use pattern matching?
like
```
if (Activator.CreateInstance(type) is not IAsyncPlugin plugin)
{
   ....
}
```

Flow.Launcher.Plugin/AllowedLanguage.cs
```diff
@@ -1,5 +1,6 @@
 ﻿namespace Flow.Launcher.Plugin
 {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
taooceros, line 3: why disable the warning here? If you would like to not include this in this pr, just leave it here and address it in the future.

Flow.Launcher/Helper/SingleInstance.cs
```diff
@@ -400,6 +400,7 @@ private class IPCRemoteService : MarshalByRefObject
             /// to ensure that lease never expires.
             /// </summary>
             /// <returns>Always null.</returns>
+            [Obsolete] // overrides an obsolete method, so also needs to be obsolete
```
taooceros, line 403: I am unsure why we mark this as Obsolete. I think this is not likely to be a public access api.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -428,10 +427,10 @@ public Application(AppxPackageHelper.IAppxManifestApplication manifestApp, UWP p
 
                 manifestApp.GetAppUserModelId(out string tmpUserModelId);
                 manifestApp.GetAppUserModelId(out string tmpUniqueIdentifier);
-                manifestApp.GetStringValue("DisplayName", out string tmpDisplayName);
-                manifestApp.GetStringValue("Description", out string tmpDescription);
-                manifestApp.GetStringValue("BackgroundColor", out string tmpBackgroundColor);
-                manifestApp.GetStringValue("EntryPoint", out string tmpEntryPoint);
+                _ = manifestApp.GetStringValue("DisplayName", out string tmpDisplayName);
```
taooceros, line 430: do we want defensive programing here to check the return value?

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -540,7 +539,7 @@ internal string LogoUriFromManifest(AppxPackageHelper.IAppxManifestApplication a
                 if (logoKeyFromVersion.ContainsKey(Package.Version))
                 {
                     var key = logoKeyFromVersion[Package.Version];
-                    app.GetStringValue(key, out string logoUri);
+                    _ = app.GetStringValue(key, out string logoUri);
```
taooceros, line 542: same as above

Flow.Launcher.Core/Plugin/PluginsLoader.cs
nachmore: It conflicts with the `RELEASE` and `DEBUG` code. Let me see if I can clean it up a bit.

Flow.Launcher.Plugin/AllowedLanguage.cs
nachmore: Seemed like a fairly self documenting class - but you're right. I'll fix it.

Flow.Launcher.Core/Plugin/JsonRPCPlugin.cs
nachmore: good catch, removed other debugging, but missed this one. Thanks!

Flow.Launcher/Helper/SingleInstance.cs
nachmore: It's actually completely obsolete (i.e. will throw if called): https://docs.microsoft.com/en-us/dotnet/api/system.marshalbyrefobject.initializelifetimeservice?view=net-6.0

I've removed the whole thing as I don't see any references to it anywhere.

taooceros: I am curious why do we have it before. The file is used for maintaining only one instance.

nachmore: Here's a good reference to it: https://github.com/Flow-Launcher/Flow.Launcher/commit/be33ac3c4f29e4d71b6684fd76e087dd076c50a4

It used to have multiple usages within https://github.com/Flow-Launcher/Flow.Launcher/blob/be33ac3c4f29e4d71b6684fd76e087dd076c50a4/Wox/Helper/SingleInstance.cs (see below), but these were removed in https://github.com/Flow-Launcher/Flow.Launcher/commit/d329b73748861a0cad7844897285c83bbe36f9d3#diff-5e380b0d6c890e4b05cd9c291ce4618cdfdf21b1cb5c5458ddc7bd25f97d7706 which only did a partial cleanup (should have removed this class at that time).

Original usage: https://github.com/Flow-Launcher/Flow.Launcher/blob/be33ac3c4f29e4d71b6684fd76e087dd076c50a4/Wox/Helper/SingleInstance.cs#L388-L410

taooceros: That's super nice. Thank you!

nachmore:
Down to [118](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/44423736/messages).

Could I please get a review of these changes? I created follow up tasks for some of the fundamental comments that were uncovered during this PR (but not directly related to these changes).

taooceros:
> Down to [118](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/44423736/messages).
> 
> Could I please get a review of these changes? I created follow up tasks for some of the fundamental comments that were uncovered during this PR (but not directly related to these changes).

Sure, just give me sometime.

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -101,9 +101,9 @@ public async Task SetHotkey(HotkeyModel keyModel, bool triggerValidate = true)
             }
         }
 
-        public void SetHotkey(string keyStr, bool triggerValidate = true)
+        public async Task SetHotkeyAsync(string keyStr, bool triggerValidate = true)
         {
-            SetHotkey(new HotkeyModel(keyStr), triggerValidate);
+            await SetHotkeyAsync(new HotkeyModel(keyStr), triggerValidate);
```
taooceros, line 106: I feel like we can simply return `SetHotkeyAsync` but this is also fine

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -157,9 +157,8 @@ public static Application[] All()
 #if !DEBUG
                     catch (Exception e)
                     {
-                        ProgramLogger.LogException($"|UWP|All|{p.InstalledLocation}|An unexpected error occured and "
-                                                        + $"unable to convert Package to UWP for {p.Id.FullName}", e);
-                        return new Application[] { };
+                        ProgramLogger.LogException($"|UWP|All|{p.InstalledLocation}|An unexpected error occured and " + $"unable to convert Package to UWP for {p.Id.FullName}", e);
```
taooceros, line 160: Let's combine both sides of the `+`.

taooceros (changes requested):
Thank you so much for making such a huge change. There's a small pieces that worth taking a look before merging. After addressing them, I think we shall be able to merge.

Let's wrap this with `Task.Run` and make the method `Async`
https://github.com/Flow-Launcher/Flow.Launcher/blob/30669e8c257ce02f08cf763dcd1efaf99950ed81/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs#L421

nachmore:
> Sure, just give me sometime.

Sorry if it sounded like I was rushing you - I am blown away by the speed of CRs in this project, it's incredible! Thank you for the super quick support...

taooceros:
> > Sure, just give me sometime.
> 
> 
> 
> Sorry if it sounded like I was rushing you - I am blown away by the speed of CRs in this project, it's incredible! Thank you for the super quick support...
> 
> 
> 
> 

Na you are good. It's always easier for me to review than you writing the code, and it will be super nice to hear feedback quickly when contributing.
Let's finish the remaining trivial stuff and merge this one.

Flow.Launcher/HotkeyControl.xaml.cs
nachmore: We can't return it directly as we want to await the result (since there is a 500ms delay in the other `SetHotkeyAsync`)

taooceros: Oh i mean return the task from the other function.

nachmore: I tried:

```csharp
        public Task SetHotkeyAsync(string keyStr, bool triggerValidate = true)
        {
            return SetHotkeyAsync(new HotkeyModel(keyStr), triggerValidate);
        }
```

but then you can't await this function (since it's not `async`). If we want this to work as a pass through then it looks like:

```csharp
        public async Task<Task> SetHotkeyAsync(string keyStr, bool triggerValidate = true)
        {
            return SetHotkeyAsync(new HotkeyModel(keyStr), triggerValidate);
        }
```

but that's awkward (and would need some kind of double await if you wanted to await this) :)

taooceros: Lol i think you understand async/await a little bit wrong. Any Task can be awaited, regardless whether the method has async signature. It is true that if we remove the async here we are not able to await function here, but since this method is only an overload of the other, we don't actually need to await anything here. 

However, in both case, the one level Task will be able to be awaited.

nachmore: heh. That was my expectation going in, but when I tried the first one it errored out - though I think I had my warnings dialed up too high which caused a warning to show as an error. I've tried again and it works cleanly *shrug*.

taooceros: lol that's common. Just note that what affects whether a function can be awaited is the return type (`Task, Task<T>, ValueTask<T>`), but not the async modifier. The async modifier only allows to use `await` inside the function.

taooceros:
> Thank you so much for making such a huge change. There's a small pieces that worth taking a look before merging. After addressing them, I think we shall be able to merge.
> 
> Let's wrap this with `Task.Run` and make the method `Async`
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/30669e8c257ce02f08cf763dcd1efaf99950ed81/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs#L421

I think this may not be useful and we are ready to merge.

PR #1351, opened 2022-08-30
Allow plugins to use web URL's for icon paths

Garulf:
This PR allows Flow Launcher to accept remote URL's provided in the `IcoPath` field. 

This greatly speeds up results.


Flow.Launcher.Infrastructure/Image/ImageLoader.cs
```diff
@@ -113,7 +114,25 @@ private static ImageResult LoadInternal(string path, bool loadFullImage = false)
                 {
                     return new ImageResult(ImageCache[path], ImageType.Cache);
                 }
-
+                if (path.StartsWith("http://") || path.StartsWith("https://"))
+                {
+                    // Download image from url
+                    using (WebClient client = new WebClient())
```
taooceros, line 120: Could you use HttpClient instead and try to make this method async? (It's not a trivial refactor, but I think it will be a good place to understand how async work).

Flow.Launcher.Infrastructure/Image/ImageLoader.cs
```diff
@@ -113,7 +114,25 @@ private static ImageResult LoadInternal(string path, bool loadFullImage = false)
...
+                    {
+                        using (MemoryStream stream = new MemoryStream(client.DownloadData(path)))
```
taooceros, line 122: Also please use the latest using declaration syntax to reduce brace🤣.

Flow.Launcher.Infrastructure/Image/ImageLoader.cs
```diff
@@ -113,7 +114,25 @@ private static ImageResult LoadInternal(string path, bool loadFullImage = false)
...
+                        {
+                            var image = new BitmapImage();
+                            image.BeginInit();
+                            image.CacheOption = BitmapCacheOption.OnLoad;
+                            image.StreamSource = stream;
+                            image.EndInit();
+                            image.Freeze();
```
taooceros, line 129: I am curious what if we remove this line?

Flow.Launcher.Plugin/Result.cs
```diff
@@ -131,7 +131,7 @@ public string PluginDirectory
             set
             {
                 _pluginDirectory = value;
-                if (!string.IsNullOrEmpty(IcoPath) && !Path.IsPathRooted(IcoPath))
+                    if (!string.IsNullOrEmpty(IcoPath) && !Path.IsPathRooted(IcoPath) && !IcoPath.StartsWith("http://") && !IcoPath.StartsWith("https://"))
                 {
                     IcoPath = Path.Combine(value, IcoPath);
```
taooceros, line 136: A remark to fix this one

taooceros (changes requested):
Nice change, but requires a bit (maybe quite a bit🤣) modification to make it perfect.

Flow.Launcher.Infrastructure/Image/ImageLoader.cs
Garulf: It crashes the entire app XD

taooceros:
https://stackoverflow.com/questions/36610442/bitmapimage-how-do-i-wait-for-bitmapimage-to-be-initialized-before-returning-im

I will adjust the code a bit before marking ready.

jjw24:
@Garulf or @taooceros please resolve conflict

jjw24:
Please update PR description on what tests have been done.

Flow.Launcher.Infrastructure/Http/Http.cs
```diff
@@ -117,7 +117,7 @@ public static async Task DownloadAsync([NotNull] string url, [NotNull] string fi
         public static Task<string> GetAsync([NotNull] string url, CancellationToken token = default)
         {
             Log.Debug($"|Http.Get|Url <{url}>");
-            return GetAsync(new Uri(url.Replace("#", "%23")), token);
+            return GetAsync(new Uri(url), token);
```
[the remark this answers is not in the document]
jjw24: what do you mean? the old code is also creating a url too no?

Flow.Launcher.Plugin/Result.cs
```diff
@@ -62,7 +62,13 @@ public string IcoPath
             {
                 if (!string.IsNullOrEmpty(PluginDirectory) && !Path.IsPathRooted(value))
                 {
-                    _icoPath = Path.Combine(value, IcoPath);
+                    string absPath = Path.Combine(value, IcoPath);
+                    // Only convert relative paths if its a valid path
+                    if (File.Exists(absPath))
```
jjw24, line 67: Is there another way to check this path without using File.Exists? Doesn't seem necessary to use it here.

Garulf: This way allows us to improve how Flow auto appends the plugins path to icon paths that are relative and still accept URI such as http, https, ftp, etc. instead of checking every URI under the sun to see if it's valid let's just detect if that file even exists and leave it alone if it doesn't.

jjw24: I mean this is an unnecessary call because we already checking that the path exists or not at https://github.com/Flow-Launcher/Flow.Launcher/blob/e5948a70c41362bde75d82c91c9a422b352bbb35/Flow.Launcher.Infrastructure/Image/ImageLoader.cs#L183-L202

(GetThumbnailResult is called by LoadInternalAsync)

We just need to determine that it is not a URI.

Flow.Launcher.Plugin/Result.cs
```diff
@@ -140,9 +146,15 @@ public string PluginDirectory
             set
             {
                 _pluginDirectory = value;
-                if (!string.IsNullOrEmpty(IcoPath) && !Path.IsPathRooted(IcoPath))
+                    if (!string.IsNullOrEmpty(IcoPath) && !Path.IsPathRooted(IcoPath))
                 {
-                    IcoPath = Path.Combine(value, IcoPath);
```
jjw24, line 145: @taooceros @Garulf do you guys know why we are setting IcoPath while setting PluginDirectory?

Flow.Launcher.Plugin/Result.cs
Garulf: We *can* detect URI and still keep original function that adds the plugins root directory to relative paths. This way just seemed to be the simplest way to accomplish both.

Should I refactor this to detect URI's and relative paths?

Flow.Launcher.Plugin/Result.cs
Garulf: The original code was set up to add the plugin's root path to any relative paths. This would mangle any URI strings set in IcoPath because it would be treated as a relative path.

For example:
`http://test.org/file.zip` would become `c:\<pluginDir>\http://test.org/file.zip`

Flow.Launcher.Plugin/Result.cs
Garulf: We're not really concerned with if the path exists or not. What we're really concerned with is the path really a relative path or not.

The built-in method for detecting relative paths isnt very smart.

jjw24: @Garulf @taooceros I reworked IcoPath, PluginDirectory and ImageLoader https://github.com/Flow-Launcher/Flow.Launcher/pull/1351/commits/83ec8099a78cf64c446b53cefa52ce6e611c7dbf, let me know if ok.

taooceros:
I remove the await Task.Run outside. However, it sometimes makes the debugger stop when fetching image fails (which throws an error that is caught outside). I wonder whether we should make the method more robust by using return code🤔

jjw24 (approved):

PR #1352, opened 2022-08-30
Rounded icon mask field

Garulf:
Allows plugins to apply a rounded mask to a result icon.

Plugin need only provide the field `RoundedIcon` with a Boolean. Defaults to `False`

Some API's provide a square icon but intend to be displayed with rounded edges. This will help display images correctly.

Before:
image

After:
image

### DIFF
--- a/Flow.Launcher/ViewModel/ResultViewModel.cs
+++ b/Flow.Launcher/ViewModel/ResultViewModel.cs
@@ -84,6 +84,19 @@ namespace Flow.Launcher.ViewModel
             }
         }
 
+        public double IconRadius
+        {
+            get
+            {
+                if (Result.RoundedIcon)
+                {
+                    return IconXY / 2;
+                }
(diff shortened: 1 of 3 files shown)


onesounds:
I think it's ok.
There are many services that mask with circular icons, so it will be used often in the long run.

Flow.Launcher/ViewModel/ResultViewModel.cs
```diff
@@ -84,6 +84,19 @@ public Visibility ShowIcon
             }
         }
 
+        public double IconRadius
+        {
+            get
+            {
+                if (Result.RoundedIcon)
```
taooceros, line 91: I feel like this is a bit hacky🤣
~~And also you may need a OnPropertyChanged?~~ (maybe not because fody handles that)

onesounds (approved):

PR #1353, opened 2022-09-01
Implement JSONRPC V2 

taooceros:
## Tasks:
- [x] Implement JSONRPC Plugin Base (for both v1 and v2), and forward current parser to v2
- [x] Implement a full jsonrpc protocol based on https://github.com/microsoft/vs-streamjsonrpc
- [x] Pass the `Query` object to plugin as argument
- [x] Implement a python server
- [x] Implement python client to send request
- [x] Initialization Method
- [x] Wait System.Text.Json Support in https://github.com/microsoft/vs-streamjsonrpc (They have merged the PR, so we only need to wait for a release)
- [x] ~~Finish Python Script example with `ijson`~~
- [x] ~~Implement API call (we need a new identifier)~~

## Motivation:
Currently our jsonrpc plugins are very limited. They are not able to visit any api that might receive a result in Flow. They are not able to cache in memory as flow run every single query in different instance.

## This PR:

- One Instance of plugin per lifetime
- Communication based on StdIn/StdOut
- Standard JsonRPC protocol with the help of https://github.com/microsoft/vs-streamjsonrpc
- Most API in Flow are callable by JsonRPC Plugins (except some that are not json serializable)
- Performance is much better (almost feel like a native plugin)
- User can reload plugin by `F5` so the development is no harder than before

## Sample Python Code
[jsonrpcv2.zip](https://github.com/Flow-Launcher/Flow.Launcher/files/11858870/jsonrpcv2.zip)

## Limitation

- The JsonRPC client in python is hand-written, without any unit test currently. I couldn't find a suitable library for the job we want. Most JsonRPC library in Python are for HTTP Protocol.
- Exception Handling can be confusing sometimes. For example, `TypeError` is used to detect whether flow sends the correct argument to the server, but this errors can also be raised inside the method.

## Future Work
- [ ] Support Node.js/Deno directly (We can use the library for vscode lsp easily to create the sample)


jjw24:
Please list in the description the tests you have done, i.e. python, js and ts plugins

jjw24:
Do we need to keep V1 around?

Garulf:
> Do we need to keep V1 around?

Ideally no. 

However what would we do with "legacy" plugins?

jjw24:
Just mark those methods obsolete, point to new ones to maintain backwards compatibility. Set a comment in the obsolete attribute to remove after a certain period.

jjw24:
>User can reload plugin by F5 so the development is no harder than before

Sorry what does this mean?

Garulf:
> >User can reload plugin by F5 so the development is no harder than before
> 
> 
> 
> Sorry what does this mean?

Because v2 plugins are always running if you change any code a refresh is needed to exit the plugin and rerun.

Flow.Launcher.Core/Plugin/JsonRPCModelContext.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Flow.Launcher.Core.Plugin
+{
+    // TODO: After Upgrading to .Net 7, adding Source Generating Context for IAsyncEnumerable JsonRPCMessage
```
jjw24, line 5: We are already on .Net 7

taooceros: oops that's an old one, we should remove it

Flow.Launcher.Core/Flow.Launcher.Core.csproj
```diff
@@ -57,6 +57,7 @@
     <PackageReference Include="FSharp.Core" Version="7.0.300" />
     <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="2.3.2" />
     <PackageReference Include="squirrel.windows" Version="1.5.2" NoWarn="NU1701" />
+    <PackageReference Include="StreamJsonRpc" Version="2.15.29" />
```
VictoriousRaptor, line 60:
```suggestion
    <PackageReference Include="StreamJsonRpc" Version="2.15.29" />
```

[v2.16.36](https://github.com/microsoft/vs-streamjsonrpc/releases/tag/v2.16.36)

taooceros: yes, let's change it

Flow.Launcher.Core/Plugin/JsonRPCModelContext.cs
jjw24, line 5:
```suggestion
```

jjw24: remove obsolete comment

jjw24:
what do we need to do to get current plugins to start using v2 ?

Flow.Launcher.Core/Plugin/JsonRPCV2Models/JsonRPCPublicAPI.cs
```diff
@@ -0,0 +1,326 @@
...
+            _api.RestartApp();
+        }
+
+        /// <summary>
+        /// Run a shell command
+        /// </summary>
+        /// <param name="cmd">The command or program to run</param>
+        /// <param name="filename">the shell type to run, e.g. powershell.exe</param>
+        /// <exception cref="FileNotFoundException">Thrown when unable to find the file specified in the command </exception>
+        /// <exception cref="Win32Exception">Thrown when error occurs during the execution of the command </exception>
```
jjw24, line 50: @taooceros why are we duplicating a lot of the same method summaries here from IPublicAPI? Can we just keep them in the interface only.

Flow.Launcher.Core/Plugin/ExecutablePlugin.cs
```diff
@@ -27,14 +28,14 @@ public ExecutablePlugin(string filename)
...
-            _startInfo.ArgumentList[0] = request.ToString();
+            _startInfo.ArgumentList[0] = JsonSerializer.Serialize(request, RequestSerializeOption);
             return ExecuteAsync(_startInfo, token);
         }
 
         protected override string Request(JsonRPCRequestModel rpcRequest, CancellationToken token = default)
         {
             // since this is not static, request strings will build up in ArgumentList if index is not specified
-            _startInfo.ArgumentList[0] = rpcRequest.ToString();
+            _startInfo.ArgumentList[0] = JsonSerializer.Serialize(rpcRequest, RequestSerializeOption);
```
jjw24, line 38: Why we changing this?

jjw24 (changes requested):

Flow.Launcher.Core/Plugin/JsonRPCV2Models/JsonRPCPublicAPI.cs
taooceros: The problem is we cannot use `IPublicAPI` directly because there's an event handler there.

taooceros: Oh you are saying the summary I guess I just copy them from there... I can remove them

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -21,16 +21,19 @@ public static List<PluginPair> Plugins(List<PluginMetadata> metadatas, PluginsSe
...
             var pythonEnv = new PythonEnvironment(metadatas, settings);
+            var pythonV2Env = new PythonV2Environment(metadatas, settings);
             var tsEnv = new TypeScriptEnvironment(metadatas, settings);
             var jsEnv = new JavaScriptEnvironment(metadatas, settings);
             var pythonPlugins = pythonEnv.Setup();
+            var pythonV2Plugins = pythonV2Env.Setup();
             var tsPlugins = tsEnv.Setup();
             var jsPlugins = jsEnv.Setup();
             
             var executablePlugins = ExecutablePlugins(metadatas);
```
jjw24, line 32: Should we not also instantiate the ExecutablePluginV2?

jjw24 (changes requested):

Flow.Launcher.Core/Plugin/PluginsLoader.cs
taooceros: Yeah true

Flow.Launcher.Core/Plugin/PortableSettings.cs
```diff
@@ -0,0 +1,415 @@
...
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Documents;
+using Flow.Launcher.Infrastructure.Storage;
+using Flow.Launcher.Plugin;
+
+namespace Flow.Launcher.Core.Plugin
+{
+    public class PortableSettings
```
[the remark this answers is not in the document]
taooceros: Ideally because I don't feel like this may be only used by jsonrpc plugin. Portable means it is applicable to any kinds of plugin but sure if you want to change the name. I have a hard time to figure out names.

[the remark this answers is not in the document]
jjw24: Hmm portable to me means this settings can be moved anyway, but in actuality this class is used for  all non-native plugins right? So all JsonRPC plugins will use it, I don't suppose native plugins will want to set up their settings window this way? Or am I missing a part of the equation?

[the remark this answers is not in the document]
taooceros: Non-native plugin doesn't have to be JsonRPC. Though that's probably the case unless we have some special plugins.

jjw24 (approved):
Good work thank you for implementing V2.

Before we can publicly announce this new feature, we will need to:
1. Add documentation
2. Add the example plugin you attached in this PR as HelloWorld plugin
3. Add a way to automatically transition/convert existing plugins to V2
So people can follow creating V2 easily.

PR #1360, opened 2022-09-02
Preview Panel

onesounds:
## Features
- Preview toggle (F1 Key) for image/full file path display.
- Option for always on preview. (if turn off, not showing preview when open query bar) Default is off. (for Simple design impression)
- Adjust Themes for Preview
- Video/Image/Music file show large preview, others will showing small icon (need plugin update to support)

## Sub Features
- Recover Darker theme
- Adjust Loading Image

## Testcase
- [x] F1 should not work when empty query state. 
- [x] Always Preview option (general in setting) should work when start flow.
- [x] Be visible in all width sizes / Themes
- [x] Works well any last query mode. 
- [x] Glyph Icon should display with glyph mode
- [x] video/image file should large (max width), others should display small size (96x96)
- [x] Memory 
- [x] Loading Icon should be gray round box, Files without icons are the same as before.
- [x] UWP apps icon should be 96x96 / transparency
- [x] media (jpg, gif, png, avi, mp4, wmv...) preview should be display large image
- [x] The media file should display normal regardless of the case of the extension. (e.g .gif/GIF)

## Future Plans
- [ ] Add High res Icons
- [ ] Add File Type information
- [ ] Customizable Preview hotkey
- [ ] Customizable File Types
- [ ] Click to Copy location or file name
- [ ] LadUpdated/DateAdded/LastUsed info
- [ ] Small Buttons
- [ ] Custom Preview Area for plugins


onesounds:
yep still in progress. it need video thumb  / customcontrol panel for plugins

onesounds:
This PR is dirty because I tried many things. (webview, videothumb, opengraph...)
In now. it's not good to merge/reviewing. I'll organize(with rollback) code or I'll make new PR when I get free time.

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -266,6 +271,20 @@ internal static Result CreateFileResult(string filePath, Query query, int score
...
+            { 
+                return false; 
+            }
+            else
+            {
+                return MediaExtensions.Contains(extension.ToLowerInvariant());
+            }
+        }
+
+        public static readonly string[] MediaExtensions = { ".jpg", ".png", ".avi", ".mkv", ".bmp", ".gif", ".wmv", ".mp3", ".flac", ".mp4" };
```
jjw24, line 287: @taooceros @VictoriousRaptor I think this is probably better moved to the plugin project, the idea is that plugin does not need to determine and provide whether the result path is a media or not and let Flow determine it, i.e. flow controls whether to display large preview image.

Let me know what you think, we can update it next round.

jjw24 (approved):

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
VictoriousRaptor: agree. but plugin should still have control to  whether to use large preview. programs should use small preview. just move this function.

PR #1365, opened 2022-09-06
Display Drive Space

onesounds:
**Features**
- When Type c:\ or d:\, Display Free space size and Full Size with progresbar
- Progressbar will be red when over 90% usage.

**Test Case**
- The progress bar should be displayed as usage percent.
- It should work properly for each drive.
- If this command is not used, the progress bar must not be displayed.


onesounds:
> Possible for other plug-ins to use this design?

I guess so. if 'progress'(nullable int, 0-100) in result. automatically showing.

Garulf:
Amazing!

Flow.Launcher/Converters/ProgressForegroundConverter.cs
```diff
@@ -0,0 +1,32 @@
...
+            double progress = (double)value;
+            string foreground = "#26a0da";
+
+            if (progress >= 90)
+            {
+                foreground = "#da2626";
+            }
+
+            return foreground;
+        }
```
jjw24, line 25: @Garulf @onesounds @taooceros Do we want to enable plugins to use progress bar? If so we will need to make this bar colorization customizable by plugin, otherwise any progress above 90 will be red.

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -86,19 +123,32 @@ internal static Result CreateOpenCurrentFolderResult(string path, bool windowsIn
...
 
-
             var subtitleFolderName = folderName;
+            var subtitle = $"Use > to search within {subtitleFolderName}, " +
+                           $"* to search for file extensions or >* to combine both searches.";
+
+
+
+
+
```
jjw24, line 138:
```suggestion
```

jjw24: formatting

Flow.Launcher/Converters/ProgressForegroundConverter.cs
Garulf: Probably best we don't hard code red here and allow the color to be adjusted in some manner. 

Best case scenario it's themed and easily overridden by the plugin.

jjw24:
@Garulf @onesounds @taooceros just changed the code a bit to allow plugins to use progress bar by providing the Result.ProgressBar value and optionally setting the color themselves. Let me know if you guys ok with the change, if all good we can merge in.

onesounds:
> @Garulf @onesounds @taooceros just changed the code a bit to allow plugins to use progress bar by providing the Result.ProgressBar value and optionally setting the color themselves. Let me know if you guys ok with the change, if all good we can merge in.

cool. I tested it and there's no problem. I think this structure is better. let's merge.

Garulf:
I have not tested but I have looked at the code. I see no issues

Garulf (approved):

PR #1367, opened 2022-09-07
Search window's position settings

onesounds:
image

- Remember Last Position (Default)
- Mouse Screen Center (same old thing)
- Mouse Screen Center Top (New)
- Mouse Screen Left Top
- Mouse Screen Right Top
- Position Reset on Tray (Turning off any one while using multiple monitors causes a location problem.)

**Features**
- Change RememberLastPosition(bool) to LauncherPosition(string)
- It is difficult to respond to screen rotation, so I think the function of always positioning it at the top will help.

**Test Case**
- When you move the window outside the screen and press Reset Position, the window must be visible to the user.
- Each function shall be aligned in a certain position regardless of the monitor resolution.

**What to do later**
- The ability to check the number of monitors and then open them at a specific monitor location. (ex> Monior 2 Screen Center)


onesounds:
This pr should merge after clock PR merged.

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
```diff
@@ -199,7 +199,7 @@ public bool HideNotifyIcon
         }
         public bool LeaveCmdOpen { get; set; }
         public bool HideWhenDeactive { get; set; } = true;
-        public bool RememberLastLaunchLocation { get; set; }
+        public string LauncherPosition { get; set; } = "RememberLastLaunchLocation";
```
taooceros, line 202: I think you have defined the enum? Why not use them?


```suggestion
        public LauncherPositions LauncherPosition { get; set; } = LauncherPositions.RememberLastLaunchLocation;
```

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -182,25 +182,41 @@ private void OnLoaded(object sender, RoutedEventArgs _)
 
         private void InitializePosition()
         {
-            if (_settings.RememberLastLaunchLocation)
+            if (_settings.LauncherPosition == "RememberLastLaunchLocation")
```
taooceros, line 185: Should use a switch for this

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -418,23 +446,38 @@ private void UpdatePosition()
             if (_animating)
                 return;
 
-            if (_settings.RememberLastLaunchLocation)
+            if (_settings.LauncherPosition == "RememberLastLaunchLocation")
```
taooceros, line 449: same switch for this

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -472,6 +515,15 @@ public double WindowTop()
             return top;
         }
 
+        public double WindowRight()
```
taooceros, line 518: what's this for?

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
```diff
@@ -382,6 +382,33 @@ public List<ColorScheme> ColorSchemes
...
 
+
+
+        public class LauncherPosition
+        {
+            public string Display { get; set; }
+            public LauncherPositions Value { get; set; }
+        }
+
+        public List<LauncherPosition> LauncherPositions
```
taooceros, line 393: This is clever, we may use this structure for all the combobox. Probably we want to cache this result? But that may break some behavior when changing language.

Flow.Launcher/MainWindow.xaml.cs
onesounds: for Right (screen) position. it using in top-right option.

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
onesounds: I just copied and pasted the existing structure. 😋

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
onesounds: applied it.

Flow.Launcher/MainWindow.xaml.cs
onesounds: I changed the part to call InitializePosition().

Flow.Launcher/MainWindow.xaml.cs
onesounds: Applied it.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -182,15 +182,28 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
+                case LauncherPositions.MouseScreenCenterTop:
+                    Left = WindowLeft();
+                    Top = 10;
+                    break;
+                case LauncherPositions.MouseScreenLeftTop:
+                    Left = 10;
+                    Top = 10;
+                    break;
+                case LauncherPositions.MouseScreenRightTop:
+                    Left = WindowRight();
```
taooceros, line 204: We may want to change name of all three method here.

taooceros (dismissed):

Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -435,6 +435,12 @@ private void OnPluginSettingKeydown(object sender, KeyEventArgs e)
                 pluginFilterTxb.Focus();
         }
 
+        private void LauncherPositionSelectedIndexChanged(object sender, SelectionChangedEventArgs e)
```
taooceros, line 438: what's this method for?

onesounds: I don't know either. It seems like a code that VS automatically makes. When I'm making something, I sometimes see this blank code.

jjw24:
@taooceros @onesounds could we merge dev in please.

onesounds:
- I confirmed the positioning problem from QuickSizeAdjust feature. (It touched positioning when Adjust Size)
- Fixed conflicts and issues
- Position Reset has small wrong positioning. (Positioning timing issue.)
- I'll fix this part more and merge it.

jjw24:
@onesounds let me know when pr is good to go again.

jjw24:
One thing from my testing prior to your recent commits was that on a fresh install, the query window starts up on the top left corner. I wonder if we should have `Mouse focused screen center` as a default instead?

onesounds:
> One thing from my testing prior to your recent merge Dev commits @onesounds, was that on a fresh install, the query window starts up on the top left corner. I wonder if we should have `Mouse focused screen center` as a default instead?

OK. I noticed little bugs. I'll mention to you when I think it's OK to review

jjw24:
> I think remember last position should default and it should be center when fresh install

This will have the query window stay on the same monitor screen when dual or more monitors are in use, wouldn't the mouse focus option be better for people who use multiple monitors (especially if they are new to flow and don't know you can change those options yet)?

onesounds:
A state that can be placed in the desired position should be the default.

> > I think remember last position should default and it should be center when fresh install
> 
> This will have the query window stay on the same monitor screen when dual or more monitors are in use, wouldn't the mouse focus option be better for people who use multiple monitors (especially if they are new to flow and don't know you can change those options yet)?

Well... OK. I think you are right too. default will be 'mouse focused screen center'.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -275,10 +317,17 @@ private void ToggleGameMode()
                 _viewModel.GameModeStatus = true;
             }
         }
+        private async void PositionReset()
+        {
+           _viewModel.Show();
+           await Task.Delay(300); // If don't give a time, Positioning will be weired.
```
jjw24, line 323:
```suggestion
           await Task.Delay(300); // If don't give a time, Positioning will be weird.
```

jjw24 (approved):
👍 👍 👍

PR #1368, opened 2022-09-07
Remember Setting window Size and Location

onesounds:
- Remember Setting Window size and location
- Remove Plugin SettingPanel's Max height


Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Win32;
 using ModernWpf;
 using System;
+using System.Diagnostics;
```
taooceros, line 15: Is this one needed?

taooceros (dismissed):

Flow.Launcher/SettingWindow.xaml.cs
onesounds: it don't need. removed.

onesounds:
Fix 0 by 0 positioning (when first launch) to centering and load position issue. (issue from WindowStartupLocation="Center" in xaml.)

jjw24 (approved):

PR #1369, opened 2022-09-07
Shortcuts to adjust width size, number of results shown and game mode

onesounds:
image

- Quick Adjust Width Size by Ctrl + Plus/Minus in query box. 
- Quick Adjust Max Result Shown by Ctrl + [ , Ctrl + ] in query box. 
- Add Tooltip 
- Adjustment adhere to maximum and minimum allowed
- Ctrl + F12 to change to Game Mode


onesounds:
- Add Ctrl+F12 to Toggle Game Mode.

onesounds:
I changed it because it was more intuitive to change it like this.
- Plus Minus : Max Result
- square brackets : Width

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -504,16 +505,76 @@ private void OnKeyDown(object sender, KeyEventArgs e)
                         _viewModel.LoadContextMenuCommand.Execute(null);
                         e.Handled = true;
                     }
+                    if (specialKeyState.CtrlPressed)
+                    {
+                        
+                        _settings.WindowSize = _settings.WindowSize + 100;
```
taooceros, line 511: I am curious why does this work. I suspect we need to set the property in the viewmodel to trigger change?

taooceros:
@onesounds please take a look on my latest change

onesounds:
> @onesounds please take a look on my latest change

It's well organized and the code works well.

onesounds:
but why the build faild?

Garulf:
Are we still using <kbd>Ctrl</kbd>+ Arrow keys?

taooceros:
> Are we still using <kbd>Ctrl</kbd>+ Arrow keys?
> If so this may interfere with normal text box shortcuts. 

I remove them.

jjw24:
1. When you use the shortcut for expanding or shrinking the query window width, the new width doesn't seem to get reflected back in the settings window.
2. Expanding the width of the query window does not respect the maximum width in the Settings, consequently you can keep expanding the width forever
3. Using the shortcut to expand the result list doesn't respect the maximum allowed results displayed, consequently can keep expanding forever.

onesounds:
> 1. When you use the shortcut for expanding or shrinking the query window width, the new width doesn't seem to get reflected back in the settings window.
> 2. Expanding the width of the query window using the shortcut does not respect the maximum width in the Settings, consequently you can keep expanding the width forever
> 3. Using the shortcut to expand the result list doesn't respect the maximum allowed results displayed, consequently can keep expanding forever.

Fixed

onesounds:
Need to make sure the subtitle sentence is appropriate. @Garulf

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -21,10 +21,11 @@
 using ISavable = Flow.Launcher.Plugin.ISavable;
 using System.IO;
 using System.Collections.Specialized;
+using CommunityToolkit.Mvvm.Input;
 
 namespace Flow.Launcher.ViewModel
 {
-    public class MainViewModel : BaseModel, ISavable
+    public partial class MainViewModel : BaseModel, ISavable
```
jjw24, line 28: @taooceros why is this changed to partial class?

global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "6.0.100",
-        "rollForward": "latestFeature"
+        "version": "6.0.*",
+        "rollForward": "latestPatch"
```
jjw24, line 4: @taooceros what's this change for?

Flow.Launcher/Flow.Launcher.csproj
```diff
@@ -114,4 +115,14 @@
...
+
+  <Target Name="RemoveDuplicateAnalyzers" BeforeTargets="CoreCompile">
+    <!-- Work around https://github.com/dotnet/wpf/issues/6792 -->
+
+    <ItemGroup>
+      <FilteredAnalyzer Include="@(Analyzer->Distinct())" />
+      <Analyzer Remove="@(Analyzer)" />
+      <Analyzer Include="@(FilteredAnalyzer)" />
+    </ItemGroup>
+  </Target>
```
jjw24, line 127: What's this change?

jjw24: @taooceros What's this change for us specifically?

jjw24:
> Need to make sure the subtitle sentence is appropriate. @Garulf

Updated the wording, hopefully ok.

Flow.Launcher/ViewModel/MainViewModel.cs
taooceros: Because I would like to use the mvvm toolkit to simplify the code of creating RelayCommand, which requires partial to generate additional code.

global.json
taooceros: Oh it shouldn't be there (I am testing different version of dotnet and see which one can help with the bug. However it is not needed because workaround is found)...Though it makes sense to apply the latest patch since that generally includes some security fixes.

taooceros:
> @taooceros What's this change for us specifically?

I couldn't reply directly. This is a workaround for source generator. Details here https://github.com/dotnet/wpf/issues/6792.

If that's being fixed, we can remove it.

taooceros:
> > @taooceros What's this change for us specifically?
> 
> I couldn't reply directly. This is a workaround for source generator. Details here [dotnet/wpf#6792](https://github.com/dotnet/wpf/issues/6792).
> 
> If that's being fixed, we can remove it.

I think it is fixed (in 6.0.401).

jjw24 (approved):

VictoriousRaptor:
Technically it's Ctrl + "=" and Ctrl + "-".

jjw24:
👍 https://github.com/Flow-Launcher/Flow.Launcher/pull/1480

onesounds:
@jjw24  @VictoriousRaptor 

You guys are too programmers. Why don't you consider the average user? 
This is my intended to make it easier for the general user to memorize hotkey. 
I changed the shortcut once in the middle for this PR. 

+:increase count.
-: decrease count.
[ : decrease size
] : increase size

It is a key chosen so that it can be intuitively understood.

"Technically it's Ctrl + "=" and Ctrl + "-"." <-  This can be talked about. This is because it is not a strong suggestion.
"OK I changed it" <- NOoooooooo

image

**Photoshop.** 
They are not idiots who don't even know the keycode.

image
**memopad**

image
**firefox**

I have never seen a case of writing = and - instead of + and - when matching these shortcuts. What stupid program writes shortcuts like this? flow launcher? 😂

jjw24:
Lol looking at those screenshots, realised I never paid much attention to how those shortcuts are described, makes sense. Will keep it as it is.

VictoriousRaptor:
In fact I found it when was looking at my keyboard and realized that + needs a shift combined to input. And I opened powertoys,tried adding some hotykeys, found that they use "=" to describe that key. And I thought every software use "=". Really never paid attention to that.

onesounds:
good. You guys sins are forgiven.

VictoriousRaptor:
So maybe for a single key, = is correct but for pairs like our new shortcut, + and - is a more user friendly choice.

PR #1380, opened 2022-09-14
Fixed QueryTextbox Binding event

Sparrkle:
Fixed an error when typing QueryTextBox in specific language.
For example, if the user enters "안녕하세요" and presses the Enter key immediately, it will be searched as "안녕하세".

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -561,7 +561,10 @@ namespace Flow.Launcher
         private void QueryTextBox_KeyUp(object sender, KeyEventArgs e)
         {
             if(_viewModel.QueryText != QueryTextBox.Text)
-                _viewModel.QueryText = QueryTextBox.Text;
+            {
+                BindingExpression be = QueryTextBox.GetBindingExpression(System.Windows.Controls.TextBox.TextProperty);
+                be.UpdateSource();
+            }
         }
     }
 }
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.ComponentModel;
 using System.Threading.Tasks;
 using System.Windows;
@@ -20,6 +20,7 @@ using Flow.Launcher.Infrastructure;
 using System.Windows.Media;
 using Flow.Launcher.Infrastructure.Hotkey;
 using Flow.Launcher.Plugin.SharedCommands;
+using System.Windows.Data;
 
 namespace Flow.Launcher
 {
@@ -554,6 +555,12 @@ namespace Flow.Launcher
             {
                 ModernWpf.ThemeManager.Current.ApplicationTheme = ModernWpf.ApplicationTheme.Dark;
             }
+        }
+
+        private void QueryTextBox_KeyUp(object sender, KeyEventArgs e)
+        {
+            BindingExpression be = QueryTextBox.GetBindingExpression(System.Windows.Controls.TextBox.TextProperty);
+            be.UpdateSource();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


onesounds:
korean works well.  I think there would have been similar problems with Chinese or Japanese.
(I've been through this problem for a long time, but I couldn't find a solution. In the end, I was mainly used in English.)
If there is no problem in other languages, it is better to apply it.

taooceros:
Thanks for the fix! Could you explains a bit why `PropertyChange` is not enough for triggering change to source for other language (is it because of ime)?

Sparrkle:
In the case of Hangul, initial, neutral, and final letters are combined to form one letter. 
However, it is assumed `PropertyChanged` event will be triggered when the next initial is passed because only the initial and neutral can be entered.

taooceros:
I am worried by whether this may trigger too much unnecessary result when user use input method like for Chinese. Before the text is typed in, the query is triggered and the querytext will not change.

https://github.com/Flow-Launcher/Flow.Launcher/blob/c81dd32d9951fa4cc8851eecfb01fb1befd24f5c/Flow.Launcher/ViewModel/MainViewModel.cs#L329-L336

Please also add a equality check here (note for `requery`, which is handled in a different way).

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -559,8 +560,7 @@ public void InitializeColorScheme()
 
         private void QueryTextBox_KeyUp(object sender, KeyEventArgs e)
         {
-            BindingExpression be = QueryTextBox.GetBindingExpression(System.Windows.Controls.TextBox.TextProperty);
-            be.UpdateSource();
+            _viewModel.ChangeQueryText(QueryTextBox.Text);
```
taooceros, line 563: Then are we still using Binding? It seems if this looks more like a solution, then we are only bind `OneWay`.

taooceros: Oh wait a moment, you can't use this method, because that will move the cursor to the end. So maybe it will be better to just add a if check for your previous change.

Sparrkle: Oh, I checked too. so modified different way.

Sparrkle:
fixed a bug not triggered while the backspace was pressed.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -555,5 +557,11 @@ public void InitializeColorScheme()
                 ModernWpf.ThemeManager.Current.ApplicationTheme = ModernWpf.ApplicationTheme.Dark;
             }
         }
+
+        private void QueryTextBox_KeyUp(object sender, KeyEventArgs e)
+        {
+            if(_viewModel.QueryText != QueryTextBox.Text)
+                _viewModel.QueryText = QueryTextBox.Text;
```
taooceros, line 564: lol do you want to change the binding mode to oneway?

Sparrkle: It looks like a oneway but it works twoway.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -561,7 +561,10 @@ public void InitializeColorScheme()
         private void QueryTextBox_KeyUp(object sender, KeyEventArgs e)
         {
             if(_viewModel.QueryText != QueryTextBox.Text)
-                _viewModel.QueryText = QueryTextBox.Text;
+            {
+                BindingExpression be = QueryTextBox.GetBindingExpression(System.Windows.Controls.TextBox.TextProperty);
+                be.UpdateSource();
```
taooceros, line 566: I will approve but just curious any difference between this two?

taooceros (approved):

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -561,7 +561,10 @@ public void InitializeColorScheme()
...
+            }
```
taooceros, line 567: Sorry I mean these two

Sparrkle: I just tested it and it works the same :'( There is no difference.

PR #1381, opened 2022-09-14
Special character error in ole db query

Sparrkle:
Throw exception when type "＃, ＆, ＊, ＠".

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
@@ -1,4 +1,4 @@
-using Flow.Launcher.Infrastructure.Logger;
+﻿using Flow.Launcher.Infrastructure.Logger;
 using Flow.Launcher.Plugin.Explorer.Search.QuickAccessLinks;
 using Microsoft.Search.Interop;
 using System;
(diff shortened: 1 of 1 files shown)


Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/IndexSearch.cs
```diff
@@ -17,7 +17,7 @@ internal static class IndexSearch
     {
 
         // Reserved keywords in oleDB
-        private const string reservedStringPattern = @"^[`\@\#\^,\&\/\\\$\%_;\[\]]+$";
+        private const string reservedStringPattern = @"^[`\@\＠\#\＃\＊\^,\&\＆\/\\\$\%_;\[\]]+$";
```
taooceros, line 20: I am curious why these additional character need escape? They are not really special character for regex.

Sparrkle: if those characters are entered alone, an error in "db".
Its so weird..

Sparrkle: In Korean, special characters are sometimes entered with function keys. There is a problem that is recognized the same even if # and # are different.

taooceros (approved):

PR #1387, opened 2022-09-16
Fix opening url twice when clicking on exception report link

jjw24:
To repro:

1. Cause an exception
2. Ctrl + click on the exception report window link
3. Observe the link is opened twice in two browser tabs
(If you can debugging locally, you can install GitHub Quick Launcher and type query 'gr' and exception window will pop up. This exception is not an issue with the plugin, but because flow is not run from the expected Local AppData location.)

image

### DIFF
--- a/Flow.Launcher/ReportWindow.xaml.cs
+++ b/Flow.Launcher/ReportWindow.xaml.cs
@@ -77,7 +77,6 @@ namespace Flow.Launcher
             };
             link.Inlines.Add(url);
             link.NavigateUri = new Uri(url);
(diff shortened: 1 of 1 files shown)


taooceros (approved):

Garulf:
... I need to update Flox on that I guess. Whoops.

jjw24:
> ... I need to update Flox on that I guess. Whoops.

I wouldn't suggest you do that because this is run from dev environment, which means it's in a random workspace that is not your standard flow installation location, you will waste effort figuring out where the dev has flow folder placed.

PR #1388, opened 2022-09-16
Spanish & Spanish Latin America languages release for v1.9.5

jjw24:
These are all language files for Spanish & Spanish Latin America added to the dev branch at the time of this PR, copied across for v1.9.5 release.

I purposely left out the language files in Windows Settings plugin because the changes are too drastic.

Expected:
Should not have any differences in the language files between this branch and dev branch.

Tested:
Selected both Spanish and Spanish Latin America and observed that the language has changed on each screen including plugin settings.

### DIFF
--- a/Flow.Launcher/Languages/es-419.xaml
+++ b/Flow.Launcher/Languages/es-419.xaml
@@ -1,0 +1,290 @@
+﻿<?xml version="1.0"?>
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:system="clr-namespace:System;assembly=mscorlib">
+    <!--  MainWindow  -->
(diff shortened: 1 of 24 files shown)


jjw24:
I am just playing it safe, I don't know how git will handle the differences there, this is just for a release so don't want to make it too complicated.

taooceros:
Will it be hard to merge back? I am only concerning that.

jjw24:
Yes my concern as well that's why I left the plugin alone, if you take a look I think there is already a default Spanish language file for master branch called 'es-EM', which is renamed (sort of) to 'es' in Dev.

taooceros:
> Yes my concern as well that's why I left the plugin alone, if you take a look I think there is already a default Spanish language file for master branch called 'es-EM', which is renamed (sort of) to 'es' in Dev.

ok let's do that in the future. Hope it won't be too hard to merge back XD.

taooceros (approved):

jjw24:
Shouldn't be, changes are clean so far

PR #1391, opened 2022-09-19
check query for null

### LINKED ISSUE #1366: Null exception: Object reference not set to an instance of an object
Flow Launcher version: 1.9.4
OS Version: Microsoft Windows NT 10.0.22000.0
IntPtr Length: 8
x64: True

Python Path: 
Date: 09/07/2022 01:06:27
Exception:
System.NullReferenceException: Object reference not set to an instance of an object.
   at Flow.Launcher.Core.Plugin.PluginManager.ValidPluginsForQuery(Query query) in C:\projects\flow-launcher\Flow.Launcher.Core\Plugin\PluginManager.cs:line 178
   at Flow.Launcher.ViewModel.MainViewModel.QueryResults() in C:\projects\flow-launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 624
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__140_0(Object state)
[2022-09-07.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/9500213/2022-09-07.txt)

taooceros:
Defensive coding. The error suggests there's something here that can be null, which can only be the query. However, I am not sure when it will happen since the reporter doesn't provide way to consistent reproduce the error.

fix #1366

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -166,18 +166,18 @@ namespace Flow.Launcher.Core.Plugin
 
         public static ICollection<PluginPair> ValidPluginsForQuery(Query query)
         {
-            if (NonGlobalPlugins.ContainsKey(query.ActionKeyword))
-            {
-                var plugin = NonGlobalPlugins[query.ActionKeyword];
-                return new List<PluginPair>
-                {
-                    plugin
-                };
-            }
-            else
-            {
+            if (query is null)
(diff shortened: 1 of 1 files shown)


jjw24:
Is it better to log an error than or to return empty? What's the behaviour when an empty is return, does the query box display anything?

taooceros:
I think it will just do nothing because there's nothing to enumerate and query for.

jjw24 (approved):

PR #1410, opened 2022-09-25
Spanish translation #1409,fix when modifying the language,LastQueryMode's Display no immediate translation

### LINKED ISSUE #1409: [Dev] Spanish translation (FL 1.10.0_ES)_3
In "Flow Launcher Settings"/"Plugins"

Description on priority mouseover, not translated.
String "Change Plugin Results Priority", not found in Crowdin
220924 FL 1 10 Complementos_Cuadro emergente prioridad_Sin traducir

NeedJustWord:
Close #1409

### DIFF
--- a/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
+++ b/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
@@ -125,25 +125,44 @@ namespace Flow.Launcher.ViewModel
         #region general
 
         // todo a better name?
-        public class LastQueryMode
+        public class LastQueryMode : BaseModel
         {
             public string Display { get; set; }
             public Infrastructure.UserSettings.LastQueryMode Value { get; set; }
         }
+
+        private List<LastQueryMode> _lastQueryModes = new List<LastQueryMode>();
         public List<LastQueryMode> LastQueryModes
         {
             get
             {
-                List<LastQueryMode> modes = new List<LastQueryMode>();
-                var enums = (Infrastructure.UserSettings.LastQueryMode[])Enum.GetValues(typeof(Infrastructure.UserSettings.LastQueryMode));
-                foreach (var e in enums)
+                if (_lastQueryModes.Count == 0)
                 {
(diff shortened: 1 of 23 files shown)


Flow.Launcher/ViewModel/SettingWindowViewModel.cs
```diff
@@ -125,25 +125,44 @@ public void Save()
...
+                if (_lastQueryModes.Count == 0)
                 {
-                    var key = $"LastQuery{e}";
-                    var display = _translater.GetTranslation(key);
-                    var m = new LastQueryMode { Display = display, Value = e, };
-                    modes.Add(m);
+                    _lastQueryModes = InitLastQueryModes();
                 }
-                return modes;
+                return _lastQueryModes;
```
taooceros, line 143: Thanks for the fix!
Though, I didn't see why this can fix the issue. It seems like you only add a cache for that (which is super reasonable though)?
Would you please explain a bit?

taooceros:
Just as a quick note, you don't have to modify all language file to add new translation string, just modify the base one (en.xaml) and the ones you do want to translate (se.xaml, zh-cn.xaml, zh-tw.xaml for this case) and others will be added automatically by the integration.

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
NeedJustWord: 1.LastQueryMode add INotifyPropertyChanged interface
public class LastQueryMode : BaseModel

2.when modifying the language,call UpdateLastQueryModeDisplay();
        private void UpdateLastQueryModeDisplay()
        {
            foreach (var item in LastQueryModes)
            {
                item.Display = _translater.GetTranslation($"LastQuery{item.Value}");
            }
        }

        public string Language
        {
...
            set
            {
 ...
                UpdateLastQueryModeDisplay();//when modifying the language,
            }
        }

taooceros (approved):

Issue #1416, opened 2022-09-26
An option to automatically clear logs 


VictoriousRaptor:
**Is your feature request related to a problem? Please describe.**

The logs are not small. Log folder can be quite big after several months. And old logs are useless most of the time.

**Describe the solution you'd like**

An option to automatically clear old logs, or clear the folder after reaching a certain size.

**Describe alternatives you've considered**

Maybe a button to clear old logs.

**Additional context**

Remzi1993:
This should also be opt-in rather than opt-out. Logs should be turned off by default and there should be an option to turn on logs with a checkbox and beneath the checkbox a link to the logs directory so users can easily find the logs directory and upload it here.

onesounds:
We will not turn off the log by default. (Because it's a problematic program. 😂) I added a button to check the log folder in about panel last time. I will add a function to erase the log.

VictoriousRaptor:
It would be better to show the size of log folder, to let users decide when
they should clear the logs.

DB P ***@***.***> 于 2022年9月27日周二 09:03写道：

onesounds:
> It would be better to show the size of log folder, to let users decide when they should clear the logs. DB P ***@***.***> 于 2022年9月27日周二 09:03写道：

Nice idea. I'll add it to button label.

Remzi1993:
> We will not turn off the log by default. (Because it's a problematic program. 😂) I added a button to check the log folder in about panel last time. I will add a function to erase the log.

Haha 🤣 Maybe after it reaches stability 😅

PR #1418, opened 2022-09-27
Add 'Clear Log' Button in about

onesounds:
image

https://github.com/Flow-Launcher/Flow.Launcher/issues/1416

### DIFF
--- a/Flow.Launcher/Languages/en.xaml
+++ b/Flow.Launcher/Languages/en.xaml
@@ -162,6 +162,8 @@
     <system:String x:Key="devtool">DevTools</system:String>
     <system:String x:Key="settingfolder">Setting Folder</system:String>
     <system:String x:Key="logfolder">Log Folder</system:String>
+    <system:String x:Key="clearlogfolder">Clear Logs</system:String>
+    <system:String x:Key="clearlogfolderMessage">Are you sure you want to delete the entire log?</system:String>
     <system:String x:Key="welcomewindow">Wizard</system:String>
 
     <!--  FileManager Setting Dialog  -->
--- a/Flow.Launcher/SettingWindow.xaml.cs
+++ b/Flow.Launcher/SettingWindow.xaml.cs
@@ -269,6 +269,23 @@ namespace Flow.Launcher
         {
             PluginManager.API.OpenDirectory(Path.Combine(DataLocation.DataDirectory(), Constant.Logs, Constant.Version));
         }
+        private void ClearLogFolder(object sender, RoutedEventArgs e)
+        {
+            var confirmResult = MessageBox.Show(InternationalizationManager.Instance.GetTranslation("clearlogfolderMessage"),InternationalizationManager.Instance.GetTranslation("clearlogfolder"), MessageBoxButton.YesNo);
+            if (confirmResult == MessageBoxResult.Yes)
+            {
+                DirectoryInfo Di = new DirectoryInfo(Path.Combine(DataLocation.DataDirectory(), Constant.Logs, Constant.Version));
+                FileInfo[] files = Di.GetFiles();
+                foreach (FileInfo file in files)
+                {
+                    file.Delete();
+                }
+            }
+            else
+            {
+            }
+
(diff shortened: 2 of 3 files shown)


Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -269,6 +269,23 @@ private void OpenLogFolder(object sender, RoutedEventArgs e)
         {
             PluginManager.API.OpenDirectory(Path.Combine(DataLocation.DataDirectory(), Constant.Logs, Constant.Version));
         }
+        private void ClearLogFolder(object sender, RoutedEventArgs e)
+        {
+            var confirmResult = MessageBox.Show(InternationalizationManager.Instance.GetTranslation("clearlogfolderMessage"),InternationalizationManager.Instance.GetTranslation("clearlogfolder"), MessageBoxButton.YesNo);
+            if (confirmResult == MessageBoxResult.Yes)
+            {
+                DirectoryInfo Di = new DirectoryInfo(Path.Combine(DataLocation.DataDirectory(), Constant.Logs, Constant.Version));
+                FileInfo[] files = Di.GetFiles();
```
taooceros, line 278: Maybe use EnumerateFiles?

onesounds: I don't know what it is. anyway, applied. 😅

onesounds:
image
- Directory size in label

Flow.Launcher/Languages/en.xaml
```diff
@@ -162,6 +162,8 @@
     <system:String x:Key="devtool">DevTools</system:String>
     <system:String x:Key="settingfolder">Setting Folder</system:String>
     <system:String x:Key="logfolder">Log Folder</system:String>
+    <system:String x:Key="clearlogfolder">Clear Logs</system:String>
+    <system:String x:Key="clearlogfolderMessage">Are you sure you want to delete the entire log?</system:String>
```
VictoriousRaptor, line 166: "Are you sure you want to delete all logs" or "clear all logs" maybe better.

onesounds: "Are you sure you want to delete all logs" applied.

Flow.Launcher/SettingWindow.xaml.cs
taooceros: Ah it's just a newer api that has better performance (not noticable Though) by using state machine to allow process the result lazily. https://stackoverflow.com/questions/5669617/what-is-the-difference-between-directory-enumeratefiles-vs-directory-getfiles

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
```diff
@@ -555,6 +555,39 @@ public FamilyTypeface SelectedResultFontFaces
...
+        {
+            get 
+            { 
+                DirectoryInfo dirInfo = new DirectoryInfo(Path.Combine(DataLocation.DataDirectory(), Constant.Logs, Constant.Version));
+                long size = 0;
+                foreach (FileInfo fi in dirInfo.GetFiles("*", SearchOption.AllDirectories))
+                {
+                    size += fi.Length;
+                }
+            return string.Format(_translater.GetTranslation("clearlogfolder")) + " (" + FormatBytes(size) + ")" ;
```
taooceros, line 568: https://stackoverflow.com/a/22111211

Try this instead? (it looks more elegant XD)

taooceros: Also please use string interpolation instead of string.format

taooceros, line 568: Why do we need a string.Format here? Correct me if wrong (i am reviewing in mobile). But it seems like this is only string concatenations?

onesounds: mistake. fixed.

jjw24 (approved):
👍

Issue #1425, opened 2022-09-28
BUG: Segoe UI font looks bad in some languages


VictoriousRaptor:
### Checks

- [X] I have checked that this issue has not already been reported.

### Problem Description

Some text blocks like `TabMenu` has property `FontFamily` explicitly set to "Segoe UI", which looks bad in some languages like Chinese and Japanese (see screenshots below). Just simply delete the property `FontFamily` and let the program use system default font will fix it. It looks fine in Chinese, but not sure if it causes issues in other languages.

### Screenshots

Segoe UI vs system default font, Chinese
![Segoe UI vs system default font, Chinese](https://user-images.githubusercontent.com/10308169/192784188-c4bfcce2-8d19-4334-a4b1-14d42d12a0e7.png)

### Flow Launcher Version

v1.9.4-v1.9.5

### Windows Build Number

10.0.19044.2006

onesounds:
Fixed. https://github.com/Flow-Launcher/Flow.Launcher/pull/1427
Thanks.

PR #1427, opened 2022-09-28
Remove font-family style for Chinese

onesounds:
**What's the PR**
- issue from https://github.com/Flow-Launcher/Flow.Launcher/issues/1425
- Removed font-family tabmenu / other windows

**Before**
image

**After**
image

**Checklist**
- There should be no impact on other language.
- It will use the default OS font, so I think there will be no problem.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -166,18 +166,18 @@ namespace Flow.Launcher.Core.Plugin
 
         public static ICollection<PluginPair> ValidPluginsForQuery(Query query)
         {
-            if (NonGlobalPlugins.ContainsKey(query.ActionKeyword))
-            {
-                var plugin = NonGlobalPlugins[query.ActionKeyword];
-                return new List<PluginPair>
-                {
-                    plugin
-                };
-            }
-            else
-            {
+            if (query is null)
+                return Array.Empty<PluginPair>();
+            
+            if (!NonGlobalPlugins.ContainsKey(query.ActionKeyword))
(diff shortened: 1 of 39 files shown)


VictoriousRaptor:
Segoe UI is not only used in tabmenu but also elsewhere. Please fix them as well (if you are busy I can do this later, I have to sleep now:( ). i ve tested that it also affects Japanese, maybe Korean as well.

VictoriousRaptor:
BTW it should affect some English texts on non-english locale systems. English texts in system default font on non-english systems, like Chinese, is slightly different from Segoe UI.

I think it doesn't matter cuz most people will choose their system language.

onesounds:
When I checked, there were no more cases where Segoe UI was designated.
I checked the other Asian fonts, and it was not a problem.

If you know how to, Do as you wish.

VictoriousRaptor:
> When I checked, there were no more cases where Segoe UI was designated. I checked the other Asian fonts, and it was not a problem.
> 
> If you know how to, Do as you wish.

There are some cases in plugins. I've created a PR to your fork.

> BTW it should affect English texts on non-english locale systems. English texts in system default font on non-english systems, like Chinese, is slightly different from Segoe UI.
> 

This is because there is no semibold weight in default Chinese font (MS YaHei UI).

onesounds:
I missed the plugins. thanks.

> BTW it should affect English texts on non-english locale systems. English texts in system default font on non-english systems, like Chinese, is slightly different from Segoe UI.

I checked it after changing OS to another language. It didn't seem particularly problematic. It doesn'thave to be Segoe UI.

PR #1431, opened 2022-09-29
Fix Blurry Context menu in query box

onesounds:
**What's the PR**
- It's a small difference, but the font in the Context menu that opens in Querybox looks blurry.
- Fix the issue

**Before**
image

**After**
image

**References**
- https://stackoverflow.com/questions/21538296/wpf-dropshadow-on-button-causes-blurry-text
- The same style is shared by the Tray menu, but there is no problem.


Garulf:
WPF has some weird issues with blurring fonts. 

Looks so much better.

jjw24 (approved):

PR #1433, opened 2022-09-29
Use translation provided in launcher instead of in Program plugin

VictoriousRaptor:
Some texts (e.g "edit", "OK") in Program plugin use translations provided in the plugin. No need to translate it again in the plugin.

BTW how to delete all unused translations?

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/AddProgramSource.xaml
+++ b/Plugins/Flow.Launcher.Plugin.Program/AddProgramSource.xaml
@@ -97,7 +97,7 @@
                     Margin="5,0,0,0"
                     HorizontalAlignment="Right"
                     Click="ButtonAdd_OnClick"
-                    Content="{DynamicResource flowlauncher_plugin_program_update}"
+                    Content="{DynamicResource done}"
                     Style="{DynamicResource AccentButtonStyle}" />
             </StackPanel>
         </Border>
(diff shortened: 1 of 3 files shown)


onesounds:
I have also considered this work. I thought that this should not be duplicated. but I didn't do this for a few reasons.

1. Other plug-ins also have their own String. (add, edit, delete...) Shouldn't they all be applied? if so, Shouldn't it be organized in one PR?

2. I think it is right for the plug-in to have the string used in the plug-in itself. The plug-in and flow shall be constructed separately. I also think this is rarely the case, but there are cases of porting.

Anyway, it is necessary to set rules for this pr to be applied. (Whether to allow redundant strings or not) I am not very interested in this rule because I am not a programmer. It doesn't matter in my case. but I know that programmers care about these problems. 

>BTW how to delete all unused translations?

 I'm not sure. as I know, unused string can be removed by removing the EN string

VictoriousRaptor:
> Other plug-ins also have their own String. (add, edit, delete...) Shouldn't they all be applied? if so, Shouldn't it be organized in one PR?

To find all common strings is hard and boring. My suggestion is that just leave it, until you find one somehow. 

>     2. I think it is right for the plug-in to have the string used in the plug-in itself. The plug-in and flow shall be constructed separately. I also think this is rarely the case, but there are cases of porting.

I didn't consider this case. Referencing resources of Flow does harm separation between plugins and Flow.

jjw24:
>I didn't consider this case. Referencing resources of Flow does harm separation between plugins and Flow.

Yes that's correct, sorry I am late to the party. Each plugin should be treated as self contained, with the only link is via the Plugin API, otherwise the dependencies will be hard to maintain. This should also be the case even for the plugins we pack with flow which we still have a little bit left to fix.

PR #1457, opened 2022-10-07
Grouping in Plugin Store

onesounds:
### New Version ###

image
image

**What's the PR**
- Add Grouping in pluginstore, and Order Like this.
  - New (Added in 7 days)
  - Recently Updated (Update code in 7 days)
    - If New/Recently Updated Both item, Display "New" Only.
  - Plugins (No label state)
  - Installed (Installed Item)
- Add "Update" Circle Label
  - it showing only Installed Item
  - it showing when has different version in local.
- Add 'Update' / 'Uninstall' Button in Plugin store

**Test Cases**
- The search behavior should operate appropriately.
- Regardless of language settings, it should be arranged in the following order.
   - New / Recently Updated / Plugins / installed
- Matching Color in Dark Mode
- Translatable Text
- Show Green Circle Label Tooltip
- The uninstall and update buttons shall be properly displayed.

----
<strike>
### Old Version ###
**What's the PR**
- New / Installed / Updated Label in Plugin Store. 
- Based Plugin Search PR. so this pr should merge when done plugin Search PR.
- It was implemented in the form of adding a value to the list using the for syntax. (LabelNew=true...)

**Label Type**
- New : Registered plugin store in 7 days.  
- Updated : Updated in 5 days. if it new, only showing New. (new and updated doesn't show both.)
- Installed. Installed can show both with Updated. 

image
image

**Test Cases**
- [x] Order by 'new' first. 'updated' 2nd. installed going to bottom.
- [x] Matching Color in Dark Mode
- [x] Translatable Text
- [x] Show Label Tooltip
- [x] If it installed item, Hide Install Button <- Someone would certainly raise the issue of a bug.
- [x] Perfomance
- [x] There are no case showing triple labels. max label count is two. 

**Remain Tasks**
- [x] If Item installed & updated. Show 'Update' / 'Uninstall' Button.
- [x] Fix AddedDate in manifest
- [x] Remove Temp Code (manifest url, new label comment)
</strike>


jjw24:
> If it installed item, Hide Install Button <- Someone would certainly raise the issue of a bug.

if installed and no update, should we not just hide it from store?

onesounds:
> > If it installed item, Hide Install Button <- Someone would certainly raise the issue of a bug.
> 
> if installed and no update, should we not just hide it from store?

It is an issue that has been considered. I concluded that it was better not to hide. (I put it at the bottom of the list in now.) 

1. In a typical app store, it is not hidden as an installed app.

2. If wehide it, user can think, 'Why isn't the app I use in the store?'

It is a matter of policy determination. If there are other opinions, they can be reflected at any time.

Flow.Launcher.Core/ExternalPlugins/UserPlugin.cs
```diff
@@ -12,5 +14,12 @@ public record UserPlugin
         public string UrlDownload { get; set; }
         public string UrlSourceCode { get; set; }
         public string IcoPath { get; set; }
+        public DateTime LatestReleaseDate { get; set; }
+        public DateTime DateAdded { get; set; }
+
+        /* Label Data for Plugin Store */
+        public bool LabelNew { get; set; }
+        public bool LabelInstalled { get; set; }
+        public bool LabelUpdated { get; set; }
```
taooceros, line 23: Let's move this to a viewmodel for plugin store.

Flow.Launcher/Converters/BoolToVisibilityConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Data;
+
+namespace Flow.Launcher.Converters
+{
+    public class BoolToVisibilityConverter : IValueConverter
```
taooceros, line 13: I remember these's a buildin boolconverter (used by @Sparrkle before). So let's use that.

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
```diff
@@ -261,10 +267,51 @@ public IList<UserPlugin> ExternalPlugins
...
+                if (DateAdded <= 7)
+                {
+
+                    //item.LabelNew = true; // Add New Label
+                    //item.LabelUpdated = false; // Hide Updated Label when Added New Label. New and Update doesn't show both same time.
+                }
+            }
+
+            // New first, Updated second. Installed to bottom of list.
+            list = list.OrderByDescending(x => x.LabelNew).OrderByDescending(x => x.LabelUpdated).ThenBy(y => y.LabelInstalled).ToList();
```
taooceros, line 310: I believe the second one also needs to be `ThenByDescending`

taooceros (changes requested):

Flow.Launcher/ViewModel/SettingWindowViewModel.cs
onesounds: Applied

Flow.Launcher/Converters/BoolToVisibilityConverter.cs
onesounds: I remember I saw it. but where is it?

Flow.Launcher.Core/ExternalPlugins/UserPlugin.cs
onesounds: How to? I'm not good at organizing things like this.

jjw24:
> > > If it installed item, Hide Install Button <- Someone would certainly raise the issue of a bug.
> > 
> > 
> > if installed and no update, should we not just hide it from store?
> 
> It is an issue that has been considered. I concluded that it was better not to hide. my reasons...
> 
> 1. In a typical app store, it is not hidden as an installed app.
> 2. If we hide it, user can think, 'Why isn't the app I use in the store?'
> 3. I put 'installed' items at the bottom of the list in now.
> 
> The proposals were "hide it" or" label it," but I think this is a proper compromise since I order it down and labeled it. It is a matter of policy determination. If there are other opinions, I can be reflected. I think it's okay to hide it. It doesn't matter to me.

Ok makes sense 👍

onesounds:
I noticed I was wrong plugin store label plan.
for example, update button should based on version diff.
There will be a change of plans... TT

onesounds:
- New Version : the item changed in 7 days. 
- Update : version diff && installed.
- Installed : Same.

- New Version and Update doesn't show both. Show 'Update' only in this case.

- So, old "New" label is gone. we don't need 'DateAdded' item in manifest. Sorry, @Garulf  (Actually, This item is data that does not need to remain on the list consistently because it is useful for a short period of time.)

Garulf:
At this point I'd like to keep the `DateAdded` field.

jjw24:
> Fix AddedDate in manifest

This task is no longer required?

onesounds:
> > Fix AddedDate in manifest
> 
> This task is no longer required?

It required.

onesounds:
I rewrite description and testcase things. check the top of this page please.

Flow.Launcher/Languages/en.xaml
```diff
@@ -83,13 +83,24 @@
...
+    <system:String x:Key="pluginStore_None">Plugins</system:String>
+    <system:String x:Key="pluginStore_Installed">Installed</system:String>
     <system:String x:Key="refresh">Refresh</system:String>
-    <system:String x:Key="install">Install</system:String>
+    <system:String x:Key="installbtn">Install</system:String>
+    <system:String x:Key="uninstallbtn">Uninstall</system:String>
+    <system:String x:Key="updatebtn">Update</system:String>
+    <system:String x:Key="LabelInstalledToolTip">Plug-in already installed</system:String>
+    <system:String x:Key="LabelNew">New Version</system:String>
+    <system:String x:Key="LabelNewToolTip">This plug-in has been updated within the last 7 days</system:String>
```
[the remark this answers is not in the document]
onesounds: OK.

jjw24 (approved):

jjw24:
Just added 3 commits:
1. Changed hardcoded strings to constants, this helps avoid typos and easier update in the future if needed.
2. Moved plugin install/update/uninstall calls from the View class to the ViewModel class, makes the code simpler and separates View logic from ViewModel logic.
3. Since we changed plugin_install key to installbtn and likewise for plugin_uninstall to uninstallbtn, updated all other language keys, otherwise they will be obsolete.

PR #1476, opened 2022-10-16
[Program Plugin] Support .url file

LeLocTai:
Notably Steam use .url for games shortcut.

----

**What's the PR**
- Add Function the program plug-in to search for shortcut that use the .url format, such as Steam or Epic.
- Changed Label File Suffixes -> File Type 
- Changed File Type Window.
- User can set up protocols other than Steam and Epic.
- Default Setting
  - on : exe, lnk, ms-app / steam, epic

**Test Cases**
- The icon in the url file should be displayed normally.
- The default settings for the first installation should be as follows: on : exe, lnk, ms-app / steam, epic
- It must behave according to the checkboxes.
- The file format / protocol entered in the custom entry must work.
- Even if a custom item is entered, it should not work if you uncheck the check box.
- Items entered in custom items must be saved even if they are not used.
- The custom text box should only be visible when you turn on the custom
- When you focus on each textbox, the appropriate description should be displayed.
- There should be no conflicts with users who have existing settings.


VictoriousRaptor:
What if the .url file is not a steam shortcut but a internet page? Obviously we don't want it in Program.

LeLocTai:
It is possible to filter by protocol. But to be honest I've been adding url to Program for months without this patch and never have a thought about it. My brain just filter them out. 
I hope we can merge this as is, and maybe adding a customizable protocol later. There are already many useless results anyway, and fuzzy match have always do a good job of surfacing the relevant one quickly. Having thumbnail on steam shortcut is great and eliminate the need for extra plugin. Those who want less result can always remove .url from Program settings.

onesounds:
image



image

The above screen shot is not this pr, but only preview pr and suffix are set.

- Currently, we are going to change the icon part in other PR (preview). If this pr is applied, the url file type icon will also be displayed appropriately. 
- How about adding only .url to the suffix default setting for this feature? When I test it, it seems to work the same way as this PR. This pr is focused on outputting the icon. If the preview(and icon) pr works properly, this function can be solved with suffix alone.

---
https://developer.valvesoftware.com/wiki/Steam_browser_protocol

It's my first time seeing this url file type, and I confirmed that I usually use it on Steam. The other files were link files to websites that most would not use. (docs in web or Game services that require website login) Considering the high number of downloads of the Steam plug-in, it is certainly inconvenient that the Steam game cannot be searched in windows. Although the number of downloads of the Steam Plug-in will decrease, it seems to good to be provided as a default. (There may be overlapping results with existing Steam Plug-in users.)

VictoriousRaptor:
Personally I don't think this is a good idea. Not the idea of supporting indexing steam games itself, but the way you manage these.url files. You need to put these .url files in a folder and manually manage them. Steam only let you add shortcuts to desktop iirc. If you don't want  all these shortcuts on your desktop you need to cut them to somewhere else.

LeLocTai:
Stream also create url files in the start menu. You don't have to manually do anything. Even windows builtin search can find these files, and this pr bring Flow to parity with it which is the bare minimum in term of capabilities.

My main issue with the steam plugin is it doesn't score the results, so you must use a trigger word which is very inconvenient.

VictoriousRaptor:
> Stream also create url files in the start menu. You don't have to manually do anything

It's my problem then. I just found that I never check the "Add to start menu" checkbox.

> My main issue with the steam plugin is it doesn't score the results, so you must use a trigger word which is very inconvenient.

Seems related to https://github.com/Garulf/Steam-Search/issues/30.

LeLocTai:
I'm aware of the issue and have discussed with Garulf about it. While it is simple to give the results a score boost, it is not easy to give them a score that play well with other result since fuzzy match is not accessible in python. Scoring is also not well documented in general.

Garulf:
> I'm aware of the issue and have discussed with Garulf about it. While it is simple to give the results a score boost, it is not easy to give them a score that play well with other result since fuzzy match is not accessible in python. Scoring is also not well documented in general.

I have a PR in the works that'll allow Python plugins to use Flow Launchers Fuzzy matching.

Garulf:
> really??? how's that possible🤔

Actually not hard to implement. I just haven't decided on the design.

onesounds:
@VictoriousRaptor Error.
image

taooceros (commented):
LGTM, but there's a few things to check.

Plugins/Flow.Launcher.Plugin.Program/ProgramSuffixes.xaml
```diff
@@ -65,10 +133,81 @@
...
+                    <TextBlock
+                        FontSize="14"
+                        Text="{DynamicResource flowlauncher_plugin_program_suffixes_tooltip}"
+                        TextWrapping="Wrap" />
+                </Border>
+
+
+                <!--
+                    <TextBox x:Name="tbSuffixes" Margin="0,20,0,20" />
+                -->
```
jjw24, line 157:
```suggestion
```

jjw24 (approved):
Thanks for doing the testing 👍

PR #1485, opened 2022-10-18
Fix Setting Window Designs

onesounds:
**What's the PR**
- Seperate style to other file from SettingWindow.Xaml (SettingWindow.xaml was too long)
- Change the Checkboxes to Toggle button in general / Hotkey / Proxy Tab

**Before**

image

**After**

image

**Fix Plugins Designs**
 - Fix Clickable Area
image

 - Fix Hover Color

image

 - Adjust Some Colors
 
image

Fix Store with Darkmode
image

**Why?**
- https://uxdworld.com/2018/08/13/checkbox-vs-toggle-switch-7-use-cases-of-forms-design/
- It is not recommended to mix check boxes and toggles. Organize in a toggle for a sense of unity.
- It was a todo that I tried to do in the beginning redesign, but I didn't know how to bind at the time.

**Test Case**
- Each Toggle Button must operate correctly.
- The color shall be marked accurately when you mouse over the plugin item in plugin list
- Even in dark mode, the color shall be displayed accurately.

### DIFF
--- a/Flow.Launcher.Core/ExternalPlugins/UserPlugin.cs
+++ b/Flow.Launcher.Core/ExternalPlugins/UserPlugin.cs
@@ -1,4 +1,6 @@
-﻿namespace Flow.Launcher.Core.ExternalPlugins
+﻿using System;
+
+namespace Flow.Launcher.Core.ExternalPlugins
 {
     public record UserPlugin
     {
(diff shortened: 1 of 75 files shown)


taooceros:
That looks great! Do you think we still need the off/on label?

onesounds:
> That looks great! Do you think we still need the off/on label? 

personally I think it don't need it. but I'm sure that MS designer better than me. and same style (windows 11) make feel the fresh & modern software. so... yes. we need it.

onesounds:
reworked in https://github.com/Flow-Launcher/Flow.Launcher/pull/1528

PR #1492, opened 2022-10-19
[Program Plugin] Fix UWP icon missing issue

VictoriousRaptor:
Program plugin can't get icons for some uwp programs because of complex icon filenames or path finding issue. Tried to fix it.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
@@ -603,101 +603,104 @@ namespace Flow.Launcher.Plugin.Program.Programs
 
                 string path = Path.Combine(Package.Location, uri);
 
-                // TODO: NOT sure why we needed this? Maybe for Windows 8?
-                //if (uri.Contains("\\"))
-                //{
-                //    path = Path.Combine(Package.Location, uri);
-                //}
-                //else
-                //{
-                //    // for C:\Windows\MiracastView etc
-                //    path = Path.Combine(Package.Location, "Assets", uri);
-                //}
-
-                var extension = Path.GetExtension(path);
-                if (extension != null)
+                var logoPath = TryToFindLogo(uri, path);
+                if (String.IsNullOrEmpty(logoPath))
                 {
-                    if (File.Exists(path))
+                    // TODO: Don't know why, just keep it at the moment
+                    // Maybe on older version of Windows 10?
+                    // for C:\Windows\MiracastView etc
+                    return TryToFindLogo(uri, Path.Combine(Package.Location, "Assets", uri));
+                }
+                return logoPath;
+
+                string TryToFindLogo(string uri, string path)
+                {
+                    var extension = Path.GetExtension(path);
+                    if (extension != null)
                     {
-                        return path; // shortcut, avoid enumerating files
-                    }
+                        if (File.Exists(path))
+                        {
+                            return path; // shortcut, avoid enumerating files
+                        }
 
-                    var logoNamePrefix = Path.GetFileNameWithoutExtension(uri); // e.g Square44x44
-                    var logoDir = Path.GetDirectoryName(path);  // e.g ..\..\Assets
-                    if (String.IsNullOrEmpty(logoNamePrefix) || String.IsNullOrEmpty(logoDir) || !Directory.Exists(logoDir))
-                    {
-                        // Known issue: Edge always triggers it since logo is not at uri
-                        ProgramLogger.LogException($"|UWP|LogoPathFromUri|{Package.Location}" +
-                           $"|{UserModelId} can't find logo uri for {uri} in package location (logo name or directory not found): {Package.Location}", new FileNotFoundException());
-                        return string.Empty;
-                    }
+                        var logoNamePrefix = Path.GetFileNameWithoutExtension(uri); // e.g Square44x44
+                        var logoDir = Path.GetDirectoryName(path);  // e.g ..\..\Assets
+                        if (String.IsNullOrEmpty(logoNamePrefix) || String.IsNullOrEmpty(logoDir) || !Directory.Exists(logoDir))
+                        {
+                            // Known issue: Edge always triggers it since logo is not at uri
+                            ProgramLogger.LogException($"|UWP|LogoPathFromUri|{Package.Location}" +
+                               $"|{UserModelId} can't find logo uri for {uri} in package location (logo name or directory not found): {Package.Location}", new FileNotFoundException());
+                            return string.Empty;
+                        }
 
-                    var files = Directory.EnumerateFiles(logoDir);
+                        var files = Directory.EnumerateFiles(logoDir);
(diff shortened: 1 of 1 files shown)


Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -601,75 +601,94 @@ internal string LogoPathFromUri(string uri)
...
                 // windows 8 https://msdn.microsoft.com/en-us/library/windows/apps/br211475.aspx
 
-                string path;
-                if (uri.Contains("\\"))
-                {
-                    path = Path.Combine(Package.Location, uri);
-                }
-                else
-                {
-                    // for C:\Windows\MiracastView etc
```
VictoriousRaptor, line 611: Not sure why we needed it. C:\Windows\MiracastView shouldn't be a uwp program right?

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -601,75 +601,94 @@ internal string LogoPathFromUri(string uri)
...
+                    var files = Directory.EnumerateFiles(logoDir);
+
+                    // Just ignore all qulifiers
+                    // select like logo[xxx].png
+                    // https://learn.microsoft.com/en-us/windows/uwp/app-resources/tailor-resources-lang-scale-contrast
+                    var selected = files.FirstOrDefault(file => 
+                        Path.GetFileName(file).StartsWith(logoNamePrefix) && extension == Path.GetExtension(file)
+                    );
+
+                    //var scaleFactors = new Dictionary<PackageVersion, List<int>>
```
taooceros, line 644: why do we remove them?

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
taooceros: Maybe it is in some older version of win10

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
VictoriousRaptor: I can't find any guarantee in the doc that scale factors comes last to the qualifiers (see [here](https://learn.microsoft.com/en-us/windows/uwp/app-resources/images-tailored-for-scale-theme-contrast)). And we don't acutally need scale factor in our plugin.

taooceros: true, let's remove them. We shall fetch all asset information in the manifest file.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
VictoriousRaptor: I can't even find it on my pc.

taooceros: I neither

taooceros: Check whether file exists? That shall be good.

VictoriousRaptor: Check the new commit please.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -601,92 +601,114 @@ internal string LogoPathFromUri(string uri)
...
-                    if (!string.IsNullOrEmpty(selected))
-                    {
-                        return selected;
+                        var files = Directory.EnumerateFiles(logoDir);
+
+                        // Currently we don't care which one to choose
+                        // Just ignore all qualifiers
+                        // select like logo.[xxx_yyy].png
+                        // https://learn.microsoft.com/en-us/windows/uwp/app-resources/tailor-resources-lang-scale-contrast
+                        var selected = files.FirstOrDefault(file =>
```
taooceros, line 642: maybe we should prefer $44\times44$ if possible to reduce image size? Though it's not a big deal if it is hard.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -601,92 +601,114 @@ internal string LogoPathFromUri(string uri)
...
+                            Path.GetFileName(file).StartsWith(logoNamePrefix) && extension == Path.GetExtension(file)
+                        );
+
+                        //var scaleFactors = new Dictionary<PackageVersion, List<int>>
```
taooceros, line 646: Please remove them if not needed.

taooceros (commented):
Check comment and we can merge after that. Could you share an example that the logo won't be loaded so I can give it a test?

VictoriousRaptor:
> Check comment and we can merge after that. Could you share an example that the logo won't be loaded so I can give it a test?

windows security center

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
VictoriousRaptor: Try to find closest to 44x44 in the new commit. It seems to work well.

taooceros:
@VictoriousRaptor What do you think about just use the icoPath property instead of the callback to let ImageLoader loads the image? I think it's no difference than a specific logo (we can adjust the imageloader a bit to load the actual image but currently icon also works I think).

VictoriousRaptor:
> @VictoriousRaptor What do you think about just use the icoPath property instead of the callback to let ImageLoader loads the image? I think it's no difference than a specific logo (we can adjust the imageloader a bit to load the actual image but currently icon also works I think).

Don't quite know how the current callback works. What is the purpose of this change?

taooceros:
> > @VictoriousRaptor What do you think about just use the icoPath property instead of the callback to let ImageLoader loads the image? I think it's no difference than a specific logo (we can adjust the imageloader a bit to load the actual image but currently icon also works I think).
> 
> Don't quite know how the current callback works. What is the purpose of this change?

As you have commented. Let the imageloader loads the icon?

The current callback handles that separately, but I think it is unnecessary since there is a logo image there.

VictoriousRaptor:
> > > @VictoriousRaptor What do you think about just use the icoPath property instead of the callback to let ImageLoader loads the image? I think it's no difference than a specific logo (we can adjust the imageloader a bit to load the actual image but currently icon also works I think).
> > 
> > 
> > Don't quite know how the current callback works. What is the purpose of this change?
> 
> As you have commented. Let the imageloader loads the icon?
> 
> The current callback handles that separately, but I think it is unnecessary since there is a logo image there.

I think it's fine. But current callback make the image plated so we need to beware of that.

taooceros:
yeah I am ok with the current change. Let's leave it to the future~

taooceros (approved):
LGTM! Sorry for the late testing....

PR #1486, opened 2022-10-18
Fix enable/disable/delete button in program plugin settings

VictoriousRaptor:
1. Fix #1356
2. Modified Add/Edit program source window. Allow users enable/disable source from the window.
图片
图片
3. Double click program source item to open edit window.

PR #1496, opened 2022-10-22
[Program Plugin] Fix program UID and other bugs

VictoriousRaptor:
## Changes

1. Currently uid(path in this case) of `ProgramSource` is case sensitve so this can happen, which will cause duplicate results:
图片
Fixed.
2. Fixed an issue that Flow prevents UWP apps from being completely uninstalled. See https://github.com/microsoft/PowerToys/pull/5109.
3. Refactored class `ProgramSource` and removed class `DisabledProgramSource`.
4. Some small optimizations.
5. Merged #1486 because it requires changes in this pr.
6. #1532 

## Tests
- [x] Changes in #1486 and new editing dialogue logic.
- [x] No longer blocks uninstallation of uwp programs.
- [x] Duplicate paths of ProgramSource not allowed. (for both editing and adding)
- [x] UID of win32 programs and custom program sources is case-insensitve (to lowercase when `set`).
- [x] UID of Program Sources in config file can be converted to case-insensitive.
- [x] Improved index speed for custom sources (on my pc).
- [x] Disabling programs in context menu triggers a reindex.


Plugins/Flow.Launcher.Plugin.Program/Views/Models/ProgramSource.cs
```diff
@@ -1,5 +1,91 @@
...
+        /// <param name="location">location of program source</param>
+        /// <param name="enabled">enabled</param>
+        public ProgramSource(string location, bool enabled=true)
+        {
+            Location = location;
+            Enabled = enabled;
+            UniqueIdentifier = location;
+        }
+
+        public ProgramSource(ProgramSource source)
```
taooceros, line 43: is it referencing the same class? Why do we need such a constructor to copy🤔?

VictoriousRaptor: Yes. We need it when copying instances.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -53,21 +52,20 @@ public UWP(Package package)
...
-            AppxPackageHelper _helper = new AppxPackageHelper();
             var path = Path.Combine(Location, "AppxManifest.xml");
 
             var namespaces = XmlNamespaces(path);
             InitPackageVersion(namespaces);
 
             const uint noAttribute = 0x80;
-            const Stgm exclusiveRead = Stgm.Read | Stgm.ShareExclusive;
-            var hResult = SHCreateStreamOnFileEx(path, exclusiveRead, noAttribute, false, null, out IStream stream);
+            const Stgm nonExclusiveRead = Stgm.Read | Stgm.ShareDenyNone;
```
taooceros, line 61: I see the pr you referenced completely remove the share option. Is that the same as `Stgm.ShareDenyNone`?

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -203,16 +198,7 @@ private void DisableProgram(IProgram programToDelete)
                 _win32s.FirstOrDefault(x => x.UniqueIdentifier == programToDelete.UniqueIdentifier)
                     .Enabled = false;
```
taooceros, line 199: please append a `?` here for nullreference since we are using `FirstOrDefault`. Though, shall we use `First` instead?🤔

Plugins/Flow.Launcher.Plugin.Program/Views/Models/ProgramSource.cs
```diff
@@ -1,5 +1,91 @@
...
+    /// <para>Custom user added program sources set UniqueIdentifier using their location</para>
+    /// </remarks>
+    public class ProgramSource
+    {
+        private string name;
+
+        public string Location { get; set; }
+        public string Name { get => name ?? new DirectoryInfo(Location).Name; set => name = value; }
+        public bool Enabled { get; set; } = true;
+        private string uid { get; set; }
```
taooceros, line 22: please make this a field.

Plugins/Flow.Launcher.Plugin.Program/Views/Models/ProgramSource.cs
```diff
@@ -1,5 +1,91 @@
...
+            return program != null && program.UniqueIdentifier == this.UniqueIdentifier;
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(uid);
+        }
+    }
+
+    public class DisabledProgramSource : ProgramSource
```
taooceros, line 75: I don't understand why we need an inheritance here....It only changes the mutable Property.....Could you explain your idea here?

Further, the property `Enabled` is public, so this subclass doesn't restrict people to change `Enabled` to `true`.

Plugins/Flow.Launcher.Plugin.Program/Views/ProgramSetting.xaml.cs
```diff
@@ -199,13 +199,9 @@ private void programSourceView_Drop(object sender, DragEventArgs e)
             {
                 foreach (string directory in directories)
                 {
-                    if (Directory.Exists(directory) && !ProgramSettingDisplayList.Any(x => x.UniqueIdentifier == directory))
+                    if (Directory.Exists(directory) && !ProgramSettingDisplayList.Any(x => x.UniqueIdentifier == directory.ToLowerInvariant()))
```
taooceros, line 202: Please use `string.Equal` with the option Ignore case instead.

Plugins/Flow.Launcher.Plugin.Program/Views/Commands/ProgramSettingDisplay.cs
```diff
@@ -9,20 +9,11 @@ namespace Flow.Launcher.Plugin.Program.Views.Commands
 {
     internal static class ProgramSettingDisplay
     {
-        internal static List<ProgramSource> LoadProgramSources(this List<Settings.ProgramSource> programSources)
+        internal static List<ProgramSource> LoadProgramSources(this List<ProgramSource> programSources)
```
taooceros, line 12: I don't think we need to copy the programsource🤔. I would suggest whether it is possible to do a bit more refactor by using `UnionBy`, but it's ok to leave it like this.

taooceros (commented):
Overall it is nice. Thank you so much.

Plugins/Flow.Launcher.Plugin.Program/Views/Models/ProgramSource.cs
VictoriousRaptor: I don't quite understand either. It's not my design though, I just keep it. Maybe the original purpose was not to mess up the disabled and the enabled.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
VictoriousRaptor: Yes, it's the default value.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
VictoriousRaptor: Use `First` instead. We've checked it exists.

Plugins/Flow.Launcher.Plugin.Program/Views/Commands/ProgramSettingDisplay.cs
VictoriousRaptor: > I don't think we need to copy the programsource

Yep, seems like we don't need to do a deep copy.

Plugins/Flow.Launcher.Plugin.Program/Views/Models/ProgramSource.cs
VictoriousRaptor: After some research I think `DisabledProgramSource` is only for a disabled SINGLE program. If you disable a single program from context menu, a new instance is added to the list.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -696,9 +686,14 @@ public ImageSource Logo()
...
             {
+                // TODO: Consider using infrastructure.image.imageloader?
                 if (File.Exists(path))
                 {
-                    var image = new BitmapImage(new Uri(path));
+                    var image = new BitmapImage();
+                    image.BeginInit();
+                    image.UriSource =  new Uri(path);
+                    image.CacheOption = BitmapCacheOption.OnLoad;
+                    image.EndInit();
```
taooceros, line 696: please freeze it.

taooceros (commented):
Overall thank you so much for doing the refactor! I will give everything a test recently (till now I just review the code). This part of code is really confusing and definitely should be refactor.

Plugins/Flow.Launcher.Plugin.Program/Views/Models/ProgramSource.cs
VictoriousRaptor: So originally it copies a instance when loading program sources from settings and copying them to the display list. It's been removed now. The only usage now is copying disabled sources to the displaylist (another problem though).

Plugins/Flow.Launcher.Plugin.Program/Views/Commands/ProgramSettingDisplay.cs
```diff
@@ -9,75 +9,30 @@ namespace Flow.Launcher.Plugin.Program.Views.Commands
...
-                                                         Enabled = t1.Enabled
-                                                     }
-                                              ));
+                .ForEach(t1 => ProgramSetting.ProgramSettingDisplayList.Add(new ProgramSource(t1)));
         }
 
-        internal static void SetProgramSourcesStatus(this List<ProgramSource> list, List<ProgramSource> selectedProgramSourcesToDisable, bool status)
+        internal static void SetProgramSourcesStatus(List<ProgramSource> selectedProgramSourcesToDisable, bool status)
         {
             ProgramSetting.ProgramSettingDisplayList
```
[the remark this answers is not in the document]
VictoriousRaptor: > we don't need to create a new list.

Can you explain it further?

Plugins/Flow.Launcher.Plugin.Program/Views/Models/ProgramSource.cs
VictoriousRaptor: Already removed `DisabledProgramSource`. Using `ProgramSource` instead.

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -381,6 +381,7 @@ private static string Extension(string path)
         private static IEnumerable<Win32> UnregisteredPrograms(List<ProgramSource> sources, string[] suffixes, string[] protocols)
         {
             var paths = ExceptDisabledSource(sources.Where(s => Directory.Exists(s.Location) && s.Enabled)
```
taooceros, line 374: Have you tested whether parallel give performance boost? In my very early test it is not so significant somehow

Plugins/Flow.Launcher.Plugin.Program/FileChangeWatcher.cs
```diff
@@ -1,58 +0,0 @@
-﻿using System.Collections.Generic;
-using System.IO;
-using System.Threading.Tasks;
-using Flow.Launcher.Infrastructure.Logger;
-using Flow.Launcher.Plugin.Program.Programs;
-
-namespace Flow.Launcher.Plugin.Program
-{
-    //internal static class FileChangeWatcher
```
taooceros, line 9: why is this deleted sorry?

VictoriousRaptor: It's unused. The whole class is commented long ago.

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
VictoriousRaptor: On my computer it is about 30% faster. I will do some more tests and show you the results.

VictoriousRaptor: Parallel gives up to 30% boost on my computer. Not always a significant boost but generally faster.

jjw24:
@VictoriousRaptor please resolve conflict.

jjw24:
@VictoriousRaptor please update description on what you have tested with this change.

VictoriousRaptor:
> @VictoriousRaptor please update description on what you have tested with this change.

Updated. Consider a version bump of the plugin?

jjw24:
Check telhe release pr, I think this plugin already has a feature bump already.

jjw24 (approved):
💯 👍 👍

PR #1501, opened 2022-10-27
Navigation Rework

onesounds:


onesounds:
- This task was initiated to solve memory-related problems, but was resolved in a different way.
- This task must be done some day. because it involves separating the code in the setting window,
- but after discussion with Tao, we decided to we will work when we changing avalonia framework.

PR #1509, opened 2022-11-02
Enable non-C# plugins to use StringMatcher's scoring

Garulf:
Provides a new field called `FuzzyMatchString` that will override a results score with Flow's `StringMatcher.FuzzSearch` by matching the current query with the new field. 

Otherwise the field is `null` and ignored.

- [x] Fuzzymatch results
- [x] Hide results tht dont meet cutoff
- [x] Add text highlighting based on match?
- [x] Fix scores not applying if query text is the same

### DIFF
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -111,6 +111,11 @@ namespace Flow.Launcher.Plugin
         public int Score { get; set; }
 
         /// <summary>
+        /// String to use in fuzzy score matching
+        /// </summary>
+        public string FuzzyMatchString { get; set; }
+
+        /// <summary>
         /// A list of indexes for the characters to be highlighted in Title
         /// </summary>
         public IList<int> TitleHighlightData { get; set; }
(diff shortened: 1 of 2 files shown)


taooceros:
I have a concern on this. What if the fuzzysearchtext is not equal to the title? Then the highlight may be weird.

jjw24:
Sperate conversation - @taooceros is it possible to do a callback to the plugin? E.g. python plugin -> json callback object with required data i.e. search string and method to call -> flow runs the API method -> result returned back to the plugin. Do you have something like this in the V2 JsonRPC design?

taooceros:
> Separate conversation - @taooceros is it possible to do a callback to the plugin? E.g. python plugin -> json callback object with required data i.e. search string and method to call -> flow runs the API method -> result returned back to the plugin. Do you have something like this in the V2 JsonRPC design?

that's in the original plan for V2.

jjw24:
@taooceros we need to properly take a look at this in relation to the V2 pr, maybe this is better done there.

taooceros:
> @taooceros we need to properly take a look at this in relation to the V2 pr, maybe this is better done there.

Well I should say yes, but it really depends on how to design the api. Exposing a string matching field is a much simpler way for plugins to use the fuzzy api (I feel like we can just match title so highligh works).

jjw24:
Marking draft. We will enable StringMatcher to be used by non-dotnet plugins via v2 JsonRPC

PR #1510, opened 2022-11-02
Fix memory leaks with Plugin store layout

onesounds:
## What's the PR

image
- Decode Image Size in Theme Preview 
  - Reason : if user use big size wallpaper, it will effect to flow. size decode will solve the problem.
  - I don't think it's a big problem, although the image drops to low definition.
- Use Virtualizing Panels in Plugin Store List / plugin list / ETC for memory leak
- Refactor Plugin Store UI (Changed Wrap panel from Uniform grid.)
  - Reason : in this case, item size doesn't fit with list width. Several implementations were attempted, but this condition was judged to be good because performance problems occurred and design problems were not significant.
- 'Update Label' (Green Ball) changed placement to icon's right.
    - Reason : Due to implementation problems, it could not be displayed in the upper right corner.

## Test Cases
- [ ] No excessive memory leaks should occurrence with all menu.
- [ ] Wallpaper displayed on the theme menu should be displayed in low resolution. (It has to be a level that doesn't bother you.)
- [ ] Items in the plug-in store must be displayed appropriately even if you resize the window
- [ ] The menu displayed when each item is clicked must operate normally.
- [ ]  Group (New, Installed....) and searches must operate normally.

### DIFF
--- a/Flow.Launcher/SettingWindow.xaml
+++ b/Flow.Launcher/SettingWindow.xaml
@@ -607,10 +607,10 @@
                     <ScrollViewer
                         Margin="0,0,0,0"
                         Background="{DynamicResource Color01B}"
-                        CanContentScroll="True"
+                        CanContentScroll="False"
                         VirtualizingPanel.ScrollUnit="Pixel"
                         VirtualizingStackPanel.IsVirtualizing="True">
-                        <StackPanel Margin="5,18,25,30" Orientation="Vertical">
+                        <VirtualizingStackPanel Margin="5,18,25,30" Orientation="Vertical">
                             <TextBlock
                                 Grid.Row="2"
                                 Margin="0,5,0,5"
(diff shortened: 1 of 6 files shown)


onesounds:
@taooceros 'refresh' button wrong work after merged Label PR , It seems like a simple task, but I couldn't fix it. Please add it on this pr or work on it separately. (Don't forget it!)

jjw24:
@onesounds Are the test cases completed?

jjw24 (commented):
Installing plugins via mouse click is not working. Click on the install button then use mouse to click the install plugin result, nothing happens. Using enter works.

onesounds:
> Installing plugins via mouse click is not working. Click on the install button then use mouse to click the install plugin result, nothing happens. Using enter works.

ok. I'll check. when fix done, I'll mention you again. thanks.

Flow.Launcher/SettingWindow.xaml
```diff
@@ -607,7 +607,8 @@
                     <ScrollViewer
                         Margin="0,0,0,0"
                         Background="{DynamicResource Color01B}"
-                        ScrollViewer.CanContentScroll="False">
+                        ScrollViewer.CanContentScroll="False"
```
taooceros, line 610: You will need to set CanContentScroll to true to enable virtualization. I will handle that for you

jjw24 (approved):

PR #1511, opened 2022-11-02
[DEV][Plugins Manager] Remove await in updating plugin manifest to speedup startup

VictoriousRaptor:
In dev branch, the init method of `PluginsManager` awaits the result of updating plugin manifest. With a slow internet connection (like public wifi) it could significantly slow down startup. The manifest can be updated later (since we can use Flow without internet).

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
@@ -37,7 +37,7 @@ namespace Flow.Launcher.Plugin.PluginsManager
             contextMenu = new ContextMenu(Context);
             pluginManager = new PluginsManager(Context, Settings);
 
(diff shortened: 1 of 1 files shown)


taooceros (approved):

VictoriousRaptor:
> @VictoriousRaptor what is displayed if you type pm install or go to plugin store while the plugin list is still loading?

nothing is shown. when typing pm install the progress bar is moving. wait a while it fetches the manifest and show available plugins.

jjw24:
Yeah ok. It might be rare that the manifest has not finished being fetched after user finishes typing pm install right?

VictoriousRaptor:
> Yeah ok. It might be rare that the manifest has not finished being fetched after user finishes typing pm install right?

Yes. And even if, it will automatically fetch one so I think it's fine. As for the store user migh need to manually click the refresh button.

PR #1514, opened 2022-11-03
[Dev] Fix unwanted URL protocols and invalid programs in custom sources

VictoriousRaptor:
Old source code doesn't check if an indexed program is marked as `Valid`. So .url files with http protocols (disabled by default) can be shown in result. Fixed.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
@@ -388,8 +388,7 @@ namespace Flow.Launcher.Plugin.Program.Programs
                 ShortcutExtension => LnkProgram(x),
                 UrlExtension => UrlProgram(x),
                 _ => Win32Program(x)
-            });
-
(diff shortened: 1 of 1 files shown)


Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -388,8 +388,7 @@ private static IEnumerable<Win32> UnregisteredPrograms(List<Settings.ProgramSour
                 ShortcutExtension => LnkProgram(x),
                 UrlExtension => UrlProgram(x),
                 _ => Win32Program(x)
-            });
-
+            }).Where(x => x.Valid);
```
taooceros, line 391: I think we can do a single check at the end after merging?

VictoriousRaptor: Sure. That might be better.

taooceros (approved):

PR #1516, opened 2022-11-04
Score adjustment for a match that is closer to the start of string

jjw24:
### Context
Currently scoring is the same regardless of the position of the matched string, e.g. 'hello world' and 'world hello' have the same score. This change adds an adjustment during score calculation to allow 'world hello' to score higher than 'hello world' when query is 'world'.

### How
The conditions for this adjustment to kick in is:
- the matched string has to be immediately preceding a space, and
- query must be a fully contained substring of the compare string.

Score adjustment is done by:
Counting the number of spaces from the start of the string and deduct this from the score. So when matching 'world', 'world hello' has 0 spaces, so nothing is deducted from the score. On the other hand 'crazy on fire world' has three spaces so deduct three from the score.

### Some pretty pictures
Before, notice 'Graphics Settings' should be higher:
image
 
After:
image

### What's tested
- Added new unit test covering several scenarios
- Manually tested some of the results via running the launcher
- Checked the score adjustment does not impact greatly on other non-relevant results

### DIFF
--- a/Flow.Launcher.Infrastructure/StringMatcher.cs
+++ b/Flow.Launcher.Infrastructure/StringMatcher.cs
@@ -202,7 +202,11 @@ namespace Flow.Launcher.Infrastructure
             if (allQuerySubstringsMatched)
             {
                 var nearestSpaceIndex = CalculateClosestSpaceIndex(spaceIndices, firstMatchIndex);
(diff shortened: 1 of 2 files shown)


onesounds (approved):

Garulf:
Wow amazingly fast! I have to try this first chance I get. Ty

PR #1529, opened 2022-11-11
Appveyor test executable unfound fix and add caching

taooceros:
- Recently CI fails because some weird issue that it detects the Flow.Launcher.Test.dll in path that contains an extra parent of `Any CPU`. This pr try to manually run the test to avoid that.

- Also let the package run after test step is done

- Added caching config to let the builds done faster

PR #1530, opened 2022-11-12
[Dev] Save Setting Window State (full / normal)

onesounds:
## what's the PR
- Before : When you close the setting window in full screen state, it does not go into full screen state when you reopen it.
- After : When you close the window in full screen state, it also opens to the full screen when you reopen it.

## Test Cases
- [x] When you close and open the window, it should remain in the window state.
- [x] Window state should be maintained during program shutdown and rerun.
- [x] The window position should not change.


jjw24 (approved):

jjw24:
Looks like CI failure as mentioned by tao in #1529

PR #1479, opened 2022-10-17
[Program Plugin] Search programs in PATH

VictoriousRaptor:
Resolves #116

## Changes

1. Add a option to allow Program plugin to search programs in PATH environment variable. Disabled by default (should it enabled by default?). And searching in PATH is not recursive because PATH itself is not recursive.
图片

## Tests

- [x] Works with #1476 
- [x] Can search programs (lnk, exe, ...) in PATH

PR #1546, opened 2022-11-17
[Dev][Program Plugin] Optimize program indexing performance 

VictoriousRaptor:
This is a optimization of #1479. PATH can be a subfolder of custom program sources. It makes unnecessary I/O when enabled PATH indexing because indexing for custom sources is recursive. And a custom source can be a subfolder of another.

And let file change watcher only watch files that can be indexed (`exe`, `lnk', ...).

## Changes

1. Refactored some code for readability.
3. Detect if PATH is a subfolder of a custom source.
4. Detect if a custom source is a subfolder of another.
5. Only watch specific file extensions.
6. Fix an issue where trying to open a .lnk files parent folder, it acutally opens its target's folder.

## Tests

- [x] Programs in PATH can be indexed correctly.
- [x] Programs in custom sources can be indexed correctly.
- [x] Unnecessary IO is avoided.
- [x] Creating/Deleting a file with specifix extension triggers a reindex.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
+++ b/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
@@ -61,7 +61,7 @@
         Insert file suffixes you want to index. Suffixes should be separated by ';'. (ex>bat;py)
     </system:String>
     <system:String x:Key="flowlauncher_plugin_program_protocol_tooltip">
-        Insert protocols of .url files you want to index. Protocols should be separated by ';'. (ex>ftp;netflix)
+        Insert protocols of .url files you want to index. Protocols should be separated by ';', and should end with "//". (ex>ftp://;mailto://)
     </system:String>
 
     <system:String x:Key="flowlauncher_plugin_program_run_as_different_user">Run As Different User</system:String>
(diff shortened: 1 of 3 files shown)


Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
```diff
@@ -61,7 +61,7 @@
         Insert file suffixes you want to index. Suffixes should be separated by ';'. (ex>bat;py)
     </system:String>
     <system:String x:Key="flowlauncher_plugin_program_protocol_tooltip">
-        Insert protocols of .url files you want to index. Protocols should be separated by ';'. (ex>ftp;netflix)
+        Insert protocols of .url files you want to index. Protocols should be separated by ';', and should end with "//". (ex>ftp://;mailto://)
```
VictoriousRaptor, line 64: Don't want really parse the URL. So try to tell users how to avoid situaions like `http` actually matches a URL starts with `https`.

taooceros: why not parse it? Uri.TryParse is not expensive?

jjw24 (approved):

Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
VictoriousRaptor: Because for steam games their URL scheme is `steam://app/` rather than `steam://` so parsing the URL for protocol is not enough.

Issue #1522, opened 2022-11-09
BUG: Graphical interface FL 1.10.0 (November 08) Spanish Language


dracp:
**FL 1.10 (11.08) / SETTINGS/HOTKEY**
"Built-in Shortcuts", not found in Crowdin.
Test image with file "es.xml" downloaded from Crowdin.
(...\221108 FlowLauncher Beta Portable\app-1.9.5\Languages\es.xaml)
![221109 FL 1 10 0 (11 08)_Built-in Shortcuts_No se encuentra en Crowdin](https://user-images.githubusercontent.com/102802425/200807172-335aec12-1ece-486b-b423-73f9bd775843.png)

PR #1548, opened 2022-11-17
Fix Narrow Items & Adjust Context Menu Design

onesounds:
## What's the PR
- There is a size problem in some languages. The size of some fixed controls has been changed flexibly.
- I had to use the context menu while doing this work. Therefore, the template of the context menu was also modified.
- As the context menu changed, some color issues happened, so also corrected.
- If the version is 1.0.0, it is marked as Dev. (I think it's unnecessary now...)
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/1522

## Check Lists
- You should check these with Slovak / Spanish / Russian language mode.
- with Minimun Window Width
- The color should be displayed correctly on other context menus and flyouts.

**Before**

**After**

**Before**

**After** (Adjust wrapping)

**Before** (Problem with some languages)

**After** (Put items into one button)

**Before**

**After** (Merge with update section / add dev label) 

**Before**

**After** (it should display well dark/light mode)

- Change Hardcoded text ("press key) to string / Fix Default Color (before doesn't work with theme)

**Before**

**After**
- it should works well with light/dark mode.
- Icon should works with isEnabled

**Before**

**After**

### DIFF
--- a/Flow.Launcher/Flow.Launcher.csproj
+++ b/Flow.Launcher/Flow.Launcher.csproj
@@ -89,7 +89,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="InputSimulator" Version="1.0.4" />
-    <PackageReference Include="Microsoft.Toolkit.Uwp.Notifications" Version="7.1.2" />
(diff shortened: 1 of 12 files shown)


kubalav:
Cutted off…
Explorer plugin:
obrázok_2022-11-17_184712496

onesounds:
> Cutted off… Explorer plugin
Fixed.

VictoriousRaptor (approved):

PR #1562, opened 2022-11-21
[Dev] Use path of lnk file when launching proram

VictoriousRaptor:
#1546 changed the behavior of launching .lnk programs. It directly launches the program pointed to. It wasn't intended. Got mislead by the varaible name.) which brings an issue. Start Menu shortcut of Discord can't be launched. This is because it acutally points to `blabla\Discord\Update.exe --processStart Discord.exe` and we don't parse the params. Fixed in this pr.

Known issue of program plugin (not related to this pr or 1546): can't distinguish different .lnk files pointing to the same target but with different params.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
@@ -220,8 +220,7 @@ namespace Flow.Launcher.Plugin.Program.Programs
             return Name;
         }
 
-        public static List<FileSystemWatcher> Watchers = new List<FileSystemWatcher>();
-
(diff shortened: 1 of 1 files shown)


jjw24:
>Known issue of program plugin (not related to this pr or 1546): can't distinguish different .lnk files pointing to the same target but with different params.

Could you please create an issue if not already done so.

jjw24 (approved):
👍

VictoriousRaptor:
> > Known issue of program plugin (not related to this pr or 1546): can't distinguish different .lnk files pointing to the same target but with different params.
> 
> Could you please create an issue if not already done so.

Already working on it. I will push the fix later.

jjw24:
Legend

PR #1567, opened 2022-11-24
Update wording

VictoriousRaptor:
Update wording.

### DIFF
--- a/Flow.Launcher.Core/Resource/Internationalization.cs
+++ b/Flow.Launcher.Core/Resource/Internationalization.cs
@@ -115,7 +115,11 @@ namespace Flow.Launcher.Core.Resource
             if (languageToSet != AvailableLanguages.Chinese && languageToSet != AvailableLanguages.Chinese_TW)
                 return false;
 
-            if (MessageBox.Show("Do you want to turn on search with Pinyin?", string.Empty, MessageBoxButton.YesNo) == MessageBoxResult.No)
+            // No other languages should show the following text so just make it hard-coded
+            // "Do you want to search with pinyin?"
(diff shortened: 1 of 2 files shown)


onesounds (approved):

Flow.Launcher.Core/Resource/Internationalization.cs
```diff
@@ -115,7 +115,11 @@ public bool PromptShouldUsePinyin(string languageCodeToSet)
             if (languageToSet != AvailableLanguages.Chinese && languageToSet != AvailableLanguages.Chinese_TW)
                 return false;
 
-            if (MessageBox.Show("Do you want to turn on search with Pinyin?", string.Empty, MessageBoxButton.YesNo) == MessageBoxResult.No)
+            // No other languages should show the following text so just make it hard-coded
+            // "Do you want to search with pinyin?"
+            string text = languageToSet == AvailableLanguages.Chinese ? "是否启用拼音搜索？" : "是否啓用拼音搜索？" ;
```
jjw24, line 120: This is going to show Chinese when someone is using non-chinese and accidentally click on the toggle or is this option is hidden unless Chinese lang is selected? Can't remember exactly. I still think it's better to put in xaml for consistency.

VictoriousRaptor: This only shows when someone select Chinese in settings. And this message pops up AFTER language change so technically only Chinese S/T is needed. I think there is no need to translate a string that will never be displayed in other languages.

onesounds:
memo :  I think this window should display both language in same time.
(For example, If US language user mistake click Chinese language, They can't understand what window open) I'll adjust the window next time.

PR #1581, opened 2022-11-28
Refactor UWP class in Program plugin

VictoriousRaptor:
## Changes

- Refator `UWP` and `UWP.Application`.
- Use WinRT API and mannually parse appmanifest.xml to get info of UWP Apps to avoid interops. (Thanks to PT Run and Wox)
- Use `ImageLoader` to load icons.
- Remove `ApplicationActivationHelper` interop.\
- Remove UWP icon background color.

## Tests

- [x] Apps should be correctly indexed.
- [x] Apps should be launched either elevated or unelevated correctly, base on user choice and `CanRunElevated` specified in manifest.
- [x] Icons should be shown correctly.
- [x] logging for exceptions while indexing apps.
- [x] Index performance is generally unaffected.
- [x] Different apps in the same package (like email and calendar) shows different if icons.

## Todos

- [x] Preview images (still require changes in Preview pr)
- [x] Clear unused code.


Flow.Launcher.Test/Plugins/ProgramTest.cs
```diff
@@ -8,23 +8,23 @@ namespace Flow.Launcher.Test.Plugins
     [TestFixture]
     public class ProgramTest
     {
-        [TestCase("Microsoft.WindowsCamera", "ms-resource:LensSDK/Resources/AppTitle", "ms-resource://Microsoft.WindowsCamera/LensSDK/Resources/AppTitle")]
```
jjw24, line 11: why is this entire class commented out?

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
```diff
@@ -651,111 +577,112 @@ string TryToFindLogo(string uri, string path)
...
-                }
-                else
-                {
-                    // todo use windows theme as background
-                    return image;
-                }
-            }
-
+            #region logo legacy 
+            // preserve for potential future use
```
jjw24, line 597: what can we use this for in the future

Flow.Launcher.Test/Plugins/ProgramTest.cs
VictoriousRaptor: This test is for translating strings like `ms-resource://Microsoft.WindowsCamera/LensSDK/Resources/AppTitle` to the acutal title `Camera`, using interop. We retrieve titles from winrt api now so no longer needed.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
VictoriousRaptor: this part fills the background of the icon with background color in appmanifest.Technically background + color is the full icon that app developer want to show. onesounds removed it in preview pr. I move it here to avoid conflict. Still I feel like maybe we may need to look into it later so I just commented it.

jjw24: Ok sounds good

Flow.Launcher.Test/Plugins/ProgramTest.cs
jjw24: If not needed let's delete it?

VictoriousRaptor: deleted

jjw24 (approved):

PR #1600, opened 2022-12-06
[Dev] Refactor AddProgramSource dialog and bugfix

VictoriousRaptor:
## Changes
- Refactored logic of `AddProgramSource` dialog.
- For win32 and uwp `ProgramSource` (single programs disabled in results by users), Directory textbox is readonly and the Browse button is hidden. Users shouldn't change its location mannually since it's indexed by plugin, or it may lead to unexpected behavior.
图片

- Fixed an issue that editing dir of the selected ProgramSource causes the selected item of ProgramSource listview can't be unselected.
- Removed unnecessary `LocationConverter`.

## Tests

- [x] Directory textbox is readonly and the Browse button is hidden for uwp and win32. Custom sources are unaffected.
- [x] Can Enable/Disable Program Sources.
- [x] Can Edit Location of Program Sources.
- [x] Items can be unselected after editing directory.
- [x] Location and Enabled column in listview are displayed correctly, after editing.


Plugins/Flow.Launcher.Plugin.Program/ViewModels/AddProgramSourceViewModel.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            Source = programSource;
+            enabled = Source.Enabled;
+            location = Source.Location;
+            AddBtnText = API.GetTranslation("flowlauncher_plugin_program_update");
+            IsCustomSource = Settings.ProgramSources.Any(x => x.UniqueIdentifier == Source.UniqueIdentifier);
+
+            this.PropertyChanged += (_, args) =>
+            {
+                switch (args.PropertyName)
```
taooceros, line 61: Well...I would suggest you modify the getter/setter instead...

Plugins/Flow.Launcher.Plugin.Program/ViewModels/AddProgramSourceViewModel.cs
```diff
@@ -0,0 +1,143 @@
...
+            else
+            {
+                var source = new ProgramSource(Location, Enabled);
+                Settings.ProgramSources.Insert(0, source);
+                ProgramSetting.ProgramSettingDisplayList.Add(source);
+                return (true, null);
+            }
+        }
+
+        public (bool, string) UpdateProgramSource()
```
taooceros, line 102: what do you think about using a Try style?
like `public bool TryUpdateProgramSource(out string newLocation)`

taooceros: Nevermind it seems like that style is used mostly when we expect to throw exception...

Plugins/Flow.Launcher.Plugin.Program/AddProgramSource.xaml.cs
```diff
@@ -55,51 +29,14 @@ private void BtnCancel_OnClick(object sender, RoutedEventArgs e)
 
         private void BtnAdd_OnClick(object sender, RoutedEventArgs e)
         {
-            string path = Directory.Text;
-            bool modified = false;
-            if (!System.IO.Directory.Exists(path))
+            var status = ViewModel.AddOrUpdate();
+            bool modified = status.Item1;
+            string msg = status.Item2;
```
taooceros, line 34: Let's use the auto destructor

jjw24:
I noticed:

1. The Program Source window does not show the path selecting the path from the browse button.
2. Double clicking on the headers 'Name', 'Status' or 'Location' brings up the 'Please select a program source' message window.

taooceros (dismissed):

taooceros:
Sorry I think @jjw24 is right...Let's address those problem

taooceros:
I fix the first one. The second one may require quite a bit refactor to fit more mvvm scheme.

jjw24 (approved):

Issue #1584, opened 2022-11-28
BUG: Win+R still call the system "run" out


Darkluna999:
### Checks

- [x] I have checked that this issue has not already been reported.

### Problem Description

1. Sometimes after I input `Win+R` it will have a short flash of Flow Launcher then the window of system `run` will come out.
2. And will you supply an option of `windows terminal` to run the code? powershell is not beautiful.

### Flow Launcher Version

v1.9.5

### Windows Build Number
dev Bulid 25247

Newedge-Isaiah:
Here’s a workaround:

1. Open the Registry Editor (regedit)
2. Navigate to: “HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced”

4. Right click on the Advanced folder on the left pane and select New -> String Value
5. Name it “DisabledHotkeys”
6. Set its value to “R”. This will disable the default Windows Key + R shortcut that launches Windows Run.

onesounds:
Reproducable. No problem with the old version. I think the code related to the shortcut has changed.
This issue needs to be fixed. @taooceros @jjw24 

related https://github.com/Flow-Launcher/Flow.Launcher/issues/1242

Darkluna999:
> 

sorry, it does not work, and I think it need be fixed by the developing of this software but not in the regedit

Darkluna999:
> Here’s a workaround:
> 
> 1. Open the Registry Editor (regedit)
> 2. Navigate to: “HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced”
> 3. Right click on the Advanced folder on the left pane and select New -> String Value
> 4. Name it “DisabledHotkeys”
> 5. Set its value to “R”. This will disable the default Windows Key + R shortcut that launches Windows Run.

sorry, it does not work, and I think it need be fixed by the developing of this software but not in the regedit

Newedge-Isaiah:
> > Here’s a workaround:
> > 
> > 1. Open the Registry Editor (regedit)
> > 2. Navigate to: “HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced”
> > 3. Right click on the Advanced folder on the left pane and select New -> String Value
> > 4. Name it “DisabledHotkeys”
> > 5. Set its value to “R”. This will disable the default Windows Key + R shortcut that launches Windows Run.
> 
[...]

PR #1606, opened 2022-12-08
[Dev] Fix shell win+r too slow to appear/does not appear when triggered 

taooceros:
Shell plugin delegate of win+r is too slow to be a valid hook when triggered (we should largely adjust that part of code later), this pr add a Task.Run to run the slow part.

Also the Show method is not safe when triggered by threadpool, so the temporary fix is to wrap a Dispatcher.Invoke around it.

Also potentially fix sound effect sometimes disappear due to threadpool thread execute show method.

Test
- [x] win+r works
- [x] The window is at the right position
- [x] Sound effect preserved

To reproduce an exception for the sound effect
1. Use a Task.Run wrap a `IPublicAPI.ShowMainWindow()`
2. Then you may notice that the sound disappear or an exception throw without the fix

Close https://github.com/Flow-Launcher/Flow.Launcher/issues/1584


taooceros:
@gissehel may you give this pr a test? I think we may not need the dispatcher priority if execute everything in the main thread.

PR #1673, opened 2022-12-18
fix Constant.PythonPath not being set

taooceros:
fix Constant.PythonPath not being set when pythonpath exists in setting file.


jjw24:
will be fixed in https://github.com/Flow-Launcher/Flow.Launcher/pull/1498

PR #1677, opened 2022-12-18
Search delay

taooceros:


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -689,19 +679,16 @@ private async void QueryResults()
...
-                false => QueryTask(plugin),
-                true => Task.CompletedTask
-            }).ToArray();
-
-
             try
             {
-                // Check the code, WhenAll will translate all type of IEnumerable or Collection to Array, so make an array at first
-                await Task.WhenAll(tasks);
+                await Parallel.ForEachAsync(plugins, currentCancellationToken,
```
taooceros, line 684: This is for a better concurrency so we don't really need to do a context switch

jjw24:
> If delay is defined in the plug-in, the plug-in delay is applied.

Is this tested?

taooceros:
> > If delay is defined in the plug-in, the plug-in delay is applied.
> 
> Is this tested?

I test it

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -770,19 +760,16 @@ private async void QueryResults()
...
+                    async (pair, token) =>
+                    {
+                        await Task.Delay(pair.Metadata.SearchDelay ?? Settings.SearchDelay, token);
+                        if (pair.Metadata.Disabled)
+                            return;
+                        await QueryTask(pair, token);
```
VictoriousRaptor, line 771: Should we check `pair.Metadata.Disabled` before delay?

taooceros: I agree we should, Though I don't think it's a really big deal.

jjw24:
I am running into issue with this sort of result: 

image

Is it possible to have no delay?

jjw24:
Correct me if I am wrong PRD has no query delay at all right?

taooceros:
> I am running into issue with this sort of result on the default delay:
> 
> image
> 
> Is it possible to have no delay?

I don't think this makes sense. How fast do you type?

taooceros:
> Another one I am finding is using the default delay if you type `dddddddddddddddddddddddddddd` then delete it one by one, the trailing query suggestion text makes it feel really slow, like there is a bit of lag almost. On a prd release the trailing query suggestion text is removed a lot faster.

could you try run without debugger? the debugger makes thing much slower

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -738,36 +783,36 @@ private async void QueryResults()
...
                     {
-                        await Task.Delay(pair.Metadata.SearchDelay ?? Settings.SearchDelay, token);
-                        if (pair.Metadata.Disabled)
+                        if(token.IsCancellationRequested)
+                            return;
+
+                        await Task.Delay(pair.Metadata.SearchDelay
+                                         ?? (string.IsNullOrEmpty(query.ActionKeyword)
+                                             ? Settings.GlobalSearchDelay
+                                             : 0));
```
VictoriousRaptor, line 811: Is it possible for plugins to notify that should use user specified delay value?

VictoriousRaptor: nvm

VictoriousRaptor: And why use delay only for global searches?

VictoriousRaptor:
Should we delay when typing the FIRST character?

taooceros:
you mean this?
image

taooceros:
What about removing the autosuggestion whenever query change?

taooceros:
> Should we delay when typing the FIRST character?

huh maybe not

Jack251970:
Closed for #1677.

Issue #1678, opened 2022-12-18
Enhance the 'Clock' feature to show seconds in the time.


z1nc0r3:
**Is your feature request related to a problem? Please describe.**

- There is no certain way to view seconds in Windows 11 clock.

**Describe the solution you'd like**

- Since there is a new feature called 'Clock' in the launcher, it'll be good to add an enhancement to show seconds also. Just add the hh:mm:ss format to the list.

**Describe alternatives you've considered**

- No, alternatives.

**Additional context**

- No additional context.

onesounds:
We can add seconds, but there may be some little memory problems. (It won't be a big problem.) It's not a feature with many users, so I'll just add the format hh:mm:ss.

PR #1680, opened 2022-12-18
Add "hh:mm:ss tt" to clock formats.

z1nc0r3:
#1678 Added hh:mm:ss tt to the clock formats to show seconds in the clock.

### DIFF
--- a/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
+++ b/Flow.Launcher/ViewModel/SettingWindowViewModel.cs
@@ -479,7 +479,7 @@ namespace Flow.Launcher.ViewModel
             "h:mm tt",
             "hh:mm tt",
             "hh:mm:ss tt",
-            "HH:mm:ss tt"
+            "HH:mm:ss"
         };
 
         public List<string> DateFormatList { get; } = new()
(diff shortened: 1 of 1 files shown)


jjw24 (commented):
Hi there, thanks for the pr.

As @onesounds mentioned, there maybe a slight impact on memory, could you please test how much impact there will be. This can be showing a screenshot of before and after using this format.

We will also make it clear to users that this may impact memory. @onesounds what's the best way to do this? Tooltip or warnin text etc?

jjw24 (changes requested):
See comments please.

onesounds:
I'll test and then if there are lot memory using, I'll add description part.(aka tooltip)

onesounds:
I have confirmed that resource usage does not increase in particular.
I think it would be okay not to add an explanation.

VictoriousRaptor (approved):

jjw24:
Haven't had a chance to take a look, will do so soon

z1nc0r3 (commented):
`hh:mm:ss tt` and `HH:mm:ss tt` are not the same.

VictoriousRaptor:
> `hh:mm:ss tt` and `HH:mm:ss tt` are not the same.

I know, but tt in `HH:mm:ss tt` is redundant.

PR #1691, opened 2022-12-19
Improve folder editor experience

### LINKED ISSUE #1633: Open folder in visual studio code

Is it possible to show `open with code` command when I select some file or directory?

When we select the file/folder and press right we get this list of options available. It would be great if we could add an option to open files/folders in visual studio code (if installed). 
image

It would be great if there would be `Open in Code` plugin that would be displayed like `Everything` in this image.

The alternative (and what I do now) is to use the `show context menu`  but it takes some time to load and is one step more to do.
image

fihorvat:
This is a feature that lets you define an editor for the folder.
Once we merge this issue can be resolved #1633


jjw24:
Hi @fihorvat, thanks for the PR.

Could you please update the description on what you have tested (even if just the intended change with this PR is fine) and also a before and after screenshot would be also nice.

fihorvat:
Hi @jjw24, no problem.
Now it should be okay 😄

taooceros:
A quick note is that we don't really want crowdin to import translation from our repository. Therefore, these translation you added now may be lost once merged. We may want to copy them to crowdin.

A simpler way to add translation is to only modify en.xaml and once merged, add translation with crowdin. It's ok to not include some string in some language file, which will fall back to english.

fihorvat:
@taooceros Ou I didn't know that. It really was annoying trying to translate to every single language and I thought there sure must be a better way 😅
Should I remove all the translations except en?

taooceros:
> @taooceros Ou I didn't know that. It really was annoying trying to translate to every single language and I thought there sure must be a better way 😅 Should I remove all the translations except en?

It's fine to do it like this. It would also be annoying to remove them🤣

fihorvat:
@jjw24 @taooceros Could anyone take a look at this PR when you find some time?

taooceros:
@fihorvat I just take a look. Everything looks great to me! There's a small issue (maybe not), that Volume like `D:\` won't have this option. Do you think we should add it too?

fihorvat:
@taooceros I just tried with Volume `F:\` and it works great so it should work for all volumes 

image
image

taooceros:
Na i mean for F: itself. We have a special Result for volume that display the space usage. I wonder whether we should treat that as a folder as well.

fihorvat:
@taooceros I didn't even know about that feature, that is so cool 👍 
I added the support for that as well and tested that it works 😄

taooceros (approved):

taooceros:
Thank you so much for the contribution!

shellwhale:
That's a nice feature but it misses the ability to pass command-line arguments. For example code --add allows adding a folder to a currently running instance of vs code (no new vscode window is created). 

I had to create a .bat file to wrap my parameters inside.

Issue #1640, opened 2022-12-14
Add the ability to display the current language layout


fervulp:
**Describe**
Add the ability to display the current layout, as with time
![изображение](https://user-images.githubusercontent.com/77237424/207470436-dedfc5b3-10ac-4126-9e13-59ad3fa43c32.png)

**Solution**
I would like to see the following
![изображение](https://user-images.githubusercontent.com/77237424/207471012-fe00d099-3dac-4053-9704-5d92ef332a2b.png)

**Additional context**
Many people start typing right away without paying attention to the layout, I think this is not difficult. Even if you refuse this idea, could you show how it can be implemented by yourself by changing the code.

jjw24:
@fervulp @onesounds since always start typing in English feature has been rolled out, is this display input in the search window still relevant?

VictoriousRaptor:
> @fervulp @onesounds since always start typing in English feature has been rolled out, is this display input in the search window still relevant?

Kind of. imo always start typing in English is generally enough.

onesounds:
> > @fervulp @onesounds since always start typing in English feature has been rolled out, is this display input in the search window still relevant?
> 
> Kind of. imo always start typing in English is generally enough.

yep I think it's enough. If fervulp has not other feedback, let's close the issue.

PR #1693, opened 2022-12-20
Always Start in English Layout

onesounds:
## What's the PR
- Add settings that always start in English when you start flow.
  -  If the user is using a Native/English layout, Sometimes the input will start in an unwanted language.
- It always starts in English, regardless of your language status, so you can avoid typing incorrectly.
- This setting can be turned on and off; the default is off.

## Test Cases
- [x] Korean / Japanese / Chinese (pinyin)
- [x] When you turn the flow back on in the native keyboard layout state, typing must begin in English.

### DIFF
--- a/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
+++ b/Flow.Launcher.Infrastructure/UserSettings/Settings.cs
@@ -147,6 +147,7 @@ namespace Flow.Launcher.Infrastructure.UserSettings
         /// </summary>
         public bool ShouldUsePinyin { get; set; } = false;
         public bool AlwaysPreview { get; set; } = false;
+        public bool AlwaysStartEn { get; set; } = false;
 
         [JsonInclude, JsonConverter(typeof(JsonStringEnumConverter))]
         public SearchPrecisionScore QuerySearchPrecision { get; private set; } = SearchPrecisionScore.Regular;
--- a/Flow.Launcher/Converters/BoolToIMEConversionModeConverter.cs
+++ b/Flow.Launcher/Converters/BoolToIMEConversionModeConverter.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Globalization;
+using System.Windows.Data;
+using System.Windows.Input;
+
+namespace Flow.Launcher.Converters
+{
+    internal class BoolToIMEConversionModeConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value is bool v)
+            {
+                if (v)
+                {
+                    return ImeConversionModeValues.Alphanumeric;
+                }
+                else
+                {
+                    return ImeConversionModeValues.DoNotCare;
+                }
+            }
+            return ImeConversionModeValues.DoNotCare;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/Flow.Launcher/Converters/BoolToIMEConversionModeConverter.cs
+++ b/Flow.Launcher/Converters/BoolToIMEConversionModeConverter.cs
@@ -28,4 +28,28 @@ namespace Flow.Launcher.Converters
             throw new NotImplementedException();
         }
     }
+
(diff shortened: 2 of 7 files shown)


VictoriousRaptor (approved):

onesounds:
> @onesounds @VictoriousRaptor when will users want to turn this option on?

It is necessary for a person who speaks a language other than English.
Since each language user uses both native language and English at the same time, there is an individual English state for each language. This language state is difficult to recognize input before it starts. So we need settings to start English state.Although it can start in the native language, it was judged that it is not general.

onesounds:
https://github.com/Flow-Launcher/Flow.Launcher/issues/1640

This is why this request occurs. I judged that this function was better than that proposal. (I want to include that function, but that function is harder to implement than I thought. I have to import the ime32.dll to find out. current ime state)

This function is difficult to explain to people who only speak English, and it is difficult to test. This is why other programs don't have this function. I believe that this function is very important and worth applying.

jjw24:
So if I have Korean IME and this option on, I type in search it will start in english but will it also change my language IME to English as well ? For example afterwards I go to website will I be typing English IME of my native IME?

onesounds:
> So if I have Korean IME and this option on, I type in search it will start in english but will it also change my language IME to English as well ? For example afterwards I go to website will I be typing English IME of my native IME?

This is the difficulty of this function. Suppose a user using English, Korean, and Japanese IME.
In this case, IME can be switched through win+space.  This pr does not change this IME. 

However, in the case of Korean or JAPANESE IME, there are two LAYOUTs. (LAYOUT is a sub-item of IME.) It is in NATIVE and English. This PR's goal is always use ENGLISH LAYOUT when flow start. 

The ime that was in use is not changed. Layout is reflected only in the input window of flow.
In other words, when I open the flow while writing a document in Chinese, it starts with an English input, and when I close the flow and return to the original process, it is in the Chinese input state.

jjw24:
Thanks for the explanation, let me take this pr for a spin.

VictoriousRaptor:
> @onesounds @VictoriousRaptor when will users want to turn this option on?

Users can now assure that IME is in English so no issue with typing action keywords.

VictoriousRaptor (approved):

VictoriousRaptor (approved):

sirito97:
I have tried this option and it is not working with Arabic, isn't it supposed to always search in English regardless of the selected language?

VictoriousRaptor:
> I have tried this option and it is not working with Arabic, isn't it supposed to always search in English regardless of the selected language?

It's designed for CJK IME. IME is complex and not well documented. We will take a look.

VictoriousRaptor:
@sirito97 How do you type Arabic? I just did some test and seems like Arabic input by Microsoft is not a IME but a keyboard layout, so our current method doesn't work. Sorry about that.

sirito97:
> @sirito97 How do you type Arabic? I just did some test and seems like Arabic input by Microsoft is not a IME but a keyboard layout, so our current method doesn't work. Sorry about that.

@VictoriousRaptor Sorry for late reply, I really have no idea what is IME but here's a screenshot of my language input
Screenshot_20230124_072347

VictoriousRaptor:
> > @sirito97 How do you type Arabic? I just did some test and seems like Arabic input by Microsoft is not a IME but a keyboard layout, so our current method doesn't work. Sorry about that.
> 
> @VictoriousRaptor Sorry for late reply, I really have no idea what is IME but here's a screenshot of my language input Screenshot_20230124_072347

it's difficult to explain what IME (Input Method Editor) is in a few words for people who don't use it. If you are curious please google it. May I ask you how do you switch between English and Arabic input? That can tell if you are using an IME.

sirito97:
@VictoriousRaptor I switch by clicking left alt + shift here's a screenshot too
image

sirito97:
> > @VictoriousRaptor I switch by clicking left alt + shift here's a screenshot too image
> 
> It's not an IME then. Sorry we are not able to fix it.

It's ok, I managed to fix it by ticking this option

Sing303:
Not working for Russian language

1. Turned on "English Mode"
2. Switching to Russian language
3. Opened Flow Launcher. 

The behaviour is .
 - Language is still Russian

Expected behaviour
 - Regardless of the current layout, English characters are printed

PR #1704, opened 2022-12-22
Remove await Task.Run for IconDelegate

taooceros:
Let's trust plugin developer when they use that for icon
This will solve the icon blink for uwp and some other plugin that use IconDelegate.

Tested:
- [x] VScode Plugin doesn't have icon flickering

### DIFF
--- a/Flow.Launcher/ViewModel/ResultViewModel.cs
+++ b/Flow.Launcher/ViewModel/ResultViewModel.cs
@@ -179,7 +179,7 @@ namespace Flow.Launcher.ViewModel
             {
                 try
                 {
-                    var image = await Task.Run(() => icon()).ConfigureAwait(false);
+                    var image = icon();
                     return image;
                 }
                 catch (Exception e)


jjw24:
@taooceros please list what's tested.

jjw24:
>Let's trust plugin developer when they use that for icon

What do they need to do?

VictoriousRaptor:
> > Let's trust plugin developer when they use that for icon
> 
> What do they need to do?

No change needed. Currently two ways to provide an icon: path to the image or a delegate.

VictoriousRaptor:
Just found that this is not the actual problem. UWP uses IcoPath now. Blinking when typing with always preview enabled is because selected item is rapidly changing when typing.

jjw24:
Shall we close this then?

VictoriousRaptor:
> Shall we close this then?

it's up to tao

taooceros:
> Shall we close this then?

Let's keep the behavior the same as before. The vscode plugin is still using the Icon delegate.

taooceros:
> @taooceros please list what's tested.

I will test once I get back to my comp

VictoriousRaptor (approved):

Issue #1732, opened 2022-12-29
index with localized name of file/folder


u-Sir:
**Is your feature request related to a problem? Please describe.**

As non-English user, some windows files/folders display localized in windows explorer, but can't search by localized name in Flow launcher. e.g. `control panel`- `控制面板`(Chinese)
just like start menu, support search in Pinyin
![图片](https://user-images.githubusercontent.com/93647031/209914752-901f8b89-c7ad-49cc-a117-dd3ce86fc978.png)

something may help :
https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shgetlocalizedname
http://archives.miloush.net/michkap/archive/2007/01/18/1487464.html

```powershell
    $FilePath = "C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Administrative Tools\Windows Defender Firewall with Advanced Security.lnk"
    $ChildObj = Get-ChildItem $FilePath 
    $FldrName = $ChildObj.DirectoryName 
    $FileName = $ChildObj.Name 
    $Shell    = New-Object -ComObject Shell.Application 
    $Folder   = $Shell.Namespace($FldrName) 
    $File     = $Folder.ParseName($FileName) 
    echo $($Folder.GetDetailsOf($File,0))
```

![图片](https://user-images.githubusercontent.com/93647031/209914956-75164528-2ce2-4d09-a319-31ec4a5e052a.png)

jjw24:
Available in pre-release.

VictoriousRaptor:
> Available in pre-release.

Folders are not available yet.

u-Sir:
> > Available in pre-release.
> 
> Folders are not available yet.

ok, for folders check this:
http://archives.miloush.net/michkap/archive/2007/01/18/1487464.html

VictoriousRaptor:
Though we can get localized folder names, Everything doesn't allow you to search localized names. If we want use localized name as search terms, we need to create a mapping from localized name to actual path which is difficult.

PR #1736, opened 2022-12-30
Use localized name for shell link programs

VictoriousRaptor:
Related to #1732 

Core codes are from https://github.com/microsoft/PowerToys/pull/19149

Changes:
- Add functionality to search in localized program names for .lnk files
- Catch COMexception in ShellLinkHelper so if failed to get description program still can be indexed
- Refactor result matching logic in Win32 class
- Fix exception message argument in UWP class
- some other minor fixes

Tests
- [x] Search for program name, localized name, file name, target file name, description are working


Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -63,48 +69,103 @@ public class Win32 : IProgram, IEquatable<Win32>
             Enabled = false
         };
 
+        private static MatchResult Match(string query, List<string> candidates)
+        {
+            if (candidates.Count == 0)
+                return null;
+
+            List<MatchResult> matches = new List<MatchResult>();
+            foreach(var candidate in candidates)
```
taooceros, line 78: let's use linq to simplify this. I believe we can write it in one line

taooceros (changes requested):

Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -69,28 +69,15 @@ public class Win32 : IProgram, IEquatable<Win32>
             Enabled = false
         };
 
-        private static MatchResult Match(string query, List<string> candidates)
+        private static MatchResult Match(string query, IReadOnlyCollection<string> candidates)
```
taooceros, line 72: I only change this part, other is formatting

taooceros (approved):

PR #1748, opened 2022-12-31
Add prompt for Log Off command

z1nc0r3:
Add a prompt for the **Log Off command** _(System commands plugin)_ which prevents users from accidentally logging off from their system.


jjw24:
Hey @z1nc0r3, just a fyi with language files, you only need to add to en.xaml, the rest will default to en if no key available. Crowdin will also add the key to the rest of the language files.

jjw24 (approved):

z1nc0r3:
> Hey @z1nc0r3, just a fyi with language files, you only need to add to en.xaml, the rest will default to en if no key available. Crowdin will also add the key to the rest of the language files.

Thank you for the information.

PR #1750, opened 2022-12-31
Allow plugin name to be searchable by Plugin Indicator

jjw24:
Often times when I download a new plugin, I don't know the action keyword to trigger it until I check in settings. This change allows searching for the name of the plugin via Plugin Indicator to find the action keyword.

Before:
image

After:
image

Tested:
- Plugin name is searchable
- Action keyword is searchable
- Selecting the result will change to the actual action keyword
- Autocomplete will change to the actual action keyword


VictoriousRaptor:
Should we give this kind of result a low score? Or it may annoy users that knows the action keyword.

jjw24:
Hmm good point. What would be an appropriate score or should I rewrite it to use the score from StringMatcher?

VictoriousRaptor:
> Hmm good point. What would be an appropriate score or should I rewrite it to use the score from FuzzySearch?

Not familiar with this part. I guess a score from FuzzySearch should be fine. Or is it possible to provide an option to turn name search off?

jjw24:
I changed it to use score from fuzzy search so the results are aligned to the rest of plugins in flow:

image

VictoriousRaptor (approved):

PR #1760, opened 2023-01-03
Stop spell check on push

VictoriousRaptor:
Stop checking on push to avoid checking merge commits on dev branch. Commits of PRs can be checked on synchronized.

### DIFF
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -65,7 +65,7 @@ jobs:
     outputs:
       followup: ${{ steps.spelling.outputs.followup }}
     runs-on: ubuntu-latest
-    if: (contains(github.event_name, 'pull_request') && github.head_ref != 'l10n_dev') || github.event_name == 'push'
+    if: (contains(github.event_name, 'pull_request') && github.head_ref != 'l10n_dev')
     concurrency:
       group: spelling-${{ github.event.pull_request.number || github.ref }}
       # note: If you use only_check_changed_files, you do not want cancel-in-progress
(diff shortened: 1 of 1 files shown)


.github/workflows/spelling.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     outputs:
       followup: ${{ steps.spelling.outputs.followup }}
     runs-on: ubuntu-latest
-    if: (contains(github.event_name, 'pull_request') && github.head_ref != 'l10n_dev') || github.event_name == 'push'
+    if: (contains(github.event_name, 'pull_request') && github.head_ref != 'l10n_dev')
```
jjw24, line 68: How does this pull request event work, would it only get triggered when a pr is created? If the case then it probably wouldn't be triggered again by subsequent pushes to the pr.

I wonder maybe a better way is to just  github.head_ref != dev && github.head_ref != master, so it doesn't check pushes to dev and master branches.

.github/workflows/spelling.yml
```diff
@@ -103,21 +103,21 @@ jobs:
...
+#     permissions:
+#       contents: write
+#     if: (success() || failure()) && needs.spelling.outputs.followup && github.event_name == 'push'
+#     steps:
+#     - name: comment
+#       uses: check-spelling/check-spelling@main
+#       with:
+#         checkout: true
+#         spell_check_this: check-spelling/spell-check-this@main
+#         task: ${{ needs.spelling.outputs.followup }}
```
jjw24, line 120: Why we commenting these out?

.github/workflows/spelling.yml
VictoriousRaptor: It is also triggered on synchronize event (any push to the pr head triggers it).

.github/workflows/spelling.yml
VictoriousRaptor: This is bot comment for push checks. Check on push is disabled so this is not needed anymore.

.github/workflows/spelling.yml
VictoriousRaptor: simply ignoring dev or master is enough though. but if you push to a fork the check is also triggered. I want to just check in the pr.

VictoriousRaptor: oh forget to comment on push at the beginning of the file

jjw24 (approved):

Issue #1504, opened 2022-10-31
Flow launcher terminates silently


jjw24:
Flow terminates silently after usage for a while, flow icon no longer in the taskbar and itself can not be triggered.

1. 
  [2022-09-15.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/9606557/2022-09-15.txt)
Log for when it crashed. When clicking FL icon, it still does not open

_Originally posted by @tejaskerai in https://github.com/Flow-Launcher/Flow.Launcher/issues/1268#issuecomment-1252152090_

2. > It looks like it happened again, and trying to relaunch it doesn't work [2022-10-21.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/9851242/2022-10-21.txt)

tejaskerai:
Thanks for opening this issue. Thought this might be some useful information; while using FL, I've noticed that this mostly occurs after coming out of hibernation or a reboot, but also sometimes it will open and then crash and will not open again

VictoriousRaptor:
> Thanks for opening this issue. Thought this might be some useful information; while using FL, I've noticed that this mostly occurs after coming out of hibernation or a reboot, but also sometimes it will open and then crash and will not open again

Next time when it happens, please check Event Viewer. It may contain some useful information.

jjw24:
> Thanks for opening this issue. Thought this might be some useful information; while using FL, I've noticed that this mostly occurs after coming out of hibernation or a reboot, but also sometimes it will open and then crash and will not open again

You wouldn't happen to be using Windows with Parallel would you?

tejaskerai:
Nope, not using parallels.

tejaskerai:
> Next time when it happens, please check Event Viewer. It may contain some useful information.

The issue occurred today again. After coming out of hibernation and using flowlauncher, it opened and immediately crashed.
Attached Event viewer Log
![image](https://user-images.githubusercontent.com/38558558/199951413-2bb9c7c5-8ce0-4b1d-bce3-1aa37b9c0786.png)

tejaskerai:
Another log that may be useful
[Report.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/9937454/Report.txt)

VictoriousRaptor:
What plugins are you using? That dll is not shipped with flow so I guess it's related to a plugin?

tejaskerai:
[...]

PR #1762, opened 2023-01-04
Safer Updater

taooceros:
might close https://github.com/Flow-Launcher/Flow.Launcher/issues/1504

Note:
I am not so sure why the autoupdater may crash flow itself...I guess it is because the Elapsed callback is triggered from the main thread, but I cannot find evidence to support it, because based on document, this happens only when SynchronizingObject is set which is not we have done.

Though, I have suppressed the error of general exception from updater, so now at least it will be safer to use the autoupdater.

### DIFF
--- a/Flow.Launcher.Core/Updater.cs
+++ b/Flow.Launcher.Core/Updater.cs
@@ -33,7 +33,7 @@ namespace Flow.Launcher.Core
 
         public async Task UpdateAppAsync(IPublicAPI api, bool silentUpdate = true)
         {
-            await UpdateLock.WaitAsync();
+            await UpdateLock.WaitAsync().ConfigureAwait(false);
             try
             {
                 if (!silentUpdate)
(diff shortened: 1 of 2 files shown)


jjw24:
Done a ln update test to ensure working as expected?

Flow.Launcher/App.xaml.cs
```diff
@@ -84,7 +85,7 @@ private async void OnStartupAsync(object sender, StartupEventArgs e)
 
                 Current.MainWindow = window;
                 Current.MainWindow.Title = Constant.FlowLauncher;
-                
+
                 HotKeyMapper.Initialize(_mainVM);
 
                 // happlebao todo temp fix for instance code logic
```
jjw24, line 91:
```suggestion
                // todo temp fix for instance code logic
```

jjw24:
Can you please fix spelling errors 'Uodate', 'settigns'.

Flow.Launcher.Core/Updater.cs
```diff
@@ -88,9 +88,13 @@ public async Task UpdateAppAsync(IPublicAPI api, bool silentUpdate = true)
                     UpdateManager.RestartApp(Constant.ApplicationFileName);
                 }
             }
-            catch (Exception e) when (e is HttpRequestException or WebException or SocketException || e.InnerException is TimeoutException)
+            catch (Exception e)
             {
-                Log.Exception($"|Updater.UpdateApp|Check your connection and proxy settings to github-cloud.s3.amazonaws.com.", e);
+                if ((e is HttpRequestException or WebException or SocketException || e.InnerException is TimeoutException))
+                    Log.Exception($"|Updater.UpdateApp|Check your connection and proxy settings to github-cloud.s3.amazonaws.com.", e);
```
jjw24, line 94: Looking at this, I wonder if we need to notify users when these type of errors occur?

Flow.Launcher/App.xaml.cs
```diff
@@ -130,20 +131,17 @@ private void AutoStartup()
...
-                    timer.Start();
-
-                    // check updates on startup
+                    // check update every 5 hours
+                    var timer = new PeriodicTimer(TimeSpan.FromHours(5));
                     await _updater.UpdateAppAsync(API);
+
+                    while (await timer.WaitForNextTickAsync())
+                        // check updates on startup
+                        await _updater.UpdateAppAsync(API);
```
jjw24, line 144: is this a double up of `await _updater.UpdateAppAsync(API);`? Doesnt it run once timer is instantiated?

jjw24:
@taooceros have you tested updating with this change?

Flow.Launcher.Core/Updater.cs
taooceros: I think we did? In the later part when silentupdate is false

jjw24 (approved):

PR #1781, opened 2023-01-11
Always show/search description in program results if enabled

VictoriousRaptor:
Always show/search description in program results if description is enabled in settings. If name and desc are equal the desc is omitted.

This pr is done because an uwp app named "Xbox" but with a description "XboxPcApp". Currently in this case the result title is "XboxPcApp" when description is enabled. In this pr the title will be "Xbox: XboxPcApp".

P.S: The diff view is messy somehow. git blame is fine. Most of the changes are deletion.

Change:
- Always show description if enabled, unless name and description are equal.

Test:
- programs can be searched via name and description

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs
@@ -378,15 +378,10 @@ namespace Flow.Launcher.Plugin.Program.Programs
                 MatchResult matchResult;
 
                 // We suppose Name won't be null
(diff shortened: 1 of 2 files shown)


jjw24:
I seem to be able to search description already with dev branch build:

image

VictoriousRaptor:
> I seem to be able to search description already with dev branch build:
> 
> image

That's in release. Currently if description starts with app name, like the example I mentioned, we only fuzzy search the description and show the description as result title. That's enough for searching but the result title might be confusing.

jjw24:
ok I see 👍

jjw24 (approved):

PR #1785, opened 2023-01-12
Test CI when only changing content in .github folder

VictoriousRaptor:


taooceros:
is this needed?

taooceros:
> > Shall we close?
> 
> yeah not an important feature

Does it work?

PR #1786, opened 2023-01-12
Expand builtin shortcuts when needed

VictoriousRaptor:
Explorer shortcut in #1275 is laggy. No need to calculate the shortcut when the shortcut is not in the query.


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -841,15 +841,20 @@ private Query ConstructQuery(string queryText, IEnumerable<CustomShortcutModel>
                 queryBuilder.Replace('@' + shortcut.Key, shortcut.Expand());
             }
 
+            string customExpanded = queryBuilder.ToString();
```
taooceros, line 844: Although I feel like this is somewhat weird....I consider it's ok

PR #1787, opened 2023-01-12
New results model

taooceros:
Summary of the change
1. Use [Dynamic Data](https://github.com/reactivemarbles/DynamicData#the-observable-list) to replace our buggy complex ResultCollection
2. Stop Triggering ResultListBox Visibility when results changes from empty to non-empty or vise versa

Effect:
Surprisingly the second change resolves #586, but will requires further test.
Though, it also introduce some style inconsistent. @onesounds will notice it easily.

image

More weirdly, if we switch to context menu and switch back, the style will become normal as current release.


jjw24:
@taooceros please list your tests

@onesounds please check if ok:
>Though, it also introduce some style inconsistent. @onesounds will notice it easily.

onesounds:
OK. That margin from 'MiddleSeparator' logic. that logic from 'BaseSeparatorStyle' in base.xaml 
Since the separator in the middle is used in common with the History and Context menu, it is separated from the Result.

image

A simple solution is to change the code as follows.

image

However, in this case, the Separator is not displayed in the History view. (Ctrl+H)
I think we can do more research, do you have any good ideas? I have some problems because I don't understand the History part well. (Previously, there was a margin problem in the history menu.)

P.S. please mention the issue that I really need in Discord separately. There(github) are so many notifications that I can't check them all.

Issue #1771, opened 2023-01-09
Let Flow remove illegal (forbidden) characters (\ / : * ? " ' < > |) in the query string


raindropsfromsky:
A user often copies the title of an article/book from internet, and then pastes it in the Flow to check if that file exists.

However, the problem with that approach is that often the text string contains non-Windows characters such as a column (:).
Earlier, Flow was able to ignore such characters. 
But now looks for a positive match.
As a result, Flow hides even when there are matching results, just because they cannot have the : character.
If the user deletes such "non-Windows" characters, the results show up.

Therefore, it would be great if Flow ignores such characters.

taooceros:
I am confused about what do you mean by "Non-Windows characters". I don't think a `:` will prevent results showing (maybe will prevent results from explorer).

z1nc0r3:
You probably mean the **forbidden characters** for file/dir names like `\ / : * ? " ' < > |`. And yes, FL considers these characters for searching files/dirs and will not give results correctly when the query contains these illegal characters. 

I think we can just replace these forbidden chars with "" in the explorer plugin. So it'll not consider these chars for searching files/dirs.

taooceros:
> You probably mean the **forbidden characters** for file/dir names like `\ / : * ? " ' < > |`. And yes, FL considers these characters for searching files/dirs and will not give results correctly when the query contains these illegal characters.
> 
> I think we can just replace these forbidden chars with "" in the explorer plugin. So it'll not consider these chars for searching files/dirs.

yeah that should be the case. I will check recently

raindropsfromsky:
> You probably mean the **forbidden characters** for file/dir names like `\ / : * ? " ' < > |`. And yes, FL considers these characters for searching files/dirs and will not give results correctly when the query contains these illegal characters.
> 
> I think we can just replace these forbidden chars with "" in the explorer plugin. So it'll not consider these chars for searching files/dirs.

Yes, that's what I meant. Thanks for the correction! 
I have corrected the subject line of my original issue.

Yes, FL should simply delete (or ignore) these characters and search for the rest.

raindropsfromsky:
BTW even **Everything** has the same issue: It cannot ignore illegal (forbidden) characters in the search string. 

So regardless of which engine is used (Explorer or Everything), the result would be the same: There is a false negative. That is, although there are matching results, they are not reported.

Therefore the best strategy is for FL to pre-process the string and remove all forbidden characters.
[...]

PR #1790, opened 2023-01-13
Filter the query string in Explorer plugin for forbidden chars

z1nc0r3:
#1771 Filter the query string in Explorer plugin for possible forbidden characters.


taooceros (commented):
I think there's another place we check for these characters (in the windows index). Could you please take a check there and see whether we should move those characters check here? Actually I am curious will this character stop the everything api?

z1nc0r3:
@taooceros thanks for the info. I'll take a look.

z1nc0r3:
It seems like this issue needs a few more changes. I'll close this PR for now.

PR #1817, opened 2023-01-19
Fix bugs in explorer plugin

### LINKED ISSUE #1796: [BUG] Variable %LocalAppData% in path cause exception
I have backup folder contains **%LocalAppData%** in path. That variable is not used literally as %LocalAppData% but as absolute path.

Example
This is wrong

> E:\Backup\CloneApp\Backup\Windows Taskbar\C:\Users\account\AppData\Roaming\Microsoft\Internet Explorer\Quick Launch\User Pinned\TaskBar\TaskBar\Windows Sandbox.lnk

This is correct

> E:\Backup\CloneApp\Backup\Windows Taskbar\%AppData%\Microsoft\Internet Explorer\Quick Launch\User Pinned\TaskBar\TaskBar\Windows Sandbox.lnk

That cause exception and I can't find these files.

### LINKED ISSUE #1802: [BUG] - Explorer Plugin crash when first input is a colon ":"
Explorer Plugin immediately crash when this first input is a colon, this happen only when Explorer plugin is ON.
Same result with FR, US and ZH layout.

Not really a major issue, just wanted to let you know.

### Log file :
[2023-01-15.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/10420173/2023-01-15.txt)

### Exception message :
Flow Launcher version: 1.11.0
OS Version: Microsoft Windows NT 10.0.19044.0
IntPtr Length: 8
x64: True

Python Path: C:\Python310\pythonw.exe
Node Path: 
Date: 01/15/2023 13:26:02
Exception:
Explorer Exception: 
Websites: https://github.com/Flow-Launcher/Flow.Launcher
Author: Jeremy Wu
Version: 2.1.0
Flow.Launcher.Core.ExternalPlugins.FlowPluginException: Object reference not set to an instance of an object.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Flow.Launcher.Plugin.Explorer.Search.SearchManager.<>c__DisplayClass4_0.<SearchAsync>b__1(AccessLink excludedPath) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\SearchManager.cs:line 120
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at System.Collections.Generic.HashSet`1.RemoveWhere(Predicate`1 match)
   at Flow.Launcher.Plugin.Explorer.Search.SearchManager.SearchAsync(Query query, CancellationToken token) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\SearchManager.cs:line 119
   at Flow.Launcher.Plugin.Explorer.Main.QueryAsync(Query query, CancellationToken token) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Main.cs:line 88
   at Flow.Launcher.Core.Plugin.PluginManager.<>c__DisplayClass22_0.<<QueryForPluginAsync>b__0>d.MoveNext() in C:\projects\flow-launcher\Flow.Launcher.Core\Plugin\PluginManager.cs:line 191
--- End of stack trace from previous location ---
   at Flow.Launcher.Infrastructure.Stopwatch.DebugAsync(String message, Func`1 action) in C:\projects\flow-launcher\Flow.Launcher.Infrastructure\Stopwatch.cs:line 40
   at Flow.Launcher.Core.Plugin.PluginManager.QueryForPluginAsync(PluginPair pair, Query query, CancellationToken token) in C:\projects\flow-launcher\Flow.Launcher.Core\Plugin\PluginManager.cs:line 213
   --- End of inner exception stack trace ---
   at Flow.Launcher.Core.Plugin.PluginManager.QueryForPluginAsync(PluginPair pair, Query query, CancellationToken token) in C:\projects\flow-launcher\Flow.Launcher.Core\Plugin\PluginManager.cs:line 213
   at Flow.Launcher.ViewModel.MainViewModel.<>c__DisplayClass135_0.<<QueryResults>g__QueryTask|2>d.MoveNext() in C:\projects\flow-launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 821
--- End of stack trace from previous location ---
   at Flow.Launcher.ViewModel.MainViewModel.QueryResults() in C:\projects\flow-launcher\Flow.Launcher\ViewModel\MainViewModel.cs:line 822
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)

VictoriousRaptor:
Changes:
- Fix #1802 (Fix null subtitle when creating disk result)
- Fix #1796 (Only expand environment var when path starts with %)
- Use case-insensitive comparator for path results
- Use api for subpath check

Tests:
- A single ":" when excluded path is not empty won't cause exception. (1802)
- Unit test for environment path expansion
- Unit test for subpath check
- Unit test for result comparator

### DIFF
--- a/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
@@ -1,9 +1,9 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
-#if !DEBUG
+#pragma warning disable IDE0005
 using System.Windows;
-#endif
+#pragma warning restore IDE0005
 
 namespace Flow.Launcher.Plugin.SharedCommands
 {
--- a/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
@@ -254,7 +254,7 @@ namespace Flow.Launcher.Plugin.SharedCommands
         /// <returns></returns>
         public static bool PathContains(string parentPath, string subPath, bool allowEqual = false)
         {
-            var rel = Path.GetRelativePath(parentPath, subPath);
+            var rel = Path.GetRelativePath(parentPath.EnsureTrailingSlash(), subPath);
             return (rel != "." || allowEqual)
                    && rel != ".."
                    && !rel.StartsWith("../")
--- a/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
-using System.Windows;
 
 namespace Flow.Launcher.Plugin.SharedCommands
 {
@@ -243,16 +242,17 @@ namespace Flow.Launcher.Plugin.SharedCommands
         }
 
         /// <summary>
-        /// Returns if <paramref name="subPath"/> is a sub path of <paramref name="basePath"/>.
+        /// Returns if <paramref name="parentPath"/> contains <paramref name="subPath"/>.
         /// From https://stackoverflow.com/a/66877016
         /// </summary>
-        /// <param name="subPath"></param>
-        /// <param name="basePath"></param>
+        /// <param name="parentPath">Parent path</param>
+        /// <param name="subPath">Sub path</param>
+        /// <param name="allowEqual">If <see langword="true"/>, when <paramref name="parentPath"/> and <paramref name="subPath"/> are equal, returns <see langword="true"/></param>
         /// <returns></returns>
-        public static bool IsSubPathOf(string subPath, string basePath)
+        public static bool PathContains(string parentPath, string subPath, bool allowEqual = false)
         {
-            var rel = Path.GetRelativePath(basePath, subPath);
-            return rel != "."
+            var rel = Path.GetRelativePath(parentPath, subPath);
+            return (rel != "." || allowEqual)
                    && rel != ".."
                    && !rel.StartsWith("../")
                    && !rel.StartsWith(@"..\")
--- a/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
@@ -261,5 +261,15 @@ namespace Flow.Launcher.Plugin.SharedCommands
                    && !rel.StartsWith(@"..\")
                    && !Path.IsPathRooted(rel);
         }
+        
+        /// <summary>
+        /// Returns path ended with "\"
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns></returns>
+        public static string EnsureTrailingSlash(this string path)
+        {
+            return path.TrimEnd(Path.DirectorySeparatorChar) + Path.DirectorySeparatorChar;
+        }
     }
 }
--- a/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
@@ -1,6 +1,9 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
+#if !DEBUG
+using System.Windows;
+#endif
 
 namespace Flow.Launcher.Plugin.SharedCommands
 {
--- a/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
@@ -241,5 +241,22 @@ namespace Flow.Launcher.Plugin.SharedCommands
 
             return path;
         }
+
+        /// <summary>
+        /// Returns if <paramref name="subPath"/> is a sub path of <paramref name="basePath"/>.
+        /// From https://stackoverflow.com/a/66877016
+        /// </summary>
+        /// <param name="subPath"></param>
+        /// <param name="basePath"></param>
+        /// <returns></returns>
+        public static bool IsSubPathOf(string subPath, string basePath)
+        {
+            var rel = Path.GetRelativePath(basePath, subPath);
+            return rel != "."
+                   && rel != ".."
+                   && !rel.StartsWith("../")
+                   && !rel.StartsWith(@"..\")
+                   && !Path.IsPathRooted(rel);
+        }
     }
 }
--- a/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
@@ -208,22 +208,16 @@ namespace Flow.Launcher.Plugin.SharedCommands
         ///</summary>
         public static string GetPreviousExistingDirectory(Func<string, bool> locationExists, string path)
         {
-            var previousDirectoryPath = "";
             var index = path.LastIndexOf('\\');
             if (index > 0 && index < (path.Length - 1))
             {
-                previousDirectoryPath = path.Substring(0, index + 1);
-                if (!locationExists(previousDirectoryPath))
-                {
-                    return "";
-                }
+                string previousDirectoryPath = path.Substring(0, index + 1);
+                return locationExists(previousDirectoryPath) ? previousDirectoryPath : "";
             }
             else
             {
                 return "";
             }
-
-            return previousDirectoryPath;
         }
 
         ///<summary>
--- a/Flow.Launcher.Test/FilesFoldersTest.cs
+++ b/Flow.Launcher.Test/FilesFoldersTest.cs
@@ -44,7 +44,7 @@ namespace Flow.Launcher.Test
 
         [TestCase(@"c:\foo", @"c:\foo", true)]
         [TestCase(@"c:\foo\", @"c:\foo", true)]
-        [TestCase(@"c:\foo\", @"c:\foo", true)]
+        [TestCase(@"c:\foo", @"c:\foo\", true)]
         public void TestPathContainsWhenEqual(string parentPath, string path, bool expectedResult)
         {
             Assert.AreEqual(expectedResult, FilesFolders.PathContains(parentPath, path, true));
--- a/Flow.Launcher.Test/FilesFoldersTest.cs
+++ b/Flow.Launcher.Test/FilesFoldersTest.cs
@@ -1,0 +1,53 @@
+﻿using Flow.Launcher.Plugin.SharedCommands;
+using NUnit.Framework;
+
+namespace Flow.Launcher.Test
+{
+    [TestFixture]
+    
+    public class FilesFoldersTest
+    {
+        // Testcases from https://stackoverflow.com/a/31941905/20703207
+        // Disk
+        [TestCase(@"c:", @"c:\foo", true)]
+        [TestCase(@"c:\", @"c:\foo", true)]
+        // Slash
+        [TestCase(@"c:\foo\bar\", @"c:\foo\", false)]
+        [TestCase(@"c:\foo\bar", @"c:\foo\", false)]
+        [TestCase(@"c:\foo", @"c:\foo\bar", true)]
+        [TestCase(@"c:\foo\", @"c:\foo\bar", true)]
+        // File
+        [TestCase(@"c:\foo", @"c:\foo\a.txt", true)]
+        [TestCase(@"c:\foo", @"c:/foo/a.txt", true)]
+        [TestCase(@"c:\FOO\a.txt", @"c:\foo", false)]
+        [TestCase(@"c:\foo\a.txt", @"c:\foo\", false)]
+        [TestCase(@"c:\foobar\a.txt", @"c:\foo", false)]
+        [TestCase(@"c:\foobar\a.txt", @"c:\foo\", false)]
+        [TestCase(@"c:\foo\", @"c:\foo.txt", false)]
+        // Prefix
+        [TestCase(@"c:\foo", @"c:\foobar", false)]
+        [TestCase(@"C:\Program", @"C:\Program Files\", false)]
+        [TestCase(@"c:\foobar", @"c:\foo\a.txt", false)]
+        [TestCase(@"c:\foobar\", @"c:\foo\a.txt", false)]
+        // Edge case
+        [TestCase(@"c:\foo", @"c:\foo\..\bar\baz", false)]
+        [TestCase(@"c:\bar", @"c:\foo\..\bar\baz", true)]
+        [TestCase(@"c:\barr", @"c:\foo\..\bar\baz", false)]
+        // Equality
+        [TestCase(@"c:\foo", @"c:\foo", false)]
+        [TestCase(@"c:\foo\", @"c:\foo", false)]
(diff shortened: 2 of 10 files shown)


jjw24:
Please update description with what you have tested :)

Plugins/Flow.Launcher.Plugin.Explorer/Search/EnvironmentVariables.cs
```diff
@@ -44,40 +62,26 @@ internal static bool IsEnvironmentVariableSearch(string search)
...
-            if (splitSearch.Length == 2 && envStringPaths.ContainsKey(exactEnvStringPath))
-            {
-                var queryPartToReplace = $"%{exactEnvStringPath}%";
-                var expandedPath = envStringPaths[exactEnvStringPath];
-                // replace the %envstring% part of the query with its expanded equivalent
-                return environmentVariablePath.Replace(queryPartToReplace, expandedPath);
-            }
-
-            return environmentVariablePath;
+            return Environment.ExpandEnvironmentVariables(environmentVariablePath);
```
jjw24, line 65: maybe just update TranslateEnvironmentVariablePath call with Environment.ExpandEnvironmentVariables? Doesnt look like it needs to be wrapped inside a method.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -105,19 +107,21 @@ when ActionKeywordMatch(query, Settings.ActionKeyword.IndexSearchActionKeyword)
...
-                if (e is EngineNotAvailableException)
-                    throw;
-
                 throw new SearchException(engineName, e.Message, e);
             }
-            
+
             results.RemoveWhere(r => Settings.IndexSearchExcludedSubdirectoryPaths.Any(
-                excludedPath => r.SubTitle.StartsWith(excludedPath.Path, StringComparison.OrdinalIgnoreCase)));
+                excludedPath => IsSubPathOf(r.SubTitle, excludedPath.Path)));
```
jjw24, line 124: what's the issue with this check?

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -8,14 +8,15 @@
...
+using System.IO;
 
 namespace Flow.Launcher.Plugin.Explorer.Search
 {
     public class SearchManager
     {
-        internal static PluginInitContext Context;
+        internal PluginInitContext Context;
 
-        internal static Settings Settings;
```
jjw24, line 18: Why are we changing this from static?

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -30,12 +31,13 @@ private class PathEqualityComparator : IEqualityComparer<Result>
...
             {
-                return x.Title == y.Title && x.SubTitle == y.SubTitle;
+                return x.Title.Equals(y.Title, StringComparison.OrdinalIgnoreCase) 
+                    && string.Equals(x.SubTitle, y.SubTitle, StringComparison.OrdinalIgnoreCase);
             }
 
             public int GetHashCode(Result obj)
             {
-                return HashCode.Combine(obj.Title.GetHashCode(), obj.SubTitle?.GetHashCode() ?? 0);
+                return HashCode.Combine(obj.Title.ToLowerInvariant(), obj.SubTitle?.ToLowerInvariant() ?? "");
```
jjw24, line 40: Is the GetHashCode not required? 

Why is the return type from the null check changed?

Plugins/Flow.Launcher.Plugin.Explorer/Search/EnvironmentVariables.cs
VictoriousRaptor: Sure.

jjw24:
@z1nc0r3 just a FYI the issue with ':' is addressed here already.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
VictoriousRaptor: a path string starts with another doesn't necessarily mean it's contains the other one (unless ending with '\' is guaranteed). like `c:\program files` and `c:\program`

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
VictoriousRaptor: search manager is Singleton but with these static members initialized in constructor. imo it's confusing

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -252,5 +256,16 @@ internal static bool IsEnvironmentVariableSearch(string search)
                    && search != "%%"
                    && !search.Contains('\\');
         }
+
+        // https://stackoverflow.com/a/66877016
+        internal static bool IsSubPathOf(string subPath, string basePath)
```
jjw24, line 261: Please move this method to FilesFolders class, which can then be shared by other plugins and flow.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
VictoriousRaptor: it's not required. using hashcode.combine is to generating hash code based on multiple values, not to hash their hash code again.

VictoriousRaptor:
> @z1nc0r3 just a FYI the issue with ':' is addressed here already.

that's a bit different. that pr was to ignore queries with ':'. this one is to fix the crash issue.

jjw24:
The only other thing I picked up is the use of %USERNAME% in the form of `c:\users\%USERNAME%\downloads`, I dont know how many other variables that used during mid-path.

Flow.Launcher.Test/FilesFoldersTest.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Flow.Launcher.Test
+{
+    [TestFixture]
+    
+    public class FilesFoldersTest
+    {
+        // Testcases from https://stackoverflow.com/a/31941905/20703207
+        // Disk
+        [TestCase(@"c:", @"c:\foo", true)]
```
jjw24, line 12: This failed the test. Using `c:` works in Win Explorer but not supported in Explorer plugin, is this test still applicable?

Plugins/Flow.Launcher.Plugin.Explorer/Search/EnvironmentVariables.cs
```diff
@@ -8,12 +8,30 @@ namespace Flow.Launcher.Plugin.Explorer.Search
...
+        private static Dictionary<string, string> _envStringPaths = null;
+        private static Dictionary<string, string> EnvStringPaths
+        {
+            get
+            {
+                if (_envStringPaths == null)
+                {
+                    _envStringPaths = LoadEnvironmentStringPaths();
+                }
+                return _envStringPaths;
```
jjw24, line 20:
```suggestion
                return _envStringPaths ?? LoadEnvironmentStringPaths();
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -13,29 +13,33 @@ namespace Flow.Launcher.Plugin.Explorer.Search
...
 
         public SearchManager(Settings settings, PluginInitContext context)
         {
             Context = context;
             Settings = settings;
         }
 
-        private class PathEqualityComparator : IEqualityComparer<Result>
+        /// <summary>
+        /// Note: Assuming all diretories end with "\".
```
jjw24, line 27:
```suggestion
        /// Note: A path that ends with "\" and one that doesn't will not be regarded as equal.
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -105,19 +109,21 @@ when ActionKeywordMatch(query, Settings.ActionKeyword.IndexSearchActionKeyword)
...
+                throw;
+            }
             catch (Exception e)
             {
-                if (e is OperationCanceledException)
-                    return results.ToList();
-
-                if (e is EngineNotAvailableException)
-                    throw;
-
```
jjw24, line 115: Why this change?

VictoriousRaptor: Just for clarity.

VictoriousRaptor:
> The only other thing I picked up is the use of %USERNAME% in the form of `c:\users\%USERNAME%\downloads`. https://www.elevenforum.com/t/complete-list-of-environment-variables-in-windows-11.11212/ I suppose though you can use %USERPROFILE% to achieve the same as using %USERNAME%
> 
> Perhaps add an exclusion for %USERNAME%?

Just found something interesting. If you create a folder called `%appdata%` in some folder like `D:\foo`, then try to cd `D:\foo\%appdata%` in cmd or type it in the location bar of windows explorer, it doesn't work. But you can still enter it by double clicking. Guess we need to reconsider how to deal withi it.

Flow.Launcher.Test/FilesFoldersTest.cs
VictoriousRaptor: It passes my local test. Weird. Should I disable it for CI?

Plugins/Flow.Launcher.Plugin.Explorer/Search/EnvironmentVariables.cs
```diff
@@ -1,83 +1,88 @@
...
         }
 
-        internal static Dictionary<string, string> LoadEnvironmentStringPaths()
+        public static bool HasEnvironmentVar(string search)
         {
-            var envStringPaths = new Dictionary<string, string>(StringComparer.InvariantCultureIgnoreCase);
+            // "c:\foo %appdata%\" returns false
+            var splited = search.Split(Path.DirectorySeparatorChar);
+            return splited.Any(dir => dir.StartsWith('%') && 
+                                        dir.EndsWith('%') &&
```
taooceros, line 38: I don't think we need endswith? Some may like to write `%appdata%\flowlauncher\` balblabla. No need to enforce expansion to go deeper

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -105,19 +109,21 @@ when ActionKeywordMatch(query, Settings.ActionKeyword.IndexSearchActionKeyword)
                 await foreach (var search in searchResults.WithCancellation(token).ConfigureAwait(false))
                     results.Add(ResultManager.CreateResult(query, search));
             }
+            catch (OperationCanceledException)
+            {
+                return results.ToList();
```
taooceros, line 114: Just return an empty list here. Don't bother the Tolist

taooceros:
> > The only other thing I picked up is the use of %USERNAME% in the form of `c:\users\%USERNAME%\downloads`. https://www.elevenforum.com/t/complete-list-of-environment-variables-in-windows-11.11212/ I suppose though you can use %USERPROFILE% to achieve the same as using %USERNAME%
> > Perhaps add an exclusion for %USERNAME%?
> 
> Just found something interesting. If you create a folder called `%appdata%` in some folder like `D:\foo`, then try to cd `D:\foo\%appdata%` in cmd or type it in the location bar of windows explorer, it doesn't work. But you can still enter it by double clicking. Guess we need to reconsider how to deal with it.
> 
> Edit: Only expand when matching `...\%blabla%\...`.

What about check whether the path exists before expanding environmental variable?

Plugins/Flow.Launcher.Plugin.Explorer/Search/EnvironmentVariables.cs
VictoriousRaptor: It matches spilt path (%appdata% and flowlauncher), not the whole path.

taooceros: oh sorry I read it wrongly thanks

VictoriousRaptor:
> > > The only other thing I picked up is the use of %USERNAME% in the form of `c:\users\%USERNAME%\downloads`. https://www.elevenforum.com/t/complete-list-of-environment-variables-in-windows-11.11212/ I suppose though you can use %USERPROFILE% to achieve the same as using %USERNAME%
> > > Perhaps add an exclusion for %USERNAME%?
> > 
> > 
> > Just found something interesting. If you create a folder called `%appdata%` in some folder like `D:\foo`, then try to cd `D:\foo\%appdata%` in cmd or type it in the location bar of windows explorer, it doesn't work. But you can still enter it by double clicking. Guess we need to reconsider how to deal with it.
> > Edit: Only expand when matching `...\%blabla%\...`.
> 
> What about check whether the path exists before expanding environmental variable?

For relative path like %homepath% (relative to %homedrive%) we need to expand it before checking existence.

VictoriousRaptor:
> What about check whether the path exists before expanding environmental variable?

Expanding is before checking the path is incomplete (user inputting and without trailing '\\'). For this bug this pr is enough imo. Even Windows Explorer can't deal with this situation perfectly.

taooceros:
> Expanding is before checking the path is incomplete (user inputting and without trailing '\'). For this bug this pr is enough imo. Even Windows Explorer can't deal with this situation perfectly.

lol ok

VictoriousRaptor:
> > Expanding is before checking the path is incomplete (user inputting and without trailing ''). For this bug this pr is enough imo. Even Windows Explorer can't deal with this situation perfectly.
> 
> lol ok

It can be done but I don't really want to do it. In this pr @kubalav 's path should be correctly parsed. but paths like d:\%appdata% still can't be handled correctly.

Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
+#if !DEBUG
```
taooceros, line 4: why this?

VictoriousRaptor: It was to neutralize warning. `System.Windows` is only used for release. Just realized I should neutralize the warning instead.

Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
```diff
@@ -241,5 +237,33 @@ public static string ReturnPreviousDirectoryIfIncompleteString(string path)
...
+                   && !rel.StartsWith(@"..\")
+                   && !Path.IsPathRooted(rel);
+        }
+        
+        /// <summary>
+        /// Returns path ended with "\"
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns></returns>
+        public static string EnsureTrailingSlash(this string path)
```
jjw24, line 264: nice, this will be helpful because through out explorer there are dup code to add trailing slash.

jjw24 (approved):

Issue #1805, opened 2023-01-16
Enhancement - add Everything setting Match Path


vital-net:
I'd like an option in the everything settings to always search also in the path (Match path), not only the name of the file.

This is not possible right now in Flow Launcher, so I have to use the Everything application instead.

Thank you

TheBestPessimist:
sounds like related to https://github.com/Flow-Launcher/Flow.Launcher/discussions/1800

PR #1848, opened 2023-01-25
Everything Search: Match Path 

TheBestPessimist:
This option makes searching take into account the full path of a file at once, instead of "chunks of the path delimited by "/" (folder separator)".

For example:

I have a folder `D:/all/wallpapers/artstation/pics and nested folders here`

In `Everything`, if i enable `Search` -> `Match Path` (<kbd>Ctrl</kbd>+<kbd>U</kbd>) i can simply type `wall arts` and it shows me all the correct files. It matches the folder `D:/all/wallpapers/artstation/` and everything underneath it.

With FL, before this feature, i have to type `wall*/arts` to have the same. Typing `wall*arts` does not work either.

Basically i have to:
1. add `*` to instruct "match all characters after `wall` until folder separator
2. add `/` (folder separator) because  `*` character does not match it

With this feature, it is enough to type `wall arts`, just like you would type in `Everything`.  Space character works as a `.*` regexp over the full path of a file.

Moreover, this gives you even more flexibility, because now you can type characters from the middle of the path as well, and they will match, for example `all tati` brings you the same result.

Screenshot of the feature in action: 
image

Closes: https://github.com/Flow-Launcher/Flow.Launcher/issues/1805
Closes: https://github.com/Flow-Launcher/Flow.Launcher/discussions/1800

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
@@ -13,13 +13,12 @@ using Flow.Launcher.Plugin.Explorer.Exceptions;
 
 namespace Flow.Launcher.Plugin.Explorer.Search.Everything
 {
-
     public static class EverythingApi
     {
-
         private const int BufferSize = 4096;
 
         private static SemaphoreSlim _semaphore = new(1, 1);
+
         // cached buffer to remove redundant allocations.
         private static readonly StringBuilder buffer = new(BufferSize);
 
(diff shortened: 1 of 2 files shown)


Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
```diff
@@ -39,7 +39,7 @@ public enum StateCode
         /// Gets or sets a value indicating whether [match path].
         /// </summary>
         /// <value><c>true</c> if [match path]; otherwise, <c>false</c>.</value>
-        public static bool MatchPath
+        public static bool MatchPath // TODO these seem to be unused, and they're related exactly to the feature i want: make everything search through the whole path. How should these be used? Should they be removed instead? IMHO, it seems better to use full path search by default, so i'd remove this completely
```
TheBestPessimist, line 42: @taooceros  do you have any opinions here?

taooceros: Thanks! I don't notice it actually🤣. Do we want it to be default or as an option for user to change

taooceros: Actually I would say let's make it default true. I like that feature. Please use that variable though but just make it default true

TheBestPessimist: Guide me here please:
> Actually I would say let's make it default true. I like that feature. Please use that variable though but just make it default true

what do you mean by this?

Moreover, something is unclear to me: why are all those variables statics? Shouldn't there be only 1 instance of `EverythingAPI` (shouldn't `EverythingAPI` be a singleton)?

taooceros: Sorry don't care about it.

Static class is a singleton in some way but if you want to make it singleton I would be happy to merge.

TheBestPessimist: For now, i think we can remove those static completely, as they're unused. I also pushed a commit which adds a flag to `EverythingSearchOption` about full path search. I think it's better than to have it hardcoded

TheBestPessimist: I've gone ahead and removed them.

taooceros (approved):

taooceros:
Thank you! I also really like this feature!

TheBestPessimist:
@taooceros when can i get the new plugin versions? will they auto update, or do you need to publish a new FL version? (ie 1.12.2)

taooceros:
try the new prerelease

Garulf:
This is such a cool way to do it. Awesome!

Issue #1803, opened 2023-01-16
BUG: Autocomplete on Program plugin should only be the name of the app


jjw24:
### Checks

- [X] I have checked that this issue has not already been reported.

### Problem Description

When show description is enabled for Program plugin, tab to autocomplete will complete the query as name + description, this in turn results in the Program plugin result disappearing from the list.

### To Reproduce

1. Enable show description in Program plugin
2. Search Chrome
3. Press tab on the result
4. See that the autocompleted result is "chrome: Google Chrome"

### Screenshots

![image](https://user-images.githubusercontent.com/26427004/212656276-6073e285-bf34-47d8-97c0-2cf0703f74df.png)

### Flow Launcher Version

v1.11.0

### Windows Build Number

Irrelevant

PR #1850, opened 2023-01-25
Fix autocomplete on Program plugin

z1nc0r3:
Fixed the issue mentioned in #1803 which autocompletes the _name + program description_ when the **Show Description** is enabled in the Program plugin.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
@@ -83,6 +83,7 @@ namespace Flow.Launcher.Plugin.Program.Programs
         public Result Result(string query, IPublicAPI api)
         {
             string title;
+            string titleOnly;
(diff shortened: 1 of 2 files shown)


Plugins/Flow.Launcher.Plugin.Program/Programs/Win32.cs
```diff
@@ -83,6 +83,7 @@ private static MatchResult Match(string query, IReadOnlyCollection<string> candi
         public Result Result(string query, IPublicAPI api)
         {
             string title;
+            string titleOnly;
```
taooceros, line 86: I think this variable name doesn't provide any information. and is this variable needed? In either case the value is the same as `resultName`.

z1nc0r3: Yeah, no need for an extra variable. I just noticed that. We can use `resultName` as `AutoCompleteText`.

PR #1891, opened 2023-02-08
Update nuget

deBabbbe:


taooceros:
Please don't update squirrel. The later version doesn't work well.

Issue #1888, opened 2023-02-08
PROPOSAL. Language file. Singular/Plural


dracp:
I believe that different strings should be used to express the term "Plugin" and "Hotkey", singular or plural.

**Plugin/Plugins**

![230208 _FL 1 13 0 6954_Propuesta_Plugin_Plugins](https://user-images.githubusercontent.com/102802425/217555127-9cb2a028-e964-481e-86a4-44a1f50b759f.png)

**Hotkey/Hotkeys**

![230208 _FL 1 13 0 6954_Propuesta_Hotkey_Hotkeys](https://user-images.githubusercontent.com/102802425/217555228-b9fb6410-4dc0-4072-b1d7-571dca1780bb.png)

PR #1895, opened 2023-02-10
Change Singular Hotkey/Plugin title area to Plural (Hotkeys/Plugins)

onesounds:
## What's the PR
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/1888
- Change Hotkey/Plugin title area to plural (Hotkeys / plugins)
- Fix Hotkey tab's singular texts : hotkeys / shortcuts
- Other languages have added key values and pasted existing text.
 
image

image

image

## Test Case
- There should be no typos in the key value.

### DIFF
--- a/Flow.Launcher/Languages/it.xaml
+++ b/Flow.Launcher/Languages/it.xaml
@@ -1,5 +1,8 @@
-﻿<?xml version="1.0"?>
-<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:system="clr-namespace:System;assembly=mscorlib">
+﻿<?xml version="1.0" ?>
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:system="clr-namespace:System;assembly=mscorlib">
     <!--  MainWindow  -->
     <system:String x:Key="registerHotkeyFailed">Impossibile salvare il tasto di scelta rapida: {0}</system:String>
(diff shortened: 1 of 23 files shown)


deefrawley:
Did we want to add the change 'Themes' tab to 'Appearance'?

onesounds:
> Did we want to add the change 'Themes' tab to 'Appearance' to this PR as well?
I forgot. I'll reflect it in this pr.

Flow.Launcher/Languages/it.xaml
```diff
@@ -80,6 +83,7 @@
     <system:String x:Key="searchplugin_Noresult_Title">No results found</system:String>
     <system:String x:Key="searchplugin_Noresult_Subtitle">Please try a different search.</system:String>
     <system:String x:Key="plugin">Plugin</system:String>
+    <system:String x:Key="plugins">Plugins</system:String>
```
taooceros, line 86: Just update `en.xaml` should be fine. This is fairly annoying to update all files

onesounds: You will not understand this. People with low IQs usually work like this.

taooceros: Na I just want to inform you the fact so next time you don't have to do it🥺

onesounds: If I follow you, doesn't every other language display "plugins"?.
I don't want to affect other languages, so I copied all the 'plugin' values into the 'plugins'.
If I didn't want to influence other languages, isn't this the right way to do it?

taooceros: The issue is crowdin won't automatically load translation....and I think it's ok to let translator revises the string if we change the source string. Sometimes crowdin can detect the change from `plugin->plugins` and will revise the translation in other language as well.

taooceros: No no need for that. I am just saying it won't be flushed into crowdin

PR #1897, opened 2023-02-10
Custom Browser managing UI Improvement in Bookmarks Plugin

onesounds:
## What's the PR
While I was testing the related pr myself, I found that bookmark registration was not clear to beginners. There is no error in trying to register the wrong file, and you have to go through cumbersome work to test the behavior of the right bookmark you registered.

- Added guide Texts in custom browser popup window
- Added Edit Button. It will be disable when non-select item situation
- Added Browse Button in Path Select Textbox.

## Test Cases
- [ ] 'The Browse button should operate normally. This button should only allow you to select folders.
- [ ] The Edit button must be disabled in the correct situation.
- [ ] Guide text contains no miss typos and statements are correct.

## Memo
- The folder selection dialog is available only in windows.forms. It was not available as usual, so I found out that I had to add a separate reference to the plug-in. I need to make sure what I did was right.
- When attempting to register, review whether the bookmark location is correct, and if there is a problem, it would be better not to register. (There should also be messages such as 'There is no bookmark data in that location' and 'We have verified and registered bookmark data')

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
+++ b/Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
@@ -64,14 +64,16 @@
     <Compile Remove="Bookmark.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="System.Windows.Forms">
+      <HintPath>..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.8\System.Windows.Forms.dll</HintPath>
+    </Reference>
+  </ItemGroup>
+
   <Target Name="CopyDLLs" AfterTargets="Build">
     <Message Text="Executing CopyDLLs task" Importance="High" />
-    <Copy
-      SourceFiles="$(TargetDir)\runtimes\win-x64\native\SQLite.Interop.dll"
-      DestinationFolder="$(TargetDir)\x64" />
-    <Copy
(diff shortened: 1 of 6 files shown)


Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
```diff
@@ -64,14 +64,16 @@
     <Compile Remove="Bookmark.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="System.Windows.Forms">
+      <HintPath>..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.8\System.Windows.Forms.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
taooceros, line 71: this should not be needed. If you want to use windows form, just add `<UseWindowsForms>true</UseWindowsForms>`.

onesounds: Fixed

taooceros (approved):

onesounds:
@VictoriousRaptor 'auto reload' works fine. but...  we forgot the chrome / edge / firefox checkbox. 😅 
Please make this check box action a trigger for auto reload, too.

onesounds:
> > @VictoriousRaptor 'auto reload' works fine. but... we forgot the chrome / edge / firefox checkboxes. 😅 (it doesn't work with auto reload) Please make this check box action a trigger for auto reload, too.
> 
> Ahh my bad. Please review again.

I see it's working fine. 👏
This PR is mine, so either you or TAOO should review it.

PR #1950, opened 2023-03-03
Everything: update RunCount on file action

TheBestPessimist:
When the user selects `Everything` as search provider, on every "Path Action" (Open File, Open Folder, Open File as Admin) we increase Everything's `RunCount` counter.

When the user chooses `SortOption = RunCount`, this will bring much more meaningful results to the user, because they are sorted such that the files/folders which are accessed most are always at the top of the search.

---

Initially discussed here: https://discord.com/channels/727828229250875472/764905163705155585/1080438121491071077


TheBestPessimist:
Testing: don't forget to change Sorting by `RunCount` desc:
image

Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
```diff
@@ -145,13 +143,25 @@ public static bool FileExists(this string filePath)
             return File.Exists(filePath);
         }
 
+        /*
+         ❓ 
+        todo does it make sense to remove Flow.Launcher.Plugin.IPublicAPI.OpenDirectory (used as Context.API.OpenDirectory(path)),
```
taooceros, line 148: I think this is legacy...That one respect the setting and allow more plugins to utilize the custom explorer feature

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
```diff
@@ -172,5 +172,19 @@ private static void CheckAndThrowExceptionOnError()
                     throw new ArgumentOutOfRangeException();
             }
         }
+
+        public static async Task IncrementRunCounterAsync(string fileOrFolder)
+        {
+            try
+            {
+                await _semaphore.WaitAsync(TimeSpan.FromSeconds(1));
```
taooceros, line 180: This might causes issue. Better keep semaphore out of try. I remember the case that it doesn't wait but release, which causes exception.

TheBestPessimist:
# Comments during development and replies to discord, regarding the wrong ordering of items.

> Sorting is from the Score property
> Result is a hash set.

All scores are 0.

Even though results is a hashset, it is backed by a linked list, so it is ordered. 
In C#, unlike Java, the `HashSet` is backed b a linked list. In Java, if we wanted to have ordering, we are using a different type: `LinkedHashSet`. When a simple `HashSet` was mentioned, i expected that it does not preserve ordering of elements, but I was wrong.

So at least from these 2 things, the results should be displayed in the order received from the list. (which is the correct order).

image


---

_will update this comment as i have more findings_

Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
TheBestPessimist: so what you're suggesting is to remove `FilesFolders.OpenPath`, correct?

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
TheBestPessimist: Can you explain a bit more? I fail to see what the problem is. If you have a SO link which explains the issue, give me that instead.

taooceros: https://stackoverflow.com/questions/24139084/semaphoreslim-waitasync-before-after-try-block

taooceros:
here you go https://github.com/Flow-Launcher/Flow.Launcher/blob/ea6622f6d544f935baf0adc469f6e818831f2aeb/Flow.Launcher/ViewModel/MainViewModel.cs#L1086-L1100

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
TheBestPessimist: Interesting. Thank you for the link. I've never had experience with async/await before and the points there make a lot of sense, in hindsight.

TheBestPessimist: fixed!

Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
TheBestPessimist: Added deprecation notice.

TheBestPessimist:
> here you go

Thank you very much, this is it! 
This is the thing which messes up with my expected sorting.

Bad:
image
Good:
image

And the bonus comes from `_userSelectedRecord.GetSelectedCount`.
image

At this point i believe i should stop my pursuit of "fixing this", as I have the feeling that it is impossible.

TheBestPessimist:
When this is merged, could you please accept the commits and also the description of the MR (ie, not squash)?
I believe having those in the future would be helpful

taooceros:
> When this is merged, could you please accept the commits and also the description of the MR (ie, not squash)? I believe having those in the future would be helpful

Generally we won't squash pr except for some special case.

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -236,33 +221,18 @@ internal static Result CreateFileResult(string filePath, Query query, int score
                 {
                     try
                     {
+                        // TODO Why do we check if file exists here, but not in the other if conditions?
```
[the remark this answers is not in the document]
taooceros: Sorry for the late response. I am not sure, which I believe is a logic flaw.

Flow.Launcher.Plugin/Result.cs
```diff
@@ -178,7 +178,7 @@ public override int GetHashCode()
         /// <inheritdoc />
         public override string ToString()
         {
-            return Title + SubTitle;
+            return Title + SubTitle + Score;
```
taooceros, line 181: what's the reason for this? I think you also need to update the hashcode if it is for replacing old result?

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
```diff
@@ -132,10 +146,12 @@ public static async ValueTask<bool> IsEverythingRunningAsync(CancellationToken t
 
                     var result = new SearchResult
                     {
+                        // todo the types are wrong. Everything expects uint everywhere, but we send int just above. pls fix
                         FullPath = buffer.ToString(),
                         Type = EverythingApiDllImport.Everything_IsFolderResult(idx) ? ResultType.Folder :
                             EverythingApiDllImport.Everything_IsFileResult(idx) ? ResultType.File :
-                            ResultType.Volume
+                            ResultType.Volume,
+                        Score = (int)EverythingApiDllImport.Everything_GetResultRunCount( (uint)idx)
```
taooceros, line 154: I am wondering whether we should give score for other sorting (i.e. `-idx * c` for some constant)

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/SortOption.cs
```diff
@@ -27,7 +27,6 @@ public enum SortOption : uint
         ATTRIBUTES_DESCENDING = 16u,
         FILE_LIST_FILENAME_ASCENDING = 17u,
         FILE_LIST_FILENAME_DESCENDING = 18u,
-        RUN_COUNT_ASCENDING = 19u,
```
taooceros, line 30: well...it is silly though; Will someone want never clicked result???🤣

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -77,11 +78,12 @@ internal static Result CreateFolderResult(string title, string subtitle, string
                 CopyText = path,
                 Action = c =>
                 {
+                    // open folder
                     if (c.SpecialKeyState.CtrlPressed || (!Settings.PathSearchKeywordEnabled && !Settings.SearchActionKeywordEnabled))
                     {
                         try
                         {
-                            Context.API.OpenDirectory(path);
+                            _openFolder(path);
```
taooceros, line 86: is `_openFolder` a function? Let's preserve the naming convention that function should be Pascal Case

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -133,27 +129,22 @@ internal static Result CreateDriveSpaceDisplayResult(string path, string actionK
                 Score = 500,
                 ProgressBar = progressValue,
                 ProgressBarColor = progressBarColor,
-                Action = c =>
+                Action = _ =>
                 {
-                    Context.API.OpenDirectory(path);
+                    _openFolder(path);
```
taooceros, line 134: Same

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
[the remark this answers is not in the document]
taooceros: Let don't check it, and move everything to `_openFile/_openFileAsAdmin` or we can give it an argument `runAsAdmin`. I will let you decide.

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -274,17 +244,12 @@ internal static Result CreateFileResult(string filePath, Query query, int score
...
-                    FullPath = filePath,
-                    WindowsIndexed = windowsIndexed
-                }
+                ContextData = new SearchResult { Type = ResultType.File, FullPath = filePath, WindowsIndexed = windowsIndexed }
             };
             return result;
         }
 
-        public static bool IsMedia(string extension)
+        private static bool _isMedia(string extension)
```
taooceros, line 252: Same naming convention

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -296,10 +261,47 @@ public static bool IsMedia(string extension)
             }
         }
 
-        public static readonly string[] MediaExtensions =
+        private static void _openFile(string filePath)
```
taooceros, line 264: Same naming convention

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -296,10 +261,47 @@ public static bool IsMedia(string extension)
...
+        {
+            _incrementEverythingRunCounterIfNeeded(filePath);
+            FilesFolders.OpenPath(filePath);
+        }
+
+        private static void _openFolder(string folderPath, string fileNameOrFilePath=null)
```
taooceros, line 270: Same

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -296,10 +261,47 @@ public static bool IsMedia(string extension)
...
+        {
+            _incrementEverythingRunCounterIfNeeded(folderPath);
+            Context.API.OpenDirectory(Path.GetDirectoryName(folderPath), fileNameOrFilePath);
+        }
+
+        private static void _openFileAsAdmin(string filePath)
```
taooceros, line 276: Same

taooceros (changes requested):
Thanks for the nice pr! Please check the naming convention. I believe C# suggest using Pascal Case for all type of function (either private and public). I don't think it will gain any benefit to distinguish this two.

Flow.Launcher.Plugin/Result.cs
TheBestPessimist: It's not replacing any old result. I updated `ToString` to have an easier time debugging. Afaics, `ToString` was not used anywhere important.

I don't know how it is in .NET world , but in JVM world, `ToString` has nothing to do with `equals` and `hashCode`. I expect in .NET to be the same.

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
TheBestPessimist: sorry, i don't understand what you're saying

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/SortOption.cs
TheBestPessimist: exactly! that's why i removed that option.

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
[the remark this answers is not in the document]
TheBestPessimist: done

Flow.Launcher.Plugin/Result.cs
taooceros: Yeah it should have nothing to do with equals and hashcode, but I think a long time ago flow uses tostring for the usercount. Anyway I think it's fine.

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs
TheBestPessimist: Feedback copied from discord.

Your changes look like this:

- In Everything API: if we use RunCount, then we use the score from RunCount times 5. if we dont use RunCount, then the score remains 0 (0 * 5 = 0)
so if i would have an entry with RunCount 1000, then the score would be 5000, and if i would have an entry with RunCount 0, then score would be 0
- in Search Manager, we ignore the existing score completely, and instead we simply set a score between [-49, +50], based on the position of the items in the array received from Everything API) (`Score = -i + 50`)

this second part looks like a brain fart (maybe you wanted `Score = Score + (-i + 50)`), but even ignoring that, i am not convinced it's the right way to sort. Using my RunCount 1000 example, this particular scoring would not really fix or change anything in the ordering of the items..

Right now, my suggestion is to remove the commit

TheBestPessimist: commit reverted

taooceros (approved):

jjw24:
@TheBestPessimist flow has mechanism to adjust score based on user selection and bring up the more selected results higher, and I see you have come across it already. As far as I can see, this PR uses Everything's runcount as score now, which means the files and folders that you have most frequently selected in Everything prior to flow will now translate rather than having to repeat the selection of those files and folders in flow several times in order to get those results higher on top. Is this the correct understanding of how this change will work?

TheBestPessimist:
@jjw24 
> will now translate _(from E.RunCount to FL.Score)_
>rather than having to repeat the selection of those files and folders in flow several times _(so that FL.Score is increased)_
> in order to get those results higher on top. 

That is correct: if you used Everything before, then you get all the benefits of RunCount sorting in Flow.
Moreover, from now on if you continue using Everything the RunCount will be increased too, which makes the results better overall.
Thirdly, from my personal usage (of the FL beta builds), this interferes minimally with FL internal scoring, so my concerns above and on Discord that somehow FL interferes with RunCount were overblown.

jjw24:
Let me know if u still need clarification on how the scoring is applied. The short of it is that after plugins return their result scores we do then apply another adjustment before listing does results in flow. This adjustment is for how many times users have triggered individual results so frequently used results are displayed higher up.

TheBestPessimist:
> Let me know if u still need clarification on how the scoring is applied

no need, thank you!

PR #1983, opened 2023-03-14
Add Grid Splitter for preview resizing

onesounds:
resize

## What's the PR
- Changed the size of the preview area to be resizable with mouse dragging.

## Test Case
- [ ] Each area should display correctly when preview is turned on and off.

## Memo
- The size is reset when turned on and off. As long as the process is alive, the size you changed will be retained.
- Is it a good idea to save this size? I'm having a little trouble with this.


Garulf:
Wow!

deefrawley (approved):

Issue #1872, opened 2023-02-01
BUG: Explorer's Ctrl+Enter does not open containing folder  + Wizard's description


jjw24:
### Checks

- [X] I have checked that this issue has not already been reported.

### Problem Description

1. Explorer's Ctrl+Enter does not open containing folder but the actual folder itself. I expect when pressing Ctrl + Enter it should have the same behaviour as opening the containig folder via the folder context menu's open containing folder function.
2. Update Wizard's wording to be plugin specific:

![image](https://user-images.githubusercontent.com/26427004/216158273-283e5b51-a6a8-446b-bd6a-c7cb33672dc4.png)

Update to 'Open A File/Folder's Containing Folder'

### To Reproduce

See description

### Flow Launcher Version

1.12.1

### Windows Build Number

Win 10

### Error Log

<details>

```shell

Replace this line with the important log contents.

```
</details>

PR #2004, opened 2023-03-24
Explorer default enter keypress action: open in file manager option

### LINKED ISSUE #1957: Change Explorer plugin's default behavior for "Enter" key on folders

- Search for a folder (via explorer plugin)
- Hit _Enter_ and flowlauncher 'browses' into the folder, offering the options "Open in default file manager" and all the subfolders present.

I perceive this behavior to be counterintuitive: my expected default Enter-Key action on a folder would be to open the folder rightaway. (Similarly like files or apps are executed rightaway, or bookmarks are opened rightaway).

Moreover, the current behavior is redundant: Hitting _Tab_ produces the exact same effect as hitting _Enter_. 
However, _Tab_ seems perfectly intuitive to me in this regard - in line with the autocomplete behavior across other plugins and also in-line with the behavior of various commandlines and editors.

Change the default behavior of Enter-Key for folders to open the folder right away. 
Leave behavior of Tab-Key unchanged.

Add an option to the explorer plugin, offering users a choice on the behavior of Enter-key, eg:

_Default Enter-Key behavior for folders:_
[ ] Hit _Enter_ to open folder in Default File Manager
[ ] Hit _Enter_ to autocomplete Folder Name and present subfolders

**Additional context**

The flow.launcher global option "Open Results Modifier Key" is also somehow related, as it indeed does directly open the folder (but with an additional key - obviously).

VictoriousRaptor:
## Changes

- Closes #1957 and #1872
- Add `ToModifierKeys` for `ActionContext` to simplify special key state check.
- Fix style of "Search in full path" option
- Add an option to control the default behavior of opening a folder result
  - Auto complete folder name and expand children in Flow (default and current behavior)
  - Open in default file manager
- Fix opening zip in explorer issue
- Hotkey `Ctrl+Shift+Enter` to open folder result in default file manager. Wizard guide added.

## Tests
- Open in default file manager option works.
- Opening a zip file result in Flow triggers the associated program rather than Windows explorer.
- `Ctrl+Enter` opens the parent folder correctly.
- `Ctrl+Shift+Enter` to open folder result in default file manager.

**TODO:** Update docs

### DIFF
--- a/Flow.Launcher/Languages/en.xaml
+++ b/Flow.Launcher/Languages/en.xaml
@@ -349,7 +349,7 @@
     <system:String x:Key="HotkeyLeftRightDesc">Item Navigation</system:String>
     <system:String x:Key="HotkeyShiftEnterDesc">Open Context Menu</system:String>
     <system:String x:Key="HotkeyCtrlEnterDesc">Open A File/Folder's Containing Folder</system:String>
-    <system:String x:Key="HotkeyCtrlShiftEnterDesc">Run as Admin</system:String>
+    <system:String x:Key="HotkeyCtrlShiftEnterDesc">Run as Admin / Open Folder in Default File Manager</system:String>
     <system:String x:Key="HotkeyCtrlHDesc">Query History</system:String>
     <system:String x:Key="HotkeyESCDesc">Back to Result in Context Menu</system:String>
     <system:String x:Key="HotkeyTabDesc">Autocomplete</system:String>


Plugins/Flow.Launcher.Plugin.Explorer/Settings.cs
```diff
@@ -32,6 +32,7 @@ public class Settings
 
         public bool ShowWindowsContextMenu { get; set; } = true;
 
+        public bool DefaultOpenInFileManager { get; set; } = false;
```
jjw24, line 35: should this be called "DefaultOpenFolderInFileManager" or just "OpenFolderInFileManager"

jjw24 (approved):
👍

PR #2031, opened 2023-04-07
Fix image cache and hash correspondence

VictoriousRaptor:
Related to #1727. Mapping between hash and (path, isfullimage) is incorrect.


taooceros (approved):
oh thanks! how much mess did I leave with that change🤣

PR #2045, opened 2023-04-11
add Terminal to shell plugin settings

### LINKED ISSUE #2043: [open shell commands in new terminal tab]

I'd like for shell commands to run in windows terminal and open in new tabs instead of new command prompt windows

I'd like for shell commands to run in windows terminal and open in new tabs instead of new command prompt windows

just use it without this feature

ThinkSalat:
fixes #2043 

Adds Terminal to the list of shells available for running shell commands.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Shell/Settings.cs
+++ b/Plugins/Flow.Launcher.Plugin.Shell/Settings.cs
@@ -36,6 +36,7 @@ namespace Flow.Launcher.Plugin.Shell
         Cmd = 0,
         Powershell = 1,
         RunCommand = 2,
+        Terminal = 3,
 
     }
 }
--- a/Plugins/Flow.Launcher.Plugin.Shell/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Shell/Main.cs
@@ -234,6 +234,22 @@ namespace Flow.Launcher.Plugin.Shell
                     }
                     break;
                 }
+                
+                case Shell.Terminal:
+                {
+                    info.FileName = "wt.exe";
+                    if (_settings.LeaveShellOpen)
+                    {
+                        info.ArgumentList.Add("-NoExit");
+                        info.ArgumentList.Add(command);
+                    }
+                    else
+                    {
+                        info.ArgumentList.Add("-Command");
+                        info.ArgumentList.Add(command);
+                    }
+                    break;
+                }
 
                 case Shell.RunCommand:
                 {


Plugins/Flow.Launcher.Plugin.Shell/Settings.cs
```diff
@@ -36,6 +36,7 @@ public enum Shell
         Cmd = 0,
         Powershell = 1,
         RunCommand = 2,
+        Terminal = 3,
```
TheBestPessimist, line 39: 1. Would it make sense for this to be named `Windows Terminal` ?
2. Do we need translations for this?

ThinkSalat: It's just called Terminal, and flow launcher is windows only, so I don't think the distinction is necessary. As for translations, the other options aren't translated, so it kind of be adding too much responsibility to this PR.

ThinkSalat:
> When I test it I got the following error 图片

Looking into it

Azakidev:
Hi,

This happens because the arguments given are for pwsh, not for windows terminal
It should specify the shell that will open and use the correct arguments if needed

It would look something like this for powershell and cmd respectively:
```cs
case Shell.TerminalPWSH:
{
    info.FileName = "wt.exe";
    info.ArgumentList.Add(pwsh)
    if (_settings.LeaveShellOpen)
    {
        info.ArgumentList.Add("-NoExit");
        info.ArgumentList.Add("-c");
        info.ArgumentList.Add(command);
    }
    else
    {
        info.ArgumentList.Add("-c");
        info.ArgumentList.Add(command);
    }
    break;
}

case Shell.TerminalCMD:
{
    info.FileName = "wt.exe";
    info.ArgumentList.Add(cmd)
    if (_settings.LeaveShellOpen)
    {
        info.ArgumentList.Add("/k");
        info.ArgumentList.Add(command);
    }
    else
    {
        info.ArgumentList.Add("/c");
        info.ArgumentList.Add(command);
    }
    break;
}
```

Ideally it would automatically use the default profile and none of this would be necessary, but I'm not quite sure if it can as of now

Azakidev:
After looking just a bit more
It's pretty much how Powertoys run does it.
[As can be seen in their implementation.](https://github.com/microsoft/PowerToys/blob/main/src/modules/launcher/Plugins/Microsoft.Plugin.Shell/Main.cs)

Hope this helps!

VictoriousRaptor:
Close as #3225 implements it.

Issue #1772, opened 2023-01-09
 Notification platform is unavailable error


raindropsfromsky:
### Checks

- [X] I have checked that this issue has not already been reported.

### Problem Description

Please open new issue in: https://github.com/Flow-Launcher/Flow.Launcher/issues/new
1. upload log file: D:\Downloads\Flow-Launcher-Portable\FlowLauncher\app-1.11.0\UserData\Logs\1.11.0\2023-01-10.txt
2. copy below exception message

Flow Launcher version: 1.11.0
OS Version: Microsoft Windows NT 10.0.22621.0
IntPtr Length: 8
x64: True

Python Path: 
Node Path: 
Date: 01/10/2023 01:21:05
Exception:
System.Runtime.InteropServices.COMException (0x803E0105): The notification platform is unavailable. (0x803E0105)
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|20_0(Int32 hr)
   at ABI.Windows.UI.Notifications.IToastNotificationManagerStaticsMethods.CreateToastNotifier(IObjectReference _obj, String applicationId)
   at Windows.UI.Notifications.ToastNotificationManager.CreateToastNotifier(String applicationId)
   at Microsoft.Toolkit.Uwp.Notifications.ToastNotificationManagerCompat.CreateToastNotifier()
   at Microsoft.Toolkit.Uwp.Notifications.ToastContentBuilder.Show(CustomizeToast customize)
   at Microsoft.Toolkit.Uwp.Notifications.ToastContentBuilder.Show()
   at Flow.Launcher.Notification.Show(String title, String subTitle, String iconPath) in C:\projects\flow-launcher\Flow.Launcher\Notification.cs:line 35
   at Flow.Launcher.PublicAPIInstance.<>c__DisplayClass13_0.<ShowMsg>b__0() in C:\projects\flow-launcher\Flow.Launcher\PublicAPIInstance.cs:line 97
   at System.Windows.Threading.Dispatcher.Invoke(Action callback, DispatcherPriority priority, CancellationToken cancellationToken, TimeSpan timeout)
   at System.Windows.Threading.Dispatcher.Invoke(Action callback)
   at Flow.Launcher.PublicAPIInstance.ShowMsg(String title, String subTitle, String iconPath, Boolean useMainWindowAsOwner) in C:\projects\flow-launcher\Flow.Launcher\PublicAPIInstance.cs:line 95
[...]

PR #2047, opened 2023-04-12
Interim fix for notification platform unavailable error

jjw24:
Since we will be moving to a new UI framework in the future, this is an interim fix for notification platform unavailable error caused by ModernWpfUI version bump from https://github.com/Flow-Launcher/Flow.Launcher/pull/1773. 

ModernWpfUI v0.9.5 introduced [change around WinRT usage](https://github.com/Kinnara/ModernWpf/pull/451) that causes notification platform unavailable error for flow on some machines.

Context:

Tests:
Win 10:
- Notification shown for plugin data reload
- Notification shown for plugin already installed
- Notifications shown for plugin install process
- Plugin Store plugin install pop up is shown
- Resizing preview window
- Increasing/Decreasing width of search window via hotkey
- Increasing/Decreasing maximum displayed result rows via hotkey
Win 11:
Same tests as above
 
Close https://github.com/Flow-Launcher/Flow.Launcher/issues/1772

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -1,6 +1,7 @@
 crowdin
 DWM
 workflows
+Wpf
 wpf
 actionkeyword
 stackoverflow
--- a/Flow.Launcher/Flow.Launcher.csproj
+++ b/Flow.Launcher/Flow.Launcher.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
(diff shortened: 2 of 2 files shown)


taooceros:
I feel like this error only appears in debug? Further, I feel like the root cause is the assemblyloader, in which I think I have made some weird issue but haven't yet got time to resolve.

Though, this might be a good workaround for now if this is consistently reproduced.

jjw24:
It happens for me only in debug, but from the linked issue it happens to some users, and we have had reports that the plugins not installing or only a zip file is downloaded, which is a high probability from the error.

I will take a look at the assembly loader.

jjw24:
@taooceros, yep looks like the change made by the assembly loader has caused this issue, see comment https://github.com/Flow-Launcher/Flow.Launcher/issues/1772#issuecomment-1507584968

taooceros:
Yes. I remember it is very tricky to fix anyway. The temporary workaround is good enough.

taooceros (approved):

PR #2071, opened 2023-04-19
Revert Ctrl+Enter description

VictoriousRaptor:
The `Ctrl+Enter` description has been modified in 9e5e213 and reverted in 2616b70. Reset this entry in localization files to commit e20ac838 to avoid translating it again.


onesounds (approved):

jjw24:
These are reverted by Crowdin because we changed the English version. Do we know if changing it like this is permanent or will Crowdin erase the change on next sync? @taooceros

jjw24:
I think changing on Crowdin make more sense. Last time we changed here it caused some annoyance, but confirm with tao as I am not familiar with Crowdin.

jjw24:
@VictoriousRaptor although, the English content has changed so I don't know how accurate those old translations are, maybe this is the correct intention to have them re-translated?

PR #2082, opened 2023-04-22
Add external preview (QuickLook) support

VictoriousRaptor:
Refined version of #859. Too many merging conflict so I just create a new branch instead. Compatible with our preview function.

## Changes
- Add a property `FilePath` for `Result.Preview` to indicate what file should QuickLook preview.
  - Explorer Plugin support
- An option to choose using QuickLook or not. Default is false.

## Behavior
- Preview Hotkey is used to toggle QuickLook in Flow.
- **If built-in preview is already open, QuickLook won't be triggered before user closes built-in preview. It's to deal with "Always Preview" option and switching between results that supports QuickLook or not.**
- When a result has non-null `Preview.FilePath` try to use QuickLook to preview.
- Flow's preview and QuickLook don't open at the same time. If a result has non-null `Preview.FilePath` pressing hotkey only triggers QuickLook.
- When QuickLook is unavailable, a notification is sent to tell users to check status.
- QL is forcibly closed when hiding flow.
- If Flow "thinks" QL is open, a lost focus won't close Flow. This is to make sure users can interact with QL's UI controls. **But when QL is closed by clicking its close button, opening QL again and interact with it will close Flow.**

## Problems 
- ~~Should "Always Preview" open QuickLook?~~ No
- ~~When "Always Preview" is on (which means internal preview is on), selecting another result that can use QuickLook triggers QuickLook. Looks bad.~~
- ~~When "Use QuickLook" is on but QuickLook is not launched or installed what kind of warning should be send to users?~~ Send a notification
- ~~No QuickLook API to set on top. (https://github.com/Flow-Launcher/Flow.Launcher/pull/859#issuecomment-1028867907)~~ Manually set it.
- ~~Switching from external preview result to builtin result doesn't properly trigger builtin preview.~~

## TODO
- [x] Roughly tested but need some more tests and code review. Generally looking good so far.
- [x] A better glyph in settings.

## Screenshot
image

## Summary by CodeRabbit

- **New Features**
  - Added QuickLook plugin for file previews, enabling functionalities like toggling, opening, closing, and switching previews.
  - Introduced settings for always using the preview feature.

- **Enhancements**
  - Improved preview functionality with new properties and default values for better user experience.
  - Extended support for file previews in various result types (folders, drives, files).

- **Bug Fixes**
  - Fixed visibility bindings in the main window for more consistent UI behavior.

- **Documentation**
  - Updated language resource files with new translations related to QuickLook.


Garulf:
Just got a chance to test run this. It looks very good! I have a few opinions on some of your problems listed:

> Should "Always Preview" open QuickLook?

I don't think this works well with quick look as both windows overlap.

> When "Use QuickLook" is on but QuickLook is not launched or installed what kind of warning should be send to users?

I think we can resolve this by greying out this option if a QuickLook installation is not detected. This can be tricky though as QuickLook is available in many different forms including the Windows app store.

Instead of tracking all this down we could send a pipe message and check its status to see if quick look is available and determine to enable the option.

> No QuickLook API to set on top. (https://github.com/Flow-Launcher/Flow.Launcher/pull/859#issuecomment-1028867907)

This was also an issue with the original PR.

jjw24:
QuickLook functionality should be provided fully by a plugin. To achieve this we need to use this #1013, and this pr is a working POC already.

Garulf:
> QuickLook functionality should be provided fully by a plugin. To achieve this we need to use this #1013, and this pr is a working POC already.

I don't think @VictoriousRaptor was around for that. But yes as a plugin would be ideal. 

Also I would use this PR in favor of my older code. This is more fleshed out.

jjw24:
Maybe best that @VictoriousRaptor just continue work as is with this one so we can keep the momentum, once I helped onesound I will come back to finishing off that pr 1013 and migrate the finished code here into a plugin.

Garulf:
Just did some testing and noticed that if a plugin provides an URL for Flow Launcher to download QuickLook is unable to view this file.

taooceros:
I think it might be hard for a plugin to have a quicklook functionality. Because there need to be a contract to communicate about the file path for previewing. This can only be done either statically by adding an api to Result.cs or dynamically dispatch some data.

Though, I would say this might be a great extensibility. Maybe we can add a field that is dynamic for additional information? I don't know whether that works for jsonrpc though.

Garulf:
I think we should release this feature for the time being and come back to addign it as an extension.

This could be a very helpful feature!

onesounds:
works on the first run, but after trying to move the quicklook pane position, the quicklook pane has a problem (the preview doesn't open properly after that) I have to reopen the quicklook for it to work.

Aside from the window position situation I mentioned, it seems to be working fine. I thought it should be to work independently of the preview settings, but it seems to work fine as it is.

VictoriousRaptor:
> works on the first run, but after trying to move the quicklook pane position, the quicklook pane has a problem (the preview doesn't open properly after that) I have to reopen the quicklook for it to work.

Trying to fix this issue.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -610,6 +568,137 @@ public string PreviewHotkey
...
+                else
+                {
+                    // Should use internal preview for selected result
+                    if (PreviewVisible)
+                    {
+                        Results.SelectedItem?.LoadPreviewImage();
+                    }
+                    else
+                    {
+                        CloseExternalPreview(); // Forcibly close, ideally should only close when it's on
```
VictoriousRaptor, line 684: We should detect QuickLook status here. If it's on then close it and show builtin preview.

jjw24:
@VictoriousRaptor is this draft or ready for review?

VictoriousRaptor:
> @VictoriousRaptor is this draft or ready for review?

It's ready.

bluckholl:
Hi,
Is this will support text files with other extensions, like. sql and. md?
Thanks a lot for your great work 🙏

taooceros:
> Hi, @VictoriousRaptor, Is this will support text files with other extensions, like. sql and. md? Any estimation when it will be available on pre release? If I want to try using this, I need to download this branch and build solution? Thanks a lot for your great work 🙏

You can use the artifact build, but I think they expire now.

VictoriousRaptor:
> Hi, @VictoriousRaptor, Is this will support text files with other extensions, like. sql and. md? Any estimation when it will be available on pre release? If I want to try using this, I need to download this branch and build solution? Thanks a lot for your great work 🙏

Sorry I didn't see this message earlier. I just updated the branch and you can download the artifact build now.

taooceros:
what's left for this one?

jjw24 (commented):
This is good to demonstrate it can work, but I am not fond of having Quicklook specific code in the codebase for several reasons but main one is this doesn't extend/enable flow to be more capable.

I would still say use this PR https://github.com/Flow-Launcher/Flow.Launcher/pull/1013 to allow plugins to capture keyboard shortcuts so they can do more, and move this Quicklook specific code into its own plugin.

taooceros:
> This is good to demonstrate it can work, but I am not fond of having Quicklook specific code in the codebase for several reasons but main one is this doesn't extend/enable flow to be more capable.
> 
> 
> 
> I would still say use this PR https://github.com/Flow-Launcher/Flow.Launcher/pull/1013 to allow plugins to capture keyboard shortcuts so they can do more, and move this Quicklook specific code into its own plugin.

That makes sense. Though before we can capture the keystroke and have a relatively sophisticated model for registering hook for plugins, I think it's ok to temporarily make this inside the core?

jjw24:
> That makes sense. Though before we can capture the keystroke and have a relatively sophisticated model for registering hook for plugins, I think it's ok to temporarily make this inside the core?

Not that complicated, it's using the same model as what we currently have.

My main concern is to release this then having to take it away and have users install a plugin instead.

onesounds:
Is this done? I haven't tested it. If it works, let's just include it.
In my opinion, this functionality does not need to be separated into a plugin. (most programs that support quicklook do. include fluent search.) Just include it as default code. I don't think it's complicated enough to be a plugin, and it don't requires any configuration. Of course, since we're using all the default features as plugins, it would make sense to go with plugins, but it's not required. 
We should not delay features that work.

jjw24:
Last time I tested it it has some bugs and there is no support for downloading and installing Quicklook for user.

VictoriousRaptor:
Maybe the best idea is to use a separate hotkey to trigger QuickLook. Resolving conflicts with QL and our preview is annoying.

jjw24:
Can't this be detected from user settings, preview -> built-in or QuickLook

VictoriousRaptor:
> Can't this be detected from user settings, preview -> built-in or QuickLook

This is the current implementation but logic to deciding to use which preview is not clear. I spent quite a lot of time on it.

I don't have time to review this PR this week so if we want to merge now it requires more detailed review from the team.

jjw24:
Updated this PR with (see commits):
- support plugin with external preview via functionality interface.
- updated preview visibility logic.
- added a QuickLook plugin for testing, but will be a published standalone plugin and not merging with this PR.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -622,7 +622,7 @@ private async void OnDeactivated(object sender, EventArgs e)
                 if (_settings.UseAnimation)
                     await Task.Delay(100);
 
-                if (_settings.HideWhenDeactivated)
+                if (_settings.HideWhenDeactivated && !_viewModel.ExternalPreviewVisible)
```
VictoriousRaptor, line 625: why?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -780,10 +738,169 @@ public string VerifyOrSetDefaultHotkey(string hotkey, string defaultHotkey)
...
+
+            switch (PluginManager.UseExternalPreview())
+            {
+                case true
+                    when ExternalPreviewVisible && CanExternalPreviewSelectedResult(out var path):
+                    SwitchExternalPreview(path, false);
+                    break;
+
+                case true
+                    when !ExternalPreviewVisible && Settings.AlwaysPreview && CanExternalPreviewSelectedResult(out var path):
```
VictoriousRaptor, line 888: When I enable AlwaysPreview QuickLook automatically pops whenever I type. Not usable imo.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -780,10 +738,169 @@ public string VerifyOrSetDefaultHotkey(string hotkey, string defaultHotkey)
...
+
+        private void HideInternalPreview()
+        {
+            ResultAreaColumn = ResultAreaColumnPreviewHidden;
+        }
+
+        public void ResetPreview()
+        {
+            switch (Settings.AlwaysPreview)
+            {
```
VictoriousRaptor, line 861: Why using a switch for a bool value?

Flow.Launcher/MainWindow.xaml.cs
jjw24: This is for when you click away from the mouse it doesnt hide both the search window and preview. Not sure which is user-friendlier, hide when deactivated or do not.

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: Yeah i know, there is the option to switch AlwaysPreview off. The real problem is QuickLook doesnt remember last position, but nothing we can do.

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: True, will update.

Flow.Launcher/ViewModel/MainViewModel.cs
VictoriousRaptor: So AlwaysPreview shouldn't control external preview.

jjw24: I want to align the behaviour for both internal and external so logic is a bit more straightforward, and also not just build logic for QuickLook as in the future we may integrate with another preview tool that works well with always preview option. For now user can just switch it off.

VictoriousRaptor: Is it possible to set a "always external preview" in preview plugin?

jjw24: Hmm, let's not not add additional logic when there is already a button that does exactly that.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -871,7 +871,7 @@ public void ResetPreview()
         {
             if (Settings.AlwaysPreview)
             {
-                ShowPreview();
+                ShowInternalPreview();
```
VictoriousRaptor, line 874: When `AlwaysPreview` is true, specify internal preview.

Plugins/Flow.Launcher.Plugin.QuickLook/Flow.Launcher.Plugin.QuickLook.csproj
```diff
@@ -0,0 +1,51 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
VictoriousRaptor, line 1: If we merge is this plugin going to be a default one?

jjw24: No I will take it out before merge. There is already a QL plugin repo in the org.

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: This is because QL is not suitable for using with always preview right?

VictoriousRaptor: > This is because QL is not suitable for using with always preview right?

Yep

jjw24: What happens when we have other preview tools that is suitable?

VictoriousRaptor: > What happens when we have other preview tools that is suitable?

Well maybe pass always preview to plugin and let plugin decide? If plugin returns false then use internal preview.

jjw24: Done. Let me know if good to merge.

jjw24:
@coderabbitai pause

jjw24:
> **Actionable comments posted: 7**

None applicable.

jjw24 (approved):

PR #2101, opened 2023-04-29
Add Profile support on Bookmarks for Edge browser

dantesg:
This allows user to open bookmarks taking into consideration which profile the bookmark is on.
I use Edge so it's well tested on it. I assume it should also work on Chrome. I don't use Firefox on a daily basis.

Here's what it looks like:
image

### DIFF
--- a/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
@@ -243,6 +243,14 @@ namespace Flow.Launcher.Plugin
         public void OpenUrl(Uri url, bool? inPrivate = null);
 
         /// <summary>
+        /// Opens the URL with the given Uri object. 
+        /// The browser and mode used is based on what's configured in Flow's default browser settings.
+        /// Non-C# plugins can use this method.
+        /// </summary>
+        /// <param name="profileArg">Optional Profile name</param>
+        public void OpenUrl(string url, bool hasProfile, string profileArg);
+        
+        /// <summary>
         /// Opens the URL with the given string. 
         /// The browser and mode used is based on what's configured in Flow's default browser settings.
         /// Non-C# plugins should use this method.
--- a/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
@@ -240,22 +240,14 @@ namespace Flow.Launcher.Plugin
         /// Opens the URL with the given Uri object. 
         /// The browser and mode used is based on what's configured in Flow's default browser settings.
         /// </summary>
-        public void OpenUrl(Uri url, bool? inPrivate = null);
-
-        /// <summary>
-        /// Opens the URL with the given Uri object. 
-        /// The browser and mode used is based on what's configured in Flow's default browser settings.
-        /// Non-C# plugins can use this method.
-        /// </summary>
-        /// <param name="profileArg">Optional Profile name</param>
-        public void OpenUrl(string url, bool hasProfile, string profileArg);
+        public void OpenUrl(Uri url, bool? inPrivate = null, string additionalArgs = "");
         
         /// <summary>
         /// Opens the URL with the given string. 
         /// The browser and mode used is based on what's configured in Flow's default browser settings.
         /// Non-C# plugins should use this method.
         /// </summary>
-        public void OpenUrl(string url, bool? inPrivate = null);
+        public void OpenUrl(string url, bool? inPrivate = null, string additionalArgs = "");
 
         /// <summary>
         /// Opens the application URI with the given Uri object, e.g. obsidian://search-query-example
--- a/Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
+++ b/Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
@@ -1,4 +1,4 @@
-using Microsoft.Win32;
+﻿using Microsoft.Win32;
 using System;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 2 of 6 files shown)


Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -85,6 +76,24 @@ public List<Result> Query(Query query)
...
+            string profileArg = GetProfileArg(c);
+            context.API.OpenUrl(c.Url, true, profileArg);
+            return true;
+        }
+
+        private static string GetProfileArg(Bookmark c)
+        {
+            try
+            {
+                return c.Source.Split('(', ')')[1];
```
taooceros, line 90: what does this split mean?

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
```diff
@@ -93,7 +93,15 @@ public static void OpenInBrowserTab(this string url, string browserPath = "", bo
...
                 {
                     psi.FileName = browserPath;
-                    psi.Arguments = (inPrivate ? $"{privateArg} " : "") + url;
+                    if (inPrivate)
+                    {
+                        psi.Arguments = (inPrivate ? $"{privateArg} " : "") + url;    
+                    }
+                    else
+                    {
+                        psi.Arguments = (!String.IsNullOrEmpty(profileArg) ? $"--profile-directory=\"{profileArg}\" " : "") + url;
```
taooceros, line 102: Will firefox works for this? Though I don't even believe firefox has a good profile system (it has a nice container feature though but that doesn't require this)

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
dantesg: this is just a fancy way to extract the value between those chars

Flow.Launcher.Plugin/SharedCommands/SearchWeb.cs
dantesg: no, I think you're right

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -242,6 +242,14 @@ public interface IPublicAPI
         /// </summary>
         public void OpenUrl(Uri url, bool? inPrivate = null);
 
+        /// <summary>
+        /// Opens the URL with the given Uri object. 
+        /// The browser and mode used is based on what's configured in Flow's default browser settings.
+        /// Non-C# plugins can use this method.
+        /// </summary>
+        /// <param name="profileArg">Optional Profile name</param>
+        public void OpenUrl(string url, bool hasProfile, string profileArg);
```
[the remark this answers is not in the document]
dantesg: that was my first approach but turns out that if I do that, we face ambiguous calls because OpenURL with string and string(default) matches the other OpenURL with only one string.
I didn't want to refactor this to use proper objects because it would go beyond the scope of this PR as well.

dantesg:
@taooceros @VictoriousRaptor I've been using this since then and no issues at all

VictoriousRaptor (changes requested):
Please clarify when opening a bookmark from non-default browser set in Flow, on your current build. Haven't tested it yet but it looks like because API doesn't provide an option to specify browser so wrong profile args are sent to default browser.

~~ I've refactored the OpenUrl API for better future compatibility.  ~~

On second thought we shouldn't touch API in this PR.

dantesg:
> Please clarify when opening a bookmark from non-default browser set in Flow, on your current build. Haven't tested it yet but it looks like because API doesn't provide an option to specify browser so wrong profile args are sent to default browser.
> 
> ~~ I've refactored the OpenUrl API for better future compatibility. ~~
> 
> On second thought we shouldn't touch API in this PR.


@VictoriousRaptor You're correct. Haven't notice this before.
If user uses Chrome and Edge to bookmark and profiles it can run into the following situation.
1. Edge default browser
2. Chrome profile 11 bookmarks google.com
3.1. Flow.Launcher launches google.com from Chrome bookmark -> on default browser -> Edge -> targeting profile 11 - It'll create profile 11 data folder and start Edge from there.
3.2. If profile 11 already exists on Edge, it'll just use that data folder instead and launch Edge from there (which is not correct if you rely on sessions from Chrome).

It's not a scenario I would run into but it's worth considering. Perhaps in the future we could extend the bookmark plugin to also keep track of browser absolute path somehow but for now it only spawns the default browser as-is on current implementation.

Currently if I have a bookmark on Firefox and open it, it'll also open on default browser Edge, not Firefox. 

Happy to close this PR as not suitable to everyone if you think the workflow above can/will be too confusing.

I want to hear your thoughts on this one before putting more work into this one.

VictoriousRaptor:
> It's not a scenario I would run into but it's worth considering. Perhaps in the future we could extend the bookmark plugin to also keep track of browser absolute path somehow but for now it only spawns the default browser as-is on current implementation.
> 
> Currently if I have a bookmark on Firefox and open it, it'll also open on default browser Edge, not Firefox.

So this is what the default browser settings's about. API only provides the function to open a url in default browser.

A reasonable solution is to implement a browser selection feature in Bookmarks plugin to open the bookmark in the browser and profile as is.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
taooceros: Oh I see, so you are getting the content between the parenthesis?

Jack251970:
Implemented in #4402

DavidGBrett:
> Implemented in #4402

Unfortunately I don't think my changes there fully address this PR.
I only added configuration options to the default browser setting, allowing you to set a profile or other flag for all calls.

This PR seems to be more about letting the bookmarks plugin match the browser and profile that created the bookmark to the one that opens it, unless I'm misunderstanding it.

There hasn't been much activity, but if its still a requested feature then perhaps this could be converted into an issue instead.

Jack251970:
> > Implemented in #4402
> 
> Unfortunately I don't think my changes there fully address this PR. I only added configuration options to the default browser setting, allowing you to set a profile or other flag for all calls.
> 
> This PR seems to be more about letting the bookmarks plugin match the browser and profile that created the bookmark to the one that opens it, unless I'm misunderstanding it.
> 
> There hasn't been much activity, but if its still a requested feature then perhaps this could be converted into an issue instead.

Created an issue for tracking in #4419

PR #2103, opened 2023-04-30
Hide auto startup programs and uninstallers

VictoriousRaptor:
Try to hide auto startup shortcuts and uninstallers.


taooceros:
hm why hide autostartup programs? Do they create duplicate?

PR #2122, opened 2023-05-07
Close Firefox bookmark database connection after loading 

### LINKED ISSUE #2120: FlowLauncher.exe locks places.sqlite in Firefox profiles folder even if Browser Bookmarks plugin is disabled.
FlowLauncher.exe locks places.sqlite in Firefox profiles folder even if Browser Bookmarks plugin is disabled. And it prevents to backup Firefox profile.

We can uncheck "Firefox" from "Load Browser From" as a workaround, but it is unexpected behaviour.

VictoriousRaptor:
~~Close #2120~~

Test:
~~- `places.sqlite` is not used by Flow process after loading bookmarks.~~

BTW: why disabled plugins still init?


VictoriousRaptor:
I thought the db connection is not closed but it's not. Can't reproduce the issue.

PR #2123, opened 2023-05-07
Reduce CPU and disk usage during Program plugin init

VictoriousRaptor:
## Changes

- Make use of cache. Add expiration time (30 hours) to program cache on disk. 
  - 24 hours is too short. If I turn on my PC about the same time everyday I'd have a 50% to index every time. 30 hours is basically bi-daily.
  - Note that it doesn't immediately refresh cache when it expires. i.e. If you keep Flow running for more than 30 hours it doesn't reindex. Next reindex is done on next startup. This is intended because this PR is focused on reducing CPU and disk usage during init.
- Stop indexing on init when cache on disk existing and not expired. Significantly reduce CPU and disk usage.
- Save cache right after indexing.

## Tests

- Cache can be saved and loaded.
- No indexing on init when cache is not expired.
- Indexing is done on a first-time startup.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Main.cs
@@ -110,6 +110,8 @@ namespace Flow.Launcher.Plugin.Program
             var win32S = Win32.All(_settings);
             _win32s = win32S;
             ResetCache();
+            _win32Storage.Save(_win32s);
+            _settings.LastIndexTime = DateTime.Now;
         }
 
         public static void IndexUwpPrograms()
@@ -117,6 +119,8 @@ namespace Flow.Launcher.Plugin.Program
             var applications = UWP.All(_settings);
             _uwps = applications;
             ResetCache();
+            _uwpStorage.Save(_uwps);
+            _settings.LastIndexTime = DateTime.Now;
         }
 
         public static async Task IndexProgramsAsync()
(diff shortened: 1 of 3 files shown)


taooceros:
I remember one reason the reindex is performed at startup is that if flow is closed, and someone else install a program during it, and open flow, the new installed program will not shown up.

VictoriousRaptor:
> I remember one reason the reindex is performed at startup is that if flow is closed, and someone else install a program during it, and open flow, the new installed program will not shown up.

Yeah truly. But for many users they just set Flow to auto startup and leave it running all day so any installation should be detected and triggering an reindex.

taooceros:
Though I agree reindex every time startup is kindly very expensive....since now we have file watcher maybe this change it good.

taooceros:
> > I remember one reason the reindex is performed at startup is that if flow is closed, and someone else install a program during it, and open flow, the new installed program will not shown up.
> 
> 
> 
> Yeah truly. But for many users they just set Flow to auto startup and leave it running all day so any installation should be detected and triggering an reindex. 

Yes I agree. Reindex at startup somehow make restart a painful thing to do.

VictoriousRaptor:
Maybe an option to choose force index on startup (default should be false)? And an option to set expiration time?

jjw24:
> I remember one reason the reindex is performed at startup is that if flow is closed, and someone else install a program during it, and open flow, the new installed program will not shown up.

would file watcher trigger an update since it will have detected change?

jjw24:
> I remember one reason the reindex is performed at startup is that if flow is closed, and someone else install a program during it, and open flow, the new installed program will not shown up.

@taooceros would file watcher trigger an update since it will have detected change?

VictoriousRaptor:
> > I remember one reason the reindex is performed at startup is that if flow is closed, and someone else install a program during it, and open flow, the new installed program will not shown up.
> 
> @taooceros would file watcher trigger an update since it will have detected change?

No. File watcher only watches changes when Flow is running.

taooceros:
You are right...it's too annoying when debugging....

taooceros (approved):

PR #2137, opened 2023-05-14
Plugin Store search (should) work after refresh button clicked

### LINKED ISSUE #2049: BUG: Plugin Store can not be searched after clicking refresh button
### Checks

### Problem Description

Plugin Store can not be searched after clicking refresh button

image

### To Reproduce

1. In the Plugin Store search "conv", you will see results come up
2. Click refresh button 
3. Search again for "github", no response, looks like search is not triggered

### Screenshots

_No response_

### Flow Launcher Version

1.14.0

### Windows Build Number

Win 10

### Error Log

<details>

```shell

Nothing in logs

```
</details>

ngtr6788:
Fixes #2049.

Hello everyone! This is my first time contributing to Flow Launcher, and I am very new at C# in general. My goal is (hopefully) get all the latest plugins into the store and show all the most up-to-date installed plugins, and allow the search function to work even after the refresh button is clicked. Can anyone check that the implementation is correct, and if not, let me know if there are any fixes? 

Thanks so much!

### DIFF
--- a/Flow.Launcher/SettingWindow.xaml.cs
+++ b/Flow.Launcher/SettingWindow.xaml.cs
@@ -9,6 +9,7 @@ using Flow.Launcher.ViewModel;
 using ModernWpf;
 using ModernWpf.Controls;
 using System;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Security.Policy;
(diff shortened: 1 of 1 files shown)


taooceros:
Thank you so much for the fix! Would you please check whether it is possible to fix this bug while keeping using Command instead of a method callback? It doesn't seem to matter here but it might be better to keep the code consistent.

ngtr6788:
> Thank you so much for the fix! Would you please check whether it is possible to fix this bug while keeping using Command instead of a method callback? It doesn't seem to matter here but it might be better to keep the code consistent.

Well, turns out it is possible. I just updated the PR :). You can take a look.

VictoriousRaptor (approved):

PR #2143, opened 2023-05-19
Consistent handling of clipboard copying

Garulf:
In this commit, several changes were made to improve the copying functionality in the application.

1. In the `MainViewModel.cs` file, an empty string was replaced with a `CopyText` field. This field is used to store the text to be copied. The logic for copying was also updated. Instead of solely checking if the input string was empty, the code now determines whether the input represents a file or folder. If it does, the file drop list is set on the clipboard with the appropriate path. If not, the input text is set as the data object on the clipboard. This change allows for better handling of different types of content to be copied.

2. In the `MainWindow.xaml.cs` file, the `OnCopy` method was modified. Previously, it only checked if there was no selection in the `QueryTextBox` before calling the `ResultCopy` method with an empty string. Now, it extracts the `CopyText` field from the selected result and passes it to the `ResultCopy` method. This change ensures that the correct text is copied when there is a selected result.

3. In the `PublicAPIInstance.cs` file, the `CopyToClipboard` method was updated. Instead of directly setting the data object on the clipboard, it now utilizes the `ResultCopy` method. This change ensures consistency in the copying functionality across different parts of the application.

Overall, these changes enhance the copying feature in the application by introducing a `CopyText` field, improving handling of file and folder copying, and utilizing the `ResultCopy` method consistently.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -59,9 +59,11 @@ namespace Flow.Launcher
 
         private void OnCopy(object sender, ExecutedRoutedEventArgs e)
         {
-            if (QueryTextBox.SelectionLength == 0)
+            var result = _viewModel.Results.SelectedItem?.Result;
+            if (QueryTextBox.SelectionLength == 0 && result != null)
             {
-                _viewModel.ResultCopy(string.Empty);
+                string copyText = result.CopyText;
+                _viewModel.ResultCopy(copyText);
 
             }
             else if (!string.IsNullOrEmpty(QueryTextBox.Text))
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -68,7 +68,7 @@ namespace Flow.Launcher
             }
             else if (!string.IsNullOrEmpty(QueryTextBox.Text))
             {
-                _viewModel.ResultCopy(QueryTextBox.SelectedText);
+                System.Windows.Clipboard.SetText(QueryTextBox.SelectedText);
             }
         }
         
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -1114,37 +1114,30 @@ namespace Flow.Launcher.ViewModel
         {
             if (string.IsNullOrEmpty(stringToCopy))
             {
-                var result = Results.SelectedItem?.Result;
-                if (result != null)
-                {
-                    string copyText = result.CopyText;
-                    var isFile = File.Exists(copyText);
-                    var isFolder = Directory.Exists(copyText);
-                    if (isFile || isFolder)
-                    {
-                        var paths = new StringCollection
-                        {
-                            copyText
-                        };
-
-                        Clipboard.SetFileDropList(paths);
-                        App.API.ShowMsg(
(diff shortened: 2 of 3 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1114,37 +1114,30 @@ public void ResultCopy(string stringToCopy)
...
-                        App.API.ShowMsg(
-                            $"{App.API.GetTranslation("copy")} {App.API.GetTranslation("textTitle")}",
-                            App.API.GetTranslation("completedSuccessfully"));
-                    }
-                }
-
                 return;
             }
+            var isFile = File.Exists(stringToCopy);
+            var isFolder = Directory.Exists(stringToCopy);
```
VictoriousRaptor, line 1120: What about just checking the strings' format to see if it's a file/dir?

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -59,9 +59,11 @@ public MainWindow()
...
-            if (QueryTextBox.SelectionLength == 0)
+            var result = _viewModel.Results.SelectedItem?.Result;
+            if (QueryTextBox.SelectionLength == 0 && result != null)
             {
-                _viewModel.ResultCopy(string.Empty);
+                string copyText = result.CopyText;
+                _viewModel.ResultCopy(copyText);
 
             }
             else if (!string.IsNullOrEmpty(QueryTextBox.Text))
```
VictoriousRaptor, line 69: When copying `QueryTextBox` it pops up an notification. I think in this case just copy the text without checking if it's a file is enough.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1114,37 +1114,30 @@ public void ResultCopy(string stringToCopy)
         {
             if (string.IsNullOrEmpty(stringToCopy))
             {
```
VictoriousRaptor, line 1116: And please update the docstring of this function.

VictoriousRaptor:
Tested and working.

VictoriousRaptor (approved):

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -59,14 +59,16 @@ public MainWindow()
...
             {
-                _viewModel.ResultCopy(QueryTextBox.SelectedText);
+                System.Windows.Clipboard.SetText(QueryTextBox.SelectedText);
```
jjw24, line 71: Why are we using this Clipboard method and not the viewmodel one?

Garulf: This way avoids wasting time parsing if the text is a file and doesn't send a notification. 

I don't think anyone just copying text from the query wants or needs this.

jjw24: >Copy selected text raises a notification, which is quite unnecessary. Maybe the right way is to remove notification for copy texts?

Yes agree no need for notification. I missed the method being ResultCopy where this is copying query test. 

>And BTW does this PR affects clipboard manager plugins like ClipboardR?

I haven't used the plugin, does it just access copied data or does it also monitor ctrl+c?

jjw24: I have created an issue in the ClipboardR repo to let the author know about the modification.

jjw24: >And BTW does this PR affects clipboard manager plugins like ClipboardR?

I created an issue to let the author know about this change.

PR #2149, opened 2023-05-22
Fix: Add admin permissions when non-admin doesn't work

### LINKED ISSUE #1396: BUG: WindowsSettings fail silently when opening item that requires elevation
### Checks

### Problem Description

WindowsSettings fail silently when opening item that requires elevation. An example is "Edit group policy"

### To Reproduce

_No response_

### Screenshots

_No response_

### Flow Launcher Version

v1.9.4

### Windows Build Number

10.0.19044.1889

### Error Log

ERROR - Flow.Launcher.Plugin.WindowsSettings.Helper.ResultHelper.DoOpenSettingsAction - can't open settings
EXCEPTION OCCURS: System.ComponentModel.Win32Exception (740): The requested operation requires elevation.
   at bool System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at bool System.Diagnostics.Process.Start() x 2
   at bool Flow.Launcher.Plugin.WindowsSettings.Helper.ResultHelper.DoOpenSettingsAction(WindowsSetting entry) in C:/projects/flow-launcher/Plugins/Flow.Launcher.Plugin.WindowsSettings/Helper/ResultHelper.cs:line 179

ngtr6788:
Fixes #1396.

What I am trying to do in this PR is that we run the process as normal, and if the process throws an "elevated permission error" (belonging to Win32Exception), it tries again with elevated permissions, and if that doesn't work, we throw. My (hopeful) reasoning is that we don't want to elevate permissions right from the get-go and only elevate when we need to.

Again, I'm pretty new to the codebase and C# in general, so feedback is greatly appreciated!

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.WindowsSettings/Helper/ResultHelper.cs
+++ b/Plugins/Flow.Launcher.Plugin.WindowsSettings/Helper/ResultHelper.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


taooceros:
Thanks!

taooceros (approved):

PR #2152, opened 2023-05-22
Fix bookmark plugin file monitoring issue

### LINKED ISSUE #2120: FlowLauncher.exe locks places.sqlite in Firefox profiles folder even if Browser Bookmarks plugin is disabled.

### LINKED ISSUE #2138: BUG: Bookmarks plugin cause memory leak
### Checks

### Problem Description

- From https://discord.com/channels/727828229250875472/727828229250875476/1107668589718282270
- Related to #2120

Summary: Flow eating up gigs of RAM after running continously for serveral days to weeks.

Hey, i can't open an issue on github atm, but if you're interested i think i managed to pinpoint what causes the growing memory usage.
It turns out that it's the default `Browser Bookmarks` plugin. The interesting thing is that it slowly eats RAM even if the plugin is turned off, and only stopped when completely removing the plugin (i didn't test if turning off individual browsers in settings has the same effect).
So it seems that there's a bit more to this issue: https://github.com/Flow-Launcher/Flow.Launcher/issues/2120

`Win10`, `Flow v1.15.0`, `Firefox ESR branch`, also i remember few other browsers being enabled by default in plugin so additionally there's also: `Opera` +  `Chrome` installed on the system, but basically never used.

I noticed that having `Browser Bookmarks` either enabled or disabled it sometimes grabs another chunk of RAM when i opened any bookmark in Firefox, but there was no other action that would(?) modify places.sqlite (adding/editing/removing bookmakrs).
When `Browser Bookmarks` was enabled/disabled after flow launcher restart, memory usage slowly grew throughout the day of using Firefox, from the normal 160-180MB after restart to 500MB~ at the end of the day and 1000MB~ at day3, after removing `Browser Bookmarks` i'm at  167,2MB on day2.

I can't tell exactly if there's a bigger issue in Flow that would act the same way, but on a significantly slower rate and `Browser Bookmarks` just being the more obvious suspect, but at the end of the day it's still weird that it even does anything when the plugin is disabled.

### To Reproduce

1. ...
2. ...
3. ...

### Screenshots

_No response_

### Flow Launcher Version

1.15.0

### Windows Build Number

Windows 10

### Error Log

<details>

```shell

```
</details>

VictoriousRaptor:
Fixes #2120 and maybe fixes #2138.

## Changes
- Don't load bookmarks or monitor bookmark files when plugin is disabled
- Remove `LastAccessed` filter to avoid unwanted `RegisterBookmarkFile()` calls

## Tests
- Loading bookmark won't create unwanted recursive calls. Previously because `LastAccessed` filter of the file watcher, an attempt to load bookmarks invokes a second call to load.
- Manual data reload creates file watchers and bookmark changes are correctly monitored.
- Bookmarks are loaded on query when plugin gets enabled later.


Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -22,16 +23,27 @@ public class Main : ISettingProvider, IPlugin, IReloadable, IPluginI18n, IContex
...
+        }
+
+        private static void LoadBookmarksIfEnabled()
+        {
+            if (context.CurrentPluginMetadata.Disabled)
+            {
+                // Don't load or monitor files if disabled
+                // Note: It doesn't start loading or monitoring if enabled later, you need to manually reload data
+                return;
+            }
```
VictoriousRaptor, line 43: We shouldn't call init() when plugin is disabled.

VictoriousRaptor: Yeah but it's not the point of this PR.

VictoriousRaptor: > should we call init every time plugin is reenabled?

Technically we should but some plugins may need to update. like if we don't dispose file watchers when disabling plugins, creating new ones after re enabling may lead to some issues.

JohnTheGr8 (commented):
I agree with @taooceros that the plugin manager should keep track of what plugins have been initialized or not (with a flag). This allows us to skip initializing disabled plugins on start-up. When a plugin is enabled, we can check the flag to determine if its `Init` method should be called or not.

This check does not belong in the `Init` of individual plugins...

VictoriousRaptor:
> I agree with @taooceros that the plugin manager should keep track of what plugins have been initialized or not (with a flag). This allows us to skip initializing disabled plugins on start-up. When a plugin is enabled, we can check the flag to determine if its `Init` method should be called or not.
> 
> This check does not belong in the `Init` of individual plugins...

Then we need some API change. init method passes context object to plugin, which is necessary for localized plugin names and descriptions.

JohnTheGr8:
> 
> Then we need to change all plugins. 

Why do we?

JohnTheGr8:
> i18n plugins uses plugincontext.API to retrieve localized plugin names. if we simply stop invoking init method of disabled plugins the plugin context is null and raises errors.

What I said was that we should invoke their `Init` method if they get enabled later on...

JohnTheGr8:
> Yeah. Anyway should I remove the check in this PR? It's going off-topic and checking it in `Init` is not a good idea.

This isn't off-topic, you have made a change that introduces a different issue (when the plugin is enabled later, as you noted) 😉

Anyway, I'm okay with keeping this as a temporary solution, but let's add a quick flag inside the bookmarks plugin that tracks if bookmarks have been loaded or not. At the top of `Query` check this flag and load bookmarks if necessary...

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -22,20 +23,41 @@ public class Main : ISettingProvider, IPlugin, IReloadable, IPluginI18n, IContex
...
             Main.context = context;
-            
+
             _settings = context.API.LoadSettingJsonStorage<Settings>();
 
-            cachedBookmarks = BookmarkLoader.LoadAllBookmarks(_settings);
+            LoadBookmarksIfEnabled();
 
-            _ = MonitorRefreshQueue();
+            initialized = true;
```
JohnTheGr8, line 37: `initialized = true;` should be moved inside `LoadBookmarksIfEnabled` at the end of the method.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -22,20 +23,41 @@ public class Main : ISettingProvider, IPlugin, IReloadable, IPluginI18n, IContex
...
+            cachedBookmarks = BookmarkLoader.LoadAllBookmarks(_settings);
+            _ = MonitorRefreshQueueAsync();
         }
 
         public List<Result> Query(Query query)
         {
+            if (!initialized)
+            {
+                LoadBookmarksIfEnabled();
+                initialized = true;
```
JohnTheGr8, line 58: similarly, this line should be moved to the end of `LoadBookmarksIfEnabled`

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -88,35 +110,49 @@ public List<Result> Query(Query query)
...
         {
             var directory = Path.GetDirectoryName(path);
-            if (!Directory.Exists(directory))
+            if (!Directory.Exists(directory) || !File.Exists(path))
+            {
                 return;
-            var watcher = new FileSystemWatcher(directory!);
-            if (File.Exists(path))
+            }
+            if (context.CurrentPluginMetadata.Disabled)
```
JohnTheGr8, line 143: this check is not necessary, `RegisterBookmarkFile` will not be called unless the plugin is enabled.

JohnTheGr8 (changes requested):

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
JohnTheGr8: to be clear, I mean this check:

```csharp
if (context.CurrentPluginMetadata.Disabled)
{
    return;
}
```

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -143,8 +172,8 @@ public void ReloadData()
         public static void ReloadAllBookmarks()
         {
             cachedBookmarks.Clear();
-
-            cachedBookmarks = BookmarkLoader.LoadAllBookmarks(_settings);
+            DisposeFileWatchers();
```
JohnTheGr8, line 175: is there a reason you dispose all file watchers and re-initialize them on every bookmark reload?

VictoriousRaptor: if user unchecks a browser then we don't need to monitor the browser anymore.

JohnTheGr8: yeah but `ReloadAllBookmarks` will also get called when either of the monitored files changes...

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -88,35 +107,45 @@ public List<Result> Query(Query query)
...
+            if (fileMonitorSemaphore.CurrentCount < 1)
+            {
+                return;
+            }
+            await fileMonitorSemaphore.WaitAsync();
             var reader = refreshQueue.Reader;
             while (await reader.WaitToReadAsync())
             {
-                await Task.Delay(2000);
+                await Task.Delay(5000);
```
JohnTheGr8, line 122: Do we need this delay? I've tested without it and had no issues...

VictoriousRaptor: > Do we need this delay? I've tested without it and had no issues...

IDK to be honest. May be for cases like modifiying mutiple bookmarks in a short time?

JohnTheGr8: That should be taken care of by the bounded channel we use as a queue.

@taooceros do we remove the delay?

taooceros: I think it is for multiple event triggering. When I test it, it seems like there will be multiple event being triggered in some cases. I am not sure about the execution order, so maybe the channel bound won't help. However, it shouldn't be a large issue. I think the delay can be very short or even removed. It's not gonna make a big deal (even reindex twice is fast).

taooceros: @VictoriousRaptor maybe a very short delay (500ms) will suffice. Have you experienced that this method got called twice?

VictoriousRaptor: > @VictoriousRaptor maybe a very short delay (500ms) will suffice. Have you experienced that this method got called twice?

I can't remember my exact test result but after removing `LastAccessed` notify filter in  d0289b49b669140b4e365817b0a452fe2f7e98ad it's no longer called twice. Sorry but VS is broken somehow and I can't test it now.

JohnTheGr8:
@VictoriousRaptor let's remove that delay and get the branch up-to-date with dev

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
taooceros: Ah sounds good. Then let's remove it.

JohnTheGr8 (approved):
Let's get another pair of eyes on this before merging

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -20,20 +21,38 @@ public class Main : ISettingProvider, IPlugin, IReloadable, IPluginI18n, IContex
...
+            }
+
+            cachedBookmarks = BookmarkLoader.LoadAllBookmarks(_settings);
+            _ = MonitorRefreshQueueAsync();
+            initialized = true;
         }
 
         public List<Result> Query(Query query)
         {
+            if (!initialized)
```
jjw24, line 51:
```suggestion
            // For when the plugin being previously disabled and is now renabled 
            if (!initialized)
```

JohnTheGr8 (approved):

PR #2174, opened 2023-06-06
Improvements for re-queries

JohnTheGr8:
This quick PR introduces the following changes:

1. the `IsReQuery` property is added to the `Query` class, which will be set to `true` whenever a Query is re-executed.
2. the `Ctrl + R` key binding is added for quickly re-executing the current query.

This enables `Ctrl + R` to act as a "refresh" mechanism. When users type a query and the shown results seem out-of-date (for example, due to caching), they can hit `Ctrl + R` to get fresh results.

### Example use-case

The Github plugin caches all query results to avoid rate-limits (and because the Github API is pretty slow). When the user tries to list the issues for a repository, they are likely to get cached results, and therefore recently created issues may be missing. `Ctrl + R` can be used to signal the plugin to skip the cache and load the repository's issues from the API.

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -210,7 +210,7 @@ namespace Flow.Launcher.ViewModel
         {
             if (SelectedIsFromQueryResults())
             {
-                QueryResults(reQuery: true);
+                QueryResults(isReQuery: true);
             }
         }
 
@@ -504,8 +504,8 @@ namespace Flow.Launcher.ViewModel
         /// but we don't want to move cursor to end when query is updated from TextBox
         /// </summary>
         /// <param name="queryText"></param>
-        /// <param name="reQuery">Force query even when Query Text doesn't change</param>
-        public void ChangeQueryText(string queryText, bool reQuery = false)
+        /// <param name="isReQuery">Force query even when Query Text doesn't change</param>
+        public void ChangeQueryText(string queryText, bool isReQuery = false)
         {
             Application.Current.Dispatcher.Invoke(() =>
             {
@@ -519,9 +519,9 @@ namespace Flow.Launcher.ViewModel
                     QueryTextCursorMovedToEnd = false;
 
                 }
-                else if (reQuery)
+                else if (isReQuery)
                 {
-                    Query(reQuery: true);
+                    Query(isReQuery: true);
                 }
                 QueryTextCursorMovedToEnd = true;
             });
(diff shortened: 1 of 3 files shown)


taooceros:
Interesting, but I also remember that f5 acts like reload plugin data? Will the two feature overlap?

Garulf:
What happens to the old F5 behavior?

Is this moved to a different hotkey?

taooceros:
Sounds very useful. I am only a little bit worrying whether users may get confused between these two.

Garulf:
I doubt many users even know about F5

jjw24:
The reload data function provides an interface for plugins, when implemented they can write their own method of reloading their data. It is a call to all plugins that implemented the interface. 

It is used for users that want to manually trigger a reload to update their plugin data. For example I don't think the PluginsManager has it implemented, if not I actually want to add it so we can press F5 and have the manifest manually updated.

It is async operation so from my experience all plugins finish very fast within a sec or two, so I have had a chance to look at the code in this PR but I think instead of introducing a new hotkey, can the plugin not just use the reload interface to refresh cache?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -747,6 +756,8 @@ private async void QueryResults()
             if (currentCancellationToken.IsCancellationRequested)
                 return;
 
+            // Update the query's IsForced property to true if this is a re-query
+            query.IsForced = reQuery;
```
TheBestPessimist, line 760: Suggestion to rename to `IsReQuery` from `IsForced`. You used `reQuery` everywhere, except for the flag name, both in the code and in descriptions and comments.

JohnTheGr8: yeah point taken, I myself was on the fence about how to name it ;)

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -716,7 +725,7 @@ private void QueryHistory()
 
         private readonly IReadOnlyList<Result> _emptyResult = new List<Result>();
 
-        private async void QueryResults()
+        private async void QueryResults(bool reQuery = false)
```
jjw24, line 728: Should we call these isReQuery so it's consistent with the query property?

JohnTheGr8: I used `reQuery` to match the previously named parameters in:
* `ChangeQueryText(string queryText, bool reQuery = false)` 
* `IPublicAPI.ChangeQuery(string query, bool requery = false)`

think I should change it?

jjw24: Ideally yes, but will this break any plugins that use this method?

JohnTheGr8: probably not during runtime but it could cause build issues. I can update all except the `IPublicApi` method...

jjw24: yeah let's do that.

PR #1116, opened 2022-04-06
merge shobu13/windowsTerminalBasicSupport To Dev

shobu13:
add a basic terminal for Windows Terminal, 
with the current version, you can set Windows Terminal as your default terminal emulator, but if you launch a command using Flow, Windows Terminal don't use your the matching profile, but a default profile instead.

So i've added a "use windows terminal" property to specify usage of WT, and a Textbox where you can input the Profile you want to use. 

Then, on query process, the plugin call WT, passing the Profile as -p option, followed by the actual query.

PR #2183, opened 2023-06-12
plugin/shell: support Pwsh (PowerShell 7.x)

JohnTheGr8:
Add `Pwsh` as a shell option. Powershell users would probably prefer to use Pwsh over the old `(Windows) Powershell` that ships with Windows.

This raises everyone's favorite problem: how to name each case to distinguish between v5 (`Windows Powershell`) and v7.x (`Powershell (Core)` or `Pwsh`). Thoughts welcome...

( changes partly picked from #1116 )

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Shell/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Shell/Main.cs
@@ -237,6 +237,19 @@ namespace Flow.Launcher.Plugin.Shell
                     break;
                 }
 
+                case Shell.Pwsh:
+                {
+                    info.FileName = "pwsh.exe";
+                    if (_settings.LeaveShellOpen)
+                    {
+                        info.ArgumentList.Add("-NoExit");
+                    }
+                    info.ArgumentList.Add("-Command");
+                    info.ArgumentList.Add(command);
+
+                    break;
+                }
+
                 case Shell.RunCommand:
                 {
                     var parts = command.Split(new[]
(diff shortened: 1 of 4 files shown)


JamesNZL:
*yes please*.

+1 for `pwsh`—imo, best to match what already exists (ie `pwsh.exe` for v7, vs `powershell.exe` for v5).

taooceros:
Maybe let's also add an option to add custom shell? Other shell is also pretty good like nushell. One concern of this is that  pwsh is not installed in windows by default. I think a textbox to customize shell might be better?

JohnTheGr8:
> Maybe let's also add an option to add custom shell? Other shell is also pretty good like nushell. 

Sure, but let's do that in a different PR. Could be added along with Windows Terminal support.

> One concern of this is that pwsh is not installed in windows by default.

That's a good point... should we handle this or trust that the user knows not to pick `Pwsh` when it is not already installed?

> I think a textbox to customize shell might be better?

Can't just do that, each shell has different CLI options for no-exit, passing the sub-command to execute, etc

JohnTheGr8:
> > I think a textbox to customize shell might be better?
> 
> Can't just do that, each shell has different CLI options for no-exit, passing the sub-command to execute, etc

to add to this: I think it would be best if we add Windows Terminal support and implement profile selection, as discussed in #1116. Then users can configure all their preferences from Terminal (what shell to use, etc) and Flow just uses that without complicating things on our end...

taooceros (approved):

PR #2185, opened 2023-06-12
AppVeyor deployment enhancements

### LINKED ISSUE #2086: BUG: WinGet release uploads with the wrong hash
### Checks

### Problem Description

WinGet release uploads with the wrong hash, most likely due to dual release steps in the appveyor.yml on merge to master.

https://github.com/microsoft/winget-pkgs/pull/103060#discussion_r1173218090 

### To Reproduce

1. ...
2. ...
3. ...

### Screenshots

_No response_

### Flow Launcher Version

0

### Windows Build Number

0

### Error Log

<details>

```shell

```
</details>

JohnTheGr8:
Starting a PR following the discussion in #2086 

with this modified setup:

1. on commit to master (ie when a release PR is merged), AppVeyor will:
  a. create a draft release on the main repository
  b. publish a release on the `Prereleases` repository (otherwise it falls behind when a stable version is published)
2. the draft release gets reviewed & tested, release notes are written
3. when the app is ready, the draft release can be published (manually). This should:
  a. trigger the WinGet release action
  b. create the git tag, which will trigger an AppVeyor build that will only deploy the `Flow.Launcher.Plugin` package to nuget (when it has been changed)

Some notes:

* releases are **only published once** (manually), **never force-updated** from a CI job... should resolve #2086 
* **during step 2**, hot-fixes can be applied to the master branch, **without bumping the app version**.
* the build-on-tag is still used for pushing to nuget, because: 
  1. we don't really care if it takes a few minutes more to deploy to nuget
  2. we want deployment to nuget to be the final step in the release process
  3. nuget packages are immutable: if we published on commit to master, we wouldn't be able to support hot-fixes 

sanity testing is still TODO but let me know what you think

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -72,15 +72,6 @@ deploy:
       secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
     artifact: Squirrel Installer, Portable Version, Squirrel nupkg, Squirrel RELEASES
     draft: true
-    force_update: true
     on:
       branch: master
-
-  - provider: GitHub
-    release: v$(flowVersion)
-    auth_token:
-      secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
-    artifact: Squirrel Installer, Portable Version, Squirrel nupkg, Squirrel RELEASES
-    force_update: true
-    on:
-      APPVEYOR_REPO_TAG: true
+      APPVEYOR_REPO_TAG: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -63,6 +63,18 @@ deploy:
       branch: dev
 
   - provider: GitHub
+    repository: Flow-Launcher/Prereleases
+    release: v$(flowVersion)
+    description: 'This is a release candidate of v$(flowVersion)'
+    auth_token:
+      secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
+    artifact: Squirrel Installer, Portable Version, Squirrel nupkg, Squirrel RELEASES
+    force_update: true
+    on:
+      branch: master
+      APPVEYOR_REPO_TAG: false
+
+  - provider: GitHub
     release: v$(flowVersion)
     auth_token:
       secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,10 +4,6 @@ init:
 - ps: |
       $version = new-object System.Version $env:APPVEYOR_BUILD_VERSION
       $env:flowVersion = "{0}.{1}.{2}" -f $version.Major, $version.Minor, $version.Build
-      if ($env:APPVEYOR_REPO_BRANCH -eq "dev")
-      {
-        $env:prereleaseTag = "{0}.{1}.{2}.{3}" -f $version.Major, $version.Minor, $version.Build, $version.Revision
-      }
 - sc config WSearch start= auto # Starts Windows Search service- Needed for running ExplorerTest
 - net start WSearch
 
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -4,10 +4,6 @@ init:
 - ps: |
       $version = new-object System.Version $env:APPVEYOR_BUILD_VERSION
       $env:flowVersion = "{0}.{1}.{2}" -f $version.Major, $version.Minor, $version.Build
-      if ($env:APPVEYOR_REPO_BRANCH -eq "dev")
-      {
-        $env:prereleaseTag = "{0}.{1}.{2}.{3}" -f $version.Major, $version.Minor, $version.Build, $version.Revision
-      }
```
jjw24, line 10: Why is this removed? We need to add the revision version for prereleases.

appveyor.yml
```diff
@@ -67,20 +63,23 @@ deploy:
...
     release: v$(flowVersion)
+    description: 'This is a release candidate of v$(flowVersion)'
     auth_token:
       secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
     artifact: Squirrel Installer, Portable Version, Squirrel nupkg, Squirrel RELEASES
-    draft: true
     force_update: true
     on:
       branch: master
+      APPVEYOR_REPO_TAG: false
```
jjw24, line 75: Is this needed? Does this mean trigger on master or non-tagged builds?

appveyor.yml
```diff
@@ -67,20 +63,23 @@ deploy:
...
     release: v$(flowVersion)
     auth_token:
       secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
     artifact: Squirrel Installer, Portable Version, Squirrel nupkg, Squirrel RELEASES
-    force_update: true
+    draft: true
     on:
-      APPVEYOR_REPO_TAG: true
+      branch: master
+      APPVEYOR_REPO_TAG: false
```
jjw24, line 85: Same as above comment

appveyor.yml
```diff
@@ -67,20 +63,23 @@ deploy:
...
     on:
       branch: master
+      APPVEYOR_REPO_TAG: false
 
   - provider: GitHub
     release: v$(flowVersion)
     auth_token:
       secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
     artifact: Squirrel Installer, Portable Version, Squirrel nupkg, Squirrel RELEASES
-    force_update: true
```
jjw24, line 84: We have never had to do a hot fix during release process, but in the case we that we do have to, this would mean it will not override the release artefacts right? I suppose you need to delete the draft then commit the hot fix to master in order for the new artefacts to get uploaded.

jjw24, line 84: We have never had to do a hot fix during release process, but in the case we that we do have to, this would mean it will not override the release artefacts right? I suppose you need to delete the draft then commit the hot fix to master in order for the new artefacts to get uploaded.

appveyor.yml
JohnTheGr8: not removed, just replaced with $APPVEYOR_BUILD_VERSION, they are exactly the same

appveyor.yml
JohnTheGr8: yup, exactly right

appveyor.yml
jjw24: That's fine make sure you test it, if I remember correctly when I updated this, that pipeline variable returned only the build number part of the version.

JohnTheGr8: see line 5 above, `$env:prereleaseTag` was derived from `$env:APPVEYOR_BUILD_VERSION`

appveyor.yml
JohnTheGr8: What I want is for it to trigger on master & tag, but I need to test to make sure that's how it behaves

jjw24: If I am not mistaking, I think you are trying to cover a scenario (doing a hot fix during a release) that has never happened. We have had several times where we needed to immediately rollout a hot fix after a release, but never during.

Assuming a tag is only created after setting the draft release to latest, you can just cater for:
1. Merge to master -> draft release, so 'on master' condition
2. Setting draft release to latest -> trigger 'on tag' -> publish NuGet and Prerelease

I think this is a simpler pipeline process. Any hot fixes we just do a new release. If we need to really hot fix during a release, I can just delete draft release, then merge hot fix in, which will kick off the above process again.

You will find testing this is a painful process because our build & test takes a while, recommend fork the project and the prerelease project, comment out the unit test code and test that way. Also I might be wrong but off the top of my head I think the 'on' condition you got is 'master' or 'tag is false'.

JohnTheGr8: > If I am not mistaking, I think you are trying to cover a scenario (doing a hot fix during a release) that has never happened. We have had several times where we needed to immediately rollout a hot fix after a release, but never during.

I am not trying to cover that scenario, really... it just happens to be possible with this setup, so I'm spelling it out.

> Assuming a tag is only created after setting the draft release to latest, you can just cater for:
> 
>     1. Merge to master -> draft release, so 'on master' condition
> 
>     2. Setting draft release to latest -> trigger 'on tag' -> publish NuGet and Prerelease

so you suggest that we push to Prereleases when `APPVEYOR_REPO_TAG: true` right? 

> Also I might be wrong but off the top of my head I think the 'on' condition you got is 'master' or 'tag is false'.

yeah it could be so, I'll define a different env variable for this to be sure

jjw24: >so you suggest that we push to Prereleases when APPVEYOR_REPO_TAG: true right?

Yeah. I don't think it matters if it's pushed on merge to master or when APPVEYOR_REPO_TAG: true. If on merge to master you can set force push to true so hot fix during a release will still update the prerelease artefacts.

JohnTheGr8: release artifacts wont be exactly the same but that's no big deal

Jack251970:
#2616

PR #2202, opened 2023-06-24
fix: :bug: check autocomplete and copy text for `Result` equality

### LINKED ISSUE #2201: `Result#Equals()` does not check `AutoCompleteText` or `CopyText`
**Point to the corresponding file and line number(s) which could be improved.**

`Result.cs`, lines `162`–`165`

https://github.com/Flow-Launcher/Flow.Launcher/blob/e1c63af06184d3993128da09373307566265e425/Flow.Launcher.Plugin/Result.cs#L162C13-L162C13

Provide your alternative implementation idea:

```csharp
public override bool Equals(object obj)
{
	var r = obj as Result;

	var equality = string.Equals(r?.Title, Title) &&
                 string.Equals(r?.SubTitle, SubTitle) &&
                 string.Equals(r?.AutoCompleteText, AutoCompleteText) &&
                 string.Equals(r?.CopyText, CopyText) &&
                 string.Equals(r?.IcoPath, IcoPath) &&
                 TitleHighlightData == r.TitleHighlightData;

	return equality;
}
```

JamesNZL:
Fixes #2201.

The current equality implementation for `Result` does not check for differences in `AutoCompleteText` or `CopyText`.

### DIFF
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -172,9 +172,16 @@ namespace Flow.Launcher.Plugin
         /// <inheritdoc />
         public override int GetHashCode()
         {
-            var hashcode = (Title?.GetHashCode() ?? 0) ^
-                           (SubTitle?.GetHashCode() ?? 0);
-            return hashcode;
+            unchecked
+            {
+                // 17 and 23 are prime numbers
+                int hashcode = 17;
+                hashcode = hashcode * 23 + (Title?.GetHashCode() ?? 0);
+                hashcode = hashcode * 23 + (SubTitle?.GetHashCode() ?? 0);
+                hashcode = hashcode * 23 + (AutoCompleteText?.GetHashCode() ?? 0);
+                hashcode = hashcode * 23 + (CopyText?.GetHashCode() ?? 0);
+                return hashcode;
+            }
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


taooceros:
What about hashcode?

JamesNZL:
> What about hashcode?

...and that is why I shouldn't stay up working until 2AM :joy:

Fixed :)

VictoriousRaptor:
Consider using HashCode.Combine()?

JamesNZL:
> Consider using HashCode.Combine()?

Ah cool, I didn't know about that method.
Updated.

taooceros (approved):

Flow.Launcher.Plugin/Result.cs
```diff
@@ -170,9 +173,7 @@ public override bool Equals(object obj)
         /// <inheritdoc />
         public override int GetHashCode()
         {
-            var hashcode = (Title?.GetHashCode() ?? 0) ^
-                           (SubTitle?.GetHashCode() ?? 0);
-            return hashcode;
+            return HashCode.Combine(Title, SubTitle, AutoCompleteText, CopyText, IcoPath);
```
VictoriousRaptor, line 176: Do you have any reason for comparing icopath?

VictoriousRaptor: oh nvm. Because Equals() checks it.

PR #2216, opened 2023-07-03
Add Plugin API -> VisibilityChangedEventHandler

Odotocodot:
Added a visibility changed event handler to the public api.

Main reason for this is to allow caching in query method without the need of a timer.


taooceros:
Sounds good to me. @JamesNZL you may like this one

JamesNZL:
Ooh, love it!

taooceros:
@jjw24 Maybe we can add this to the recent release. There isn't much code change and should not break anything.

Odotocodot:
> @jjw24 Maybe we can add this to the recent release. There isn't much code change and should not break anything.

Was going ask if I did the PR early enough for this to happen 😄

jjw24:
Yeah I don't see why not. I will take a look soon.
Do you have an example implementing this?

Odotocodot:
Not currently, but I can whip one up quickly if you need?

jjw24:
Yes please, could you whip one up with quickly, I want to test to see how it works in a plugin.

Odotocodot:
Is this okay?
The changes that it would do for me -> https://github.com/Odotocodot/Flow.Launcher.Plugin.OneNote/commit/bdf213b801d6ee5731dd9d472df870c123253813 ([Whole File](https://github.com/Odotocodot/Flow.Launcher.Plugin.OneNote/blob/bdf213b801d6ee5731dd9d472df870c123253813/Main.cs))

jjw24:
Worth noting, subscribing to this event means the plugin will continue to be subscribed even if the it is disabled. This however is a shortcoming of Flow's own disable plugin mechanism so at this stage not much we can do for this feature. 

Maybe in the future disabling a plugin should also call System.IDisposable on plugins so they can unsubscribe from this feature.

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -96,7 +96,12 @@ public interface IPublicAPI
         /// </summary>
         /// <returns></returns>
         bool IsMainWindowVisible();
-        
+
+        /// <summary>
+        /// Invoked when the visibility of the main window has changed
+        /// </summary>
```
jjw24, line 102:
```suggestion
        /// <summary>
        /// Invoked when the visibility of the main window has changed. Currently, the plugin will continue to be subscribed even if it is turned off. 
        /// </summary>
```

Flow.Launcher.Plugin/VisibilityChangedEventHandler.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace Flow.Launcher.Plugin
```
jjw24, line 3: Is there any reason why this is in a file of its own instead of being placed in the EventHandler.cs file?

Odotocodot: No reason! Though I originally thought EventHandler.cs was only for key events.

jjw24: Yeah, mostly key events so far, but I would expect to find event handlers in a file with that filename.

Odotocodot: Alright! Happy for it to move 😄

Odotocodot: All done

JohnTheGr8:
fixed the build problem for you ;)

Odotocodot:
> fixed the build problem for you ;)

Thanks! Beat me too it! ;)

jjw24 (approved):

PR #2218, opened 2023-07-04
Fix artefact hash discrepancy for WinGet release

jjw24:
Context:
Currently project is rebuilt and output artefacts re-uploaded after marking a draft release as latest. This creates a discrepancy for WinGet publish process. https://github.com/Flow-Launcher/Flow.Launcher/issues/2086

Solution:
Tag is created after marking a draft release as prerelease or latest. No need to upload the artefacts again on tag.

If changes need to occurr during release process, delete draft then merge new changes in.

Related #2185

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -75,12 +75,3 @@ deploy:
     force_update: true
     on:
       branch: master
-
-  - provider: GitHub
-    release: v$(flowVersion)
-    auth_token:
-      secure: ij4UeXUYQBDJxn2YRAAhUOjklOGVKDB87Hn5J8tKIzj13yatoI7sLM666QDQFEgv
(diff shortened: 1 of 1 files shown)


JohnTheGr8:
I'm quite sure the first Github deployment job will _still_ run on tag CI builds and update the artifacts of the release. Therefore this does not solve the Winget issue...

See for example [this CI job for tag v1.15.0](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/46970268)

If you see the very end of the log, it is updating the existing v1.15.0 release artifacts on github, **twice**.

JohnTheGr8:
Perhaps a simpler temporary solution would be to switch the winget release action to manual only? And trigger it manually when both appveyor jobs have finished...

jjw24:
Ah yes, does look like the case.

Yes lets change the WinGet action to manual temporarily.

Issue #2195, opened 2023-06-21
BUG: Flow Launcher plugins not loading


postlemons:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

the plugins store doesn't show any plugins, neither does the PM command, I've tried re-installing the app, but no clue

### To Reproduce

Not sure how did this happen

### Screenshots

![image](https://github.com/Flow-Launcher/Flow.Launcher/assets/66541630/6a9d3b8c-a1e3-46e8-9cec-a2e6354ad686)

### Flow Launcher Version

1.15.0

### Windows Build Number

[Version 10.0.19045.3086]

### Error Log

<details>

[2023-06-21.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/11823771/2023-06-21.txt)

</details>

JohnTheGr8:
> 21:52:08.9258+03:00 - ERROR - PluginsManifest.UpdateManifestAsync - Http request failed
EXCEPTION OCCURS: System.Net.Http.HttpRequestException: No such host is known. (cdn.jsdelivr.net:443)
 ---> System.Net.Sockets.SocketException (11001): No such host is known.

Flow is having problems fetching plugin information data because of a network issue... there's many possible causes for this.

Can you access [this cdn.jsdelivr.net link](https://cdn.jsdelivr.net/gh/Flow-Launcher/Flow.Launcher.PluginsManifest@plugin_api_v2/plugins.json) ?

JohnTheGr8:
> is it being blocked by the firewall or what?

yes, probably. My assumption is that your system DNS is blocking that domain, but your browser is configured with DNS over HTTPS and bypasses that block...

JohnTheGr8:
try `nslookup cdn.jsdelivr.net 8.8.8.8`

JohnTheGr8:
(assuming you're located in Egypt) I'm seeing other reports of Egyptian ISPs blocking `cdn.jsdelivr.net` so this is 99% that...

JohnTheGr8:
> nice, so VPN?! you gotta be kidding me...

We'll have a solution for this soon alongside #2048 

You can use `pm install direct-url-to-plugin` in the meantime

> now jsdelivr after Patreon... what else?! Google?

😂

JohnTheGr8:
[...]

Issue #2178, opened 2023-06-09
User configurable plugins manifest URL for Plugin Store


jjw24:
Provide the option for an user-configurable manifest URL (like an option between github, jsdelivr or user-defined) which will have the following additional benefits:
1. it will be possible to have "unofficial" plugin communities, which means the whole ecosystem is a bit more open
2. it will make it possible to do full integration testing of plugins: for example, before opening a PR to the manifest repository to add a new plugin, I can point my flow-launcher to my fork repository and test that a complete install/update/etc works as expected.

jjw24:
Might also be a good idea to allow two manifest urls and merge the plugins from both urls into one list. This user can have both plugins from the flow community and ones hosted somewhere else.

manuelkoester:
Hey!
Just looking at this again. Would this be a good issue to start with as a first time contributor?
How would you like to see this implemented?

VictoriousRaptor:
> Hey! Just looking at this again. Would this be a good issue to start with as a first time contributor? How would you like to see this implemented?

Feel free to contribute if you're interested.

jjw24:
> How would you like to see this implemented?

The Plugin Store and PluginsManager plugin code is where you will do most of the changes:

1. In the backend after loading the plugins manifest from GitHub, the code would load local/network location json file plugin manifests.
2. After all plugin json files have been loaded, merge them together. This can then be used by the PluginsManager and the Plugin Store.
3. In terms of UI, I think the field where users enter the additional manifest path would be in the PluginsManager. The Plugin Store would display all plugins, those hosted from flow and ones added by user.

Issue #2048, opened 2023-04-13
BUG: Plugin store not updating in a timely manner when a new plugin ver released


deefrawley:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

I updated General Converter to v2.0 over 12 hours ago. Plugin manifest picked up the changes shortrly after, but Flow Plugin Store hasn't updated and shows version 1.4 still

Using `pm install` installs v1.4

Tried reinstalling Flow, and hitting Refresh in the Plugin Store but no update.

This must update eventually otherwise there would be no plugin updates since the last change, but it should be much quicker than this.

### Flow Launcher Version

1.14.0

### Windows Build Number

Microsoft Windows [Version 10.0.19045.2846]

### Error Log

Log shows that a manifest update occurred:

`09:56:51.9743+10:00 - INFO  - PluginsManifest.UpdateManifestAsync - Fetched plugins from manifest repo`

JohnTheGr8:
I also noticed this today and had a look... it's caused by the CDN we use in [PluginsManifest.cs#L15](https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher.Core/ExternalPlugins/PluginsManifest.cs#L15). jsDelivr caches for 12 hours.

We can switch back to a direct Github link, but Chinese users will have issues 🤷‍♂️

jjw24:
Good pick up @JohnTheGr8. I don't suppose there is any settings with it we can tweak to reduce the cache time?

JohnTheGr8:
they offer a limited-access API for purging cache that we _could_ use from the CI script... but it shouldn't be worth the trouble.

Maybe we should use a Github link by default and either:
* fall back to jsdelivr if the github link fails
* make the manifest URL configurable from settings

jjw24:
> they offer a limited-access API for purging cache that we _could_ use from the CI script... but it shouldn't be worth the trouble.
> 
> Maybe we should use a Github link by default and either:
> 
> * fall back to jsdelivr if the github link fails
> * make the manifest URL configurable from settings
> * or both of the above

This might be our last resort.

If It's a simple API call to purge, we can do that after the manifest is updated.

@taooceros do you know any other viable CDN options?

JohnTheGr8:
> This might be our last resort.

keep in mind that their docs state:

[...]

PR #2222, opened 2023-07-05
Rework how we fetch community plugin data

JohnTheGr8:
### Community Plugin Stores

Added an abstraction for **community plugin stores**, which can be configured with one or more manifest file URLs (for example, a Github link and a few CDN links). When fetching plugin data, all manifest file URLs are used until one of them succeeds. This fixes issues with geo-blocking such as #2195 without compromising performance.

For the moment we only have one plugin store (the PluginsManifest repository), but this concept can be expanded later to be fully customizable and to support unofficial plugin communities (see #2178)

### Plugin manager enhancements

Using `Ctrl+R` from a `pm install` or `pm update` query will result in fetching plugin data from the primary manifest URL, which is a github link. This will bypass any configured CDN links and force Flow to get the latest data, which helps with #2048

PS: I had a very different approach to #2200 in mind, and thus decided to open another PR instead of trying to explain

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
@@ -1,4 +1,5 @@
-﻿using Flow.Launcher.Plugin.PluginsManager.ViewModels;
+using Flow.Launcher.Core.ExternalPlugins;
+using Flow.Launcher.Plugin.PluginsManager.ViewModels;
 using Flow.Launcher.Plugin.PluginsManager.Views;
 using System.Collections.Generic;
 using System.Linq;
@@ -34,7 +35,7 @@ namespace Flow.Launcher.Plugin.PluginsManager
             contextMenu = new ContextMenu(Context);
             pluginManager = new PluginsManager(Context, Settings);
 
-            _ = pluginManager.UpdateManifestAsync();
+            await PluginsManifest.UpdateManifestAsync();
         }
 
         public List<Result> LoadContextMenus(Result selectedResult)
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
@@ -1,4 +1,4 @@
-using Flow.Launcher.Core.ExternalPlugins;
+﻿using Flow.Launcher.Core.ExternalPlugins;
 using Flow.Launcher.Plugin.PluginsManager.ViewModels;
 using Flow.Launcher.Plugin.PluginsManager.Views;
 using System.Collections.Generic;
@@ -51,9 +51,9 @@ namespace Flow.Launcher.Plugin.PluginsManager
             return query.FirstSearch.ToLower() switch
             {
                 //search could be url, no need ToLower() when passed in
-                Settings.InstallCommand => await pluginManager.RequestInstallOrUpdate(query.SecondToEndSearch, token),
+                Settings.InstallCommand => await pluginManager.RequestInstallOrUpdate(query.SecondToEndSearch, token, query.IsReQuery),
                 Settings.UninstallCommand => pluginManager.RequestUninstall(query.SecondToEndSearch),
-                Settings.UpdateCommand => await pluginManager.RequestUpdateAsync(query.SecondToEndSearch, token),
+                Settings.UpdateCommand => await pluginManager.RequestUpdateAsync(query.SecondToEndSearch, token, query.IsReQuery),
                 _ => pluginManager.GetDefaultHotKeys().Where(hotkey =>
                 {
                     hotkey.Score = StringMatcher.FuzzySearch(query.Search, hotkey.Title).Score;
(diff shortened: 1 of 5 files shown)


Plugins/Flow.Launcher.Plugin.PluginsManager/Main.cs
```diff
@@ -50,9 +51,9 @@ public async Task<List<Result>> QueryAsync(Query query, CancellationToken token)
             return query.FirstSearch.ToLower() switch
             {
                 //search could be url, no need ToLower() when passed in
-                Settings.InstallCommand => await pluginManager.RequestInstallOrUpdate(query.SecondToEndSearch, token),
+                Settings.InstallCommand => await pluginManager.RequestInstallOrUpdate(query.SecondToEndSearch, token, query.IsReQuery),
                 Settings.UninstallCommand => pluginManager.RequestUninstall(query.SecondToEndSearch),
-                Settings.UpdateCommand => await pluginManager.RequestUpdateAsync(query.SecondToEndSearch, token),
+                Settings.UpdateCommand => await pluginManager.RequestUpdateAsync(query.SecondToEndSearch, token, query.IsReQuery),
```
taooceros, line 56: why requery means use the primary url?

JohnTheGr8: I wanted to add a way to fetch fresh data (not cached data from a CDN) and this seemed a reasonable solution, if you think of `Ctrl+R` as a refresh mechanism.

This is mostly a temporary hack for #2048 honestly, I think we can rethink this as part of #2178

taooceros: I see. I previously thought `Last()` refers to cdn.

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -357,9 +337,9 @@ private bool InstallSourceKnown(string url)
             return url.StartsWith(acceptedSource) && Context.API.GetAllPlugins().Any(x => x.Metadata.Website.StartsWith(contructedUrlPart));
         }
 
-        internal async ValueTask<List<Result>> RequestInstallOrUpdate(string search, CancellationToken token)
+        internal async ValueTask<List<Result>> RequestInstallOrUpdate(string search, CancellationToken token, bool usePrimaryUrlOnly = false)
         {
-            await UpdateManifestAsync(token);
+            await PluginsManifest.UpdateManifestAsync(token, usePrimaryUrlOnly);
```
taooceros, line 342: If we fetch result every time, the ctrl+r will only refresh the result on the first one, but later one will still use the manifest in cdn? I believe we should have local cache for this.

JohnTheGr8: We don't fetch _every time_, I have added a 10-second timeout so that we don't fetch repeatedly while a user is typing. Hitting `Ctrl+R` will bypass the timeout and fetch from the primary url.

> Inserting another character will revert back to the old results

so this is only an issue if the user waits 10 seconds before inserting that extra character.

taooceros: 10 seconds seems to be pretty short? As we have the ctrl+r features, why don't we add a couple minutes timeout?

JohnTheGr8: > 10 seconds seems to be pretty short? 

it's only meant to avoid multiple http requests while a user is typing

> As we have the ctrl+r features, why don't we add a couple minutes timeout?

I don't see a reason to: fetching is very fast with the CDN links, and I've added 3 CDN links now so that fetching should be fast for users anywhere in the world (the fastest cdn link will be used in each user's case).

It takes around 80ms for me, a 2-minute timeout for a call this quick seems a bit unreasonable 😂

taooceros: If you are in China, the cdn is not so fast. It takes around 1s to download the manifest for me without proxy (with proxy it shall be fast).

image

JohnTheGr8: Okay, I can increase the timeout to a couple minutes. Does the rest of the PR look good to you?

> the other cdn takes 600ms, but sometimes it doesn't work.

yeah that's served by cloudflare, don't think they're very solid in China

taooceros: The rest looks good to me. I will take a detail look today. Previously jsdelivr has license in China, so they are able to deliver cdn that located in China, which is much much faster and stabler.

taooceros (approved):
Sorry for the delay. It looks very good to me.

JohnTheGr8:
@jjw24 do you also wanna have a look before we merge?

jjw24:
Yep, I will take a quick look as well then merge.

jjw24 (approved):

PR #2231, opened 2023-07-09
Switch winget release action to manual

JohnTheGr8:
As discussed in #2218 I have updated the winget release action to only fire manually, as a temporary solution to #2086

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,8 +1,7 @@
 name: Publish to Winget
 
 on:
-  release:
-    types: [released]
+  workflow_dispatch:
 
 jobs:
   publish:


jjw24:
Have you by any chance had a look at the GitHub action whether it will do anything if there is no new releases? Wondering if we can set it up as a cron job instead.

JohnTheGr8:
> Have you by any chance had a look at the GitHub action whether it will do anything if there is no new releases? Wondering if we can set it up as a cron job instead.

Had a quick look, didn't see a way to configure this...

jjw24:
I think this is doable based on https://github.com/vedantmgoyal2009/winget-releaser/issues/1, I will push up the change.

JohnTheGr8:
why would we want this to be a cron, though?

jjw24:
Just so I don't have to wait around for the release to finish and remember to manually run the action, it's still the workaround for until PR 2185 is finished.

JohnTheGr8:
hopefully we'll finish that PR by the next release, so you'll only need to wait this once 😛

jjw24:
Alright I have your word then, no pressure 😛 haha

jjw24 (approved):

PR #2269, opened 2023-08-06
Set main Flow Launcher project as default start up

jjw24:
Context:
Currently on a fresh clone of the project, when the solution is opened up in Visual Studio it has Flow.Launcher.Test set as startup.

Change:
This small convenience change makes it so that the main flow project is the startup project after a fresh clone.

Tests:
Clone the project down in sandbox env and open in Visual Studio, able to just press F5 to run as the project is already set as the startup project.


taooceros:
I didn't see what's changed in the diff?

jjw24:
The order of projects is changed so the main one is on top in the solution file. VS marks the first project in the list as default.

taooceros:
Ah ok

taooceros (approved):

Issue #2270, opened 2023-08-06
BUG: Windows search index throws exception when searching


Phoenix-:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

Hey. 
Got a problem similar to https://github.com/Flow-Launcher/Flow.Launcher/issues/1630
Flow shows message 'Flow.Launcher.Plugin.Explorer.Exceptions.SearchException: Invalid URI: The Authority/Host could not be parsed.' as first result

Debugged it and I'm seeing that encodedFragmentPath var in WindowsIndex.cs:59 is containing only 'file:' and URI ctor fails. 
Did not dig deeper. Can do if needed, tho.
Possible quick fix is just to wrap every result parsing with try-catch and just ignore invalid results, but thats a no-brainer from someone without knowledge on what's going on there.

Complete exception stack:
`WindowsIndex Search Exception:
 Flow.Launcher.Plugin.Explorer.Exceptions.SearchException: Invalid URI: The Authority/Host could not be parsed.
 ---> System.UriFormatException: Invalid URI: The Authority/Host could not be parsed.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind, UriCreationOptions& creationOptions)
   at Flow.Launcher.Plugin.Explorer.Search.WindowsIndex.WindowsIndex.ExecuteWindowsIndexSearchAsync(String indexQueryString, String connectionString, CancellationToken token)+MoveNext() in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\WindowsIndex\WindowsIndex.cs:line 70
   at Flow.Launcher.Plugin.Explorer.Search.WindowsIndex.WindowsIndex.ExecuteWindowsIndexSearchAsync(String indexQueryString, String connectionString, CancellationToken token)+MoveNext() in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\WindowsIndex\WindowsIndex.cs:line 70
   at Flow.Launcher.Plugin.Explorer.Search.WindowsIndex.WindowsIndex.ExecuteWindowsIndexSearchAsync(String indexQueryString, String connectionString, CancellationToken token)+MoveNext() in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\WindowsIndex\WindowsIndex.cs:line 70
[...]

PR #2271, opened 2023-08-07
Fix windows search empty paths

Phoenix-:
Fix for #2270 

Ignore 'file:' windows search results as they produce URI constructor exception.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/WindowsIndex.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/WindowsIndex.cs
@@ -47,14 +47,14 @@ namespace Flow.Launcher.Plugin.Explorer.Search.WindowsIndex
             while (await dataReader.ReadAsync(token))
             {
                 token.ThrowIfCancellationRequested();
-                if (dataReader.GetValue(0) == DBNull.Value || dataReader.GetValue(1) == DBNull.Value)
+                if (dataReader.GetValue(0) is DBNull
+                    || dataReader.GetValue(1) is not string rawFragmentPath
+                    || string.Equals(rawFragmentPath, "file:", StringComparison.OrdinalIgnoreCase))
                 {
                     continue;
                 }
                 // # is URI syntax for the fragment component, need to be encoded so LocalPath returns complete path   
-                var encodedFragmentPath = dataReader
-                    .GetString(1)
-                    .Replace("#", "%23", StringComparison.OrdinalIgnoreCase);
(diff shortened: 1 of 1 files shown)


jjw24:
Is it possible to compare `dataReader.GetValue(1)` so we can keep this check together with the db null checks?

Phoenix-:
Not sure if I should use {} around single-line blocks according to repo code style.

Phoenix-:
> Is it possible to compare `dataReader.GetValue(1)` so we can keep this check together with the db null checks?

Sure. One moment

jjw24:
> Not sure if I should use {} around single-line blocks according to repo code style.

The format you got is correct, no need {}

Phoenix-:
Changed both columns checks to type checks instead of value, combined path column checks.

Phoenix-:
Added one more check to file extension(or directory flag), just in case.

Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/WindowsIndex.cs
```diff
@@ -47,21 +47,22 @@ private static async IAsyncEnumerable<SearchResult> ExecuteWindowsIndexSearchAsy
             while (await dataReader.ReadAsync(token))
             {
                 token.ThrowIfCancellationRequested();
-                if (dataReader.GetValue(0) == DBNull.Value || dataReader.GetValue(1) == DBNull.Value)
+                if (dataReader.GetValue(0) is DBNull
+                    || dataReader.GetValue(1) is not string rawFragmentPath
+                    || string.Equals(rawFragmentPath, "file:", StringComparison.OrdinalIgnoreCase)
+                    || dataReader.GetValue(2) is not string extension)
```
jjw24, line 53: When will this not be a string?

Phoenix-: Most likely it will never be anything non-string, but we don't need anything else anyway. Code below used GetString(1) and GetString(2), so it will most likely crash if we get smth else(I believe GetString() differs from GetValue() only by hard-casting GetValue() to string). So I just filtered out this hypothetical case.

Phoenix-: To clarify: I never saw anything non-string there.

jjw24: Yeah, very good change 💯 , thanks for adding the additional checks.

jjw24 (approved):

Phoenix-:
Thank you :)

PR #2332, opened 2023-09-08
Fix exclude path check for index search

VictoriousRaptor:
Exclude path is broken because `FilesFolders.PathContains` doesn't return true when two paths are equal.

Tests:
- [x] When `c:\foo` is in exclude path, `c:\foo\x.txt` is excluded in search results.


jjw24 (approved):
Just added allowEqual parameter name to make it a bit more clearer when calling.

Issue #2289, opened 2023-08-14
BUG: Preview window(F1) stays draggable while hidden.


Sneakpeakcss:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

While in hidden state the border is still draggable, possibly blocking mouse actions if you manage to hit the perfect spot.

### To Reproduce

https://github.com/Flow-Launcher/Flow.Launcher/assets/77424331/c8ad9255-719e-4744-8578-f2f72629bfb9

### Flow Launcher Version

1.16.1

### Windows Build Number

10.0.19044.1826

onesounds:
Fixed in #2354.

PR #2354, opened 2023-09-19
Fix the issue resizable while preview turned off

onesounds:
## What's the PR
- Fix the issue resizable while preview turned off
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/2289

## Tests
- It should not be resizable when preview is off.


taooceros:
lol funny solution. Why can't we just disable it?

taooceros (approved):

onesounds:
> lol funny solution. Why can't we just disable it?

I thought this was the disable method 😅

PR #2394, opened 2023-10-16
Update nl.xaml

stefanroelofs:
Correct some Dutch translations


stefanroelofs:
I'm not sure why the Check Spelling job fails, or what I can do about it.
Thanks.

Issue #2347, opened 2023-09-16
Add new option "Close Command Prompt after key press" for "Shell" plugin


bluckholl:
For more convenient, it will be handy to have new option "Close Command Prompt after key press" so you can see the results and press any key to close.
The alternative for now is the current option "Do not close Command Prompt execution after command execution"

flooxo:
I am currently trying out how to close the terminal by displaying the text "Press any key to close", which can then be closed with a key.
Of course you could just append an exit command, but there must be a nicer solution, right?

PR #2410, opened 2023-11-05
New option CloseShellAfterPress

flooxo:
As described in #2347, it would be nice to see the result of a shell command and then closing it by pressing any key.

prototype-vision
(In this example running `> echo Hello` is expected to lead to this result)

ToDo:
- [x] Implement buttons
- [x] only selectable if `LeaveShellOpen` is not selected (contradictory options)
- [x] Add more language support
- [x] Implement logic


flooxo:
What is the best way to close the terminal after a command has been closed? I've currently tried appending the `pause` command, it works, but I'm sure there's a better solution :)

taooceros:
> What is the best way to close the terminal after a command has been closed? I've currently tried appending the `pause` command, it works, but I'm sure there's a better solution :)

I think earlier version of the shell plugin uses the same strategy as you. It might be good enough

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -203,7 +203,7 @@ private List<Result> ResultsFromHistory()
                 case Shell.Cmd:
                 {
                     info.FileName = "cmd.exe";
-                    info.Arguments = $"{(_settings.LeaveShellOpen ? "/k" : "/c")} {command} {(_settings.CloseShellAfterPress ? "& pause" : "")}";
+                    info.Arguments = $"{(_settings.LeaveShellOpen ? "/k" : "/c")} {command} {(_settings.CloseShellAfterPress ? $"&& echo {context.API.GetTranslation("flowlauncher_plugin_cmd_press_any_key_to_close")} && pause > nul /c" : "")}";
```
VictoriousRaptor, line 206: iirc pause command already prompts press any key to continue or something similar. Why do we need another prompt?

flooxo: I thought it would be nicer to display another more fitting message, because `pause` only prompts `Press any key...` 
Also, for the PowerShell another message is needed anyway, because here `pause` prompts `Press the Enter key to continue...:` (but we want that any key can be pressed) and simply to have the same behavior I used the additional message for both

VictoriousRaptor: Then we can just keep it for PowerShell. No need to `pause` in cmd.

VictoriousRaptor (changes requested):
Please revert changes in language files. Translation is done after merging to `dev` branch.

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
flooxo: Sorry, I dont understand. Why is `pause` not needed in cmd?
So only: `info.Arguments = $"{(_settings.LeaveShellOpen ? "/k" : "/c")} {command} {(_settings.CloseShellAfterPress ? $"&& echo {context.API.GetTranslation("flowlauncher_plugin_cmd_press_any_key_to_close")} /c" : "")}";` ? But then the terminal does not remain open

flooxo: @VictoriousRaptor could you please explain your comment in more detail so that I can continue working on the PR

taooceros (approved):

taooceros (approved):

VictoriousRaptor (approved):

PR #2416, opened 2023-11-10
switch back to jsonmessageformatter

taooceros:
https://github.com/microsoft/vs-streamjsonrpc/discussions/959

NewLineDelimitedMessageHandler hasn't supported SystemTextJsonFormatter yet.

Sorry for the change without testing😭I assume this is something that should not cause any issue.

PR #2418, opened 2023-11-12
use memorypack instead of binaryformatter

taooceros:
BinaryFormatter has been obsolete for a long time, so let's change it.

https://github.com/Cysharp/MemoryPack seems to be a nice choice. Good thing is that it supports circular reference. Though I think this is no longer a problem because uwpapp no longer reference back to their package.

The payload size seems to be slightly larger, but I think that's not a problem.


taooceros:
oops I also includes change in #2416 for testing, let me remove it

jjw24:
>The payload size seems to be slightly larger, but I think that's not a problem.

As in the final exe installer?

jjw24:
This change touches the core code, make sure you thoroughly test, e.g. fresh install & existing plugins.

taooceros:
> > The payload size seems to be slightly larger, but I think that's not a problem.
> 
> As in the final exe installer?

no I mean the cache. like 150k -> 200k. pretty trivial and nobody will care

taooceros:
> This change touches the core code, make sure you thoroughly test, e.g. fresh install & existing plugins.

Na not much service is using that. Only the ImageCache and Program plugin, and both seem working fine

jjw24:
@taooceros build failed

taooceros:
fixed @jjw24

jjw24 (approved):

PR #2430, opened 2023-11-18
Improve JsonRPC plugin settings

Garulf:
# Problem

- JsonRPC plugins are unable to make settings changes unless they have a `SettingsTemplate.yaml` file.
- Plugin's are unable to initiate new setting keys
- Plugin settings are only saved to disk upon exit

#  Solution

- [X] Allow plugins without `SettingsTemplate.yaml` to change their settings
- [X] Remove logic restricting settings key creation
- [X] Save plugin settings to file when updated

### DIFF
--- a/Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
+++ b/Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
@@ -83,6 +83,7 @@ namespace Flow.Launcher.Core.Plugin
                             break;
                     }
                 }
+                Save();
             }
         }
         
--- a/Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
+++ b/Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
@@ -36,6 +36,11 @@ namespace Flow.Launcher.Core.Plugin
             _storage = new JsonStorage<Dictionary<string, object>>(SettingPath);
             Settings = await _storage.LoadAsync();
 
+            if (Settings != null)
+            {
+                return;
+            }
+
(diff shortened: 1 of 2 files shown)


Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
```diff
@@ -81,6 +83,7 @@ public void UpdateSettings(IReadOnlyDictionary<string, object> settings)
                             break;
                     }
                 }
+                Save();
```
taooceros, line 86: maybe put this out of the loop. No need to save at every setting change?

Garulf: Oh whoops! 

I meant to make sure it was after the loop too!

Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
```diff
@@ -83,8 +83,8 @@ public void UpdateSettings(IReadOnlyDictionary<string, object> settings)
                             break;
                     }
                 }
-                Save();
             }
+            Save();
```
taooceros, line 88: sorry another thoughts that why didn't we keep this consistent with other plugins. `Save` only be called when manually triggered or flow being closed?

Garulf: I guess just personal preference. It feels off to not save to file on change.

taooceros: yeah ok

taooceros (approved):

PR #2448, opened 2023-11-30
Add Flow Launcher Theme Selector to Sys plugin

Odotocodot:
A plugin to quickly change the current Flow Launcher theme.

image

If someone has a better suggestion for the default action keyword rather than `flowtheme`, would love to hear it. Was a bit apprehensive in using theme 🤷‍♂️ .


jjw24:
I think this plugin would be best served by publishing to the Store. We can have a look at exposing the needed methods via the API.

taooceros:
> Great idea. I think this plugin would be best served by publishing to the Store. We can have a look at exposing the needed methods via the API.

I feel like we can allow this plugin to be built in. Thememanager is pretty internal.

jjw24:
Well yes and no, it's directly referencing the core project, ideal design is to have even the default plugins as self contained.

taooceros:
> Well yes and no, it's directly referencing the core project, ideal design is to have even the default plugins as self contained.

Some of our internal stuff uses the plugin interface for receiving query, which I think is a pretty good idea. I am not sure whether exposing everything to the interface is a very good idea. PublicAPI sometimes can be the curse as we may want to maintain backward compatibility.

jjw24:
Well I think at least this functionality is a good one to enable plugins to use. Remember the more plugins we pack the more we need to maintain and the bigger the size eventually get. More ideal to encourage plugin development.

taooceros:
> Well I think at least this functionality is a good one to enable plugins to use. Remember the more plugins we pack the more we need to maintain and the bigger the size eventually get. More ideal to encourage plugin development.

well a plugin only takes 400k size...which should be pretty neglectable. and this feature seems to be good to built in.

taooceros:
@Odotocodot isn't you also need to add the plugin folder to the solution?

jjw24:
> > Well I think at least this functionality is a good one to enable plugins to use. Remember the more plugins we pack the more we need to maintain and the bigger the size eventually get. More ideal to encourage plugin development.
> 
> well a plugin only takes 400k size...which should be pretty neglectable. and this feature seems to be good to built in.

What about putting this functionality in the Sys plugin since it has flow related functionalities like open log file and reload data etc.

Odotocodot:
Sorry ive been busy with uni! 
> > > Well I think at least this functionality is a good one to enable plugins to use. Remember the more plugins we pack the more we need to maintain and the bigger the size eventually get. More ideal to encourage plugin development.
> > 
> > 
> > well a plugin only takes 400k size...which should be pretty neglectable. and this feature seems to be good to built in.
> 
> What about putting this functionality in the Sys plugin since it has flow related functionalities like open log file and reload data etc.

So move to be inside the sys plugin?

> @Odotocodot isn't you also need to add the plugin folder to the solution?

Im not sure its needed? or im confused... its be a while

jjw24:
I would say ideally expose the methods via the API so people can build the theme plugin around it and what not but this may get complicated, so in the mean time putting it in the Sys plugin would be the other go to option.

jjw24:
>So move to be inside the sys plugin?

I would say ideally exposing the methods via the plugin API would be the best so people can build plugin around it, but this change can be complicated, so putting it in the Sys plugin is the next best option.

Odotocodot:
I've moved the theme selector plugin into the sys plugin now, though I have only done the english translation.

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
```diff
@@ -0,0 +1,95 @@
...
+                Score = score,
+                Action = c =>
+                {
+                    ThemeManager.Instance.ChangeTheme(theme);
+                    return true;
+                }
+            };
+        }
+
+        public void ReloadData() => LoadThemes();
```
jjw24, line 85: Why do we need to reload themes?

Odotocodot: That is a good point. I think the visibility changed should cover it!

taooceros:
@jjw24 is there anything else left on this pr?

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
```diff
@@ -0,0 +1,94 @@
...
+        public const string Keyword = "fltheme";
+
+        private readonly PluginInitContext context;
+        private IEnumerable<string> themes;
+
+        public ThemeSelector(PluginInitContext context)
+        {
+            this.context = context;
+            context.API.VisibilityChanged += OnVisibilityChanged;
+        }
```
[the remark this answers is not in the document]
jjw24: This is not required. context should never be null.

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
```diff
@@ -0,0 +1,94 @@
...
+            };
+        }
+
+        public void Dispose()
+        {
+            if (context != null && context.API != null)
+            {
+                context.API.VisibilityChanged -= OnVisibilityChanged;
+            }
+        }
```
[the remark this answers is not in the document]
jjw24: 👍

Plugins/Flow.Launcher.Plugin.Sys/Main.cs
```diff
@@ -407,6 +415,18 @@ private List<Result> Commands()
...
+                },
+                new Result
+                {
+                    Title = "Set Flow Launcher Theme",
+                    SubTitle = context.API.GetTranslation("flowlauncher_plugin_sys_theme_selector"),
+                    IcoPath = "Images\\theme_selector.png",
+                    Glyph = new GlyphInfo("/Resources/#Segoe Fluent Icons", "\ue790"),
+                    Action = c =>
+                    {
+                        context.API.ChangeQuery($"{ThemeSelector.Keyword} ");
```
jjw24, line 427: Actionkeyword is not supported here as it only changes the query to a keyword string. I don't think we even need to use the keyword `fltheme` perhaps.

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        public List<Result> Query(Query query)
+        {
+            if (query.IsReQuery)
+            {
+                LoadThemes();
+            }
+
+            int keywordIndex = query.Search.IndexOf(Keyword, StringComparison.Ordinal);
+            string search = query.Search[(keywordIndex + Keyword.Length + 1)..];
```
jjw24, line 30: An error occurs when nothing is entered after the fltheme keyword, probably can use query.SecondToEndSearch instead of trying to work out the user input manually.

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
```diff
@@ -0,0 +1,113 @@
...
+        private void OnVisibilityChanged(object sender, VisibilityChangedEventArgs args)
+        {
+            if (args.IsVisible && !context.CurrentPluginMetadata.Disabled)
+            {
+                LoadThemes();
+            }
+        }
+
+        private void LoadThemes() 
+            => themes = ThemeManager.Instance.LoadAvailableThemes().Select(Path.GetFileNameWithoutExtension);
```
jjw24, line 55: This is actually causing the build to fail. Needs to be `themes = ThemeManager.Instance.LoadAvailableThemes().Select(x => x.FileNameWithoutExtension);` maybe?

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        public const string Keyword = "fltheme";
+
+        private readonly PluginInitContext context;
+        private IEnumerable<string> themes;
+
+        public ThemeSelector(PluginInitContext context)
+        {
+            this.context = context;
+            context.API.VisibilityChanged += OnVisibilityChanged;
```
jjw24, line 19: This is going to cause the theme to be loaded every time flow is triggered, therefore plugin is constantly called. Why do we need to do this?

jjw24:
> @jjw24 is there anything else left on this pr?

Getting there but it's not ready yet.

Plugins/Flow.Launcher.Plugin.Sys/Main.cs
Odotocodot: I see. Its sort of a "fake" ActionKeyword its purpose was when the user query is `fltheme` they can see all their themes (and only their themes) and then search through them.
I could move this list and search functionality to a context menu on the sys plugin result perhaps?

Plugins/Flow.Launcher.Plugin.Sys/Images/theme_selector.png
Odotocodot, line 1: Should this be removed to match other sys plugin results i.e. change to FL logo

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
Odotocodot: Fixed!

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
Odotocodot: Fixed!

Plugins/Flow.Launcher.Plugin.Sys/Images/theme_selector.png
Jack251970: Yes. Now I use this icon for those theme results.

Plugins/Flow.Launcher.Plugin.Sys/Main.cs
Jack251970: This is used for expanding a second menu where you can show more items. And I think it works fine here.

Moving to context menu will make user confused since most users should prefer left click for further actions.

Plugins/Flow.Launcher.Plugin.Sys/ThemeSelector.cs
Jack251970: I do not think this is a good idea. We should load themes every time users query `ThemeSelector.Keyword` keywords so that results are always correct.

By the way, this is second menu and only if users open it, we will load those themes so it will not affect so much on performance.

Odotocodot: Alright so remove `OnVisibilityChanged` and the `query.IsReQuery` check so themes are always loaded when the user types. Therefore the `Query` method would start like this:
```csharp
public List<Result> Query(Query query)
{
    LoadThemes();
    ...
```

Jack251970:
@Odotocodot Hi, I have changed theme selection part and here is the final result.

Firstly, I use codes from `SettingsPaneThemeViewModel.cs` to fix possible drop effect issue.

Secondly, every time you change the theme, the window will not be closed but re-query the results because users may not feel good about the theme they changed so they possibly want to change theme for many times.

Thirdly, I add light/dark theme tooltip into results as their subtitles.

https://github.com/user-attachments/assets/d2031fab-e3aa-40a9-93ed-e63813dc38f3

This works fine for me and I have approved those changes. Please check if this works for you, and feel free to merge if you think it is ready.

Jack251970 (approved):

Odotocodot:
@Jack251970 Nice!
Those are very good additions 😃 ,definitely makes sense to keep the window open and update the theme.

I fixed a null exception and removed some extra code (In `ThemeSelector` `DropShadowEffect` was only ever set to false so removed the logic when setting to true).

If eveything is in order I am happy to merge!

Jack251970 (approved):
Look good to me! Thanks for your work!👍

Issue #2395, opened 2023-10-17
BUG: Browser Bookmarks plugin not working


zhj9709:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

BrowserBookmark plugin not working.

![image](https://github.com/Flow-Launcher/Flow.Launcher/assets/13991774/4bbe2b4b-cf77-4e40-8bd8-5a8b4e976ed6)

When I unchecked chrome and re-checked chrome, it started working fine.

### To Reproduce

1. alt+space to show flow launcher
2. Enter keywork b

### Screenshots

![image](https://github.com/Flow-Launcher/Flow.Launcher/assets/13991774/9d7e2d6a-6e7a-441e-8cff-565cfa09638a)

### Flow Launcher Version

v1.16.1,  v1.16.2

### Windows Build Number

Windows11 22635.2483

### Error Log

```shell
09:04:09.2956+08:00 - ERROR -  - ------------- BEGIN Flow.Launcher.Plugin.Program exception -------------
09:04:09.2956+08:00 - ERROR -  - 
Exception full name: System.IO.FileNotFoundException
Error status: UNKNOWN
Class name: Win32
Calling method: ExeProgram
Program path: C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux_1.3.17.0_x64__8wekyb3d8bbwe\msrdc.exe
InnerException number: 1
Exception message: C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux_1.3.17.0_x64__8wekyb3d8bbwe\msrdc.exe
Exception error type: HResult -2147024894
Exception thrown in called method: System.Diagnostics.FileVersionInfo GetVersionInfo(System.String)
Possible interpretation of the error: File not found when trying to load the program from C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux_1.3.17.0_x64__8wekyb3d8bbwe\msrdc.exe
Possible resolution: Not yet known
09:04:09.2956+08:00 - ERROR -  - ------------- END Flow.Launcher.Plugin.Program exception -------------
09:04:10.2123+08:00 - ERROR -  - ------------- BEGIN Flow.Launcher.Plugin.Program exception -------------
09:04:10.2123+08:00 - ERROR -  - 
Exception full name: System.Runtime.InteropServices.COMException
[...]

PR #2457, opened 2023-12-18
add a safety check for getproperty

taooceros:
add a safety check for https://github.com/Flow-Launcher/Flow.Launcher/issues/2395#issuecomment-1803229966

This seems to be the only place that we didn't check whether property exists.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.BrowserBookmark/ChromiumBookmarkLoader.cs
+++ b/Plugins/Flow.Launcher.Plugin.BrowserBookmark/ChromiumBookmarkLoader.cs
@@ -2,12 +2,14 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Text.Json;
+using Flow.Launcher.Infrastructure.Logger;
 
 namespace Flow.Launcher.Plugin.BrowserBookmark
 {
     public abstract class ChromiumBookmarkLoader : IBookmarkLoader
     {
         public abstract List<Bookmark> GetBookmarks();
+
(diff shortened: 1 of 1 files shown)


Plugins/Flow.Launcher.Plugin.BrowserBookmark/ChromiumBookmarkLoader.cs
```diff
@@ -2,12 +2,14 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Text.Json;
+using Flow.Launcher.Infrastructure.Logger;
```
jjw24, line 5: Infrastructure Logger, What's this for?

taooceros: oh mb forget about this is for plugin. Though I think somewhere else is still using Infra? I will check soon

taooceros: Yeah we ref Infra for Enum displaying...Unless we want to get rid of that it seems to ok to ref the static Log?

jjw24: hmm actually used for calling logging, when we redo logging we can move it to Plugin project so there is less/no dependency on the Infrastructure project.

jjw24 (approved):

PR #2468, opened 2024-01-03
Add translation keys for System Command Plugin

### LINKED ISSUE #2459: [Improvement] System commands plugin
### **Is your feature request related to a problem? Please describe.**

The commands in "system commands" plugin are not translated (plugin by qianlifeng)

### **Describe the solution you'd like**

Adding these strings on Crowdin for translation. Or even better and simpler, let users edit these commands.

### **Additional context**

System commands plugin

flooxo:
This PR closes #2459.

## Changes
- Add translation key for the System Command Plugin commands as only the description has been translated


VictoriousRaptor:
Personally I disagree with the idea of translating the title of the results. Typing the command in English is quite convenient imo. At least we should make both English and localized languages available.

flooxo:
Okay, I just thought, because there was no comment on the topic, that it might not be wanted, I implemented it. Otherwise, just close the PR again :)


> At least we should make both English and localized languages available.

I don't quite understand what you mean by that?

flooxo:
Thanks, now i understand. I'll take another look at it

flooxo:
What would be an elegant solution so that both languages can be used if the language is not english?
Sure, one could use `Title = $"{"Shutdown"} | {context.API.GetTranslation("flowlauncher_plugin_sys_shutdown_computer_cmd")}"` (with an additional condition that the translation is not always displayed) or what other alternative options are there?

taooceros:
> What would be an elegant solution so that both languages can be used if the language is not english? Sure, one could use `Title = $"{"Shutdown"} | {context.API.GetTranslation("flowlauncher_plugin_sys_shutdown_computer_cmd")}"` (with an additional condition that the translation is not always displayed) or what other alternative options are there?

I think you can do that when filtering the result? Though that may need some additional work.

flooxo:
image image

It is now possible to execute a system command in English as well as in the configured language, whereby only the more appropriate title is displayed.
I have solved this by determining for each result whether the English or translated title has a higher match and the one with the higher score is then used

I am interesed in your feedback :)

taooceros:
> image image
> 
> It is now possible to execute a system command in English as well as in the configured language, whereby only the more appropriate title is displayed.
> 
> I have solved this by determining for each result whether the English or translated title has a higher match and the one with the higher score is then used
> 
> I am interesed in your feedback :)

ah that's better than my earlier thought!

Plugins/Flow.Launcher.Plugin.Sys/Main.cs
```diff
@@ -77,9 +82,48 @@ public List<Result> Query(Query query)
...
+            var englishTitleMatch = StringMatcher.FuzzySearch(query.Search, pair.Key);
+            var translatedTitleMatch = StringMatcher.FuzzySearch(query.Search, pair.Value);
+
+            return englishTitleMatch.Score >= translatedTitleMatch.Score ? pair.Key : pair.Value;
+        }
+
         public void Init(PluginInitContext context)
         {
             this.context = context;
+            KeywordTitleMappings = new Dictionary<string, string>{
```
taooceros, line 106: Probably we don't want to fix translation as user are free to change language when using (though unlikely). I suppose you can have a dictionary mapping to the key of the translation?

taooceros (changes requested):

Plugins/Flow.Launcher.Plugin.Sys/Main.cs
flooxo: Sure, that would make sense. I thought that the value from the dictionary would be automatically updated, but you're right. Thanks for the hint!

Plugins/Flow.Launcher.Plugin.Sys/Main.cs
```diff
@@ -77,9 +82,50 @@ public List<Result> Query(Query query)
             return results;
         }
 
+        private string GetDynamicTitle(Query query, Result result)
+        {
+            var pair = KeywordTitleMappings
```
taooceros, line 87: why didn't use the dictionary accessor directly?

flooxo: Correct me if I'm wrong, but if you use the dictionary accessor directly, you can only perform a lookup based on the key, and you can't include the additional condition of inequality between key and value.
But acctually it might be better to use the accessor directly and make a new condition if key == value, return key and skip everything else.

taooceros: Now the value is the "key" to lookup, so it should never be equal to the key (the english version) right? or do I miss anything?

flooxo: Correct, the `value` (translation key) should never be equal to the `key` (english version), except if the translation is missing or the language is english. In this case theres no need to compare both matching scores, because the title is set to the english version anyways.

Plugins/Flow.Launcher.Plugin.Sys/Main.cs
```diff
@@ -77,9 +82,52 @@ public List<Result> Query(Query query)
             return results;
         }
 
+        private string GetDynamicTitle(Query query, Result result)
+        {
+            if (!KeywordTitleMappings.TryGetValue(result.Title, out var translationKey))
+            {
+                Log.Error($"Dynamic Title not found for: {result.Title}");
```
VictoriousRaptor, line 89:
```suggestion
                Log.Error("Flow.Launcher.Plugin.Sys.Main", $"Dynamic Title not found for: {result.Title}");
```

taooceros (approved):

VictoriousRaptor (approved):

Issue #2476, opened 2024-01-13
BUG: No plugins displaying in the plugins store tab


jonesy-b-dev:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

No plugins are displayed in the plugin store tab in the Flowlauncher settings app under the plugins section.
Multible people reporting same issue on Discord.

### To Reproduce

1. Open Flowlauncher settings app
2. Go to the plugins store

### Screenshots

![image](https://github.com/Flow-Launcher/Flow.Launcher/assets/59418915/28d721f6-50c1-4e2d-a9bd-808d844a4ec2)

### Flow Launcher Version

1.16.2

### Windows Build Number

Microsoft Windows [Version 10.0.23585.1001]

### Error Log

Log file of today
[2024-01-13.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/13927718/2024-01-13.txt)

Karillion:
I am seeing the same issue on a fresh `1.16.2` install today. My logs:

[2024-01-13.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/13928763/2024-01-13.txt)

TheFireRed:
Same here

Garulf:
Possibly related to https://github.com/Flow-Launcher/Flow.Launcher/issues/2474

Garulf:
Should be fixed with: https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/pull/307

deefrawley:
Problem resolved. Was a plugin author who had deleted their github account.

PR #2479, opened 2024-01-14
ignore null datetimes for pluginstoreitem

taooceros:
#2476 

Ignore any potential null value for struct (DateTime) which causes failed deserialization.

https://docs.microsoft.com/en-us/dotnet/api/System.Text.Json.Serialization.JsonIgnoreCondition?view=net-7.0
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values


VictoriousRaptor:
Do we still need this patch?

taooceros:
Yes let's make it safe

VictoriousRaptor:
TESTED? LGTM

VictoriousRaptor (approved):

taooceros:
> TESTED? LGTM

nope but that's by document and it doesn't hurt to do that

PR #2488, opened 2024-01-16
feat: Add sound effect volume

### LINKED ISSUE #2461: Open Sound Effect volume
The launcher open sound effect is very satisfying, but it's too quiet. Customizable volume is required.

flooxo:
This PR closes #2461

on

## Description
I added a slider for the sound effect volume similar to the implementation of the animation speed. The slider only gets displayed when the setting is enabled

### Alternatives
- It would also have been possible to only increase the volume of the file, so that the user then has to use the windows volume mixer -> imo not that user friendly
- It would also have been possible to increase the volume of the audio file on the go -> imo not that efficient

## Changes
Increase the volume so that it can be increased overall
Changed `SoundPlayer` to `MediaPlayer` as the `SoundPlayer` does not allow to modify the volume


flooxo:
as a little feature it would also be possible to change the volume icon depending on the volume, but i didn't know how to look up which icon is what

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -113,6 +112,8 @@ private void OnLoaded(object sender, RoutedEventArgs _)
                             {
                                 if (_settings.UseSound)
                                 {
+                                    animationSound.Open(new Uri(AppDomain.CurrentDomain.BaseDirectory + "Resources\\open.wav"));
```
taooceros, line 115: does it have to open everytime?

taooceros:
> as a little feature it would also be possible to change the volume icon depending on the volume, but i didn't know how to look up which icon is what

check here https://learn.microsoft.com/en-us/windows/apps/design/style/segoe-fluent-icons-font

but I feel like two icon is a bit duplicate. @onesounds do you want to take a look?

onesounds:
> > as a little feature it would also be possible to change the volume icon depending on the volume, but i didn't know how to look up which icon is what
> 
> check here https://learn.microsoft.com/en-us/windows/apps/design/style/segoe-fluent-icons-font
> 
> but I feel like two icon is a bit duplicate. @onesounds do you want to take a look?

I think simply its ok to remove icon for volume. (because it's sub category of sound effect) 

But if it has align issue, you can change sound effect's icon to e7f5 Speakers icon. Volume icon is fine.

Flow.Launcher/MainWindow.xaml.cs
flooxo: It was necessary because once the audio was played it had to be reloaded. I have just seen that it is also sufficient to simply reset the MediaPlayer to the beginning

flooxo:
Oke thanks, unfortunately i don't know how to replace the icon with a blank space so that everything is still aligned correctly. imo i think it looks better if each setting has an icon, because otherwise it would look like the icon is missing

taooceros:
@jjw24 this seems to be a nice little feature to have. Maybe we add it to 1.17.3 as well?

jjw24:
Keep it for 1.18.0, we have another feature that can be released so once we get couple more we can just go straight to 1.18.0 instead.

jjw24:
@taooceros is this one approved or still need review?

VictoriousRaptor:
Why the open.wav is also modified?

flooxo:
I have changed the audio file so that you can make it louder than it is right now. With the media player, you can only select a value between 0 and 1 as the volume.
The current volume corresponds to 0.5 using the new volume slider

taooceros:
@flooxo mind fix the conflict?

@jjw24 i think it is a relatively simple change and the code looks good to me.

taooceros:
@flooxo seems like the pipeline fails.

> C:\projects\flow-launcher\Flow.Launcher\MainWindow.xaml.cs(43,9): error CS0246: The type or namespace name 'MediaPlayer' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\flow-launcher\Flow.Launcher\Flow.Launcher_l1jpuugg_wpftmp.csproj]

VictoriousRaptor (approved):
LGTM.

onesounds:
@flooxo 

image

After the update, the issue has been reported that there is no sound, and it is not showing in the app volume control list in windows setting (win11. Users cannot hear the sound when set in flow, nor can they find flow in a Windows mixer.). Is there a part that is presumed to be the cause?

onesounds:
> @flooxo
> 
> image
> 
> After the update, the issue has been reported that there is no sound, and it is not showing in the app volume control list in windows setting (win11. Users cannot hear the sound when set in flow, nor can they find flow in a Windows mixer.). Is there a part that is presumed to be the cause?
> 
> +) I've tested this and that, but there's nothing particularly wrong with this pr. I've guessed that if I've already adjusted the sound through the mixer, it will affect after this patch, but there was no particular issue with the bar I tried. Why...

This issue occurred because the user deleted the Windows Media Player game. We should consider including separate checks and guidance on future media player presence and version.

PR #2506, opened 2024-01-24
Speed up context menu search

Garulf:
## Concern
Flow Launcher currently rebuilds the context menu each time the query changes, leading to potential slowdowns. The process becomes even more sluggish if a context menu result requires a network call.

## Solution
Implement a memory cache for built results. This optimization ensures that querying the context menu becomes nearly instantaneous, and it enables the caching of expensive results.

## Drawbacks
The introduction of a memory cache may result in increased memory usage. However, this impact should be manageable, given that the machine would need to load these results initially. The trade-off is justified by the significant improvement in context menu querying speed and the ability to cache resource-intensive results.

### Before
https://github.com/Flow-Launcher/Flow.Launcher/assets/535299/94dda425-c73e-47af-9126-2f15196e327c

### After
https://github.com/Flow-Launcher/Flow.Launcher/assets/535299/92153e6a-0461-4c2e-ab07-e82a862f6ebe

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -34,6 +34,8 @@ namespace Flow.Launcher.ViewModel
 
         private bool _isQueryRunning;
         private Query _lastQuery;
+        private Result lastContextMenuResult = new Result();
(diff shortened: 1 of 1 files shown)


taooceros:
let's also make context menu async

taooceros (approved):

PR #2508, opened 2024-01-25
Add open target for program shortcuts

### LINKED ISSUE #2390: Program Plugin - Add an "Open shortcut's Target".

When a program's shortcut shows up in the result and that I want to open that's program's folder, the "Open Containing Folder" option send me to the shortcut's location and I then have to right click on it and select "Open target's Folder" to go to the program's folder.

- I'd like to have the choice to directly select that "Open Target's Folder" option into Flow Launcher if the selected result is a shortcut.

Not sure it's possible, though.

flooxo:
This PR closes #2390.

## Description
I added a new result in the context menu for programs to open the destination folder.
Only in `win32 ` this new result is added, because `UWP` work differently and have no links, so "open containing folder" already opens the correct location automatically

## Changes
Add `open target folder` option for programs

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
+++ b/Plugins/Flow.Launcher.Plugin.Program/Languages/en.xaml
@@ -73,6 +73,7 @@
     <system:String x:Key="flowlauncher_plugin_program_run_as_administrator">Run As Administrator</system:String>
     <system:String x:Key="flowlauncher_plugin_program_open_containing_folder">Open containing folder</system:String>
     <system:String x:Key="flowlauncher_plugin_program_disable_program">Disable this program from displaying</system:String>
+    <system:String x:Key="flowlauncher_plugin_program_open_target_folder">Open target folder</system:String>
 
     <system:String x:Key="flowlauncher_plugin_program_plugin_name">Program</system:String>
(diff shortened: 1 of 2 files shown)


taooceros:
mind putting up a screenshot for the result?

VictoriousRaptor:
I wonder how many options are in context menu.

flooxo:
> mind putting up a screenshot for the result?

image
Sure! Thanks for the hint, I totally forgot

flooxo:
There are the following options:
- Run As Different User
- Run As Administrator
- Open Containing Folder
- Open Target Folder
- Disable this program from displaying
- Set as topmost in this query
- Plugin: Program

VictoriousRaptor:
I've made some changes:
- Add `Args` property so `ExecutablePath` and `LnkResolvedPath` no longer contains args.
- Only show target folder context menu for .lnk files

jjw24 (approved):

VictoriousRaptor:
Looks like `Win32.cache` can't be loaded after updating and users need to manually reindex the programs.

taooceros:
> Looks like `Win32.cache` can't be loaded after updating and users need to manually reindex the programs.

Will it throw the error? If so I guess the reindex will happen automatically. You might try the release build?

PR #2529, opened 2024-02-04
Improve context menu results

Garulf:
## Concern

`Result` should be responsible for how it's deep copied and not the context menu method.

## Solution

Move all deep copy logic into `Result`

### DIFF
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -184,6 +184,31 @@ namespace Flow.Launcher.Plugin
         }
 
         /// <summary>
+        /// Clones the current result
+        /// </summary>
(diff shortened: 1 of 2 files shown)


VictoriousRaptor:
build failed?

Garulf:
I think a test may have failed because i changed `Result`

taooceros:
We actually only need a shalow copy of the result. Do we get any library that can do that?

taooceros:
@Garulf want to get this into 1.18? I think it's pretty nice to have.

Garulf:
> @Garulf want to get this into 1.18? I think it's pretty nice to have.

Sounds good to me!

taooceros (approved):

PR #2578, opened 2024-02-16
fix node plugin, job object association, v2 context menu

taooceros:
~~- Fix Python plugin v2~~
- Fix Node plugin (we didn't redirect input before)
- Fix job process associate to the child process created by program plugin which causes the exit of flow causing the program being created exited.
- Fix JsonRPCV2 plugins setting not sent via query

~~I don't know when but somehow python plugin doesn't pass the required argument which will create start a empty python interactive. So anyone with a python plugin v2 will simply can't run flow...~~

It's not as bad as I anticipated. The argument will still being passed. It's just the place is weird.

- [x] The app opened with program plugin won't get killed after flow exit
- [x] Settings are sent
- [x] Node plugin can use https://www.npmjs.com/package/vscode-jsonrpc without issue
- [x] Context Menu

### DIFF
--- a/Flow.Launcher.Core/Plugin/ExecutablePluginV2.cs
+++ b/Flow.Launcher.Core/Plugin/ExecutablePluginV2.cs
@@ -12,15 +12,7 @@ namespace Flow.Launcher.Core.Plugin
 
         public ExecutablePluginV2(string filename)
         {
-            StartInfo = new ProcessStartInfo
-            {
-                FileName = filename,
-                UseShellExecute = false,
-                CreateNoWindow = true,
-                RedirectStandardOutput = true,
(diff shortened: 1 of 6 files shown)


taooceros:
> > Fix job process associate to the child process created by program plugin which causes the exit of flow causing the program being created exited.
> 
> I can't reproduce it. Could you describe how to reproduce it?

weird I can't reproduce either. I only reproduce once when I exit flow and rider associate it being killed. Though it might be good to only manually associate the subprocess

taooceros:
@jjw24 mind take a look on this? It should be pretty straightforward as you have reviewed other v2 before.

jjw24 (approved):

PR #2581, opened 2024-02-23
New API Instance (ReQuery)

AminSallah:
**Changes:**
- Added a new instance `ReQuery` in [PublicAPIInstance.cs](https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher/PublicAPIInstance.cs).
- Introduced a new instance `ReQuery` in [IPublicAPI.cs](https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs).
- Changed the protection level of [ReQuery](https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher/ViewModel/MainViewModel.cs) to `public`.

**Benefits:**

This update allows plugin developers to reload queries to refresh data, addressing the following use cases:

- For counters and Pomodoro timers.
- Display cached results and, when new cache data is received from an online API, use `ReQuery` to update the query.
- Simultaneously refresh data.

## Testing the ReQuery Instance

I have tested the `ReQuery` instance in [one of my plugins](https://github.com/AminSallah/Flow.Launcher.Plugin.Azan) to requery every second. 
```csharp
Task.Run(() =>
{
    Thread.Sleep(500);
    _context.API.ReQuery();
    Thread.Sleep(500);
});
```

**Result:**
demo_1

**Why not use `ChangeQuery(Query.RawQuery, true)`:**
The use of this method failed for me; it is too sticky. I cannot remove the action keyword or continue typing in the query. When triggering the flow, it changes the query on its own.

```csharp
Task.Run(() =>
{
    Thread.Sleep(500);
    // _context.API.ReQuery();
    _context.API.ChangeQuery(query.RawQuery, true);
    Thread.Sleep(500);
});
```

**Result:**
demo_2


jjw24 (approved):
👍 
Really well described PR, thank you for making the change requirement clear and easy to understand.

jjw24:
Let's hold off merging till 1.18.0 release is scheduled.

taooceros:
theoredically this can be achieved by `ChangeQueryText(..., requery=true)`. Do we want to mark the second parameter as obsolete with this new api?

AminSallah:
Hey there, I'd like to introduce an idea I've tested. However, I can't judge whether it enhances the user experience. I just want to share it.

**Problem:** By default, ReQuery sets the selected result to the first index all the time.

```csharp
Task.Run(async () =>
{
    await this._notionDataParser.CallApiForSearch();
    Context.API.ReQuery(reselect: true);
});
```

***Result***
In this example, in a different thread, it waits for the API to update the local cache and then reloads the query using the default way.

reselection

**Fix:** I introduced a new boolean, `reselect`, in [MainViewModel.cs](https://github.com/AminSallah/Flow.Launcher/blob/reselect/Flow.Launcher/ViewModel/MainViewModel.cs) and overridden the `ReQuery` method to `ReQuery(bool reselect)`.

```csharp
Task.Run(async () =>
{
    await this._notionDataParser.CallApiForSearch();
    Context.API.ReQuery(reselect: false);
});
```

***Result***
Same as before but keeps the selected result.

reselectionoff

I cannot determine whether adding this option is better or not. It might not be more beneficial for large queries, like fetching 200 results, where users might not be noticed that the query has been reloaded. Alternatively, adding this option and allowing plugin developers to choose it could be an improvement. Feel free to share your thoughts on this feature.

Note: The method override keeps `Ctrl + R` unchanged.
Note: If the result after the reload is not present, Flow by default chooses the first one.

[Link to branch](https://github.com/AminSallah/Flow.Launcher/tree/reselect) let me know if you interested to share changes.

taooceros:
interesting. I think this is easier to use compared to the UpdateResults event?

AminSallah:
It's still using [UpdateResults ](https://github.com/AminSallah/Flow.Launcher/blob/reselect/Flow.Launcher/ViewModel/ResultsViewModel.cs#L166) but passing a new parameter to it to prevent selecting the first index. I didn't attempt to come up with something new to reduce test time and potential issues that may arise.

AminSallah:
Hey @jjw24 , Please see above comment.

taooceros:
I don't like the idea of storing a variable for whether reselect. Probably it should be a parameter?

AminSallah:
I am unsure if achieving smooth functionality is possible in this case, mainly due to the involvement of the RegisterViewUpdate method, which is invoking the UpdateResultView instance. This specific method is called by the Class structor and occurs subsequent to a query change and I encountered difficulty in passing a new parameter to it
https://github.com/AminSallah/Flow.Launcher/blob/99ac5f010d6136f44480b55274e1273b67e59b12/Flow.Launcher/ViewModel/MainViewModel.cs#L124

taooceros:
> I am unsure if achieving smooth functionality is possible in this case, mainly due to the involvement of the RegisterViewUpdate method, which is invoking the UpdateResultView instance. This specific method is called by the Class structor and occurs subsequent to a query change and I encountered difficulty in passing a new parameter to it https://github.com/AminSallah/Flow.Launcher/blob/99ac5f010d6136f44480b55274e1273b67e59b12/Flow.Launcher/ViewModel/MainViewModel.cs#L124

Oh probably add another property to `ResultsForUpdate` class?

AminSallah:
How you'd suggest passing reselect param from [ReQuery method()](https://github.com/AminSallah/Flow.Launcher/blob/99ac5f010d6136f44480b55274e1273b67e59b12/Flow.Launcher/ViewModel/MainViewModel.cs#L220) to [RegisterViewUpdate()](https://github.com/AminSallah/Flow.Launcher/blob/99ac5f010d6136f44480b55274e1273b67e59b12/Flow.Launcher/ViewModel/MainViewModel.cs#L124) because calling RegisterViewUpdate() with new param causing a bug that flow not select first result by default any more after method is invoked, that's why i prefer storing a variable for whether reselect, it didn't raise any issues with me till now.

```csharp
public void ReQuery(bool reselect)
{
    if (SelectedIsFromQueryResults())
    {
        QueryResults(isReQuery: true);
        RegisterViewUpdate(reselect);
    }
}
```

taooceros:
> How you'd suggest passing reselect param from [ReQuery method()](https://github.com/AminSallah/Flow.Launcher/blob/99ac5f010d6136f44480b55274e1273b67e59b12/Flow.Launcher/ViewModel/MainViewModel.cs#L220) to [RegisterViewUpdate()](https://github.com/AminSallah/Flow.Launcher/blob/99ac5f010d6136f44480b55274e1273b67e59b12/Flow.Launcher/ViewModel/MainViewModel.cs#L124) because calling RegisterViewUpdate() with new param causing a bug that flow not select first result by default any more after method is invoked, that's why i prefer storing a variable for whether reselect, it didn't raise any issues with me till now.
> 
> ```cs
> public void ReQuery(bool reselect)
> {
>     if (SelectedIsFromQueryResults())
>     {
>         QueryResults(isReQuery: true);
>         RegisterViewUpdate(reselect);
>     }
> }
> ```

No that's not what I mean. I mean add a parameter to the `UpdateResultView` method (or add a property to the parameter with type `IEnumerable<ResultsForUpdate>`. This result would probably only makes sense when we only have one `ResultsForUpdate` (one plugin is responsible for one).

It doesn't make sense to re-register everytime query.

taooceros:
Sure let me show you what I mean.

taooceros:
@AminSallah sorry for the delay. Take a look?

AminSallah:
Yeah, now I understand, but do you prefer to add a parameter to the entire collection, needing only one to determine whether to select the first or not, rather than declaring it as a top class parameter? Also, I encountered a bug that I tried to fix, but couldn't figure out why it occurred. I double checked that the whole collection has the same `ReselectFirstResult` value.
When switching to any plugin, flow does not select a result at all in this scenario.

Bug Screenshot

taooceros:
> Yeah, now I understand, but do you prefer to add a parameter to the entire collection, needing only one to determine whether to select the first or not, rather than declaring it as a top class parameter? Also, I encountered a bug that I tried to fix, but couldn't figure out why it occurred. I double checked that the whole collection has the same `ReselectFirstResult` value. When switching to any plugin, flow does not select a result at all in this scenario.
> 
> Bug Screenshot [ Bug Screenshot ](https://private-user-images.githubusercontent.com/124622454/314523091-0227bc69-af37-450e-a565-626d121c1f2f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTA5NTQ3MTMsIm5iZiI6MTcxMDk1NDQxMywicGF0aCI6Ii8xMjQ2MjI0NTQvMzE0NTIzMDkxLTAyMjdiYzY5LWFmMzctNDUwZS1hNTY1LTYyNmQxMjFjMWYyZi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMzIwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDMyMFQxNzA2NTNaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT05YWQ4NDNkOTRiMmU5OGQzZmVlYjNiMGZlZDA0NjY5ZGVlM2I2ZmRiNTQzYWIxYTk2ODRiNmNiM2Y2OGZhNzhhJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.HYiYkIYiebPD8zCO1yEkcBWvLeY-f3fjThTswGUemN8) [ ](https://private-user-images.githubusercontent.com/124622454/314523091-0227bc69-af37-450e-a565-626d121c1f2f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTA5NTQ3MTMsIm5iZiI6MTcxMDk1NDQxMywicGF0aCI6Ii8xMjQ2MjI0NTQvMzE0NTIzMDkxLTAyMjdiYzY5LWFmMzctNDUwZS1hNTY1LTYyNmQxMjFjMWYyZi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMzIwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDMyMFQxNzA2NTNaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT05YWQ4NDNkOTRiMmU5OGQzZmVlYjNiMGZlZDA0NjY5ZGVlM2I2ZmRiNTQzYWIxYTk2ODRiNmNiM2Y2OGZhNzhhJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.HYiYkIYiebPD8zCO1yEkcBWvLeY-f3fjThTswGUemN8)

I guess it's the problem we try to avoid by selecting the first result.

taooceros:
The parameter is not added to the entire collection of results, but only the result batch, which is slightly better. In fact, I think it is needed, because the results of different plugins can arrive at different times. They may not be processed at once?

I am doubting that the original implementation will only disable reSelect  at the first results update but not in the later ones. Though I am not sure whether this is desirable or not.... This kind of reasoning can be the reason of the issue you mentioned.

taooceros:
Class parameter introduce mutable state across various method, which might eventually make it very very hard to debug (what if we trigger a normal query right after some plugin trigger requery?). Flow already have a massive mutable states that makes it very hard to modify, so I would advocate that we should make things more pure.

Note: The local function (in the query result) still capture some state but that seems to be a little bit unavoidable unless we would like to pass the channel reader to the query function (which might be a good choice though).

taooceros:
@AminSallah I found the issue...I made two parameter with the same when refactoring, which somehow create weird behavior.

AminSallah:
> The parameter is not added to the entire collection of results, but only the result batch, which is slightly better. In fact, I think it is needed, because the results of different plugins can arrive at different times. They may not be processed at once?

I noticed this just now. Currently, it's working very well. Much thanks to you @taooceros, for your assistance.

taooceros:
You are very welcome.

jjw24:
@AminSallah could you please resolve conflict.

I see this PR should be ready to go right, any additional from you @taooceros ?

taooceros:
Yeah it's good for me

taooceros (approved):

Issue #2598, opened 2024-02-29
BUG: Only bookmarks from toolbar are being indexed for Opera browser


Phoenix-:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

Bookmarks from Opera browser are not being indexed unless placed onto bookmarks toolbar.
Checked for Opera 107.0.5045.36 but I dont think its linked to any particular version.
Bookmarks file for opera differs from Chrome one, note 'custom_root' sub-object
```
{
   "roots": {
      "bookmark_bar": {
         "children": [  ]
      },
      "custom_root": {
         "pinboard": {
            "children": [  ]
         },
         "speedDial": {
            "children": [ 
               {
                  "children": [  ]
               } 
            ]
         },
         "trash": {
            "children": [  ]
         },
         "unsorted": {
            "children": [  ]
         },
         "unsyncedPinboard": {
            "children": [  ]
         },
         "userRoot": {
            "children": [ {
               "children": [ {
                  "name": "...",
                  "type": "url",
                  "url": "https://..."
               } ],
               "name": "...",
               "type": "folder"
            },
<...>
```
Current implementation of EnumerateFolderBookmark method skips any object without 'children' field.
My fix slightly changes EnumerateRoot method to check for possible 'custom_root' field and if found treat it as a root itself.
Should not break other browsers and fixes Opera.

### To Reproduce

1. Install Opera
2. Set Bookmarks plugin to enumerate its profile directory
3. Create some temp bookmark outside toolbar
4. Observe Flow does not know about its existence

### Flow Launcher Version

1.17.2

### Windows Build Number

10.0.22631.3155

### Error Log

No relevant info there as it is not an error

Phoenix-:
Fixed myself, pr is created.

PR #2599, opened 2024-02-29
Fix opera bookmarks

Phoenix-:
Fix for #2598

Take "custom_root" sub-object into account.


Phoenix-:
Nothing else should be affected. The behavior changes only when tag with name 'custom_root' is encountered,

taooceros:
I think it's ok for now. Could you add a comment for that? If we have more exception then we consider more abstraction.

VictoriousRaptor:
LGTM. BTW could you please copy the problem description part of #2598 to this PR?

VictoriousRaptor (approved):

Phoenix-:
Added description

Issue #2329, opened 2023-09-07
Copy file/folder path: built-in hotkey `ctrl-shift-c` and remove notification.


Maverickx0:
I'm used to fill the file path for some app's config or go to one deep path folder to handle something. Based above requirement, I need to get the file/folder path quickly. 

I has known that I can do this by `ctrl + o` with `alt + 4` key combination and get notification when I copy done successfully. But it has some questions:

* Two key combination is slow.
* Copy notification will get the bad sense of delayed feedback and it's none-necessary. 

**For Q1**, I think copy file/folder path for most people is necessary. So, I hope that you add one built-in hotkey like `ctrl-shift-c` for copy file/path.

**For Q2**, When I copy successfully, the search panel will be close correctly. So, we don't need to get the notification.

Lastly, thanks for the nice app you brought, I love it.

TheBestPessimist:
This was implemented in https://github.com/Flow-Launcher/Flow.Launcher/pull/2612, so we can close it 🎉

PR #2612, opened 2024-03-18
Added ctrl+shift+c

AsafMah:
Solves #2051 #2329 

When pressing ctrl+shift+c, instead of copying the file, it will only copy the path to it, as it's a commonly used shortcut.


onesounds (approved):
I haven't tested it, but it doesn't seem to be a problem. It is a shortcut that is commonly used in UX, so there is no problem applying it.

jjw24:
Let's hold off merging for 1.18.0 release

PR #2618, opened 2024-03-25
Hide main window when dragging results

VictoriousRaptor:
Hide main window when dragging results, so users can see where they're dragging to.

### DIFF
--- a/Flow.Launcher/ResultListBox.xaml.cs
+++ b/Flow.Launcher/ResultListBox.xaml.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.IO;
 using System.Windows;
 using System.Windows.Controls;
@@ -136,6 +136,8 @@ namespace Flow.Launcher
                 return;
 
             isDragging = false;
+
+            App.API.HideMainWindow();
 
             var data = new DataObject(DataFormats.FileDrop, new[]
             {


Flow.Launcher/ResultListBox.xaml.cs
```diff
@@ -137,6 +137,8 @@ private void ResultList_MouseMove(object sender, MouseEventArgs e)
 
             isDragging = false;
 
+            App.API.HideMainWindow();
+
```
VictoriousRaptor, line 141: Main window will show later when finishing dragging since we're calling `ChangeQuery()` in line 156.

onesounds (approved):

Flow.Launcher/ResultListBox.xaml.cs
VictoriousRaptor: Main window will only show later when it's a Move action now. Should we amend this? @onesounds @jjw24

jjw24: Sorry what do you mean? Does change query call make the search window visible again? When I was testing this with moving a file the search window wasn't showing after the drop was done and had to retrigger window to appear.

VictoriousRaptor: When I was testing this PR something did bring back the search window, and I thought it was ` ChangeQuery`. But now I am confused about the logic. Will test later this week.

PR #2625, opened 2024-03-31
Fix program indexing logic when cache is empty

VictoriousRaptor:
https://github.com/Flow-Launcher/Flow.Launcher/pull/2508#issuecomment-2028550756

Reindex during initializing Program Plugin, when win32 OR uwp cache is empty, instead of both are empty.

Downside: When user choose not to index uwp in plugin settings, reindex is unnecessary.

Not dev only. it's in release as well.


Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -84,16 +84,39 @@ public async Task InitAsync(PluginInitContext context)
...
+                        WatchProgramUpdate();
+                    });
+                }
+                if (!_uwps.Any())
+                {
+                    _ = Task.Run(async () =>
+                    {
+                        await Stopwatch.NormalAsync("|Flow.Launcher.Plugin.Program.Main|UWPProgram index cost", async () =>
+                        {
+                            await Task.Run(IndexUwpPrograms);
```
taooceros, line 114: Why multiple task run?

taooceros (approved):

PR #2635, opened 2024-04-06
WPFUI / Fluent Window / Theme system / Resizable Window

onesounds:
## What's the PR
- These are interrelated features, so I work on them all at once. It would be nice to separate them, but I don't know if that's possible. 

## Fluent Window
- Added WPFUI
  - I plan to gradually transition the modernwpf controls I've been using.
- [ ] Support Legacy Themes
  - Window Coloring for legacy themes
- [ ] Removed animation (window moving, still has fade-in)
  - Because It's not good with fluent window.

- [ ] Removed Legacy Shadow option
  - Now we use 'single border window'. not transparent window.
  - in this case, window has OS rendered shadow. it can't be remove.
  - If we need window without shadow for everything taskbar layout or somethings), we should make a new window for this.

## Adjust Theme Features
- Removed themes almost.
- [ ] Add Subtitle Font Style
- Added and Moved some settings from theme
  - Added : Search Bar height, Query / Title / Subtitle font size / Item Size
  - Moved : Font Styles, 
  - [ ] Todo : Transparent type (Arcrylic, Mica, None), Corner Radius.
  - Themes are now more like styling. The basic settings are changed by the user.
- Add appearance editor layout in preview area
- [ ] Reset to Default Button
- [ ] Preset from themes
  - If theme has preset font styles or sizing information, user can override them with settings if user want.

## Resizable Window
- Change the window to fully resizable.
- Removed Window Width option
- Removed Max Show Item option (not sure)
  - Height : MaxResultsToShow is automatically changed by Window Height when resize end
  - Width : Simply remember when resize ended.

Todo
- [ ] Add Subtitle Style applying
- [ ] Adjust Default Themes and Adjust Colors
- [ ] Apply Translation Strings

More Plans
- Adjust Menus (New : Preview and Lab or power user things, Proxy page moved in general)
- [ ] Change the sidebar icons


onesounds:
This PR is separated into three PRs, like the Dragon Ball that used the wish.
- #2699 
- #2735
- #2676

PR #2636, opened 2024-04-06
Show target path in hover tooltip

VictoriousRaptor:
图片

Show target path in result's hover tooltip. Continuation of #2508.


onesounds:
It would be nice if we could set what to display on the tooltip, but let's work on this later.

onesounds (approved):

PR #2644, opened 2024-04-11
Code cleanup: Flow.Launcher/{Converters,Helper}

Yusyuriv:
- Switch from block-scoped namespaces to file-scoped
- Reduce nesting by early exiting functions
- Replace `using` block with `using` declaration
- Rename a few private variables for consistency

GitHub's diff viewer behaves weirdly on this one, so I'd recommend reviewing it in an IDE. I don't know about Visual Studio, but in Rider it looks fine.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -58,7 +58,7 @@ dotnet_style_prefer_conditional_expression_over_return = true:silent
 ###############################
 # Style Definitions
 dotnet_naming_style.pascal_case_style.capitalization             = pascal_case
-# Use PascalCase for constant fields  
+# Use PascalCase for constant fields
 dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
 dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
 dotnet_naming_rule.constant_fields_should_be_pascal_case.style = pascal_case_style
@@ -134,7 +134,7 @@ csharp_preserve_single_line_statements = true
 csharp_preserve_single_line_blocks = true
 csharp_using_directive_placement = outside_namespace:silent
 csharp_prefer_simple_using_statement = true:suggestion
-csharp_style_namespace_declarations = block_scoped:silent
+csharp_style_namespace_declarations = file_scoped:silent
 csharp_style_prefer_method_group_conversion = true:silent
 csharp_style_expression_bodied_lambdas = true:silent
 csharp_style_expression_bodied_local_functions = false:silent
--- a/Flow.Launcher/Helper/DataWebRequestFactory.cs
+++ b/Flow.Launcher/Helper/DataWebRequestFactory.cs
@@ -2,68 +2,67 @@
 using System.IO;
 using System.Net;
 
-namespace Flow.Launcher.Helper
+namespace Flow.Launcher.Helper;
+
+public class DataWebRequestFactory : IWebRequestCreate
 {
-    public class DataWebRequestFactory : IWebRequestCreate
+    class DataWebRequest : WebRequest
     {
-        class DataWebRequest : WebRequest
+        private readonly Uri _uri;
+
+        public DataWebRequest(Uri uri)
         {
-            private readonly Uri m_uri;
+            _uri = uri;
+        }
 
-            public DataWebRequest(Uri uri)
-            {
-                m_uri = uri;
-            }
+        public override WebResponse GetResponse()
+        {
+            return new DataWebResponse(_uri);
+        }
+    }
 
-            public override WebResponse GetResponse()
+    class DataWebResponse : WebResponse
+    {
+        private readonly string _contentType;
+        private readonly byte[] _data;
+
+        public DataWebResponse(Uri uri)
+        {
+            string uriString = uri.AbsoluteUri;
+
+            int commaIndex = uriString.IndexOf(',');
+            var headers = uriString.Substring(0, commaIndex).Split(';');
+            _contentType = headers[0];
+            string dataString = uriString.Substring(commaIndex + 1);
+            _data = Convert.FromBase64String(dataString);
+        }
+
+        public override string ContentType
+        {
+            get { return _contentType; }
+            set
             {
-                return new DataWebResponse(m_uri);
+                throw new NotSupportedException();
             }
         }
 
-        class DataWebResponse : WebResponse
+        public override long ContentLength
         {
-            private readonly string m_contentType;
-            private readonly byte[] m_data;
-
-            public DataWebResponse(Uri uri)
+            get { return _data.Length; }
+            set
             {
-                string uriString = uri.AbsoluteUri;
-
-                int commaIndex = uriString.IndexOf(',');
-                var headers = uriString.Substring(0, commaIndex).Split(';');
-                m_contentType = headers[0];
-                string dataString = uriString.Substring(commaIndex + 1);
-                m_data = Convert.FromBase64String(dataString);
-            }
-
-            public override string ContentType
-            {
-                get { return m_contentType; }
-                set
-                {
-                    throw new NotSupportedException();
-                }
-            }
-
-            public override long ContentLength
-            {
-                get { return m_data.Length; }
-                set
-                {
-                    throw new NotSupportedException();
-                }
-            }
-
-            public override Stream GetResponseStream()
-            {
-                return new MemoryStream(m_data);
+                throw new NotSupportedException();
             }
         }
 
-        public WebRequest Create(Uri uri)
+        public override Stream GetResponseStream()
         {
-            return new DataWebRequest(uri);
+            return new MemoryStream(_data);
         }
+    }
+
+    public WebRequest Create(Uri uri)
+    {
+        return new DataWebRequest(uri);
     }
 }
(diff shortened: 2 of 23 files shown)


.editorconfig
```diff
@@ -134,7 +134,7 @@ csharp_preserve_single_line_statements = true
 csharp_preserve_single_line_blocks = true
 csharp_using_directive_placement = outside_namespace:silent
 csharp_prefer_simple_using_statement = true:suggestion
-csharp_style_namespace_declarations = block_scoped:silent
+csharp_style_namespace_declarations = file_scoped:silent
```
Yusyuriv, line 137: Changed .editorconfig here so new files are created with file-scoped namespace, but it doesn't mind that old files might stay with block-scoped namespace for a while.

Flow.Launcher/Helper/DataWebRequestFactory.cs
```diff
@@ -2,68 +2,67 @@
...
     {
-        class DataWebRequest : WebRequest
-        {
-            private readonly Uri m_uri;
-
-            public DataWebRequest(Uri uri)
-            {
-                m_uri = uri;
-            }
+        private readonly Uri _uri;
```
Yusyuriv, line 11: Renamed from `m_uri` to make it more consistent with the rest of the codebase. It's `private` so it should be safe to rename.

Flow.Launcher/Helper/DataWebRequestFactory.cs
```diff
@@ -2,68 +2,67 @@
...
+            return new DataWebResponse(_uri);
+        }
+    }
 
-            public DataWebResponse(Uri uri)
-            {
-                string uriString = uri.AbsoluteUri;
+    class DataWebResponse : WebResponse
+    {
+        private readonly string _contentType;
```
Yusyuriv, line 26: Same renaming as with `_uri` above.

Flow.Launcher/Helper/ErrorReporting.cs
```diff
@@ -4,45 +4,52 @@
...
+        return info;
+    }
+
+    public static string DependenciesInfo()
+    {
+        var info = $"""
+
+                    Python Path: {Constant.PythonPath}
+                    Node Path: {Constant.NodePath}
+                    """;
```
Yusyuriv, line 52: Same as above.

Flow.Launcher/Helper/HotKeyMapper.cs
```diff
@@ -7,97 +7,96 @@
...
 using Flow.Launcher.ViewModel;
 
-namespace Flow.Launcher.Helper
+namespace Flow.Launcher.Helper;
+
+internal static class HotKeyMapper
 {
-    internal static class HotKeyMapper
+    private static Settings _settings;
+    private static MainViewModel _mainViewModel;
```
Yusyuriv, line 15: Same renaming as with `_uri` a few files above. Also should be safe because these fields are `private`.

Flow.Launcher/Helper/SingleInstance.cs
```diff
@@ -12,375 +12,372 @@
...
+    }
+
+    /// <summary>
+    /// Creates a remote server pipe for communication.
+    /// Once receives signal from client, will activate first instance.
+    /// </summary>
+    /// <param name="channelName">Application's IPC channel name.</param>
+    private static async Task CreateRemoteService(string channelName)
+    {
+        using (NamedPipeServerStream pipeServer = new NamedPipeServerStream(channelName, PipeDirection.In))
```
Yusyuriv, line 322: Didn't touch this `using` block and the next one because they're doing async stuff, and I didn't want to accidentally break something.

Flow.Launcher/Helper/WallpaperPathRetrieval.cs
```diff
@@ -5,44 +5,43 @@
...
+        RegistryKey key = Registry.CurrentUser.OpenSubKey(@"Control Panel\Colors", true);
+        var result = key?.GetValue("Background", null);
+        if (result is string strResult)
         {
-            RegistryKey key = Registry.CurrentUser.OpenSubKey("Control Panel\\Colors", true);
-            var result = key.GetValue(@"Background", null);
-            if (result != null && result is string)
+            try
+            {
+                var parts = strResult.Trim().Split(new[] {' '}, 3).Select(byte.Parse).ToList();
```
Yusyuriv, line 38: The original has a redundant null check (because it immediately checks if it's a string after that) and then uses `.ToString()` on `object` type variable to convert it to a string even though it's already known that it's a string, just for the type system to also understand that it's a string now. I declare the string in the pattern matching if, so no need for additional conversion.

Flow.Launcher/Helper/WindowsInteropHelper.cs
```diff
@@ -8,156 +8,152 @@
...
-                    {
-                        return true;
-                    }
-                }
-            }
+        //if current active window is desktop or shell, exit early
+        if (hWnd.Equals(HWND_DESKTOP) || hWnd.Equals(HWND_SHELL))
+        {
+            return false;
+        }
```
Yusyuriv, line 81: Reduced the two levels of nesting by returning early. I'm kind of on the fence with this one, too. I do think that generally reducing nesting improves readability, but what do you think for this specific case?

Flow.Launcher/Helper/ErrorReporting.cs
```diff
@@ -4,45 +4,52 @@
...
+    public static string RuntimeInfo()
+    {
+        var info =
+            $"""
+
+             Flow Launcher version: {Constant.Version}
+             OS Version: {ExceptionFormatter.GetWindowsFullVersionFromRegistry()}
+             IntPtr Length: {IntPtr.Size}
+             x64: {Environment.Is64BitOperatingSystem}
+             """;
```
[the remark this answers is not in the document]
taooceros: I wonder whether this will make the string ends with a newline?

Flow.Launcher/Helper/SingleInstance.cs
taooceros: I think you can change it to be await using...but not a big deal.

Flow.Launcher/Helper/WindowsInteropHelper.cs
taooceros: Do you change it with rider refactor? I am a bit skeptical about whether it follows the same logic (my mind is not so clear atm).

taooceros (commented):
Nice work! It seems like even rider cannot handle the diff view well for changing blockwise namespace😅but overall it looks good

Flow.Launcher/Helper/ErrorReporting.cs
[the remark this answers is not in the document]
Yusyuriv: It will preserve the new line I added after opening the string and it will ignore the rest of unnecessary whitespace as per [documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string).
image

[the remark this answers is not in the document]
taooceros: I think regardless it is better than before. We can still use `\n` if you think that the first empty line feels weird?

Flow.Launcher/Helper/SingleInstance.cs
taooceros: Maybe could you revert back all the format change in `SingleInstance.cs`? As #2616 is also changing this file (massively), it might be annoying to merge the format change....

Flow.Launcher/Helper/ErrorReporting.cs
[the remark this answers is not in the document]
Yusyuriv: I do think it looks a little bit weird, but the alternative is this, because you can't have escape sequences in raw strings:

```csharp
        var info =
            $"""
             {"\n"}Flow Launcher version: {Constant.Version}
             OS Version: {ExceptionFormatter.GetWindowsFullVersionFromRegistry()}
             IntPtr Length: {IntPtr.Size}
             x64: {Environment.Is64BitOperatingSystem}
             """;
```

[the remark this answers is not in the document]
Yusyuriv: Or something like this:

```csharp
        var info = "\n" +
            $"""
             Flow Launcher version: {Constant.Version}
             OS Version: {ExceptionFormatter.GetWindowsFullVersionFromRegistry()}
             IntPtr Length: {IntPtr.Size}
             x64: {Environment.Is64BitOperatingSystem}
             """;
```

The second example looks better to me than the first one, but I'm still not sure if it looks better than just having it in the raw string directly.

Flow.Launcher/Helper/SingleInstance.cs
Yusyuriv: Done.

Yusyuriv:
> well it looks like github diff (with hide whitespace) is doing a much better job than rider....huh

Huh. Weird. It seems like Rider highlights diffs differently based on where you get the changes from. When I open the changes directly from the pull request, it looks like this:
image

But when I open the same changes in a local branch, the diff looks great:
image

Flow.Launcher/Helper/WindowsInteropHelper.cs
Yusyuriv: Yes, besides changing a comment

taooceros (approved):

PR #2652, opened 2024-04-16
Code cleanup: Flow.Launcher.Plugin.BrowserBookmark

Yusyuriv:
Mostly just moving from block-scoped namespaces to file-scoped, and renaming private properties to start with `_` for consistency.

There are a few other minor changes that I will address in comments.

Since GitHub's diff view doesn't work well for migrating from block-scoped to file-scoped namespaces, and as it turns out, Rider also doesn't display it well when looking directly at the pull request, I suggest pulling this branch in Rider locally and looking at the diff there as it seems to work correctly when the diff is local.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
+++ b/Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
@@ -5,140 +5,133 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 
-namespace Flow.Launcher.Plugin.BrowserBookmark
+namespace Flow.Launcher.Plugin.BrowserBookmark;
+
+public abstract class FirefoxBookmarkLoaderBase : IBookmarkLoader
 {
-    public abstract class FirefoxBookmarkLoaderBase : IBookmarkLoader
+    public abstract List<Bookmark> GetBookmarks();
+
+    private const string QueryAllBookmarks = """
+        SELECT moz_places.url, moz_bookmarks.title
+        FROM moz_places
+         INNER JOIN moz_bookmarks ON (
+             moz_bookmarks.fk NOT NULL AND moz_bookmarks.title NOT NULL AND moz_bookmarks.fk = moz_places.id
+         )
+        ORDER BY moz_places.visit_count DESC
+        """;
+
+    private const string DbPathFormat = "Data Source ={0}";
+
+    protected static List<Bookmark> GetBookmarksFromPath(string placesPath)
     {
-        public abstract List<Bookmark> GetBookmarks();
+        // Return empty list if the places.sqlite file cannot be found
+        if (string.IsNullOrEmpty(placesPath) || !File.Exists(placesPath))
+            return new List<Bookmark>();
 
-        private const string queryAllBookmarks = @"SELECT moz_places.url, moz_bookmarks.title
-              FROM moz_places
-              INNER JOIN moz_bookmarks ON (
-                moz_bookmarks.fk NOT NULL AND moz_bookmarks.title NOT NULL AND moz_bookmarks.fk = moz_places.id
-              )              
-              ORDER BY moz_places.visit_count DESC
-            ";
+        Main.RegisterBookmarkFile(placesPath);
 
-        private const string dbPathFormat = "Data Source ={0}";
+        // create the connection string and init the connection
+        string dbPath = string.Format(DbPathFormat, placesPath);
+        using var dbConnection = new SqliteConnection(dbPath);
+        // Open connection to the database file and execute the query
+        dbConnection.Open();
+        var reader = new SqliteCommand(QueryAllBookmarks, dbConnection).ExecuteReader();
 
-        protected static List<Bookmark> GetBookmarksFromPath(string placesPath)
-        {
-            // Return empty list if the places.sqlite file cannot be found
-            if (string.IsNullOrEmpty(placesPath) || !File.Exists(placesPath))
-                return new List<Bookmark>();
+        // return results in List<Bookmark> format
+        return reader
+            .Select(
+                x => new Bookmark(
+                    x["title"] is DBNull ? string.Empty : x["title"].ToString(),
+                    x["url"].ToString()
+                )
+            )
+            .ToList();
+    }
+}
 
-            var bookmarkList = new List<Bookmark>();
 
-            Main.RegisterBookmarkFile(placesPath);
-
-            // create the connection string and init the connection
-            string dbPath = string.Format(dbPathFormat, placesPath);
-            using var dbConnection = new SqliteConnection(dbPath);
-            // Open connection to the database file and execute the query
-            dbConnection.Open();
-            var reader = new SqliteCommand(queryAllBookmarks, dbConnection).ExecuteReader();
-            
-            // return results in List<Bookmark> format
-            bookmarkList = reader.Select(
-                x => new Bookmark(x["title"] is DBNull ? string.Empty : x["title"].ToString(),
-                    x["url"].ToString())
-            ).ToList();
-
-            return bookmarkList;
-        }
+public class FirefoxBookmarkLoader : FirefoxBookmarkLoaderBase
+{
+    /// <summary>
+    /// Searches the places.sqlite db and returns all bookmarks
+    /// </summary>
+    public override List<Bookmark> GetBookmarks()
+    {
+        return GetBookmarksFromPath(PlacesPath);
     }
 
-
-    public class FirefoxBookmarkLoader : FirefoxBookmarkLoaderBase
+    /// <summary>
+    /// Path to places.sqlite
+    /// </summary>
+    private string PlacesPath
     {
-        /// <summary>
-        /// Searches the places.sqlite db and returns all bookmarks
-        /// </summary>
-        public override List<Bookmark> GetBookmarks()
+        get
         {
-            return GetBookmarksFromPath(PlacesPath);
-        }
-        
-        /// <summary>
-        /// Path to places.sqlite
-        /// </summary>
-        private string PlacesPath
-        {
-            get
-            {
-                var profileFolderPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), @"Mozilla\Firefox");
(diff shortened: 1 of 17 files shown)


Plugins/Flow.Launcher.Plugin.BrowserBookmark/Models/CustomBrowser.cs
```diff
@@ -1,45 +1,44 @@
...
+public class CustomBrowser : BaseModel
 {
-    public class CustomBrowser : BaseModel
-    {
-        private string _name;
-        private string _dataDirectoryPath;
-        private BrowserType browserType = BrowserType.Chromium;
+    private string _name;
+    private string _dataDirectoryPath;
+    private BrowserType _browserType = BrowserType.Chromium;
```
Yusyuriv, line 7: Add `_` to private property name to be consistent with other private property names

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Models/CustomBrowser.cs
```diff
@@ -1,45 +1,44 @@
...
+        set
         {
-            get => _dataDirectoryPath;
-            set
-            {
-                _dataDirectoryPath = value;
-                OnPropertyChanged(nameof(DataDirectoryPath));
-            }
+            _name = value;
+            OnPropertyChanged();
```
Yusyuriv, line 15: `BaseModel` class, which this class extends, has the `[CallerMemberName]` attribute on this argument, so there's no need to pass it manually.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Views/CustomBrowserSetting.xaml
```diff
@@ -63,7 +63,6 @@
             <StackPanel Margin="26,0,26,0">
                 <StackPanel Margin="0,0,0,12">
                     <TextBlock
-                        Grid.Column="0"
```
Yusyuriv, line 66: Not a child of a Grid, this attribute doesn't do anything.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Views/SettingsControl.xaml.cs
```diff
@@ -4,119 +4,116 @@
...
+    private void Others_Click(object sender, RoutedEventArgs e)
+    {
+        CustomBrowsersList.Visibility = CustomBrowsersList.Visibility switch
         {
-            EditSelectedCustomBrowser();
-        }
+            Visibility.Collapsed => Visibility.Visible,
+            _ => Visibility.Collapsed
+        };
+    }
```
Yusyuriv, line 100: In my opinion this is a little easier to read than the original if/else. What do you think?

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
```diff
@@ -5,140 +5,133 @@
...
 {
-    public abstract class FirefoxBookmarkLoaderBase : IBookmarkLoader
-    {
-        public abstract List<Bookmark> GetBookmarks();
+    public abstract List<Bookmark> GetBookmarks();
 
-        private const string queryAllBookmarks = @"SELECT moz_places.url, moz_bookmarks.title
-              FROM moz_places
-              INNER JOIN moz_bookmarks ON (
+    private const string QueryAllBookmarks = """
```
Yusyuriv, line 14: Rename `private const` to start with a capital letter. Same with `DbPathFormat`.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
```diff
@@ -5,140 +5,133 @@
...
+
+        // return results in List<Bookmark> format
+        return reader
+            .Select(
+                x => new Bookmark(
+                    x["title"] is DBNull ? string.Empty : x["title"].ToString(),
+                    x["url"].ToString()
+                )
+            )
+            .ToList();
```
Yusyuriv, line 48: I think it's much more readable when every method call in the chain is on its own line. I'd like to hear your opinion.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
```diff
@@ -5,140 +5,133 @@
...
-            }
+            var profileFolderPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), @"Mozilla\Firefox");
+            var profileIni = Path.Combine(profileFolderPath, @"profiles.ini");
+
+            if (!File.Exists(profileIni))
+                return string.Empty;
+
+            // get firefox default profile directory from profiles.ini
+            using var sReader = new StreamReader(profileIni);
+            var ini = sReader.ReadToEnd();
```
Yusyuriv, line 78: Replaced `using` block with the `using` declaration. Since there's no block anymore, it's no longer necessary to have `ini` declared outside.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
```diff
@@ -5,140 +5,133 @@
...
+                [Profile0]
+                Name=default-release
+                IsRelative=1
+                Path=Profiles/7789f565.default-release
+
+                [General]
+                StartWithLastProfile=1
+                Version=2
+            */
+            var lines = ini.Split("\r\n").ToList();
```
Yusyuriv, line 107: This does exactly the same thing as before, but with less code.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
```diff
@@ -5,140 +5,133 @@
...
+
+            var defaultProfileFolderNameRaw = lines.FirstOrDefault(x => x.Contains("Default=") && x != "Default=1") ?? string.Empty;
```
Yusyuriv, line 109: `Where` call is redundant because `FirstOrDefault` also can accept a function for filtering.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -12,233 +11,234 @@
...
-                    Score = 5,
-                    Action = _ =>
-                    {
-                        context.API.OpenUrl(c.Url);
-                        return true;
-                    },
-                    ContextData = new BookmarkAttributes
+                )
+                .Where(r => r.Score > 0)
+                .ToList();
```
Yusyuriv, line 83: Same as in `FirefoxBookmarkReader.cs`: I think it's much more readable when every chained method call is in a separate line, and I'd like to hear your opinion on this.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
taooceros: I feel like this add newline that does not exist before? have you tested whether it still worked?

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
taooceros: Yeah I think it's fine. Though for me, I majorly like putting the first `.Select` on the same line...though I don't know how others will think about this. @Flow-Launcher/team

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
```diff
@@ -12,233 +11,234 @@
...
 
-        public List<Result> Query(Query query)
+        _cachedBookmarks = BookmarkLoader.LoadAllBookmarks(_settings);
+        _ = MonitorRefreshQueueAsync();
+        _initialized = true;
+    }
+
+    public List<Result> Query(Query query)
+    {
+        // For when the plugin being previously disabled and is now reenabled
```
taooceros, line 50: maybe a hyphen between re and enabled? I don't feel like this is a word...correct me if i am wrong

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Models/CustomBrowser.cs
taooceros: I suppose you want to say private field (they are different in c#).

Yusyuriv: Yeah, sorry, I meant field, not property.

Plugins/Flow.Launcher.Plugin.BrowserBookmark/FirefoxBookmarkLoader.cs
Yusyuriv: This does add a linebreak at the and and also reduce indenting. Luckily, SQL is not a language with significant whitespace, so I don't think it matters, and I think a raw string literal (`"""`) visually conveys the idea that the string is multiline much better than a verbatim string (`@"`). I did test if it works: I added this text when reading from the database:
image
image

Plugins/Flow.Launcher.Plugin.BrowserBookmark/Main.cs
Yusyuriv: You're correct. Added a hyphen.

VictoriousRaptor:
LGTM. Tested?

Yusyuriv:
I did launch it, and it does work as expected. Here are bookmarks from my Firefox after the changes
image

Here's a bookmark from Edge
image

It was probably unintentional, but in your last commit you added an import in two files and modified a project file. @VictoriousRaptor

VictoriousRaptor:
> It was probably unintentional, but in your last commit you added an import in two files and modified a project file. 

Fixed. Sorry about that.

VictoriousRaptor (approved):

PR #2655, opened 2024-04-18
Remove unused SharpVectors library

jjw24:


taooceros:
why is it draft? @jjw24

JohnTheGr8:
what about: https://github.com/Flow-Launcher/Flow.Launcher/blob/de8c8bde76bf5098af58fa7654ca60d191b5efc5/Flow.Launcher/MainWindow.xaml#L9

taooceros:
> what about:
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/de8c8bde76bf5098af58fa7654ca60d191b5efc5/Flow.Launcher/MainWindow.xaml#L9

I think it is not used. Let's delete it. Maybe we add it to make support for svg sometimes but fail to provide a good resolution?

taooceros (approved):

Issue #1919, opened 2023-02-18
BUG: I can't use Flow Launcher with screen readers


MesterPerfect:
### Checks

- [X] I have checked that this issue has not already been reported.

### Problem Description

I am a blind user
I use NVDA/JAWS screen readers to navigate Windows
I can't use Flow Launcher with screen readers
I found it very difficult to install Flow Launcher using an NVDA screen reader
After installing Flow Launcher, I can't open settings, browse search results, or anything

### To Reproduce

1. install Flow Launcher
2. I can't access it in the system tray

### Flow Launcher Version

v1.13.0

### Windows Build Number

windows 10 Version 22H2 (OS Build 19045.2604)

### Error Log

[2023-02-18.txt](https://github.com/Flow-Launcher/Flow.Launcher/files/10772900/2023-02-18.txt)

onesounds:
Hi,

I tested it and found something wrong. I changed it to be readable by a screen reader. It hasn't gone live yet, so we need to make sure it's working properly. The following URL is my modified version, and you can get a portable version to test.
https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/46271078/artifacts

The current UI has a lot of arbitrary changes I've made, so there are parts of it that don't read properly in screen readers (toggle switches, for example), and I don't expect you to be happy with it. I will fix the preference pane as much as I can, and if there are other things you think I should fix, please let me know and I'll do my best to incorporate them.

MesterPerfect:
> Hi,
> 
> I tested it and found something wrong. I changed it to be readable by a screen reader. It hasn't gone live yet, so we need to make sure it's working properly. The following URL is my modified version, and you can get a portable version to test. https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/46271078/artifacts
> 
> The current UI has a lot of arbitrary changes I've made, so there are parts of it that don't read properly in screen readers (toggle switches, for example), and I don't expect you to be happy with it. I will fix the preference pane as much as I can, and if there are other things you think I should fix, please let me know and I'll do my best to incorporate them.

Thank you for responding
I have downloaded your modified version
Installed by clicking on it
But I couldn't access settings to test changes you made
The right-click menu is still not compatible
I mean menu on system tray
I can attach photos and videos if you want

onesounds:
ok please attach video.

MesterPerfect:
> ok please attach video.
ok 
[...]

PR #2664, opened 2024-04-23
Support Screen Reader

onesounds:
## What's the PR
- Resolve https://github.com/Flow-Launcher/Flow.Launcher/issues/1919
- You can enter the TrayMenu from the keyboard.
- Changed Toggle Switch to be selectable with keyboard navigation.
- The title of the selected result item is screen-readable.
- This is a basic change, and we can use this as a reference to work with in the future.

## Test Case
- [x] You should be able to open and select the tray context menu with just the keyboard.
  - WIN+B -> Move to Flow Launcher Tray Icon -> Press the Context Key in Keboard -> Arrow or Tab
- Read the title of the resulting item in a screen reader (NVDA).

## Comment 
- This requires a lot of additional work, which we will continue to work on later.
  - Add screen reader properties to each setting item
  - Make context menu items readable
  - Allows settings to be set for what to read (including subtitle).

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -280,7 +280,6 @@ namespace Flow.Launcher
                 Visible = !_settings.HideNotifyIcon
             };
 
-            //var contextMenu = contextMenu;
 
             var openIcon = new FontIcon
             {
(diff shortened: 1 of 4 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -270,52 +275,57 @@ private void InitializeNotifyIcon()
...
             _notifyIcon = new NotifyIcon
             {
-                Text = Infrastructure.Constant.FlowLauncher,
+                Text = Infrastructure.Constant.FlowLauncherFullName,
                 Icon = Properties.Resources.app,
                 Visible = !_settings.HideNotifyIcon
             };
 
-            contextMenu = new ContextMenu();
+            //var contextMenu = contextMenu;
```
jjw24, line 283:
```suggestion
```

PR #2668, opened 2024-04-24
Fix MediaPlay sound issue

onesounds:
for test in now.


onesounds:
Occurred because the user deleted the windows media player game. You should consider whether to include a separate check for the media player version in the future.

PR #2669, opened 2024-04-24
Fix Icon Position in PREV/NEXT page hotkey

onesounds:
Clip_20240425_181113

## What's the PR
- Change Item order to PREV/NEXT from NEXT/PREV (page, item)
- Change the label string "Quick Adjust Window Width Size" to "Quick Adjust Window Width" (with Height)


VictoriousRaptor:
Should we swap positions of these two cards? Next page is above currently.

onesounds:
>~~이 두 카드의 위치를 교환해야 할까요? 다음 페이지는 현재 위에 있습니다.~~
> @onesounds 괜찮아? 설명을 업데이트했습니다.
 
I woke up and I'm out of my mind. You can change the order, but please order similar next prev items in the same order. You want to change it to the previous/next order, right? I'll check when I wake up completely.

VictoriousRaptor (approved):

PR #2677, opened 2024-04-29
Fix SettingWindow's LeftTop Position when first launch

onesounds:
## What's The PR
- Fixed getting stuck in the upper left corner when opening the settings window after the first launch.
- If Setting Window Position is null, center the screen.

I think I made it this stupid for a reason, but I can't remember. 
Needs to be reviewed to see if null is acceptable.

## Test Case
When you open the settings window after the first launch with no settings, it is displayed in the center of the screen.

### DIFF
--- a/Flow.Launcher/SettingWindow.xaml.cs
+++ b/Flow.Launcher/SettingWindow.xaml.cs
@@ -446,7 +446,7 @@ namespace Flow.Launcher
 
         public void InitializePosition()
         {
-            if (settings.SettingWindowTop >= 0 && settings.SettingWindowLeft >= 0)
+            if (settings.SettingWindowTop == null)
             {
                 Top = settings.SettingWindowTop;
                 Left = settings.SettingWindowLeft;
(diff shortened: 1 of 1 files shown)


VictoriousRaptor:
It's tricky to dealing with the window. Like if I close the settings window when it's minimized, the next time I open it, it stays minimized.

VictoriousRaptor:
> I think I made it this stupid for a reason, but I can't remember.

Maybe because https://github.com/Flow-Launcher/Flow.Launcher/pull/1368#issuecomment-1244957935

Don't have much time to check but why is it null on first launch?

onesounds:
> > I think I made it this stupid for a reason, but I can't remember.
> 
> Maybe because [#1368 (comment)](https://github.com/Flow-Launcher/Flow.Launcher/pull/1368#issuecomment-1244957935)
> 
> Don't have much time to check but why is it null on first launch?

image
(I didn't know too so I looked it up)
The settings window doesn't have a default value on initial launch, so it seems to have done so.

onesounds:
Fixed. (I think I wrote the code while drunk.)

jjw24 (approved):

PR #2679, opened 2024-04-30
Improve Explorer Preview Panel

onesounds:
## What's the PR

- The preview shows the file size and the dates it was created and modified.
- This item can be turned on and off by the user. All three are turned on and off at once.
- General > Show details in preview area (on/off). Default is off.

## ETC
- If we need more in the future, we may need to turn them on and off on a per-item basis.
  - Media Lengh...?
- We might need a separate settings page for the preview area. (not now...)
- Users may want to change the timeformat, but this is how we'll proceed for now.

## Test Cases
- It should be on and off depending on settings.
- Items with no file size information(ex>folder) should not be visible at all.
- Colors should display properly even if the theme changes.

## Todo
- [ ] Check ResultManager.cs logic
- [ ] Can we add folder created/modifed date? (Folder size is not considered. This is usually computationally intensive.)


VictoriousRaptor:
IMO the implementation of preview panel should be moved to explorer plugin. it's all about files.

Yusyuriv:
I moved the new functionality into the Explorer plugin itself.

I have a few issues with the way the date is currently displayed:
1. I think `yy-M-dd` date format is very confusing. It's `24-5-20` right now. I would strongly prefer `yyyy-MM-dd` (`2024-05-20`) or an option in the settings for the user to be able to customize it.
2. Same with time. `hh:mm` displays time in 12-hour format, but it doesn't specify AM/PM. At the very least it should display AM/PM (add `tt` at the end), but there's a problem: with `CultureInfo.CurrentCulture`, it does translate the date, but in my `CurrentCulture` AM/PM doesn't exist, we use 24-hour clock, so I get an empty space where AM/PM should be if I add `tt` in the format.

onesounds:
Let's add the Show Toggle option and DateTimeFormat in plugin setting.

Yusyuriv:
Time formats and date formats in `Settings.cs` probably need some adjustments.

onesounds:
There's a separator color issue, which I'll fix in a separate PR. (Previously, I didn't want to add a style to the separator color, so I made it get the color from the already assigned separator. Changing to panels in this PR is causing this color to not be imported. Need to add a separate style).

Yusyuriv (approved):

Issue #2682, opened 2024-05-01
BUG: Sound Effect broken on v1.18


indigofairyx:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

theres no longer a sound effect when opening the search box. tired changing the new volume slide. no luck.

### To Reproduce

1. ... open the search box
2. ...
3. ...

### Flow Launcher Version

1.18.0

### Windows Build Number

Microsoft Windows [Version 10.0.19045.4291]

### Error Log

<details>

```shell

Replace this line with the important log contents.

```
</details>

onesounds:
Please check if you have deleted Windows Media Player from the system.

indigofairyx:
hmm, yes i have removed WMP, a year ago. and the sound works on v1.17

i tired to leave it but it couldnt handle not being the "default" and keep popping up like a red-head-step-child. sooo oh well i guess?

realgooseman:
I confirm, I deleted WMP years ago and the opening sound effect worked fine till v1.17.
Now, no sound at all even with the setting turned on in `Appearance > Sound effect`

Is it possible to make the sound work without Windows dependencies or at least make it use the system's default player if possible (VLC in my case for example).

indigofairyx:
im not the one to answer this...

 thou. i can say i reverted back to v1.17 and the sound is working as it was, without WMP. so i doubt it that causing the bug. also windows doesn't "remove or uninstall" any features from the "additional features" list, it only hides them, WMP is still in the system, its just not trying to trying to set itself as the default anymore.

onesounds:
@realgooseman @indigofairyx 

We use a framework called WPF, and there are two functions for playing sound. There are two functions: soundplay() and mediaplay(). We used soundplay() until version 1.17, and it didn't have volume control. There was a volume adjustment request pr and we changed the function to mediaplay()(It has a volume control function.) function. But mediaplay relies on WMP to work. this is what happens. 

As @realgooseman said, we can add something to play the sound without any dependencies. But I feel it's unreasonable to put this in just to play "a single sound". Uninstalling WMP is not recommended as it will cause problems for programs other than ours. (Just curious, Why is it so common to uninstall WMP? It is intentionally made inconvenient to delete, and tells you that it may cause problems when deleted.)

We're going to take care of it by putting these warnings in for now.

[...]

PR #2683, opened 2024-05-01
Add Windows Media Player Warning

onesounds:
## What's the PR
image
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/2682

- If WMP doesn't exist on your system and can't adjust volume, it will warn and guide you beforehand.
- Check if the executable of Windows Media Player(Legacy) exists by reading registry and turn it on/off.
  - Fallback to use `SoundPlayer` to play sound if there no WMP

## More Info 
- After merging https://github.com/Flow-Launcher/Flow.Launcher/pull/2488, Multiple “sound not playing” issues reported
- I investigated and found that these are users who have uninstalled WMP.
- I looked into it and found that many people were uninstalling wmp. (Many posts offer this as a tip - the system warns you that there may be a problem, but many people ignore it)
  - soundplay() is not volume adjustable, and mediaplayer() is WMP-based, so it won't play without WMP.
  - I decided this was better than adding a library for WAV playback.
  - WMPlib is in system32 and there seems to be a way to utilize it, but I don't want to bother investigating further.

## I am upset with the following situation 🤬
- .NET uses WMP for sound play.
- .NET doesn't provide a volume property for soundplay(). Who in the world decides not to put a volume property in a sound play function?
- There are many posts telling people to delete WMP that is utilized in their system. 
- Despite the warnings on the system, people delete WMP. 
- We are the ones who suffer from all these problems.


Flow.Launcher/Languages/en.xaml
```diff
@@ -158,6 +158,7 @@
     <system:String x:Key="SoundEffectTip">Play a small sound when the search window opens</system:String>
     <system:String x:Key="SoundEffectVolume">Sound Effect Volume</system:String>
     <system:String x:Key="SoundEffectVolumeTip">Adjust the volume of the sound effect</system:String>
+    <system:String x:Key="SoundEffectWarning">Flow Launcher can't play sound because "Windows Media Player" doesn't exist on your system. If you need sound, install WMP.</system:String>
```
VictoriousRaptor, line 161:
```suggestion
    <system:String x:Key="SoundEffectWarning">Sound effect is unavailable because Windows Media Player is missing. Please install it if you need sound effect.</system:String>
```

Garulf:
Would this be better?

"Windows Media Player is unavailable and is required for sound effects. Please check your installation."

taooceros:
maybe try something like this https://github.com/mobiletechtracker/NetCoreAudio?

Yusyuriv:
Adding a new dependency to play a single sound sounds wrong to me.

onesounds:
- Change to use soundplay to play sound if there no WMP
- Change the warning message to "Volume cannot be adjusted".
- Warning is hidden by default. If there is no WMP, it wll be visible.

onesounds:
- Add disable the volume control when without WMP
  - Disabled only the control, when it should have correctly output as disabling the entire card. Apply this after completing the SPLIT SETTINGS. (I don't know if there is a disable style)
- Issue reporters have reported that this PR is working.

taooceros:
@onesounds do you want to abstract this piece out to a usercontrol?

jjw24:
I am going to enable squash and merge into one commit in case we need to get this out sooner.

onesounds:
> @onesounds do you want to abstract this piece out to a usercontrol?

not now.

jjw24 (commented):
@VictoriousRaptor can we move the check media player even higher, rather than when main window initializes maybe when we actually starting up the application? Put the check result as a constant variable maybe in https://github.com/Flow-Launcher/Flow.Launcher/blob/dev/Flow.Launcher.Infrastructure/Constant.cs or user settings class (but don't persist it to disk).

If user reinstalls the media player to fix the problem they can restart flow.

onesounds:
Need to work on UI to match split setting pr. Will be worked on after split setting merge.

Yusyuriv (dismissed):

VictoriousRaptor (approved):

Issue #2692, opened 2024-05-06
The ability to re-queue last queries.


j-a-s-o-n-g:
In Alfred, you're able to use Alt + up arrow to go to your previous query, then again to go to the one before that. I know FlowLauncher already stores like the last 10 queries in the history config file, so it should be fairly easy to implement.

Easy option: Alt+Up goes to previous queries.

Harder option: Alt+Up cycles back, Alt+Down cycles forwards.

Superstar option: In settings, set keybinds for cycling back and forward, then have those work as mentioned above.

onesounds:
We'll work on it. 

This is a bit of an old PR, but will be merged.
https://github.com/Flow-Launcher/Flow.Launcher/pull/852

Shortcut settings PRs are currently merged in DEV, so they need to be reworked accordingly.

onesounds:
Done. It is included in the next release.

PR #2696, opened 2024-05-15
Ported "Up Arrow key cycles through Query history" PR

onesounds:
## What's the PR
- Convert https://github.com/Flow-Launcher/Flow.Launcher/pull/852
- Solved https://github.com/Flow-Launcher/Flow.Launcher/issues/2692
- Pressing the up arrow key when no query will autocomplete the previous excuted query.
- ALT+UP, ALT+DOWN cycling query.
- Hotkeys are changable.

## Todo
- [x] Don't forget Strings

## Test Cases
- If you close the window and start the cycle again, you should see the most recent query first.
- The cycle stops when it acts on the earliest query.

## ETC
- Should the query cycle completely? My feeling is that the behavior of just stopping at the end is correct. (Current state)
  - I've tried different CLI. it should.


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1116,6 +1163,7 @@ public void Show()
 
         public async void Hide()
         {
+            lastHistoryIndex = 1;
```
VictoriousRaptor, line 1166: Why is this reset needed?

VictoriousRaptor: > If you close the window and start the cycle again, you should see the most recent query first.

If a user set last query style to preserve or selected that means he wants to continue the working state if so. Maybe better to reset it only when empty?

onesounds: > Why is this reset needed?

When the user closes the window, the index is initialized. Lastemptyquery is just a setting about query input, so it shouldn't have anything to do with this setting. It's kind of weird when I try it in  without doing this initialization. Get rid of it and try it. It's weird.

onesounds: > > If you close the window and start the cycle again, you should see the most recent query first.
> 
> If a user set last query style to preserve or selected that means he wants to continue the working state if so. Maybe better to reset it only when empty?

@VictoriousRaptor When the user closes the window, the index is initialized. Lastemptyquery is just a setting about query input, so it shouldn't have anything to do with this setting. It's kind of weird when I try it in  without doing this initialization. Get rid of it and try it. It's weird.

It is also not a good idea to initialize when user delete a query. Not closing the window means that the user intends to keep it that way. When using the function in other cli programs, the index is maintained even if the query is deleted. This is a feature that cli users will mainly use, so it's right to follow this UX.

VictoriousRaptor:
LGTM.

jjw24 (approved):

Issue #2595, opened 2024-02-27
When search window appears mouse cursor is triggered to highlight result where the cursor is positioned 


lotsocds:
I'm a user coming from many years using Launchy.  Flow Launcher is fabulous and I plan to stick with it.  However, there is one useability issue that slows me down, thus this enhancement request.  

The request is when the Flow Launcher window opens always highlight the top list entry if the list is not empty, never highlight the list entry under the mouse until the mouse is moved.  To illustrate, please follow the steps below which show the undesired behavior at step 6 and the desired behavior at step 11.  Why?  Because my use case involves running the same program multiple times.  I'd like to hit the Alt-Space hotkey then Enter to start the top program in the list. If the mouse happens to be in the list area when Flow Launcher opens, the wrong program will be started.  Moving the mouse off the wrong entry or remembering to hit Alt-1 will work but both are quite awkward compared to just Alt-Space, Enter.

Thanks for considering and I hope this is possible.

1. Use hotkey to open Flow Launcher.  
2. Type a couple characters to populate the list.  
3. Put the mouse in the middle of the list, e.g. on the second entry.  Don't move it again.
4. Hit Esc to close the window.  

5. Reopen Flow Launcher using the hot key.  
6. Notice that the list entry under the mouse is highlighted. (Undesired behavior)

7. Backspace to clear the entry field.
8. Hit Esc to close the window.  
9. Reopen Flow Launcher using the hot key.  
10. Type a couple characters to populate the list.
11. Notice that the list entry under the mouse is not highlighted until the mouse is moved. (Desired behavior)

taooceros:
Your step 3 already change the selection. I wouldn't say we should reselect to the top result whenever reopen the window, which might be weird.

Yusyuriv:
> I wouldn't say we should reselect to the top result whenever reopen the window

If I understood correctly, they aren't asking to reselect the top result, but to ignore the mouse cursor when the window appear animation is in action. I also had this issue a few times. This issue only happens when the animations are enabled because Flow Launcher moves slightly up when it appears, triggering the selection with the mouse cursor. When animations are disabled this doesn't happen. The correct sequence of actions to reproduce this is:

0. Enable animations in the settings if they are disabled.
1. Open Flow Launcher.
2. Type something to populate the list.
3. Leave the first result selected, hit Esc to close the window.
4. Put the mouse cursor where the third result would be if the window was still open.
[...]

PR #2697, opened 2024-05-16
Block Mouse Hover event when Show / Keyboard Navigation

onesounds:
## What's the PR
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/2595
- When the window is visible, it blocks mouse hover events. 
- Fixes a situation where the mouse is recognized when the up or down arrow key is pressed and held while the mouse is over a window.

## Test Cases
- [x] If the animation is on and the mouse is moved while the animation is running, the mouse hover is not work.
- [x] In a situation with many results, hover should not work when the mouse is moved while holding down the up or down arrow key.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -47,6 +47,7 @@ namespace Flow.Launcher
         private MainViewModel _viewModel;
         private bool _animating;
         MediaPlayer animationSound = new MediaPlayer();
+        private bool isArrowKeyPressed = false;
 
         #endregion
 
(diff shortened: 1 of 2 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -644,10 +649,12 @@ private void OnKeyDown(object sender, KeyEventArgs e)
             switch (e.Key)
             {
                 case Key.Down:
+                    isArrowKeyPressed = true;
```
VictoriousRaptor, line 652: Just did a quick test and found that holding Next Result hotkey also has this issue. IMO we should just left it. hard to fix for every hotkey.

onesounds: I agree with you. Basically, the user mainly uses the single key, but the shortcut key that combines the two will not be used. (Pressing key situation) I think we can ignore other than this key.

jjw24 (approved):

PR #2700, opened 2024-05-17
add a close overload for jsonrpcpluginv2

taooceros:
Send a request to jsonrpcv2 when flow is closing.

### DIFF
--- a/Flow.Launcher.Core/Plugin/JsonRPCPluginV2.cs
+++ b/Flow.Launcher.Core/Plugin/JsonRPCPluginV2.cs
@@ -5,7 +5,6 @@ using System.IO.Pipelines;
 using System.Threading;
 using System.Threading.Tasks;
 using Flow.Launcher.Core.Plugin.JsonRPCV2Models;
-using Flow.Launcher.Infrastructure.Logger;
 using Flow.Launcher.Plugin;
 using Microsoft.VisualStudio.Threading;
 using StreamJsonRpc;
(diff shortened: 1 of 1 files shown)


Flow.Launcher.Core/Plugin/JsonRPCPluginV2.cs
```diff
@@ -139,11 +140,24 @@ public virtual Task ReloadDataAsync()
...
+            try
+            {
+                await RPC.InvokeAsync("close");
+            }
+            catch (RemoteMethodNotFoundException e)
+            {
+            }
+            catch (Exception e)
+            {
+                Log.Exception(
```
jjw24, line 154: Is this the case where the plugin hasn't imemted the 'close' method? If so do we need to be logging it?

taooceros: This is not the case that plugin didn't implement the `close` method. That is the previous exception branch. I don't think we really need to log it, because if plugin doesn't have the close method, we can safely assume that they don't need it. Further, given that we already release jsonrpc v2 even though without documentation and as prerelease, it may be a problem to break the backward compatibility. And it doesn't gain us anything for logging it, unless we mark it as exception?

Yusyuriv:
LGTM, just remove the Logger import since you're no longer logging anything.

Flow.Launcher.Core/Plugin/JsonRPCPluginV2.cs
```diff
@@ -139,11 +139,18 @@ public virtual Task ReloadDataAsync()
...
+            try
+            {
+                await RPC.InvokeAsync("close");
+            }
+            catch (RemoteMethodNotFoundException e)
+            {
+            }
+
             RPC?.Dispose();
             ErrorStream?.Dispose();
```
Yusyuriv, line 153: Actually, sorry, one more thing: shouldn't .Dispose() here be in a `finally` block now, so they're disposed of regardless of if there are any other exceptions or not?

Yusyuriv (approved):

PR #2715, opened 2024-05-24
Add Settings.XamlStyler to root of the solution

Yusyuriv:
Add solution-wide config for XAML Styler, which is available as an extension for JetBrains Rider, Visual Studio, and as a standalone CLI tool. This pull request includes the default settings with a few small changes:

1. `ThicknessSeparator` is set to `1` (space-separated): `Margin="1,2,3,4"` turns into `Margin="1 2 3 4"`
1. `ThicknessAttributes` also includes `CornerRadius`


onesounds (approved):

onesounds:
This PR is to add a setting to the solution that simplifies that code when using xamlstyler. All users of xamlstyler will follow this setting.

PR #2740, opened 2024-05-29
Change Caching Policy

taooceros:
Change Cache to Concurrentlfu in BitFaster.Caching. The eviction policy seems to be undesirable for our cases.

https://github.com/neon-sunset/fast-cache/issues/107

Changes:
- Remove the usage property in the cache, since we are not managing the cache ourselves.

Tested:

- [x] With a relatively small cache size, the icon is not flickering when type continuous query that consistently get the same result. (like "vscode" or "powerpoint").


Flow.Launcher.Infrastructure/Image/ImageCache.cs
```diff
@@ -3,97 +3,81 @@
...
-        {
-            this.usage = usage;
-            imageSource = image;
-        }
-    }
-
     public class ImageCache
     {
-        private const int MaxCached = 150;
+        private const int MaxCached = 10;
```
VictoriousRaptor, line 14: is it enough for explorer plugin that has many results?

taooceros: Oops sorry I forget to change it back. It's for testing🥲

onesounds:
After changing maxcached, I noticed that when I return after keystrokes and context menu switching, the icons are not flickering and display well.

onesounds (approved):

Issue #2030, opened 2023-04-05
BUG: F1 preview of images does not fill the whole preview area


TheBestPessimist:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

I have a bunch of JPEG and PNG images which i use as wallpapers, so they all have resolution of 1080p or higher.

While using FL to search and preview my pictures, it seems picture preview does not fill the whole preview area: some pics are small, others are larger, etc

What I expected is that FL fills the preview area according to the "highest dimension (of the pic/preview area)" or somehow the biggest ratio of pic vs preview area.

### Screenshots

Please see here: https://cdn.discordapp.com/attachments/917500039931256933/1093227949655855134/FL_preview_issue.gif

### Flow Launcher Version

1.14.0. some beta version

### Windows Build Number

Microsoft Windows [Version 10.0.22621.1485]

ligq:
I also have this issue.

taooceros:
you mean this? https://github.com/Flow-Launcher/Flow.Launcher/blob/8f58d53b8a3a9eedc13e2be1cf7fb5f9e0950ad9/Flow.Launcher.Infrastructure/Image/ImageLoader.cs#L288-L322

TheBestPessimist:
bad bot

VictoriousRaptor:
Looks like comment doesn't remove stale label.

onesounds:
This issue will be resolved so that the image fills the preview panel area as much as possible.
However, you may find that the image resolution is low (it's set low because it affects the speed of the program), and I'll consider a separate setting for this. (but in now. this is low priority work. ) 
If you need to check high-resolution images, you're better off using quicklook, (we're working on support for quicklook, which will be merged soon.)

PR #2741, opened 2024-05-30
Adjust Preview Panel

onesounds:
## What's the PR
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/2030
- The image appears full in the preview pane.
- Vertically long images are prioritized to be vertically full.
- Specify the minimum height in the preview area for too few results
- Fix a problem where the screen center separator appears later than the result
- Fix the color of the separator in the explorer plugin's preview panel.
- Fixing a problem where the number of results is adjusted by triggering resize when the window is moved or the width is adjusted.
- When preview is on and there are enough results, temporarily adjust the number of results to fit the preview minimum height. When preview is off, change it back to the existing number of results.
  - If the number of results is very small, we'll just leave it blank.

## Test Cases
- [x] The color of the preview separator must be the same as the bottom separator of the search window, and there must be no binding errors.
- [x] The image should be displayed as horizontally or vertically as possible as the size of the preview area.
- [x] When the preview is turned on with the minimum number of results, the number of results should not change if the window is moved or the width is adjusted.
- [x] If you turn on previw when the maximum number of results to display is low, it temporarily adjusts the number by the size of the preview.


taooceros:
> Vertically long images are prioritized to be vertically full.

What does this mean? I think on dev it is vertically full?

onesounds:
> > Vertically long images are prioritized to be vertically full.
> 
> What does this mean? I think on dev it is vertically full?

 Long vertically images fill the vertical area first. (that is, there is a margin on the left and right) I don't remember what the current dev state is like. In dev, the transverse is certainly not filled within the preview area, and it is a part adjusted for this change.

Yusyuriv:
I just noticed that if an image is too small, it still gets scaled up to fill all available space. Is that okay?

PR #2742, opened 2024-05-30
Integrate native Windows context menu directly into the Explorer plugin

### LINKED ISSUE #1192: [Add action : open a new e-mail with flow search-bar result as attachment]
So you might want to send a quick mail with let say a pdf document.

When I find the item I want, I would like to send it with my favorite mail provider.

Frustration is that now I can only copy the path or the file, so I have to do more click to open the mail application.

**Additional context**
This feature exist on other products like listary if I am correct.

image

Yusyuriv:
Native context menu has `Send to > Email recipient` option, so fixes #1192?

## What's the PR
- This PR integrates the native Windows context menu directly into the Explorer plugin.

- This feature is enabled through the Explorer plugin and can be turned on or off as needed.
- You can filter out specific context menu items that you don't want to display.
- The menu is set at a slightly different height.
- Features that are unavailable or overlap with the plugin are excluded.
- The existing open context menu and unavailable menu will be maintained.

## Why do we need a feature
- It allows users to access features not currently offered by Flow Launcher through the native context menu.
- While there is already an option to open the context menu, this new integration improves search functionality and accessibility.

## TODO
- [x] Correctly free all memory in case of exceptions
- [ ] Test on different versions of Windows
  - [x] Windows 11
  - [ ] Windows 10
  - [ ] Windos 7?
- [x] Add settings option to disable the native context menu
- [ ] Show a message to user if the selected context menu action failed?
- [ ] Display some icon when a context menu item doesn't have an icon?
- [ ] ~~Combine this with the existing `Helper/ShellContextMenu.cs`?~~

## Known issues
- [x] Icons look bad because it's 16x16 icons stretched to 32x32
- [x] If you launch something via this context menu and then close Flow Launcher, the thing you launched from the context menu also closes. I feel like that shouldn't happen.
- [x] `Pin to Start` returns error code `80070005`.
- [x] `Share` returns error code `80070578`.
- [ ] Off-by-one error in `Send to` submenu
- There is a very small delay when retrieving the context menu. It's minor and can be addressed later with caching.

There might be other commands that don't work.

## Test Cases
- [ ] You must be able to exclude specific menu items using a word filter.
- [ ] There should be no errors with the features provided.

## Summary by CodeRabbit

- **New Features**
  - Added support for native context menu in the Explorer plugin.

- **Improvements**
  - Updated UI margins for better alignment and consistency.
  - Enhanced context menu logic to respect user settings for included and excluded items.

- **Settings**
  - Introduced new settings for managing Windows context menu items in the Explorer plugin.


taooceros:
Do you use the win11 context menu or the old one?

Yusyuriv:
I believe it's the old menu.

Yusyuriv:
I'll need to take a closer look at the changes to Result tomorrow.

onesounds (approved):

PR #2748, opened 2024-05-31
Small Fix in Setting window

onesounds:
# What's the PR
## Adjust Accent Button Color
- Adjust to match winui style
- Changed mouse over and color.
- Add a line at the bottom of the button in the Light Color Scheme state

## Add "User Data Location" item in about page

 - Added a button to open the location of the currently used settings folder. 
 - If it is portable, it should change to the portable location and open.
 - Added because users often ask for backups or locations.

## Toggle Switch Focus Visual 
 - Live: No focus visual
 - DEV : Focus visual entered incorrectly while making changes
 - CURRENT: Fixed the focus visual.
image

 - I found some issues with other controls as well, but I'll put this off and work on it when Upgrad the WPF UI by MS.

## Adjust Sidebar Width
 - Sometimes labels are long, like in Japanese, so I increased the length a bit. Auto Size is not supported by modernwpf, so this is what I did for now.

## Fix Fullscreen Titlebar in settingwindow
- Fixed an issue that caused the window area to be slightly larger when the window was maximized. (In this case, it feels like there is not enough margin). This issue creates a problem where the minimize/maximize buttons appear to be out of line
- Fixed by giving some margin when maximizing.

## Adjust Margin in Pages
- The items on the general page are widely spaced, (vertical) so I've narrowed it down a bit. 
- It should feel similarly spaced to the other pages.

# Test Cases
 - Open Folder behavior should be correct. (including portable mode)


Flow.Launcher/Languages/en.xaml
```diff
@@ -75,6 +75,7 @@
     <system:String x:Key="autoUpdates">Auto Update</system:String>
     <system:String x:Key="select">Select</system:String>
     <system:String x:Key="hideOnStartup">Hide Flow Launcher on startup</system:String>
+    <system:String x:Key="hideOnStartupToolTip">Flow Launcher starts out hidden in the tray.</system:String>
```
jjw24, line 78:
```suggestion
    <system:String x:Key="hideOnStartupToolTip">Flow Launcher search window is hidden in the tray after starting up.</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -279,6 +280,9 @@
     <system:String x:Key="clearlogfolder">Clear Logs</system:String>
     <system:String x:Key="clearlogfolderMessage">Are you sure you want to delete all logs?</system:String>
     <system:String x:Key="welcomewindow">Wizard</system:String>
+    <system:String x:Key="userdatapath">User Data Location</system:String>
+    <system:String x:Key="userdatapathToolTip">Installed plugins and user settings are saved in the User data folder. This location may vary depending on whether you are in portable mode or not.</system:String>
```
jjw24, line 284:
```suggestion
    <system:String x:Key="userdatapathToolTip">User settings and installed plugins are saved in the user data folder. This location may vary depending on whether it's in portable mode or not.</system:String>
```

jjw24 (approved):

PR #2749, opened 2024-06-02
Use binding for values introduced in Resizable PR

VictoriousRaptor:
Use binding for values introduced in Resizable PR to simplify code.

Tests:
- [x] Color schemes can be changed and saved.
- [x] Font can be changed and saved.

### DIFF
--- a/Flow.Launcher/SettingPages/ViewModels/SettingsPaneThemeViewModel.cs
+++ b/Flow.Launcher/SettingPages/ViewModels/SettingsPaneThemeViewModel.cs
@@ -22,6 +22,7 @@ namespace Flow.Launcher.SettingPages.ViewModels;
 
 public partial class SettingsPaneThemeViewModel : BaseModel
 {
+    private const string DefaultFont = "Segoe UI";
     public Settings Settings { get; }
 
     public static string LinkHowToCreateTheme => @"https://flowlauncher.com/docs/#/how-to-create-a-theme";
@@ -144,11 +145,13 @@ public partial class SettingsPaneThemeViewModel : BaseModel
     }
 
     public IEnumerable<int> MaxResultsRange => Enumerable.Range(2, 16);
+
     public bool KeepMaxResults
     {
         get => Settings.KeepMaxResults;
         set => Settings.KeepMaxResults = value;
     }
+
     public string ClockText => DateTime.Now.ToString(TimeFormat, CultureInfo.CurrentCulture);
 
     public string DateText => DateTime.Now.ToString(DateFormat, CultureInfo.CurrentCulture);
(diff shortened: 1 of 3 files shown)


Yusyuriv:
I don't know if it's possible to post suggestions in the code that isn't included in the PR changes, so I'll write it here: maybe get rid of `Reset_Click` completely, since all the logic is now in the view model?

```diff
-    private void Reset_Click(object sender, RoutedEventArgs e)
-    {
-        _viewModel.Reset();
-    }
```
```diff
+    [RelayCommand]
     public void Reset()
     {
         SelectedQueryBoxFont = new FontFamily(DefaultFont);
```
```diff
                             <Button
                                 Margin="8"
                                 HorizontalAlignment="Stretch"
-                                Click="Reset_Click"
+                                Command="{Binding ResetCommand}" 
                                 Content="{DynamicResource resetCustomize}" />
```

onesounds:
Works fine. when you done and need approve, mention me again.

Flow.Launcher/SettingPages/ViewModels/SettingsPaneThemeViewModel.cs
```diff
@@ -94,30 +96,9 @@ public double ResultSubItemFontSize
...
-                var key = $"ColorScheme{e}";
-                var display = InternationalizationManager.Instance.GetTranslation(key);
-                var m = new ColorScheme { Display = display, Value = e, };
-                modes.Add(m);
-            }
-
-            return modes;
-        }
-    }
+    public List<ColorSchemeData> ColorSchemes { get; } = DropdownDataGeneric<ColorSchemes>.GetValues<ColorSchemeData>("ColorScheme");
```
taooceros, line 101: This may have the issue that it may not reflect the language change in flight? Are the previous version able to handle that?

Yusyuriv: The old version also didn't handle language change. In this case it doesn't matter though, because the language change can only happen in the General tab. When the user goes back to the Appearance tab after that, it will re-render completely, including those checkboxes.

VictoriousRaptor:
> I don't know if it's possible to post suggestions in the code that isn't included in the PR changes, so I'll write it here: maybe get rid of `Reset_Click` completely, since all the logic is now in the view model?
> 
> ```diff
> -    private void Reset_Click(object sender, RoutedEventArgs e)
> -    {
> -        _viewModel.Reset();
> -    }
> ```
> 
> ```diff
> +    [RelayCommand]
>      public void Reset()
>      {
>          SelectedQueryBoxFont = new FontFamily(DefaultFont);
> ```
> 
> ```diff
>                              <Button
>                                  Margin="8"
>                                  HorizontalAlignment="Stretch"
> -                                Click="Reset_Click"
> +                                Command="{Binding ResetCommand}" 
>                                  Content="{DynamicResource resetCustomize}" />
> ```

Done.

Yusyuriv (approved):

PR #2750, opened 2024-06-02
Add theme metadata

Yusyuriv:


onesounds:
This work will be included and merged into #2735.

Issue #2320, opened 2023-09-03
BUG: Wrong positioning on multimonitor setup


real-yfprojects:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

I have a multi monitor setup with two screens 1920x1200 (screen 1) and 1900x1080.
![grafik](https://github.com/Flow-Launcher/Flow.Launcher/assets/62463991/e3c05ef3-44fb-4b88-b542-d8cd201f12be)

On monitor 2 the following settings for the *window position on monitor* result in the search bar being shown on monitor 1...

- ...on the right top for center top 
- ...on the left top for left top
- ...not at all for right top (however the current window looses focus -> its probably displayed out of bounds)

### To Reproduce

1. Configure setting
2. Set *window position* to custom monitor 2 (the same happens for monitor for mouse cursor or monitor with focused window)
3. Hit alt+enter

### Flow Launcher Version

1.16.1

### Windows Build Number

10.0.22621.2134

### Error Log

Nothing gets added to the logs.

<details>

```shell

Replace this line with the important log contents.

```
</details>

onesounds:
Hello,
https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/49954411/artifacts
Please grab the portable version from the link above, give it a try, and let us know if it solves your problem.

real-yfprojects:
Yes, it does. Thank you very much!

Jack251970:
@onesounds Has this issue been resolved?

onesounds:
Idk. I tried working on it but ended up giving up. I’m pretty sure someone else handled it, but I’m not exactly sure.

Jack251970:
Oh no. Tao has directly edited your PR😢, so the fix is reverted finally.

real-yfprojects:
Nice, thank you all!

PR #2756, opened 2024-06-04
Fix Window Positioning with Multiple Montiors

onesounds:
## Review Guide
- You can review it in 30 minutes or less. (Repeated resolution changes are required to test this pr.)
- This code affects the handling of window positioning with "remember last position"
- It is a combination of two things: bug correction and feature improvement.

## What's the PR
- Fixed handling when monitors are arranged vertically
  - Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/2320
- When you change the resolution or dpi after the program ends, adjust the resetting and main windows so that they are not located outside the screen.
  - I usually set this to "last remember position" because it's the most convenient, but it was very annoying when the screen resolution changes when using moonlight. 

## Test Cases
- [x] After the program ends, change the horizontal resolution to be narrower, and then the window position does not deviate from the screen when the program is executed.
  1.  Select the highest resolution in the monitor.
  2. Run flow and place the window (main window and settings window) in the upper right corner.
  3. Exit the program.
  4. Change the system resolution to one less than the 1 setting.
  5. Run flow and open the flow and settings windows. The location of each window should be within the screen, not outside the monitor.

- [x] Remember last position should operate as before
- [x] Other settings other than the last position setting must operate normally.
- [x] When you select Top-Left/Center-Top/Right as the custom monitor position in a vertical dual-monitor alignment, both monitors should display in their normal positions.
- [x] Both the setup window and the main window must work.

## ETC
- This pr initially tried to readjust the window position slightly to the ratio at the time of resolution change, but decided that it was difficult to adjust it to the ratio. Basically, when changing the resolution, it is readjusted so that the window is located inside, so this is not handled separately.
- However, if the resolution changes after the program is terminated, the window may be located outside according to the existing logic, so this part has been improved.

## Summary by CodeRabbit

- **New Features**
  - Enhanced screen resolution and DPI settings, allowing better handling of window positioning across different screen dimensions.
  
- **Bug Fixes**
  - Improved window initialization logic to adapt to changes in screen resolution and DPI, preventing positioning issues when screens are resized or display settings are altered.


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -48,6 +49,10 @@ public partial class MainWindow
 
         private MediaPlayer animationSoundWMP;
         private SoundPlayer animationSoundWPF;
+        private double _previousScreenWidth;
+        private double _previousScreenHeight;
+        private double _previousDpiX;
+        private double _previousDpiY;
```
onesounds, line 55: Store previously used resolution and dpi values.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -306,15 +333,15 @@ private void InitializePosition()
                         break;
                     case SearchWindowAligns.CenterTop:
                         Left = HorizonCenter(screen);
-                        Top = 10;
+                        Top = VerticalTop(screen);
```
onesounds, line 336: Because this value is a fixed value, there is a problem with vertical alignment monitor situation. Try to locate the window and set the position.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -288,41 +293,75 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
+            double currentScreenWidth = screen.WorkingArea.Width;
+            double currentScreenHeight = screen.WorkingArea.Height;
+
+            if (!(
+                _previousScreenWidth == currentScreenWidth &&
+                _previousScreenHeight == currentScreenHeight &&
+                _previousDpiX == currentDpi.X &&
+                _previousDpiY == currentDpi.Y
+            ))
+            {
```
onesounds, line 316: “If anything is different from the previous information”

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -288,41 +293,75 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
+                if (_previousScreenWidth != 0 && _previousScreenHeight != 0 && _previousDpiX != 0 && _previousDpiY != 0)
```
onesounds, line 317: Not sure why it's there...unless they're all non-zero. chatgpt wrote.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -288,41 +293,75 @@ private void OnLoaded(object sender, RoutedEventArgs _)
...
-                        Left = WindowsInteropHelper.TransformPixelsToDIP(this, screen.WorkingArea.X + _settings.CustomWindowLeft, 0).X;
-                        Top = WindowsInteropHelper.TransformPixelsToDIP(this, 0, screen.WorkingArea.Y + _settings.CustomWindowTop).Y;
-                        break;
+                    double widthRatio = currentScreenWidth / _previousScreenWidth;
+                    double heightRatio = currentScreenHeight / _previousScreenHeight;
+                    double dpiXRatio = currentDpi.X / _previousDpiX;
+                    double dpiYRatio = currentDpi.Y / _previousDpiY;
+
+                    _settings.WindowLeft *= widthRatio * dpiXRatio;
+                    _settings.WindowTop *= heightRatio * dpiYRatio;
```
onesounds, line 325: Rescales previously stored values to a ratio.

onesounds:
This function is also applied when the resolution is changed while the program is turned off. It didn't work  well, so I stopped. I'll think about the logic a little more.

Yusyuriv:
This PR breaks these settings.
image

onesounds:
> This PR breaks these settings. image

Rework.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -292,11 +293,39 @@ private void InitializePosition()
...
+                double previousDpiY = _settings.PreviousDpiY;
+
+                // Save current screen width, height, DPI X, and DPI Y
+                _settings.PreviousScreenWidth = SystemParameters.VirtualScreenWidth;
+                _settings.PreviousScreenHeight = SystemParameters.VirtualScreenHeight;
+                _settings.PreviousDpiX = GetDpiX();
+                _settings.PreviousDpiY = GetDpiY();
+
+                // If previous screen width, height, DPI X, or DPI Y are not zero
+                if (previousScreenWidth != 0 && previousScreenHeight != 0 &&
```
taooceros, line 309: what does it mean to be 0?

jjw24:
@onesounds is this ready for review?

Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -112,23 +112,45 @@ public void InitializePosition()
...
+            SetWindowPosition(WindowTop(), WindowLeft());
         }
         else
         {
-            Top = _settings.SettingWindowTop.Value;
-            Left = _settings.SettingWindowLeft.Value;
+            double left = _settings.SettingWindowLeft.Value;
+            double top = _settings.SettingWindowTop.Value;
+            AdjustWindowPosition(ref top, ref left);
+            SetWindowPosition(top, left);
```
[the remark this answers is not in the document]
jjw24: Maybe the condition in above line 113 needs to be 'and' not ''or'? Because one of them could be a null and still passes through the condition check.

onesounds:
> @onesounds is this ready for review?

1. This PR was reviewable by my standards.
2. @taooceros  said he would fix some things. and commented like this.
image
https://stackoverflow.com/questions/60872044/how-to-get-scaling-factor-for-each-monitor-e-g-1-1-25-1-5/60987055#60987055

3. I did not understand what the problem was that Taoo found and what they were trying to fix. I don't know what to do about this PR because there has been no comment from Taoo since then. 
4. This is a typical Taoo way of interrupting someone else's work, so you might as well just review and merge it and let him fix it later.

taooceros:
The problem I found exists in `dev` branch. I will try to fix that later in another pr. I will review this pr first.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -727,6 +794,13 @@
             return left;
         }
 
+        public double VerticalTop(Screen screen)
+        {
+            var dip1 = WindowsInteropHelper.TransformPixelsToDIP(this, 0, screen.WorkingArea.Y);
+            var top = dip1.Y +10;
+            return top;
+        }
```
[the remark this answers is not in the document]
taooceros: Maybe you want `Math.Min`?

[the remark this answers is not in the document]
taooceros: Sorry I am wrong.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -292,8 +297,24 @@ private void InitializePosition()
...
-                Left = _settings.WindowLeft;
+                var previousScreen = _settings.PreviousScreen;
+
+                var previousDpi = _settings.PreviousDpi;
+
+                _settings.PreviousScreen = (SystemParameters.VirtualScreenWidth, SystemParameters.VirtualScreenHeight);
+
+                var currentDpi = GetDpi();
+
+                if (previousScreen == default ||
```
taooceros, line 308: Should reset the position if no previous record is found.

taooceros (approved):

taooceros:
@onesounds I made some change to group Left and Top operation mainly without change of logic except some very obvious one. You can take a look and merge as you wish.

PR #2769, opened 2024-06-09
Localize startup messages about missing Python/NodeJS

### LINKED ISSUE #1292: Make Python download dialog translatable
Python dowload dialog is in English only. Is needed to change it to translateable.

obrázok_2022-07-24_065317867

Yusyuriv:
Fixes #1292, closes PR #1392.

## Issue
On startup, Flow was waiting until plugins are loaded before loading any localization files. If there are Python/Node plugins installed without the Python/Node runtime specified in settings, it would show a hard-coded message in English (and a few other hard-coded messages).

## Changes
What I did:
1. Moved internationalization initialization higher, above loading plugins
1. Extracted the hard-coded messages from above into `en.xaml`
1. Since now it doesn't wait for plugins to load first, and those directories are now unknown at the time of initialization, removed adding plugins' translation directories from `Internationalization` constructor
1. Made `PluginManager` call `AddPluginLanguageDirectories` and `ChangeLanguage` after the plugins have been loaded.

## Testing
- [x] The startup messages are now localized
- [x] Flow launches and seems to work correctly

## Summary by CodeRabbit

- **New Features**
  - Enhanced internationalization support with new message prompts and translations for better language consistency.

- **Bug Fixes**
  - Improved handling of failed plugin initialization with detailed internationalized error messages.

- **Refactor**
  - Refined message prompt constructions and formatting for better readability and maintainability.
  - Reordered settings and language assignments for smoother plugin loading.

- **Localization**
  - Added new translation strings for English and Russian language files related to plugin prompts and error messages.


taooceros:
> Moved internationalization initialization higher, above loading plugins

I think internationalization requires loading plugin (for loading plugin resources). Maybe that needs to be changed

Yusyuriv:
> > Moved internationalization initialization higher, above loading plugins
> 
> I think internationalization requires loading plugin (for loading plugin resources). Maybe that needs to be changed

Yes, which is why I only load Flow's language first (so plugin loading code could display localized errors) and then let PluginManager class add plugins' languages to resources after all plugins have been loaded.

Flow.Launcher.Core/Resource/Internationalization.cs
```diff
@@ -40,7 +39,7 @@ private void AddFlowLauncherLanguageDirectory()
         }
 
 
-        private void AddPluginLanguageDirectories()
+        internal void AddPluginLanguageDirectories()
```
taooceros, line 42: what do you think about making this method accepting a `IEnumerable` or something similar instead of referencing back to the PluginManager?

Yusyuriv: And pass it from PluginManager like this? Seems logical to me.
```csharp
InternationalizationManager.Instance.AddPluginLanguageDirectories(GetPluginsForInterface<IPluginI18n>());
```

Flow.Launcher.Core/ExternalPlugins/Environments/AbstractPluginEnvironment.cs
```diff
@@ -80,8 +82,7 @@ internal IEnumerable<PluginPair> Setup()
             }
             else
             {
-                MessageBox.Show(
-                    $"Unable to set {Language} executable path, please try from Flow's settings (scroll down to the bottom).");
+                MessageBox.Show(string.Format(InternationalizationManager.Instance.GetTranslation("runtimePluginUnableToSetExecutablePath"), Language));
```
VictoriousRaptor, line 85: Considering we've implemented MessageboxEx so maybe it's better to use it here.

Yusyuriv: It's implemented in Flow.Launcher project while this file is in Flow.Launcher.Core. I don't know of a way to do that (besides moving MessageBoxEx to a different project) without introducing a circular dependency, and moving MessageBoxEx to a different project would be beyond the scope of this PR.

Flow.Launcher/Languages/ru.xaml
```diff
@@ -3,6 +3,14 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:system="clr-namespace:System;assembly=mscorlib">
+    <!--  Startup  -->
+    <system:String x:Key="runtimePluginInstalledChooseRuntimePrompt">
+        Flow определил, что вы установили {0} плагины, которым требуется {1} для работы. Скачать {1}?
```
jjw24, line 8: Just fyi, if this language update hasn't been added to Crowdin, pretty sure they will get reverted back to English or overridden with updates coming from Crowdin.

jjw24 (approved):

Issue #881, opened 2021-12-09
Windows pop up message missing dark theme


kubalav:
If you press Edit button with no selected item that information window pop out without dark theme.

![obrázok_2021-12-09_110938](https://user-images.githubusercontent.com/37414585/145376598-9e05007e-a06f-4999-b5b8-9402f7c20d01.png)

onesounds:
The warning window is difficult to modify because it is the basic design of os. I have to create a new window for warning purposes and change all warnings. I'll see if there's a way, but don't expect too much. 😂

onesounds:
Surely, This may be the case in other windows, even if it is not a warning window. Let me know if you find it again.

kubalav:
YOC

kubalav:
> Surely, This may be the case in other windows, even if it is not a warning window. Let me know if you find it again.

Install, Uninstall plugin dialog too

jjw24:
this is the case for all warning windows because we use the os library to display them, you can find them if you search for 'messagebox' in the code

taooceros:
I am curious isn't wpf has implementation of messagebox? We may be able to remove the winform dependency.

onesounds:
I am currently working on creating a new messagebox window and code to make it compatible with the existing code.

Jack251970:
Please wait for 1.20.0 release (#3093).

PR #2774, opened 2024-06-11
Add MessageBoxEx

onesounds:
## What's the PR

- Improve the MessageBox design.
- Dark mode is applied.


jjw24:
Why is it called MessageBoxEx? what's 'Ex'?

onesounds:
https://www.codeproject.com/Articles/5290638/Customizable-WPF-MessageBox

https://github.com/nmoschkin/MessageBoxEx

As an extension of the messagebox. the examples I saw used that name. I thought this was an ordinary name, but  I get the same question for the second time. why? Isn't this better to maintain the existing messagebox grammar? 

Here's the name I can think of. 
- messageBox2
- messageboxV2
- SuperSaiyanMessagebox
- CustomMessageBox
- mBox2
- messagebox64
- messageboxWii
- newSuperMessageBox
- iMessageBox
- messageBoxZero
- messageBoxModel2
- messageBoxExpansionPack
- messageBoxSeasonPass
- messageBoxZwei
...

onesounds:
OK. Let me explain a little more.
1. The messagebox is not a customizable window. It is a window generated by WPF. So it's not affected by dark mode.
2. That's why people use the code under the name “MessageBoxEX”, keeping the syntax, but with their own design changes.
3. I'm not sure if it's possible to use this design and code while maintaining the “messagebox” syntax. In any case, the name needed to fit the existing syntax and be easily modifiable so that it would be easy to replace all the messagebox code in the entire code.

If you've ever used this window and checked for this issue (https://github.com/Flow-Launcher/Flow.Launcher/issues/881), you'll know why it's messageBoxEX. (taoo asked the same question. everyone too disinterested!!)

Anyway, I don't know if this code is usable or not. Someone should optimize it for me.

jjw24:
I apologise, did not click in my head at the time that it is short for extension, and yes the points explained above are correct.

jjw24 (approved):

PR #2798, opened 2024-06-23
Semantic Version

taooceros:
- Use SemanticVersioning to check about new version
- ~~Allow Prerelease~~


Jack251970 (approved):

taooceros:
Actually I think we don't really need to allow pre-release as that's in a separate repository. I will remove that.

Jack251970 (approved):

PR #2842, opened 2024-07-16
feat: Exclude certain File Types from the Index search result

### LINKED ISSUE #2144: Exclude certain File Types in the index searching results.
Hello,
It would be great to have somewhere in the settings, to have a field box you can add exclude/Ignore certain file types. 
in the search results from not being indexed or showing up in the results.

Examples are (.jpg .mov .mp3 .docx) and other types a user can add.

Thank you for our work, and this wonderful software.

z1nc0r3:
- User can add file types through Explorer Settings which should be ignored in search results.
- Partially fixes #2144 since there was another idea to filter certain folders.

image


VictoriousRaptor:
Is it possible to pass the exclude list when using everything？

z1nc0r3:
Yeah, that would be the best solution here. But it seems there are no such parameters (options) acceps from Everything API as I checked the SDK.

jjw24 (approved):

PR #2874, opened 2024-08-03
Pass MaxResult setting to EverythingApi

### LINKED ISSUE #2873: BUG: Explorer plugin with Everything index does not honor MaxResult setting
### Checks

### Problem Description

Results from **Explorer** plugin are limited to the hardcoded MaxResult (MaxCount) of 100 while using **Everything** engine. Changing the value in `Settings.json` has no effect.

Direct Enumeration and Windows Index are unaffected.

### To Reproduce

0. Disable all other plugins.
1. Edit `Flow.Launcher.Plugin.Explorer\Settings.json`.
2. Change `MaxResult` to 5.
3. Ensure that **Everything** engine is enabled and **Everything** itself is running.
4. Perform a query.

Expected result would be a list of 5 hits but the actual results will have up to 100 results.

5. Switch from **Everything** engine to **Windows Index** engine.
6. Perform same query.

Search results will be capped at 5 items as expected. 

### Screenshots

_No response_

### Flow Launcher Version

1.18.0

### Windows Build Number

Microsoft Windows [Version 10.0.19045.4651]

### Error Log

No error log associated with the issue. MaxResult setting is simply not passed to Everything API.

Ishmaeel:
`MaxResult` value in `Settings.json` was ignored for *Everything* searches. Pass the value to `EverythingApi` for all searches.

Fixes #2873.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>net7.0-windows</TargetFramework>
+    <TargetFramework>net7.0-windows10.0.19041.0</TargetFramework>
     <UseWPF>true</UseWPF>
     <UseWindowsForms>true</UseWindowsForms>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
@@ -48,6 +48,7 @@
     <PackageReference Include="System.Data.OleDb" Version="8.0.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
     <PackageReference Include="tlbimp-Microsoft.Search.Interop" Version="1.0.0" />
+    <PackageReference Include="ModernWpfUI" Version="0.9.4" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>net7.0-windows10.0.19041.0</TargetFramework>
+    <TargetFramework>net7.0-windows</TargetFramework>
     <UseWPF>true</UseWPF>
     <UseWindowsForms>true</UseWindowsForms>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
@@ -48,7 +48,6 @@
     <PackageReference Include="System.Data.OleDb" Version="8.0.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
     <PackageReference Include="tlbimp-Microsoft.Search.Interop" Version="1.0.0" />
-    <PackageReference Include="ModernWpfUI" Version="0.9.4" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
@@ -530,7 +530,7 @@ namespace Flow.Launcher.Plugin.Explorer.ViewModels
             get => Settings.MaxResult;
             set
             {
-                Settings.MaxResult = Math.Clamp(value, 0, 10000);
+                Settings.MaxResult = Math.Clamp(value, 100, 10000);
                 OnPropertyChanged();
             }
         }
--- a/Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
@@ -525,12 +525,15 @@ namespace Flow.Launcher.Plugin.Explorer.ViewModels
             }
         }
 
+        public int MaxResultLowerLimit => 100;
+        public int MaxResultUpperLimit => 100000;
+
         public int MaxResult
         {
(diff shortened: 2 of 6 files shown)


jjw24:
Thanks for the PR 👍 

Is it applicable to add a UI so users not have to edit the json file?

Ishmaeel:
> Thanks for the PR 👍
> 
> Is it applicable to add a UI so users not have to edit the json file?

I had assumed it was intentionally left as an advanced option, but sure, I'm happy to add one.

Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
Ishmaeel, line 1: Using `NumberBox` seemed to require these changes.

I admit that I do not understand the full implication of this. I'm happy to revert these and go with a plain `TextBox`.

Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
Ishmaeel, line 1: A minimum of zero effectively disables the plugin and values between 1..100 do not make much sense either. I left the minimum as zero for easier testing, minimum as the hardcoded default of 100 makes more sense.

10.000 might be a tad high, but it performed pretty well on my machine.  I tested as high as 100.000 and the launcher remained responsive, but slightly sluggish.

Ishmaeel:
I tried my hand at adding the UI but there might be some questionable choices in the PR, review and feedback appreciated.

Here's the "after":

image

z1nc0r3:
since this is _Everything_ related config, what about moving it under _Everything Settings_?

Ishmaeel:
> since this is _Everything_ related config, what about moving it under _Everything Settings_?

It's not specific to _Everything_. The setting itself is implemented for the entire plugin and already in use by _Windows Index_:

https://github.com/Flow-Launcher/Flow.Launcher/blob/dd0ae986145e22f87caf2ffba7943d485e5706cc/Plugins/Flow.Launcher.Plugin.Explorer/Search/WindowsIndex/QueryConstructor.cs#L22

Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
jjw24: Make the default and minimum both as 100 I think is the most appropriate.

Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
```diff
@@ -525,6 +525,16 @@ public string ExcludedFileTypes
             }
         }
 
+        public int MaxResult
+        {
+            get => Settings.MaxResult;
+            set
+            {
+                Settings.MaxResult = Math.Clamp(value, 0, 10000);
```
jjw24, line 533: You already placed min and max restrictions on the UI, is this still needed?

Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
Ishmaeel: Done.

Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
Ishmaeel: In my view, real validation/restriction belongs in lower levels like viewmodel or business layer. Redundant validation on UI should be for user convenience. Of course there's always risk of two going out of sync and producing a wonky experience, so, yes, it's good to have a single source of truth.

In any case, I removed the min/max from UI so the only restriction is here now. I had to add a slight binding delay though, because otherwise WPF binding would randomly fail to bind the "corrected" value back into the UI.

Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
jjw24: I think for simplicity maybe just use textbox and validate input is numeric and between min and max in the xaml.cs file.

You can also put the min max as variables in the view model which can then be accessed by the xaml.cs file as well.

Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs
jjw24: >In my view, real validation/restriction belongs in lower levels like viewmodel or business layer. Redundant validation on UI should be for user convenience. Of course there's always risk of two going out of sync and producing a wonky experience, so, yes, it's good to have a single source of truth.

Yes agree. You can put the min max as view model class variables so will keep single source of truth and remove the need to hardcode.

Ishmaeel: Converted hardcoded limits to variables but I'm not using them in codebehind. See my other comment.

Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
Ishmaeel: Reverted project target changes and changed to `TextBox`

I kept the limits in the viewmodel and only added maxlength in xaml. I added limited validation code to xaml.cs but it's only preventing the user from typing most non-digit characters. Unfortunately it won't stop pasting or entering spaces due to WPF's shortcomings. 

Turns out it is nothing but simple to implement a perfect numeric textbox without adding tons of complicated code to workaround the issues where MS dropped the ball. This version works nicely for the majority of use cases and it is the maximum amount of code I'm willing to submit for this simple feature.

jjw24:
This PR should be open right.

Ishmaeel:
> This PR should be open right.

Yes, thank you. I must have closed it accidentally when I merged it into my personal branch.

jjw24 (approved):

Plugins/Flow.Launcher.Plugin.Explorer/Flow.Launcher.Plugin.Explorer.csproj
jjw24: Tested and it works fine 👍

PR #2881, opened 2024-08-10
Fix startup localization issues

Yusyuriv:
Fixes issue from https://github.com/Flow-Launcher/Flow.Launcher/pull/2769

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -308,6 +308,7 @@ namespace Flow.Launcher.Core.Plugin
 
         public static IEnumerable<PluginPair> GetPluginsForInterface<T>() where T : IFeatures
         {
+            // Handle scenario where this is called before all plugins are instantiated, e.g. language change on startup
             return AllPlugins?.Where(p => p.Plugin is T) ?? Array.Empty<PluginPair>();
         }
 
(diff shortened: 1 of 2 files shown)


Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -308,7 +308,7 @@ public static PluginPair GetPluginForId(string id)
 
         public static IEnumerable<PluginPair> GetPluginsForInterface<T>() where T : IFeatures
         {
-            return AllPlugins.Where(p => p.Plugin is T);
+            return AllPlugins?.Where(p => p.Plugin is T) ?? Array.Empty<PluginPair>();
```
Yusyuriv, line 311: Fixes startup issue when built by Visual Studio

Flow.Launcher.Core/Resource/Internationalization.cs
```diff
@@ -58,6 +58,7 @@ internal void AddPluginLanguageDirectories(IEnumerable<PluginPair> plugins)
 
         private void LoadDefaultLanguage()
         {
+            RemoveOldLanguageFiles();
```
Yusyuriv, line 61: Removes language files loaded before any plugins were loaded. Prevents the language Flow started in from overwriting English if the user switches back to English

jjw24, line 61:
```suggestion
            // Removes language files loaded before any plugins were loaded.
            // Prevents the language Flow started in from overwriting English if the user switches back to English
            RemoveOldLanguageFiles();
```

Flow.Launcher.Core/Plugin/PluginManager.cs
jjw24, line 311:
```suggestion
            // Handle scenario where this is called before all plugins are instantiated, e.g. language change on startup
            return AllPlugins?.Where(p => p.Plugin is T) ?? Array.Empty<PluginPair>();
```

jjw24 (approved):

PR #2973, opened 2024-09-15
Improve plugin loading performance

Jack251970:
Improve plugin loading performance using concurrency

### DIFF
--- a/Flow.Launcher.Core/ExternalPlugins/Environments/AbstractPluginEnvironment.cs
+++ b/Flow.Launcher.Core/ExternalPlugins/Environments/AbstractPluginEnvironment.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Linq;
 using System.Windows.Forms;
 using Flow.Launcher.Core.Resource;
+using System.Threading.Tasks;
 
 namespace Flow.Launcher.Core.ExternalPlugins.Environments
 {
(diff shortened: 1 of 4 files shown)


Flow.Launcher.Core/ExternalPlugins/Environments/AbstractPluginEnvironment.cs
```diff
@@ -37,10 +38,10 @@ internal AbstractPluginEnvironment(List<PluginMetadata> pluginMetadataList, Plug
             PluginSettings = pluginSettings;
         }
 
-        internal IEnumerable<PluginPair> Setup()
```
taooceros, line 40: why would you not use a generic version of task?

taooceros (commented):
Could you give a small benchmark about the speed up?

Flow.Launcher.Core/ExternalPlugins/Environments/AbstractPluginEnvironment.cs
Jack251970: I have fixed this.

Jack251970:
> Looks cool! Could you give a small benchmark about the speed up?

It seems this optimization is invalid. The instance-creating functions are executed very fast, and thus concurrency are hard to  optimize their performances.

PR #2986, opened 2024-09-21
GitHub action

taooceros:
Add a github action workflow to build artifact. We could switch the entire pipeline later.

Why?
- It's faster than appveyor
- It has higher concurrency (not like appveyor when we can only build one job at a time)
- It is more popular
- appveyor has a limit of download for artifact which is pretty easy to be reached when we are testing new artifacts.


.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,110 @@
...
+    branches:
+      - dev
+      - master
+
+jobs:
+  build:
+
+    runs-on: windows-latest
+    env:
+      FlowVersion: 1.18.0
```
jjw24, line 22: This shouldn't be hard coded?

jjw24:
This is a test pr?

taooceros:
Yeah mainly for arm64

.github/workflows/dotnet.yml
taooceros: In appveyor we also hardcode this. We use that to inject into the version of release.

taooceros:
It is not clear why the trimming doesn't work....so weird

taooceros:
looks like switching to nuget restore (the same as appveyor) will make it work.....So weird, anyway

onesounds (approved):

onesounds:
@coderabbitai review it. Does this PR enable ARM builds now? I recently bought a Lunar Lake laptop—should I have gone with an ARM one instead?

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,91 @@
...
+      - dev
+      - master
+  pull_request:
+
+jobs:
+  build:
+
+    runs-on: windows-latest
+    env:
+      FlowVersion: 1.19.5
```
jjw24, line 19: This shouldn't be hardcoded.

Issue #2945, opened 2024-08-30
Select arguments when reopen


1208nn:
add a new option in setting

Now it can select all the query:

![image](https://github.com/user-attachments/assets/9b522e22-4b3a-40cb-b135-b9bce70436cc)

What the new feature I want to add work (I think I means this, I'm not mean to patch the old way but to and a new way):

![image](https://github.com/user-attachments/assets/b4befbc8-52da-40bb-889e-ec5605af2c15)

PR #3065, opened 2024-11-09
Add setting to select query text (non-action keyword) on window reopen

1208nn:
* **SettingsPaneGeneralViewModel.cs**
  - Add `SelectAllQueryOnReopen` property
  - Bind the new property to the settings in the constructor

* **MainWindow.xaml.cs**
  - Update `OnLoaded` method to select all query text if the new setting is enabled
  - Add `SelectAllQueryText` method to handle selecting all query text

* **MainWindow.xaml**
  - Bind `QueryTextBox` to the new setting to select all query text when the window is reopened

* **SettingWindow.xaml.cs**
  - Initialize the new checkbox based on the settings


taooceros:
@1208nn Sorry for the late response. Is there a reason that this is a draft pr?

1208nn:
> @1208nn Sorry for the late response. Is there a reason that this is a draft pr?

I'm sorry that I have no touch to those devices that can run VS, so this PR isn't finished as you can see and untested.

Jack251970:
@1208nn Hi, I think FL already provided this feature.

image

If you want to select all text when window is reopened, you can use `Select Last Query`. And you do not want to do that, you can use `Preserve Last Query`.

Closed.

1208nn:
Oh, sorry.

I think I didn't make it clear.

for example （see #2945）

bm github

Select github

@Jack251970

Jack251970:
Hmm, you mean just select part of query text

Jack251970:
Why do we need that?

onesounds:
Isn't this something that can be done with a custom query hotkey? It seems pointless to create it.

EnesEfeTokta (commented):
I'm not getting any error codes from the console, but the program isn't running.

Jack251970:
> I'm not getting any error codes from the console, but the program isn't running.

Have you exited existing Flow?

Jack251970 (commented):
I know what you are meaning here.

You want to select the non-action keyword text so that you can reopen window to continue search results for the same action keyword.

It makes sense.

Issue #3091, opened 2024-11-22
BUG: System.OperationCanceledException Issue in Flow.Launcher.Plugin.Program


Jack251970:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

System.OperationCanceledException
  HResult=0x8013153B
  Message=The query has been canceled via the token supplied to WithCancellation.
  Source=System.Linq.Parallel
  StackTrace:
   at System.Linq.Parallel.CancellationState.ThrowWithStandardMessageIfCanceled(CancellationToken externalCancellationToken)
   at System.Linq.Parallel.QueryTaskGroupState.QueryEnd(Boolean userInitiatedDispose)
   at System.Linq.Parallel.SpoolingTask.SpoolStopAndGo[TInputOutput,TIgnoreKey](QueryTaskGroupState groupState, PartitionedStream`2 partitions, SynchronousChannel`1[] channels, TaskScheduler taskScheduler)
   at System.Linq.Parallel.DefaultMergeHelper`2.System.Linq.Parallel.IMergeHelper<TInputOutput>.Execute()
   at System.Linq.Parallel.MergeExecutor`1.Execute[TKey](PartitionedStream`2 partitions, Boolean ignoreOutput, ParallelMergeOptions options, TaskScheduler taskScheduler, Boolean isOrdered, CancellationState cancellationState, Int32 queryId)
   at System.Linq.Parallel.PartitionedStreamMerger`1.Receive[TKey](PartitionedStream`2 partitionedStream)
   at System.Linq.Parallel.WhereQueryOperator`1.WrapPartitionedStream[TKey](PartitionedStream`2 inputStream, IPartitionedStreamRecipient`1 recipient, Boolean preferStriping, QuerySettings settings)
   at System.Linq.Parallel.UnaryQueryOperator`2.UnaryQueryOperatorResults.ChildResultsRecipient.Receive[TKey](PartitionedStream`2 inputStream)
   at System.Linq.Parallel.SelectQueryOperator`2.WrapPartitionedStream[TKey](PartitionedStream`2 inputStream, IPartitionedStreamRecipient`1 recipient, Boolean preferStriping, QuerySettings settings)
   at System.Linq.Parallel.UnaryQueryOperator`2.UnaryQueryOperatorResults.ChildResultsRecipient.Receive[TKey](PartitionedStream`2 inputStream)
[...]

PR #3092, opened 2024-11-22
Fix System.OperationCanceledException Issue

Jack251970:


Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -60,15 +60,26 @@ public async Task<List<Result>> QueryAsync(Query query, CancellationToken token)
...
+                            .WithCancellation(token)
+                            .Where(HideUninstallersFilter)
+                            .Where(p => p.Enabled)
+                            .Select(p => p.Result(query.Search, Context.API))
+                            .Where(r => r?.Score > 0)
+                            .ToList();
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        // Fix #3091: System.OperationCanceledException Issue
```
jjw24, line 78: @Jack251970 please remove this comment and link the issue to PR instead

jjw24: Please also add debug logging.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -60,15 +60,26 @@ public async Task<List<Result>> QueryAsync(Query query, CancellationToken token)
...
+                        return emptyResults;
+                    }
+                   
+                });
```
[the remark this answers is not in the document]
jjw24: @Jack251970 would you also consider to add these?

Plugins/Flow.Launcher.Plugin.Program/Main.cs
Jack251970: > What happens to flow UI/UX wise when this exception is caught? If the result is empty, we need to consider error logging so it's not swallowed.

The result should be non-empty because other QueryAsync functions should be called. Considering some unknown situations, I added the warning log output.

taooceros: It should only affect debug code. This error should be caught outside. Though it can be a little bit annoying when debugging though.

Jack251970: Get it.

jjw24: > It should only affect debug code. This error should be caught outside. Though it can be a little bit annoying when debugging though.

It's caught outside but iirc it is not descriptive enough from the logs to figure out what is causing the error (or maybe it is and I just confuse it with another issue, because I see operation cancelled issue raised more times than usual).

@Jack251970 @taooceros thoughts about logging as error for now so we can see from users how often this happens?

jjw24: >It should only affect debug code
Actually @taooceros what do you mean it only affects debug code? Does this not happen in production?

Jack251970: Please refere to (#3091) image.
This issue is captured under VS in debug mode.

It seems there are try catch statements outside the codes in production, so this exception can be handled outside in production, while in debug not (pop up a exception like that screenshot).
Thats why @taooceros says 'Though it can be a little bit annoying when debugging though.'.

Also, I think handling exceptions in the inner try-catch block can improve performance in certain scenarios.

Jack251970: > > It should only affect debug code
> 
> Actually @taooceros what do you mean it only affects debug code? Does this not happen in production?

I think @taooceros perhaps mean this.

taooceros: > > It should only affect debug code. This error should be caught outside. Though it can be a little bit annoying when debugging though.
> 
> It's caught outside but iirc it is not descriptive enough from the logs to figure out what is causing the error (or maybe it is and I just confuse it with another issue, because I see operation cancelled issue raised more times than usual).
> 
> @Jack251970 @taooceros thoughts about logging as error for now so we can see from users how often this happens?

It should only be noticeable when you are debugging flow with a debugger.

taooceros: Sure I agree we can do that. This can be quite annoying since the program plugin often encounter cancellation

jjw24: Hmm ok in that case let's keep as debug logging then.

jjw24:
@coderabbitai summary

jjw24 (approved):

PR #3093, opened 2024-11-24
Replace original MessageBox with MessageBoxEx & Improve MessageBoxEx display style & Add new public api for using MessageBoxEx

Jack251970:
In #2774, a new message box with dark theme support is introduced.
This PR replaces all System.Windows.MessageBox in Flow.Launcher & Flow.Launcher.Core namespace with the new message box.

Also improve original MessageBoxEx of OK type:
Original:
Screenshot 2024-11-24 162111
Now:
Screenshot 2024-11-24 224420

Move Yes, OK button to the left of No, Cancel button and focus on Yes, OK by default as System.Windows.MessageBox:
Original:
Screenshot 2024-11-24 162121
Now:
Screenshot 2024-11-25 120427

Test already done:
* All types of MessageBoxEx create & display & interaction.
* Proxy message box display & interaction in setting proxy panel page: `MessageBoxEx.Show(InternationalizationManager.Instance.GetTranslation(message));`. (direct call method of MessageBoxEx)
* Shutdown message box display & interaction in Flow.Launcher.Plugin.Sys: `var result = context.API.ShowMsgBox(
    context.API.GetTranslation("flowlauncher_plugin_sys_dlgtext_shutdown_computer"),
    context.API.GetTranslation("flowlauncher_plugin_sys_shutdown_computer"),`. (api call method of MessageBoxEx)


jjw24 (commented):
Hey @Jack251970, could you fix the build fail https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/51049381. Let me know if you need help.

jjw24:
See line 398 of https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/51049381#L398. It's complaining about 'no argument given that corresponds to the required parameter 'messageBoxExShow' of 'FilesFolders.RemoveFolderIfExists'. 

I can't tell why without going infront of a computer, maybe also check if you still have changes not pushed up yet?

Plugins/Flow.Launcher.Plugin.Sys/Flow.Launcher.Plugin.Sys.csproj
```diff
@@ -37,6 +37,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\Flow.Launcher.Core\Flow.Launcher.Core.csproj" />
```
jjw24, line 40: I see you brought in the core project for the default plugins, this creates a dependency on core. We need to aim to have these default plugins as self-contained plugins, where all they need will just be the plugin project, so that they are consistent with the external . net plugins. We are not there yet obviously, but I wonder is there a way to move the mssagebox code to the plugin project instead. Doing so we will also enable other .net plugins to utilise it.

Jack251970:
> See line 398 of https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/51049381#L398. It's complaining about 'no argument given that corresponds to the required parameter 'messageBoxExShow' of 'FilesFolders.RemoveFolderIfExists'.
> 
> I can't tell why without going in front of a computer, maybe also check if you still have changes not pushed up yet?

Sorry, I will check it. Perhaps I miss something.

Plugins/Flow.Launcher.Plugin.Sys/Flow.Launcher.Plugin.Sys.csproj
jjw24: >Yeah, I have considered to export this MessageBoxEx into IPublicAPIor put it into Flow.Launcher.Plugin for all 3rd plugins to use. And in that way, Plugin.Sys do not need to refer to Core project.

Yeah exactly, and more consistency across plugins.

>By the way, Do you think it's better for me to put these two files (MessageBoxEx.xaml & MessageBoxEx.xaml.cs) in Flow.Launcher.Plugin or supply the call as a function in IPublicAPI (like MessageBoxResult ShowMessageBox(string text, string title))?

I think supply the call as a function in IPublicAPI is ideal as they are also used by flow itself, if doing this makes the code complicated/tangled/messy then let's have another chat.

Jack251970:
@jjw24 Now expose MessageBoxEx to public api, and all plugins call it to use MessageBoxEx.

jjw24:
@Jack251970 could you please list in the PR description, what tests you have done so far for your changes. This will let me know what other tests I can do for this change.

Jack251970:
@jjw24 Sure, please see `Test already done`.

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -298,5 +299,81 @@ public interface IPublicAPI
...
+        /// <param name="icon">A System.Windows.MessageBoxImage value that specifies the icon to display.</param>
+        /// <param name="defaultResult">
+        /// A System.Windows.MessageBoxResult value that specifies the default result of
+        /// the message box.
+        /// </param>
+        /// <returns>
+        /// A System.Windows.MessageBoxResult value that specifies which message box button
+        /// is clicked by the user.
+        /// </returns>
+        public MessageBoxResult ShowMsgBox(string messageBoxText, string caption, MessageBoxButton button, MessageBoxImage icon, MessageBoxResult defaultResult);
```
jjw24, line 377: Maybe instead of four methods, can just do one with defaults?

Jack251970: I just follow the same style as the System.Windows.MessageBox. Do you think it is necessary?

Flow.Launcher.Plugin/SharedCommands/FilesFolders.cs
```diff
@@ -21,7 +21,8 @@ public static class FilesFolders
         /// </summary>
         /// <param name="sourcePath"></param>
         /// <param name="targetPath"></param>
-        public static void CopyAll(this string sourcePath, string targetPath)
+        /// <param name="messageBoxExShow"></param>
+        public static void CopyAll(this string sourcePath, string targetPath, Func<string, MessageBoxResult> messageBoxExShow)
```
jjw24, line 25: Can you provide defaults for the methods with the new MessageBox parameter in FilesFolders please.

Jack251970: Sure!

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
Jack251970: @jjw24 Any ideas?

jjw24: Yeah following that style is good. I just meant we can just have one method and have caption (empty string), button (null) and etc set with defauls. Let PublicAPIInstance.cs handle how to call based on what parameters are passed in.

Jack251970: Get it.

taooceros:
@onesounds is there a reason that previously the no is on left yes on right?

onesounds:
> @onesounds is there a reason that previously the no is on left yes on right?

I can't remember why either. As far as I know, this order of PR is correct. I wish I could change the order freely on the code, ("no yes" or "yes no" but I remember it was probably hard to make. (I had to make several buttons.)

jjw24:
@Jack251970 I have made the MessageBoxEx's Show method to use defaults so we can remove the other unnecessary overrides, please see my commit. Everything else looks really good. Let me know if this is OK.

Jack251970:
@jjw24 Thank you! I have forgotton to simplify functions in the MessageBoxEx.xaml.cs. I use dark mode for my system and could not wait to experience this improvement.

jjw24 (approved):
👍 very nice thank you

PR #3099, opened 2024-11-27
Fix possible null reference when query is cancelled & Add new image icon for images whose formats are not support & Other improvements

Jack251970:
This PR contains those fixes and improvements

1. Fix possible null reference when query is cancelled.
It will happen when one query is cancelled and can cause query bar not working like this:
https://github.com/user-attachments/assets/8a0a2c0f-6db6-4061-8ff6-a979408fc6ff
Throw exception from VS:
Screenshot 2024-11-25 182150
Screenshot 2024-11-25 182242

2. Add support when image format is not support.
It add support for images whose formats are not support for preview like this:
Screenshot 2024-11-27 193044

Before it is catch in outer try catch sentences, and the icon of images will be `AppMissingIcon.png`, which is quite strange for `.png` extension.
Screenshot 2024-11-27 193035

Now show the preview like this:
image

3. Introduce concurrent directionary for topmost record.
Make sure TopMostRecord is thread safe.

4. Fix modified collection enumeration issue when updating results, which can also cause the same problem like 1.
It could happen when calling `ResultsUpdated.Invoke` event in `IResultUpdate` interface so fast that main view model cannot complete update event (still execute `foreach` sentence).

6. Fix xaml convert issue and use icon loader for PluginActivationIcon.
Screenshot 2024-11-27 220729


taooceros:
@Jack251970 I don't want to swipe your interest! It is always a tough time (and discouraging) waiting people to review your prs. Just to let you know I will be a bit busy before 12/3. Afterward I will try my best to review your pr asap. If I forget afterward, ping me at discord😁. (Or maybe other people in our teams will review your pr earlier).

Jack251970:
@taooceros No worries at all! I completely understand that reviewing PRs can take time, especially with a busy schedule.

taooceros:
@onesounds how do you think about the not supported image?

Flow.Launcher/Storage/TopMostRecord.cs
```diff
@@ -1,29 +1,29 @@
...
     public class TopMostRecord
     {
         [JsonInclude]
-        public Dictionary<string, Record> records { get; private set; } = new Dictionary<string, Record>();
+        public ConcurrentDictionary<string, Record> records { get; private set; } = new ConcurrentDictionary<string, Record>();
 
         internal bool IsTopMost(Result result)
         {
-            if (records.Count == 0 || !records.ContainsKey(result.OriginQuery.RawQuery))
+            if (records.IsEmpty || (result.OriginQuery != null && !records.ContainsKey(result.OriginQuery.RawQuery)))
```
taooceros, line 15: consider separate the contains key to later logic so you don't have to repeat checking the `result.OriginQuery`

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1440,26 +1447,35 @@ public void UpdateResultView(ICollection<ResultsForUpdate> resultsForUpdates)
             }
 #endif
 
-            foreach (var metaResults in resultsForUpdates)
+            try
```
taooceros, line 1450: could you explain for the try?

Flow.Launcher/Storage/TopMostRecord.cs
Jack251970: Get it!

onesounds:
> @onesounds how do you think about the not supported image?

Blank area is not good. It's weird that there's no image at all. If there's an imgmissing icon like an appmissing icon, ok.
image
Like an example screenshot, such an "empty image" should come out anyway.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: It could happen when calling `ResultsUpdated.Invoke` event in` IResultUpdate` interface so fast that main view model cannot complete update event (still execute foreach sentence).

Take `WebSearch` plugin for example,

```
public async Task<List<Result>> QueryAsync(Query query, CancellationToken token)
{
    var results = new List<Result>();

    foreach (...)
    {
        ...

        results.Add(result);  // edit this List

        ...

        // Here you use this List to give to MainViewModel for updating
        // If you do the time-consuming work after this so fast, you will edit this list when updating the MainViewModel
        // So we need to use Try to avoid this problem
        ResultsUpdated?.Invoke(this, new ResultUpdatedEventArgs
        {
            Results = results,
            Query = query
        });

        ... // do time-consuming work
    }

    return results;
}
```

Jack251970:
@onesounds I will try to provide an image to show this message.

Flow.Launcher/ViewModel/MainViewModel.cs
taooceros: Make sense, but that also means we need additional synchronization? I remember the ConcurrentDictionary is already providing a enumerator that is thread safe?

taooceros: How about a `ConcurrentBag`?

Jack251970: The key of the problem is that this plugin has not tell the developer to do thread-safe operation in `QueryAsync` (like return a copy of the list as the `Results` in `ResultUpdatedEventArgs`). So the developer will possibly return the same `List` to the `ResultUpdatedEventArgs` and the result of the `QueryAsync` function.

If you want to fully address this problem, you need to add additional synchronization.

For example, you can change the definition of the `ResultUpdatedEventArgs` to
```
public class ResultUpdatedEventArgs : EventArgs
{
    public ConcurrentBag<Result> Results;  // thread-safe
    public Query Query;
    public CancellationToken Token { get; init; }
}
```.

Or, you need to ask the developer to return a copy of the `List` so that the plugin will not change the `List` for updating  the view.

For example,
```
public async Task<List<Result>> QueryAsync(Query query, CancellationToken token)
{
    var results = new List<Result>();

    foreach (...)
    {
        ...

        results.Add(result);

        ...

        var resultsCopy = results.Clone();  // create a copy of the results
        ResultsUpdated?.Invoke(this, new ResultUpdatedEventArgs
        {
            Results = resultsCopy,
            Query = query
        });

        ...
    }

    return results;
}
```

Anyway, I think this is the most developer-friendly solution for this problem and developers don not need to redevelop their plugins to fix this.

Jack251970: @taooceros What do you mean by `ConcurrentDictionary` and `ConcurrentBag`?

Jack251970:
@taooceros Have got a new icon for the images whose formats are not supported.

Jack251970:
@onesounds What do you think of this icon? 

image

onesounds:
It's not bad, but I don't think it's the best. Isn't there this icon?

onesounds:
! [Screenshot_20241203_141935_Samsung Internet.jpg Uploading... ]()

How about this? Images exist and can be opened, but I think this is better because only previews are not possible. The red line feels like there's something wrong with the file.

Jack251970:
Yeah, the right icon is better than my original icon with red line. And for this icon called Broken Image: 
image
What do you think of it?

I search on web and I have learnt that unsupported file format is one kind of the causes of the broken images. Perhaps this can let user know this image is broken and cannot be previewed for some reasons?

Jack251970:
@onesounds And thanks for your help!

onesounds:
> @onesounds  Yeah, the right icon is better than my original icon with red line. And for this icon called Broken Image: 
> image
> What do you think of it?
> 
> I search on web and I have learnt that unsupported file format is one kind of the causes of the broken images. Perhaps this can let user know this image is broken and cannot be previewed for some reasons?

 If the file is not normal, you can use the icon you chose. But if it's a file that the preview doesn't support, it should just be an image icon. (Windows also process previews as icons if possible, and other images as image icons.) I understand that there is no indication of broken files. I won't distinguish it because it's based on extensions. You don't show a file in the preview called a "broken file." You shouldn't choose an icon that can be misunderstood like this.

Jack251970:
Now updated the image and will show the preview like this:
image

Flow.Launcher/ViewModel/MainViewModel.cs
taooceros: Hold on, I remember this is a producer/consumer model with appropriate synchronization through the `Channel` class. Let me recheck.

taooceros: I am a little bit suspicious of this change. Is it possible for you to revert just this part so we can merge first. Then Include it in a new pr so we can discuss specifically about the potential synchronization problem.

Jack251970: @taooceros Sure! Please check!

taooceros (approved):
Sorry for the late response. Thanks!

taooceros:
@Jack251970 Could you move the reverted change to a new pr?

Jack251970:
@taooceros Moved. Please check #3115.

PR #3110, opened 2024-12-05
Add `.`, `./lib`, `./plugin` directories to path for Python plugins

Yusyuriv:
# The problem
Currently, due to Flow Launcher using embedded Python by default, plugin directory doesn't get added to path. This causes inconvenience for plugin developers. The plugin development documentation for Python suggests using `./lib` for installing dependencies and `./plugin` for actual plugin code, but developers can't import anything from there because these directories are not in Python's path. They have to add them to path manually in every plugin before they can start importing their code. For example, FlowYouTube plugin:
```py
import sys
import os

plugindir = os.path.abspath(os.path.dirname(__file__))
sys.path.append(plugindir)
sys.path.append(os.path.join(plugindir, "lib"))
sys.path.append(os.path.join(plugindir, "plugin"))

# Only now you can import and use your actual code
from plugin.main import FlowYouTube
```

# The solution
We can't simply add these paths to `PYTHONPATH` env variable before starting the plugin because we're using embedded Python. It ignores this variable. Flow Launcher users also might be using their own Python installations, so we can't use any solution that would apply only to our embedded Python. So the easiest solution I came up with was instead of launching the files directly (like `python path/to/file.py`), run some code that Flow Launcher generates, adding these paths to Python's path, and only then run the actual plugin file via `runpy`:
```cs
StartInfo.ArgumentList.Add("-c");
StartInfo.ArgumentList.Add(
    $"""
     import sys
     sys.path.append(r'{rootDirectory}')
     sys.path.append(r'{libDirectory}')
     sys.path.append(r'{pluginDirectory}')
     
     import runpy
     runpy.run_path(r'{filePath}', None, '__main__')
     """
);
```

# Backwards compatibility
I tested this change with several plugins, with them still manually adding directories to path and with them simply importing things. From my tests it seemed like they all continued working normally. Worst case scenario I can see is that either:
* a plugin will have non-existent directories added to path, which shouldn't have any effect
* or a plugin will have the same directories added to path twice, which also shouldn't have any effect

Now, to the plugins I tried running. I tested the listed plugins before removing the code marked red in the code blocks and after. They all continued working like before.

### GitHub Quick Launcher
No changes here, it's released as a `.pyz` file.

### FlowYouTube
```diff
- import sys
- import os

- plugindir = os.path.abspath(os.path.dirname(__file__))
- sys.path.append(plugindir)
- sys.path.append(os.path.join(plugindir, "lib"))
- sys.path.append(os.path.join(plugindir, "plugin"))

from plugin.main import FlowYouTube

if __name__ == "__main__":
    FlowYouTube()
```

### Search-MDI
```diff
# -*- coding: utf-8 -*-
- import sys
- import os

- plugindir = os.path.abspath(os.path.dirname(__file__))
- sys.path.append(plugindir)
- sys.path.append(os.path.join(plugindir, "lib"))
- sys.path.append(os.path.join(plugindir, "plugin"))

from plugin.main import MDI

if __name__ == "__main__":
    MDI()
```

### ChatGPT
```diff
# -*- coding: utf-8 -*-

- import sys
- import os

- parent_folder_path = os.path.abspath(os.path.dirname(__file__))
- sys.path.append(parent_folder_path)
- sys.path.append(os.path.join(parent_folder_path, "lib"))
- sys.path.append(os.path.join(parent_folder_path, "plugin"))

from plugin.main import ChatGPT


if __name__ == "__main__":
    ChatGPT()
```

### My unreleased plugin I use to test my plugin API v2 library
```diff
- import sys
- import os

- rootDirectory = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
- libDirectory = os.path.join(rootDirectory, 'lib')
- pluginDirectory = os.path.join(rootDirectory, 'plugin')

- sys.path.append(rootDirectory)
- sys.path.append(libDirectory)
- sys.path.append(pluginDirectory)

from plugin import Plugin

class MyPlugin(Plugin):
    pass
```

# Notes
While doing this I encountered a problem. v1 plugins expect requests to be sent as the third argument. I couldn't make `-c <code>` work after the request because I think it was thinking that the request JSON was file name, so I had to move `-c <code>` to be the first two arguments. However, Python stops processing arguments after it encounters `-c <code>`, so I couldn't specify `-B` after it, and I couldn't specify `-B` before because `-c <code>` takes up two arguments and the third argument must be the request. So I replaced `-B` flag usage with the PYTHONDONTWRITEBYTECODE environment variable that does the same thing.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PythonPlugin.cs
+++ b/Flow.Launcher.Core/Plugin/PythonPlugin.cs
@@ -1,4 +1,5 @@
-﻿using System.Diagnostics;
+﻿using System;
+using System.Diagnostics;
 using System.IO;
 using System.Text.Json;
 using System.Threading;
@@ -29,10 +30,6 @@ namespace Flow.Launcher.Core.Plugin
             _startInfo.EnvironmentVariables["FLOW_VERSION"] = Constant.Version;
             _startInfo.EnvironmentVariables["FLOW_PROGRAM_DIRECTORY"] = Constant.ProgramDirectory;
             _startInfo.EnvironmentVariables["FLOW_APPLICATION_DIRECTORY"] = Constant.ApplicationDirectory;
-
-
-            //Add -B flag to tell python don't write .py[co] files. Because .pyc contains location infos which will prevent python portable
-            _startInfo.ArgumentList.Add("-B");
         }
 
         protected override Task<Stream> RequestAsync(JsonRPCRequestModel request, CancellationToken token = default)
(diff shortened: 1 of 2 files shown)


Garulf:
Thanks for making this, I'll try and take a look this weekend when I have more time.

jjw24:
LGTM. We will want to update how those path are handled in the [ci here](https://github.com/Flow-Launcher/Flow.Launcher.PluginsManifest/blob/6a705af6b642597d9df9537530df9ff8d7ac0ce7/ci/src/test-python.py#L16) before we remove them from documentation going forward.

jjw24:
@Garulf do you still want to do more testing?

cibere:
Would it make sense to also add the win32 package paths to path? Incase they are using the pywin32 package? If they are using the package, this would be needed (so the question is should the plugin dev do it or flow), since pywin32 uses a pth file that doesn't get used with the lib folder

jjw24:
I believe so, thoughts @Yusyuriv ?

Yusyuriv:
I don't think we should add package-specific things to the path because most of the plugin devs won't be using those packages. I added `lib` and `plugin` directories because that's a de facto standard and those two directories are mentioned in the docs. But I'm not opposed to it if others decide that it's needed. @Garulf?

Garulf:
Sorry for the late reply. I think pywin32 can be very useful for a Python plugin developer. However its unique packaging can cause a lot of friction. 

I think it would be very helpful to add a path for this package as well.

jjw24:
Hi @Yusyuriv have you got time to add this in or did you want to separate this change out as a todo later issue?

Yusyuriv:
Sure, I'll add it sometime this week.

Yusyuriv:
Done. As far as I know, these are the only two directories that needed to be added:
* `lib/win32/lib`
* `lib/win32`

jjw24 (approved):

PR #3114, opened 2024-12-07
add CopyText and AutoCompleteText to some results in the sys plugin

cibere:
Adds the `CopyText` and `AutoCompleteText` keys to the `Open recycle bin`, `Open Flow Launcher's log location`, and `Open the location where Flow Launcher's settings are stored` results.

## ~~Disclaimer~~
~~This code is untested (idk how to test it)~~
*Elven helped me figure out how to test it through AppVeyor*


cibere:
Ok, I managed to test it through appveyor, and it seems like everything works except copying the log location and userdata folder. Though I'm not sure why it isn't working

Yusyuriv:
I think we should have a way to opt out of this behavior. Right now [it always tries to copy files](https://github.com/Flow-Launcher/Flow.Launcher/blob/5f46ee746feff61197acb6ff3e46966389ff85da/Flow.Launcher/MainWindow.xaml.cs#L147C1-L148C69). Maybe add another property to Result? Something like `bool CopyTextAsPlainText`?

jjw24:
> I think we should add a way to opt out of this behavior. Right now [it always tries to copy files](https://github.com/Flow-Launcher/Flow.Launcher/blob/5f46ee746feff61197acb6ff3e46966389ff85da/Flow.Launcher/MainWindow.xaml.cs#L147C1-L148C69). Maybe add another property to Result? Something like `bool CopyTextAsPlainText`?

What do you mean? It copies query text if it's highlighted, or else the text passed from plugin. Which behaviour are you referring to?

cibere:
> > I think we should add a way to opt out of this behavior. Right now [it always tries to copy files](https://github.com/Flow-Launcher/Flow.Launcher/blob/5f46ee746feff61197acb6ff3e46966389ff85da/Flow.Launcher/MainWindow.xaml.cs#L147C1-L148C69). Maybe add another property to Result? Something like `bool CopyTextAsPlainText`?
> 
> 
> 
> What do you mean? It copies query text if it's highlighted, or else the text passed from plugin. Which behaviour are you referring to?

The part of the code where if the text from the plugin is a valid path, flow will copy the file/dir it points to instead of the path

jjw24:
`Ctrl+Shift+C` hotkey does that already doesn't it? It copies text only from path.

cibere:
> `Ctrl+Shift+C` hotkey does that already doesn't it? It copies text only from path.

Oh, that does seem to work

jjw24:
Have you tested the copy and autocomplete works for Recycle bin command?

Yusyuriv:
> `Ctrl+Shift+C` hotkey does that already doesn't it? It copies text only from path.

Oh, I completely forgot about that one. Yes, it does work.

> Have you tested the copy and autocomplete works for Recycle bin command?

It does autocomplete and copy the text `shell:RecycleBinFolder`, but since it's not a file path, the only thing Flow suggests after autocompleting it is searching about it online.

cibere:
`shell:RecycleBinFolder` seems to open a virtual folder that includes the recycle bins of all of the user's current drives, and that virtual folder doesn't seem to have a filepath besides the shell shortcut and a guid.

I see 3 main options:
1. Keep the changes I made to the open recycle bin result
2. Revert the changes I made to the open recycle bin result
3. Keep the `CopyText` key, but remove the `AutoCompleteText` key

I personally like 3 the most since `CopyText` works just fine, and as Yusyuriv mentioned, the auto complete just prompts you to search `shell:recyclebinfolder` online.

Yusyuriv:
I also think option 3 is the best, but I'd like to hear others' opinion.

jjw24:
Option 3 sounds good too.

jjw24 (approved):

PR #3115, opened 2024-12-08
Fix possible result update issue for plugins with IResultUpdate interface

Jack251970:
# Issue
This issue could happen when calling `ResultsUpdated.Invoke` event in `IResultUpdate` interface so fast that main view model cannot complete update event (still execute foreach sentence).

# Example
Take `WebSearch` plugin for example, I change its `QueryAsync` function to this:
```
public async Task<List<Result>> QueryAsync(Query query, CancellationToken token)
{
    var results = new List<Result>();

    for (var i = 0; i < 1000; i++)
    {
        var result = new Result
        {
            Title = $"{i}",
            SubTitle = $"{i}",
            Score = i
        };

        results.Add(result);

        // here we update the results very fast and call results update event
        ResultsUpdated?.Invoke(this, new ResultUpdatedEventArgs
        {
            Results = results,
            Query = query
        });
    }

    return results;
}
```

And I can catch the exception in `UpdateResultView` function of `MainViewModel`:
Screenshot 2024-12-08 091833

## Solution

The key of the problem is that this plugin has not tell the developer to do thread-safe operation in `QueryAsync` (like return a copy of the list as the `Results` in `ResultUpdatedEventArgs`). So the developer will possibly return the same `List` to the `ResultUpdatedEventArgs` and the result of the `QueryAsync` function. I think there are two solutions for this.

### Solution 1

Update the `IResultUpdate` documents and ask developers to return the copy of the list as the parameter of `ResultUpdatedEventArgs`.

Take the above example, I can change its `QueryAsync` function to this:

```
var resultsCopy = results.ToList();

ResultsUpdated?.Invoke(this, new ResultUpdatedEventArgs
{
    Results = resultsCopy,
    Query = query
});
```

### Solution 2

Wrap the update codes in `UpdateResultView` with try catch, it is what my PR does.

### Solution 3

For example, you can change the definition of the `ResultUpdatedEventArgs` so that developer won't return the same `List` like this:

```
public class ResultUpdatedEventArgs : EventArgs
{
    public ConcurrentBag<Result> Results;  // thread-safe
    public Query Query;
    public CancellationToken Token { get; init; }
}
```

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -1447,26 +1447,35 @@ namespace Flow.Launcher.ViewModel
             }
 #endif
 
-            foreach (var metaResults in resultsForUpdates)
+            try
             {
-                foreach (var result in metaResults.Results)
+                foreach (var metaResults in resultsForUpdates)
                 {
-                    if (_topMostRecord.IsTopMost(result))
+                    foreach (var result in metaResults.Results)
                     {
-                        result.Score = int.MaxValue;
-                    }
-                    else
-                    {
-                        var priorityScore = metaResults.Metadata.Priority * 150;
-                        result.Score += _userSelectedRecord.GetSelectedCount(result) + priorityScore;
+                        if (_topMostRecord.IsTopMost(result))
+                        {
+                            result.Score = int.MaxValue;
+                        }
+                        else
+                        {
+                            var priorityScore = metaResults.Metadata.Priority * 150;
+                            result.Score += _userSelectedRecord.GetSelectedCount(result) + priorityScore;
+                        }
                     }
                 }
+
+                // it should be the same for all results
+                bool reSelect = resultsForUpdates.First().ReSelectFirstResult;
+
+                Results.AddResults(resultsForUpdates, token, reSelect);
             }
-
-            // it should be the same for all results
-            bool reSelect = resultsForUpdates.First().ReSelectFirstResult;
-
-            Results.AddResults(resultsForUpdates, token, reSelect);
+            catch (InvalidOperationException e)
+            {
+                // Plugin with IResultUpdate interface can somtimes throw this exception
+                // Collection was modified; enumeration operation may not execute
+                Log.Exception($"{nameof(MainViewModel)}.{nameof(UpdateResultView)}|UpdateResultView failed", e);
+            }
         }
 
         #endregion
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -233,8 +233,11 @@ namespace Flow.Launcher.ViewModel
 
                     var token = e.Token == default ? _updateToken : e.Token;
 
-                    PluginManager.UpdatePluginMetadata(e.Results, pair.Metadata, e.Query);
-                    if (!_resultsUpdateChannelWriter.TryWrite(new ResultsForUpdate(e.Results, pair.Metadata, e.Query,
+                    // make a copy of results to avoid plugin change the result when updating view model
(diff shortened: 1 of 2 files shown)


Jack251970:
I do not know why I have caught exception in `UpdateResultView` but still could throw exception in
```
            void continueAction(Task t)
            {
#if DEBUG
                throw t.Exception;
#else
                Log.Error($"Error happen in task dealing with viewupdate for results. {t.Exception}");
                _resultsViewUpdateTask =
                    Task.Run(updateAction).ContinueWith(continueAction, TaskContinuationOptions.OnlyOnFaulted);
#endif
            }
```

Seems that Solution 2 fails to work.

taooceros:
However, what I am confused is that IResultsUpdate should only append a job to the channel

https://github.com/Flow-Launcher/Flow.Launcher/blob/304f98e9edb6a7f058cf358bdcedb6d05e455531/Flow.Launcher/ViewModel/MainViewModel.cs#L229-L241

and then the consumer should be a single thread?

https://github.com/Flow-Launcher/Flow.Launcher/blob/304f98e9edb6a7f058cf358bdcedb6d05e455531/Flow.Launcher/ViewModel/MainViewModel.cs#L188-L205

taooceros:
on the other hand looks like you didn't commit the correct change?

Jack251970:
> on the other hand looks like you didn't commit the correct change?

Sorry, I don not know your meaning.

Jack251970:
> I do not know why I have caught exception in `UpdateResultView` but still could throw exception in
> 
> ```
>             void continueAction(Task t)
>             {
> #if DEBUG
>                 throw t.Exception;
> #else
>                 Log.Error($"Error happen in task dealing with viewupdate for results. {t.Exception}");
>                 _resultsViewUpdateTask =
>                     Task.Run(updateAction).ContinueWith(continueAction, TaskContinuationOptions.OnlyOnFaulted);
> #endif
>             }
> ```
> 
> Seems that Solution 2 fails to work. So I reverted fix in `MainViewModel` and fixed possible issue in WebSearch plugin.

Here I describe my finding about the issue of `Solution 1`.

Jack251970:
Sorry that I am not quite familiar with such synchronization problem and I just happen to find it.

taooceros:
Oh now I understand what's going wrong. The ResultsUpdateEvent didn't return a new list, which causes issue when the plugin is modifying the same underlying list.

taooceros:
A better solution is to create a copy here

https://github.com/Flow-Launcher/Flow.Launcher/blob/304f98e9edb6a7f058cf358bdcedb6d05e455531/Flow.Launcher/ViewModel/MainViewModel.cs#L237C20-L241C22

Jack251970:
Now the only change of this PR is all about `Solution 1`, which I find it can work perfectly. I use a copy of `list` in plugin to give to `Flow.Launcher` so that `Flow.Launcher` can handle this `list` whatever it likes.

Jack251970:
> A better solution is to create a copy here
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/304f98e9edb6a7f058cf358bdcedb6d05e455531/Flow.Launcher/ViewModel/MainViewModel.cs#L237C20-L241C22

Let me have a try.

Jack251970:
@taooceros Now handle the copy operation in `ResultsUpdated` event handler. Thanks for your suggestions!

Jack251970:
@taooceros PTAL?

taooceros:
@Jack251970 have you tested whether it works? I can merge this now if that's the case!

Jack251970:
@taooceros I have tested, and it can address this issue.

taooceros (approved):

taooceros:
> @taooceros I have tested, and it can address this issue.

Sounds good! Thanks for looking into this!

PR #3117, opened 2024-12-08
ci: fix permissions + migrate to fresher action (Pr Assign)

NoPlagiarism:
- added permissions, so it should fix `Resource not accessible by integration`
- migration from https://github.com/thomaseizinger/assign-pr-creator-action (archived from oct. 2023) to https://github.com/toshimaru/auto-author-assign (last commit 6 months ago)

### DIFF
--- a/.github/workflows/pr_assignee.yml
+++ b/.github/workflows/pr_assignee.yml
@@ -1,8 +1,5 @@
 name: Assign PR to creator
 
-# Due to GitHub token limitation, only able to assign org members not authors from forks.
-# https://github.com/thomaseizinger/assign-pr-creator-action/issues/3
-
 on:
   pull_request:
     types: [opened]
--- a/.github/workflows/pr_assignee.yml
+++ b/.github/workflows/pr_assignee.yml
@@ -1,7 +1,7 @@
 name: Assign PR to creator
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened]
     branches-ignore:
       - l10n_dev
--- a/.github/workflows/pr_assignee.yml
+++ b/.github/workflows/pr_assignee.yml
@@ -9,11 +9,14 @@ on:
     branches-ignore:
       - l10n_dev
 
+permissions:
+  pull-requests: write
+
 jobs:
   automation:
     runs-on: ubuntu-latest
     steps:
     - name: Assign PR to creator
-      uses: thomaseizinger/assign-pr-creator-action@v1.0.0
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_assignee.yml
```diff
@@ -1,19 +1,19 @@
...
+  pull-requests: write
+
 jobs:
   automation:
     runs-on: ubuntu-latest
     steps:
     - name: Assign PR to creator
-      uses: thomaseizinger/assign-pr-creator-action@v1.0.0
+      uses: toshimaru/auto-author-assign@v2.1.1
       with:
```
taooceros, line 18: do we still need the token?

jjw24: It will use the GITHUB_TOKEN by default if not specified, so we can remove it.

.github/workflows/pr_assignee.yml
```diff
@@ -1,19 +1,19 @@
 name: Assign PR to creator
 
-# Due to GitHub token limitation, only able to assign org members not authors from forks.
-# https://github.com/thomaseizinger/assign-pr-creator-action/issues/3
-
 on:
   pull_request:
```
jjw24, line 4: According to the documentation this needs to be pull_request_target

jjw24:
```suggestion
  pull_request_target:
```

.github/workflows/pr_assignee.yml
jjw24: I can't make the suggestion commit because that line hasn't changed and I'm on my phone, so let's ignore for now, doesn't make much of difference.

jjw24 (approved):
Thanks for the fix 👍

.github/workflows/pr_assignee.yml
NoPlagiarism: Then maybe milestone should go to `pull_request_target` too?

jjw24: It's specified in the auto assign action readme + issue, no mention of it for the milestone action.

NoPlagiarism: It is mentioned here

https://github.com/andrefcdias/add-to-milestone/blob/399d78cf88a6d47d4b600e2e0b1f4f0b2bdfbe7c/src/runner.ts#L94

jjw24: Yeah ok, could you make update it with a new PR please. Also please remove the GitHub token line for auto assign while you are at it.

PR #3118, opened 2024-12-08
Add preserve last action keyword options

### LINKED ISSUE #3116: Add "Preserve Last Action Keyword" to "Last Query Style" option in settings
Someone in Discord asked for this option. I think this makes perfect sense for people that often need to make a lot of different searches with the same plugin.

How it should work:

1. The plugin keyword is `foo`
2. User types in `foo bar`
3. User selects the first result, Flow Launcher hides
4. User opens Flow Launcher again
5. The search bar contains `foo ` (with the space at the end), user can immediately begin typing new search in the same plugin

Yusyuriv:
Closes #3116. My IDE also removed a bunch of unnecessary whitespace when I edited these files. I thought it's not a big deal and that whitespace should have been removed anyway, so I didn't exclude those changes from the PR.

# What it is
These changes introduce two new options to Last Query Style in settings: "Preserve Last Action Keyword" and "Select Last Action Keyword". They work just like "Preserve/Select Last Query", but only for action keywords, not for whole queries. With these options, when there is an action keyword, the next time the user opens Flow Launcher, the query is set to `keyword ` (with a space after it, so the user can immediately begin typing). If the last query didn't have an action keyword, the query is set to an empty string.

# Testing
I've tested:
* That the previous options still work after my changes
* New options work as expected:
  * Preserve Last Action Keyword before the window hides
    image
  * Preserve Last Action Keyword after the window hides and re-appears
    image
  * Select Last Action Keyword before the window hides
    image
  * Select Last Action Keyword after the window hides and re-appears
    image

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -771,7 +771,7 @@ namespace Flow.Launcher.ViewModel
         public string Image => Constant.QueryTextBoxIconImagePath;
 
         public bool StartWithEnglishMode => Settings.AlwaysStartEn;
-        
+
         #endregion
 
         #region Preview
@@ -833,7 +833,7 @@ namespace Flow.Launcher.ViewModel
         }
 
         private void HidePreview()
-        {            
+        {
             if (PluginManager.UseExternalPreview())
                 CloseExternalPreview();
 
@@ -912,7 +912,7 @@ namespace Flow.Launcher.ViewModel
                     break;
             }
         }
-        
+
         private void UpdatePreview()
         {
             switch (PluginManager.UseExternalPreview())
@@ -1400,6 +1400,16 @@ namespace Flow.Launcher.ViewModel
                     if (Settings.UseAnimation)
                         await Task.Delay(100);
                     LastQuerySelected = false;
+                    break;
+                case LastQueryMode.ActionKeywordPreserved or LastQueryMode.ActionKeywordSelected:
+                    var newQuery = _lastQuery.ActionKeyword;
+                    if (!string.IsNullOrEmpty(newQuery))
+                        newQuery += " ";
+                    ChangeQueryText(newQuery);
+                    if (Settings.UseAnimation)
+                        await Task.Delay(100);
+                    if (Settings.LastQueryMode == LastQueryMode.ActionKeywordSelected)
+                        LastQuerySelected = false;
                     break;
                 default:
                     throw new ArgumentException($"wrong LastQueryMode: <{Settings.LastQueryMode}>");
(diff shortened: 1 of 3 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1401,6 +1401,16 @@ public async void Hide()
...
+                case LastQueryMode.ActionKeywordPreserved or LastQueryMode.ActionKeywordSelected:
+                    var newQuery = _lastQuery.ActionKeyword;
+                    if (!string.IsNullOrEmpty(newQuery))
+                        newQuery += " ";
+                    ChangeQueryText(newQuery);
+                    if (Settings.UseAnimation)
+                        await Task.Delay(100);
+                    if (Settings.LastQueryMode == LastQueryMode.ActionKeywordSelected)
+                        LastQuerySelected = false;
+                    break;
```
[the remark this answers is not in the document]
Yusyuriv: I believe it can never be null. Initially it gets assigned an instance of Query in this ViewModel's constructor, after that it always contains the last query.

taooceros:
Looks cool! I wonder maybe it would be interesting to allow custom hotkey to perform this action (so sometimes if we (don't) want to delete the query we could use a different hotkey).

Yusyuriv:
Something like pressing `Ctrl + NumpadPlus` (just an example) instead of `Enter` if you want to perform the action, but want the query to be preserved despite having "Empty Query" selected in settings? And a hotkey for the opposite, for emptying the query after performing the action when you have it set to "Preserve Last Query" in settings?

If that's what you meant, this does sound like it could be useful to some people, but that feature is outside the scope of this PR.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1401,6 +1401,16 @@ public async void Hide()
                         await Task.Delay(100);
                     LastQuerySelected = false;
                     break;
+                case LastQueryMode.ActionKeywordPreserved or LastQueryMode.ActionKeywordSelected:
+                    var newQuery = _lastQuery.ActionKeyword;
+                    if (!string.IsNullOrEmpty(newQuery))
+                        newQuery += " ";
+                    ChangeQueryText(newQuery);
+                    if (Settings.UseAnimation)
+                        await Task.Delay(100);
```
jjw24, line 1410: Why do we wait here?

jjw24: Ok because using animation, think previously other options had some issues when animation is in process.

jjw24 (approved):

Flow.Launcher/ViewModel/MainViewModel.cs
Yusyuriv: I'm actually not sure. It's mentioned here https://github.com/Flow-Launcher/Flow.Launcher/blob/0506f095b34bc10d34eaa54dff171ff5f803ef08/Flow.Launcher/ViewModel/MainViewModel.cs#L1390-L1403 and I added it in my code because this was present in all three other `case` branches, so it probably solves some old issue I don't know about.

PR #3144, opened 2024-12-23
Add options for setting result on top & keeping result order

Jack251970:
1. Fix user result set max value issue

If one plugin set the score of one result to int.MaxValue, `FL` should not add selected count & priority score.

Generally, this plugin want this result to the top of the query list.

2. Add support for removing selected count from score
 
If one plugin want to force the sequence of its results, it can set the `AddSelectedCount` to false so that `FL` will not add selected count.

Like one plugin with many query items which needs to allow user to set the sequence manually (order by datatime, etc.).

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -1471,9 +1471,9 @@ namespace Flow.Launcher.ViewModel
                 {
                     if (_topMostRecord.IsTopMost(result))
                     {
-                        result.Score = int.MaxValue;
+                        result.Score = Result.MaxScore;
                     }
-                    else if (result.Score != int.MaxValue)
+                    else if (result.Score != Result.MaxScore)
                     {
                         var priorityScore = metaResults.Metadata.Priority * 150;
                         result.Score += result.AddSelectedCount ?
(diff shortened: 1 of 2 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1473,10 +1473,12 @@ public void UpdateResultView(ICollection<ResultsForUpdate> resultsForUpdates)
                     {
                         result.Score = int.MaxValue;
                     }
-                    else
+                    else if (result.Score != int.MaxValue)
```
jjw24, line 1476: Just wanted to note here, at some point I think it's best to provide an actual max value- standardise the scoring system so to not allow randomly assigning any score by plugins. 

Also this int max value is not documented anywhere so unless devs look into this Result class they probably won't know that they can do this.

Any how just thought I mention these, but they are out side the scope of this change.

Jack251970: @jjw24 Now I export the max score in the `Result` class so developers can see this maximum score.

jjw24 (approved):

jjw24:
👍 thanks for adding this long awaited option.

PR #3151, opened 2024-12-28
[API] Escape ActionKeywords duplicates

NoPlagiarism:


taooceros:
should we warn this with an error or something? I think technically plugin should call whether this actionkeyword is available before calling add (correct me if I am wrong as I add those thing way too long ago).

taooceros:
We don't really support duplicate keyword...I don't know whether we should though

NoPlagiarism:
> warn this with an error

I guess it should be mentioned in logs 🤔 

> support duplicate keyword

This PR is not to add support, but to escape it in the first place. If there are 2 identical keywords, nothing bad happens

taooceros:
I was a little bit not sure whether we want to do this. For example a plugin add two action keywords for two different purposes. However because of this change, when user delete one then both will be deactivated, which might not be an ideal scenario.

Generally I think people solves this kind of thing with some reference counting, but I am not sure whether we want to do that.

jjw24:
> We don't really support duplicate keyword...I don't know whether we should though

Probably separate conversation to this change. But to answer your question, currently I think we should allow duplicates. The result might look messy but this means having duplicate actionkeywords is not a show stopper. Thinking in the long term when we have hundreds of plugins it's just not feasible for plugin devs to find a non-used action keyword.

jjw24:
> > warn this with an error
> 
> I guess it should be mentioned in logs 🤔
> 
> > support duplicate keyword
> 
> This PR is not to add support, but to escape it in the first place. If there are 2 identical keywords, nothing bad happens

What is the current behaviour? IIRC flow will error out and disable the plugins right?

With this change does it mean it will trigger both plugins' results after the action keyword is added?

NoPlagiarism:
Stop. Let me get this straight to skip any misunderstandings.

This PR is to escape per-plugin duplicates

This does not interferes with other plugins. It only looks up current plugin before adding new action keyword to it

taooceros:
> Stop. Let me get this straight to skip any misunderstandings.
> 
> This PR is to escape per-plugin duplicates
> 
> This does not interferes with other plugins. It only looks up current plugin before adding new action keyword to it
> 
> P.S: I can rename PR & commit to be more specific if needed

No I understand the purpose of this pr. However, as I mentioned, it impose the risk that when a plugin add 2 duplicate action keyword, but when it delete one, then both will be deleted.

taooceros:
@NoPlagiarism what do you think?

NoPlagiarism:
IG, you're right. I'll close PR for now

PR #3161, opened 2025-01-02
Add System Language Item

Jack251970:
# Add support of system language item to help FL automatically select the system language as display language

image

* Test already done
   * Change to "System" language
   * Switch Windows display language to English & Chinese
   * After switching the language, restart FL and the display language of FL changes


Flow.Launcher.Core/Resource/AvailableLanguages.cs
```diff
@@ -63,5 +62,38 @@ public static List<Language> GetAvailableLanguages()
             };
             return languages;
         }
+
+        public static string GetSystemTranslation(string languageCode)
+        {
+            return languageCode switch
+            {
+                "en" => "System",
+                "zh-cn" => "系统",
```
jjw24, line 71: Is there a way/website to get these translated strings, for example when adding a new language?

Jack251970: > Is there a way/website to get these translated strings, for example when adding a new language?

I have not considered this before.

> Or is it just through google translate?

Yeah, it is a convenient way to provide translation for `System`.

I find in `AvailableLanguage.cs` there are the list of all languages, so I put this translation here and when adding a new language, developers will notice this function as well and add new translation for this.

Jack251970: Also, because the languages of FL are limited, so using language code to get the translation of `System` is a good choice to show the correct text.

Jack251970: @jjw24 Any ideas?

taooceros: > Also, because the languages of FL are limited, so using language code to get the translation of `System` is a good choice to show the correct text.

I think it might be better to use the resource for this, so next time when we are adding a new language we won't forget about this.

Flow.Launcher.Core/Resource/Internationalization.cs
```diff
@@ -68,8 +69,18 @@ private void LoadDefaultLanguage()
...
             languageCode = languageCode.NonNull();
-            Language language = GetLanguageByLanguageCode(languageCode);
-            ChangeLanguage(language);
+
+            // Get actual language if language code is system
+            var isSystem = false;
+            if (languageCode == Constant.SystemLanguageCode)
+            {
+                languageCode = GetSystemLanguageCode();
+                isSystem = true;
```
taooceros, line 78: why do we need this? I think we are checking whether the language code is system instead?

Jack251970: If we have set the current language to system, and the `_settings.Language` is `Constant.SystemLanguageCode`. So when we call `ChangeLanguage` function, we need to get the actual language code for changing the language.

Flow.Launcher.Core/Resource/AvailableLanguages.cs
Jack251970: @taooceros From my view, it is not a good idea to use resource for this.

If our system is in Chinese and current language in FL is English, the loaded dictionary is `en.xaml`, and we need to load `zh-cn.xaml`. (By the way, `GetTranslation()` of current version cannot work)

And in `AvailableLanguage.cs`, there are the list of all languages, so next time when we are adding a new language we also won't forget about this.

taooceros: I see. That makes sense. I agree system should follow the system language.

Flow.Launcher.Core/Resource/Internationalization.cs
taooceros: Sorry I mean why do we need `isSystem`.

Jack251970: @taooceros In `private void ChangeLanguage(Language language, bool isSystem)`, the `Language language` is the actual language item, so we cannot know whether language is system or not.

So `isSystem` is used for setting storage here `Settings.Language = isSystem ? Constant.SystemLanguageCode : language.LanguageCode;`.

E.g. If the FL language is system and your windows is in English, here `Language language` is English language item but we cannot store `en` in the settings.

taooceros: Then shouldn't we parse the system language before invoking `ChangeLanguage`?

Jack251970: Cannot get your meaning. Why should we parse the system language before invoking `ChangeLanguage`? I just encapsulate all codes inside the `Internationalization.cs`.

jjw24: > Sorry I mean why do we need `isSystem`.

@taooceros looks like we do need it, it's mainly for storing it the system code in the settings file- line 116. It loads the actual language from system code, but we persist the system code not the code of the actual language.

Flow.Launcher.Core/Resource/Internationalization.cs
```diff
@@ -167,7 +177,35 @@ private void LoadLanguage(Language language)
...
+            list.Insert(0, new Language(Constant.SystemLanguageCode, AvailableLanguages.GetSystemTranslation(GetSystemLanguageCode())));
+            return list;
+        }
+
+        private string GetSystemLanguageCode()
+        {
+            var availableLanguages = AvailableLanguages.GetAvailableLanguages();
+
+            // Retrieve the language identifiers for the current culture
+            var currentCulture = CultureInfo.CurrentCulture;
```
jjw24, line 190: There is a slight issue with this method. If you have let's say Korean selected, then switch to system, this CultureInfo.CurrentCulture actually returns Korean. So, it's not returning the actual language of system.

Jack251970: That makes sense and now move it to initialization.

jjw24:
I just renamed the GetSystemLanguageCode to be more descriptive about its intended usage.

jjw24 (approved):
👍

PR #3122, opened 2024-12-11
Replace DllImport and flags with CSWin32 & Remove useless DllImport and flags & Fix right Win key click issue

Jack251970:
# PR
* Replace DllImport & flags with CSWin32 to improve code quality and fix possible issue like getting localized string of certain application like `Windows Performance Recorder`.
* Remove useless DllImport & flags in `SingleInstance.cs`.
* Fix right Win key click issue (Fix VIRTUAL_KEY.VK_RWIN check issue).
* Fix clipboard action under STA thread issue (see more in https://learn.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-oleinitialize).

# Test already done
    * Sleep & Empty Recycle Bin action in `Flow.Launcher.Plugin.Sys` plugin.
    * All changes in `Flow.Launcher.Plugin.ProcessKiller` plugin.
    * All changes in `Flow.Launcher.Plugin.Program` plugin.
    * All changes in `ThumbnailReader.cs` in `Flow.Launcher.Plugin.Infrasture`.
    * `API.CopyToClipboard` & `new BuiltinShortcutModel("{clipboard}", "shortcut_clipboard_description", () => Win32Helper.StartSTATaskAsync(Clipboard.GetText).Result)`.

PR #3165, opened 2025-01-03
Fix Copy Action Issue

Jack251970:
Revert "Fix clipboard action under sta thread issue" to fix copy action issue in #3122.


jjw24 (approved):
Awesome, thank you.

PR #3170, opened 2025-01-05
Add New API for ProgressBoxEx to Show Progress & Add Progress Display for Plugin Downloading & Improve DownloadUrl Api Function

Jack251970:
# Content
* Add new api of showing progress: `IProgressBoxEx ShowProgressBox(string caption, Action forceClosed = null)`.
* Add progress displaying & cancellation when downloading plugin

# Screenshots
## Interface
image
## Cancel Download
https://github.com/user-attachments/assets/aa4d345d-b189-41b4-b9ab-b4505b4d0e73
## Finish Download
https://github.com/user-attachments/assets/b76c9ddd-6f3d-4cea-b7b8-4457619a1b1a

# Test
* `ProgressBoxEx` usage in PluginManager plugin (progress displaying & cancellation)


taooceros:
Mind sharing some screenshot (or a short video) to demonstrate the behavior?

Jack251970:
> Mind sharing some screenshot (or a short video) to demonstrate the behavior?

Sure, please check updated desc.

taooceros:
@onesounds what do you think?

onesounds:
Designwise, there is nothing to point out. It looks good.

Flow.Launcher/ProgressBoxEx.xaml.cs
```diff
@@ -0,0 +1,117 @@
...
+        private new void Close()
+        {
+            if (_isClosed)
+            {
+                return;
+            }
+
+            base.Close();
+            _isClosed = true;
+        }
```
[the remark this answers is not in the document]
taooceros: Why do we need the overload? I don't think this is a good idea.

taooceros:
I feel like the code related to progress report is way to complicated and should requires some encapsulation. Would you mind do something like this https://stackoverflow.com/a/46497896 (and maybe use the `IProgress<T>` interface instead of creating a new one).

Flow.Launcher/ProgressBoxEx.xaml.cs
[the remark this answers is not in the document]
Jack251970: > Why do we need the overload? I don't think this is a good idea.

Now remove this.

Jack251970:
> I feel like the code related to progress report is way to complicated and should requires some encapsulation. Would you mind do something like this https://stackoverflow.com/a/46497896 (and maybe use the `IProgress<T>` interface instead of creating a new one).

I have considered to use delegate to report the progress. (This indeed can remove the IProgressBoxEx)

But in `PluginManager` plugin, it can cause the issue that the progress window cannot be closed if catching exception in
```
catch (HttpRequestException e)
{
    // force close progress box
    await prgBox?.CloseAsync();

    // show error message
    Context.API.ShowMsgError(
        string.Format(Context.API.GetTranslation("plugin_pluginsmanager_downloading_plugin"), plugin.Name),
        Context.API.GetTranslation("plugin_pluginsmanager_download_error"));
    Log.Exception("PluginsManager", "An error occurred while downloading plugin", e);

    return;
}
catch (Exception e)
{
    // force close progress box
    await prgBox?.CloseAsync();

    // show error message
    Context.API.ShowMsgError(Context.API.GetTranslation("plugin_pluginsmanager_install_error_title"),
        string.Format(Context.API.GetTranslation("plugin_pluginsmanager_install_error_subtitle"),
            plugin.Name));
    Log.Exception("PluginsManager", "An error occurred while downloading plugin", e);

    return;
}
```

So I need to get the progress window in the exception sentence, like using `await prgBox?.CloseAsync();` to force the FL to close progress box as you can see above.

taooceros:
I don't really understand sorry. Why does delegate cannot handle that? I think we can use some special value (like -1) to represent error on the delegate so I don't really think that's a problem?

Jack251970:
> I don't really understand sorry. Why does delegate cannot handle that? I think we can use some special value (like -1) to represent error on the delegate so I don't really think that's a problem?

@taooceros Sorry, it seems that I encountered the thread issue last time. Now update and please check.

taooceros:
> > I don't really understand sorry. Why does delegate cannot handle that? I think we can use some special value (like -1) to represent error on the delegate so I don't really think that's a problem?
> 
> @taooceros Sorry, it seems that I encountered the thread issue last time. Now update and please check.

Totally understand the thread issue. Maybe you can make the ProgressBar MVVM so we didn't modify the UI state directly but rather updating the view via the property (which by WPF spec is thread safe)?

taooceros:
As the stackoverflow indicated, could you make the ProgressReport part an extension method of Http (maybe inside the Http class). That would also allow other parts of code to utilize this feature if they wants. Then interact the progress report with the ProgressWindow via delegate or something instead of handling the stream itself, which makes this part of code too complicated.

Jack251970:
> As the stackoverflow indicated, could you make the ProgressReport part an extension method of Http (maybe inside the Http class). That would also allow other parts of code to utilize this feature if they wants. Then interact the progress report with the ProgressWindow via delegate or something instead of handling the stream itself, which makes this part of code too complicated.

I see. I improve the `DownloadUrl` api funcion. Please check.

jjw24:
@taooceros is there anything additional you still want to review? Are we happy to merge this?

taooceros:
Sorry I was a bit stuck by too many things lately (as I travel back to China for new year).

I will recheck this today.

Jack251970:
@taooceros Any time to recheck it?

taooceros:
> @taooceros Any time to recheck it?

Sorry was stuck by a paper submission and traveling before. Now I am back. Will check now

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
```diff
@@ -190,6 +199,34 @@ internal async Task InstallOrUpdateAsync(UserPlugin plugin)
             }
         }
 
+        private async Task DeleteFileAndDownloadMsgBoxAsync(string prgBoxTitle, string downloadUrl, string filePath, CancellationTokenSource cts)
```
taooceros, line 202: what does this name mean🤔

taooceros: From my opinion this name is too long and confusing. Maybe it's better to make `delete file` optional as a parameter `override` and `msgbox` optional as a parameter `showProgress?

taooceros:
Overall it looks cool to me, thanks! One little concern is the parameter with type `Func<Action<void>, Task>`, which makes sense but is a bit confusing, but I guess it's unavoidable when we are using callback. I will check the functionality soon and approve afterward.

Plugins/Flow.Launcher.Plugin.PluginsManager/PluginsManager.cs
Jack251970: It makes sense! Using parameters can make it more readable.

taooceros (approved):

PR #3174, opened 2025-01-09
Use System.Text.Json to replace Newtonsoft.Json for better performance

Jack251970:
Use System.Text.Json to replace Newtonsoft.Json for better performance & Improve code quality

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.PluginsManager/Utilities.cs
+++ b/Plugins/Flow.Launcher.Plugin.PluginsManager/Utilities.cs
@@ -71,9 +71,8 @@ namespace Flow.Launcher.Plugin.PluginsManager
 
                 if (pluginJsonEntry != null)
                 {
-                    using StreamReader reader = new StreamReader(pluginJsonEntry.Open());
-                    string pluginJsonContent = reader.ReadToEnd();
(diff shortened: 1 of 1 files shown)


Plugins/Flow.Launcher.Plugin.PluginsManager/Utilities.cs
```diff
@@ -72,12 +71,10 @@ internal static UserPlugin GetPluginInfoFromZip(string filePath)
...
                 {
-                    using (StreamReader reader = new StreamReader(pluginJsonEntry.Open()))
-                    {
-                        string pluginJsonContent = reader.ReadToEnd();
-                        plugin = JsonConvert.DeserializeObject<UserPlugin>(pluginJsonContent);
-                        plugin.IcoPath = "Images\\zipfolder.png";
-                    }
+                    using StreamReader reader = new StreamReader(pluginJsonEntry.Open());
+                    string pluginJsonContent = reader.ReadToEnd();
+                    plugin = JsonSerializer.Deserialize<UserPlugin>(pluginJsonContent);
```
taooceros, line 76: I remember there's an overload for stream directly so we don't need to read it as string before deserializing it

Jack251970: Yeah, please check.

taooceros (approved):

PR #3175, opened 2025-01-09
Introduce Dependency Injection for Better Development Experience

Jack251970:
Introduce dependency injection for accessing the `API` property in anywhere inside the project.

This can improve development experience. E.g. We can put the `MessgaeBoxEx` to `Flow.Launcher` project and we can access the design page like this:

image

### DIFF
--- a/Flow.Launcher/PublicAPIInstance.cs
+++ b/Flow.Launcher/PublicAPIInstance.cs
@@ -34,16 +34,14 @@ namespace Flow.Launcher
     public class PublicAPIInstance : IPublicAPI
     {
         private readonly Settings _settings;
-        private readonly SettingWindowViewModel _settingsVM;
         private readonly MainViewModel _mainVM;
 
         #region Constructor
 
-        public PublicAPIInstance()
+        public PublicAPIInstance(Settings settings, MainViewModel mainVM)
         {
-            _settings = Ioc.Default.GetRequiredService<Settings>();
-            _settingsVM = Ioc.Default.GetRequiredService<SettingWindowViewModel>();
-            _mainVM = Ioc.Default.GetRequiredService<MainViewModel>();
+            _settings = settings;
+            _mainVM = mainVM;
             GlobalHotkey.hookedKeyboardCallback = KListener_hookedKeyboardCallback;
             WebRequest.RegisterPrefix("data", new DataWebRequestFactory());
         }
@@ -87,7 +85,7 @@ namespace Flow.Launcher
         {
             PluginManager.Save();
             _mainVM.Save();
-            _settingsVM.Save();
+            _settings.Save();
             _ = ImageLoader.Save();
         }
 


taooceros:
I was thinking about unifying the access to `IPublicAPI`. Currently this is quite messy (some uses `PluginManager.API` some uses other technique).

Very earlier I was thinking about integrating dependency injection in flow, but that seems to be pretty complicated. Maybe we can do that gradually specifically for the PublicAPI?

I am not sure. What do others think about global state access?

Jack251970:
@taooceros I agree that dependency injection is the most elegant way to access `API` from anywhere. But considering the current architecture of the FL codes , `App.API` should not be deprecated (so many codes rely on this). So this PR exposes `App.API` for other projects to call it.

Jack251970:
I mean it can serve as a transitional solution for unifying the call of `ProgressBoxEx`.

Jack251970:
> I was thinking about unifying the access to `IPublicAPI`. Currently this is quite messy (some uses `PluginManager.API` some uses other technique).
> 
> Very earlier I was thinking about integrating dependency injection in flow, but that seems to be pretty complicated. Maybe we can do that gradually specifically for the PublicAPI?
> 
> I am not sure. What do others think about global state access?

Now I implement dependency injection model for this, and can you please take a look of this?

Flow.Launcher/PublicAPIInstance.cs
```diff
@@ -33,19 +34,26 @@ public class PublicAPIInstance : IPublicAPI
...
+            _mainVM = Ioc.Default.GetRequiredService<MainViewModel>();
             GlobalHotkey.hookedKeyboardCallback = KListener_hookedKeyboardCallback;
             WebRequest.RegisterPrefix("data", new DataWebRequestFactory());
         }
 
+        public void Initialize()
+        {
+            // We need to initialize Updater not in the constructor because we want to avoid
+            // recrusive dependency injection
+            _updater = Ioc.Default.GetRequiredService<Updater>();
```
taooceros, line 54: I remember this is one of the reason I forfeit DI before. Not elegant but maybe good enough.

Flow.Launcher/SettingWindow.xaml.cs
```diff
@@ -145,7 +151,7 @@ private double WindowLeft()
         var screen = Screen.FromPoint(System.Windows.Forms.Cursor.Position);
         var dip1 = WindowsInteropHelper.TransformPixelsToDIP(this, screen.WorkingArea.X, 0);
         var dip2 = WindowsInteropHelper.TransformPixelsToDIP(this, screen.WorkingArea.Width, 0);
-        var left = (dip2.X - this.ActualWidth) / 2 + dip1.X;
+        var left = (dip2.X - ActualWidth) / 2 + dip1.X;
```
taooceros, line 154: You sure this is correct? I think it's very instance dependent

taooceros:
Will this create conflict with #2774 ?

Flow.Launcher/SettingWindow.xaml.cs
Jack251970: I believe this is only necessary under the occasions as follows:

Disambiguation:
If a local variable or parameter has the same name as a class property or field, you need this. to distinguish the instance member from the local scope.

Accessing the current instance explicitly:
When passing the current object as a parameter to a method or event handler.

So I think I can remove this here.

Jack251970:
> Will this create conflict with #2774 ?

@taooceros From my view, it will not cause any conflicts. PublicAPI instance is created before other instances. And messageboxEx api function should work.

Flow.Launcher/SettingWindow.xaml.cs
taooceros: Sorry I think I read it wrongly. Somehow when I read it I think I saw somewhere you add a static.

I will recheck later

Flow.Launcher.Core/Updater.cs
```diff
@@ -17,28 +17,31 @@
...
 using System.Text.Json.Serialization;
 using System.Threading;
+using CommunityToolkit.Mvvm.DependencyInjection;
 
 namespace Flow.Launcher.Core
 {
     public class Updater
     {
-        public string GitHubRepository { get; }
+        private readonly IPublicAPI API = Ioc.Default.GetRequiredService<IPublicAPI>();
```
taooceros, line 26: Since Updater is part of the Dependency Injection, maybe we shouldn't access IPublicAPI like this, but rather through a constructor or a property injection.

https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/ioc#constructor-injection

taooceros:
I remember we have multiple cyclic reference, how's that being handled here?

Jack251970:
> I remember we have multiple cyclic reference, how's that being handled here?

I just do not let it initialize in the constructor.

taooceros (changes requested):

taooceros:
> I just do not let it initialize in the constructor and use Ioc.Default.GetRequiredService instead.

Alright. I think eventually we need to redesign a little bit. Those are pretty disgusting and hard to maintain.

Flow.Launcher.Core/Updater.cs
Jack251970: Is https://github.com/Flow-Launcher/Flow.Launcher/pull/3175/commits/39d7187713432d7bcf8c11911ff4a54c3d64c1c3 ok for you?

Flow.Launcher/App.xaml.cs
```diff
@@ -50,49 +76,43 @@ private async void OnStartupAsync(object sender, StartupEventArgs e)
...
-                _settings.WMPInstalled =  WindowsMediaPlayerHelper.IsWindowsMediaPlayerInstalled();
-
                 AbstractPluginEnvironment.PreStartPluginExecutablePathUpdate(_settings);
 
-                _alphabet.Initialize(_settings);
-                _stringMatcher = new StringMatcher(_alphabet);
-                StringMatcher.Instance = _stringMatcher;
-                _stringMatcher.UserSettingSearchPrecision = _settings.QuerySearchPrecision;
+                var stringMatcher = Ioc.Default.GetRequiredService<StringMatcher>();
+                StringMatcher.Instance = stringMatcher;
```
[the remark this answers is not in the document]
Jack251970: Ok, I remove instance for StringMatcher, Internalization and Theme. And I add TODO to remove InternalizationManager.Instance and ThemeManager.Instance in future.

Flow.Launcher/App.xaml.cs
```diff
@@ -50,49 +76,43 @@ private async void OnStartupAsync(object sender, StartupEventArgs e)
         {
             await Stopwatch.NormalAsync("|App.OnStartup|Startup cost", async () =>
             {
-                _portable.PreStartCleanUpAfterPortabilityUpdate();
+                Ioc.Default.GetRequiredService<Updater>().Initialize(Launcher.Properties.Settings.Default.GithubRepo);
```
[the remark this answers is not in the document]
Jack251970: Still in `private async void OnStartupAsync(object sender, StartupEventArgs e)`?

Flow.Launcher.Core/Updater.cs
Jack251970: Improved in https://github.com/Flow-Launcher/Flow.Launcher/pull/3175/commits/d4f90999c76ae3467c55fc71fda0aa4cb14d8e96

Jack251970:
> [@Jack251970 ](/Jack251970)
> ` [Add circular dependency comments](/Flow-Launcher/Flow.Launcher/pull/3175/commits/39d34b1a037e2e1a1ed3419faebeb39f693ea98d)`

Yeah, so I add comments in https://github.com/Flow-Launcher/Flow.Launcher/pull/3175/commits/39d34b1a037e2e1a1ed3419faebeb39f693ea98d.

Flow.Launcher.Core/Updater.cs
```diff
@@ -22,23 +22,26 @@ namespace Flow.Launcher.Core
 {
     public class Updater
     {
-        public string GitHubRepository { get; }
+        private readonly IPublicAPI API;
```
taooceros, line 25: should the name be `api` rather than `API` as this is a filed rather than a property?

taooceros:
@Jack251970 I make a few more service to be injected via constructor when possible. Please take a brief look.

I think the status of this pr is fine to merge.

taooceros (dismissed):

Jack251970:
@taooceros Hi, I fix unit test build issue and remove usage of _settingsVM._settings, could you please check if this works for you?

taooceros (approved):

PR #3178, opened 2025-01-10
New API for Customizing Equal Rules for Results & Fix Result Clone Issue

Jack251970:
# New API for Customizing Equal Rules for Results
`GetTitleKey` and `GetSubTitleKey` are the actions to get the key of the title or the subtitle.

These keys will be used when FL checks whether the result is the topmost record.

Or FL calculates the hashcode of the result for user selected records.

This is useful if one plugin needs to change the title or the subtitle of one record dynamically.

E.g. The records of plugin `Edge Workspace` have subtitles which can change based on the tab number of the edge workspace, but FL should regard these records as the same. (Full issue in https://github.com/cspotcode/Flow.Launcher.Plugin.EdgeWorkspaces/issues/3)

> Because #3112 is in the 1.20.0 milestone, so I have not change the codes related to this.

# Fix Result Clone Issue
I find that some properties are not cloned when calling `Clone` function in the `Result`, so I fix this clone issue.

### DIFF
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -203,6 +203,17 @@ namespace Flow.Launcher.Plugin
                 TitleHighlightData = TitleHighlightData,
                 OriginQuery = OriginQuery,
                 PluginDirectory = PluginDirectory,
+                ContextData = ContextData,
+                PluginID = PluginID,
+                TitleToolTip = TitleToolTip,
+                SubTitleToolTip = SubTitleToolTip,
+                PreviewPanel = PreviewPanel,
+                ProgressBar = ProgressBar,
+                ProgressBarColor = ProgressBarColor,
+                Preview = Preview,
+                AddSelectedCount = AddSelectedCount,
+                GetTitleKey = GetTitleKey,
+                GetSubTitleKey = GetSubTitleKey,
             };
         }
 
--- a/Flow.Launcher.Plugin/Result.cs
+++ b/Flow.Launcher.Plugin/Result.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Runtime;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
@@ -13,7 +12,6 @@ namespace Flow.Launcher.Plugin
     /// </summary>
     public class Result
     {
-
         private string _pluginDirectory;
 
         private string _icoPath;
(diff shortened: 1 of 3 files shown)


taooceros:
I actually think we don't need an action to handle this. If plugin want to specify the equality of two result, they just need one Key, so we can add an identification field that is nullable?

Jack251970:
> I actually think we don't need an action to handle this. If plugin want to specify the equality of two result, they just need one Key, so we can add an identification field that is nullable? That field can be long probably (so plugin can calculate hashcode or something directly)?
> 
> One problem is the `GetHashCode` function used in c# is not stable across different start of executable (so not persistable) for security reason, so plugin may need to implemenet an id mechanism themselves instead of relying on the GetHashCode. However, if they have the need of changing title/subtitle without making the item differently, I think that's a valid assumption they can store some id related to the result.

Yeah. I see. What about introducing `RecordKey` property to the `Result` class for FL to identify one result? (If one plugin does not specific this, FL just uses `Title` and `SubTitle` to identify this result.)

From my view, `Title` and `SubTitle` of the `Result` should be the properties for display, and `RecordKey` can be the property for identification.

That way one plugin can specify any unique key, and FL will use it, ignoring title and subtitle. Plugin can get full control and can modify title or subtitle without breaking `TopmostRecord` and `UserSelectedRecord`.

cspotcode:
`RecordKey` is also easier for non-C# plugins because it avoids extra RPC calling between C# and the external process.  And in the case of Edge Workspaces, the Edge browser already assigns each workspace a unique UUID, so we can use that value for `RecordKey`.

taooceros:
> > I actually think we don't need an action to handle this. If plugin want to specify the equality of two result, they just need one Key, so we can add an identification field that is nullable? That field can be long probably (so plugin can calculate hashcode or something directly)?
> > One problem is the `GetHashCode` function used in c# is not stable across different start of executable (so not persistable) for security reason, so plugin may need to implemenet an id mechanism themselves instead of relying on the GetHashCode. However, if they have the need of changing title/subtitle without making the item differently, I think that's a valid assumption they can store some id related to the result.
> 
> Yeah. I see. What about introducing `RecordKey` property to the `Result` class for FL to identify one result? (If one plugin does not specific this, FL just uses `Title` and `SubTitle` to identify this result.)
> 
> From my view, `Title` and `SubTitle` of the `Result` should be the properties for display, and `RecordKey` can be the property for identification.
> 
> That way one plugin can specify any unique key, and FL will use it, ignoring title and subtitle. Plugin can get full control and can modify title or subtitle without breaking `TopmostRecord` and `UserSelectedRecord`.

I think this is cool. Sorry got trapped by many things lately. Let's do that.

PR #3203, opened 2025-01-26
Improve context menu item action response & Fix origin query null exception

Jack251970:
# 1. Feature: Improve context menu item action response

For those actions in the FL and Program plugin (Pin / Unpin record to the topmost; Disable one program), when you click those records, like pin / unpin it to the topmost without closing the FL, the search bar will be empty and you must need to click `esc` to return back to the query list before any other actions.

Now with new API `BackToQueryList` (#3087), we can improve those behavious.

What's more, we need to remove the context menu cache feature, it can cause the cache of the original wrong context menu.

By the way, I do not think context menu caching is a good idea because the plugin with dynamic context menu will have a bad experience if user open the context menu of the same record twice.

## Test

Original:

Screenshot 2025-01-26 195656

Now:

https://github.com/user-attachments/assets/b3b3f752-7222-4e3d-8b7f-f9daf3003779

After you clicked the one record to pin / unpin it to the topmost, it will automatically return back to the query list.

# 2. Fix: Fix origin query null exception

**Note: This fix can make the test of the feature above work, so it is included in this PR!**

When user select the context menu item directly of one item from query list, the origin query is null. (No idea why this happens but it happens) And we need to add null check for this.

Screenshot 2025-01-26 202331

### DIFF
--- a/Flow.Launcher.Plugin/Query.cs
+++ b/Flow.Launcher.Plugin/Query.cs
@@ -1,22 +1,10 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text.Json.Serialization;
+﻿using System.Text.Json.Serialization;
 
 namespace Flow.Launcher.Plugin
 {
     public class Query
     {
         public Query() { }
-
-        [Obsolete("Use the default Query constructor.")]
-        public Query(string rawQuery, string search, string[] terms, string[] searchTerms, string actionKeyword = "")
-        {
-            Search = search;
-            RawQuery = rawQuery;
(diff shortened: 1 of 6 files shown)


jjw24 (approved):

jjw24:
@Jack251970 after a result is cancelled from topmost, it seems to still show up at the top until a restart of flow. Could you please check if this is happening on your end.

Jack251970:
> @Jack251970 after a result is cancelled from topmost, it seems to still show up at the top until a restart of flow. Could you please check if this is happening on your end.

Hello, fixed in #3257.

Issue #3187, opened 2025-01-17
BUG: Settings window is not responsive after pc wakes from sleep.


j-a-s-o-n-g:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I am using the latest version of Flow Launcher.

### Problem Description

First of all, thank you so much for FlowLauncher's existence, I can't imagine a world where this doesn't exist for me. Much love <3

So this happens on a pretty consistent basis, and doesn't always reproduce with every system sleep. However it seems the longer the computer sleeps the higher the chances I encounter this issue. 

It seems whenever I wake my computer from the previous day, FL's settings window never responds to user input if I ever need to pull it up. The left nav doesn't click, the right side options don't click, nothing. I don't even think the close buttons or anything works. It's just there as a static backdrop. The only way to fix this is to quit the process of FL and re-open it, then it works again.

Also to note:
- After waking the computer, FL's core program works fine. I can pull it up, type in most stuff as normal and it works. This is just the settings window if I ever try to access it.
- I have a two monitor setup but I have FL settings window on my primary monitor because I am aware settings window on the second monitor is even worse where it's wacky even if I don't put my computer to sleep.

### To Reproduce

1. Work normally with flowlauncher
2. Put computer to sleep overnight
3. Wake computer
4. Trigger Flow Launcher, type in "settings" press enter.
5. Settings window appears but no clicks register.

### Flow Launcher Version

1.19.4

### Windows Build Number

Windows 11 Home 22631.4602

### Error Log

No error info produced.

Jack251970:
I have experienced this for many times, and I recorded more details in #3139.

Jack251970:
I find that if you open settings with FL search, the settings window will be responsive:

![Image](https://github.com/user-attachments/assets/a1852841-0aa4-4ad2-927e-dcf88d19bd19)

Jack251970:
> I find that if you open settings with FL search, the settings window will be responsive:
> 
> ![Image](https://github.com/user-attachments/assets/a1852841-0aa4-4ad2-927e-dcf88d19bd19)

This can serve as a temporary solution for this.

jjw24:
Many thanks to Jack251970, this is now fixed and available in the prerelease https://github.com/Flow-Launcher/Prereleases/releases.

Issue #3169, opened 2025-01-04
BUG: General settings and plugins store appear blank


d3faulter:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

Sometimes when opening settings the "General" and "Plugin Store" page appear blank (forgot to check the other settings pages, so I don't know wether or not those display correctly).

Other times only "Plugin Store" page appears blank and I am unable to search for any plugins in there as well as not getting any suggestions or previews using "pm install" in launcher. Writing out name in full does successfully install plugins.

Issue is persistent and a restart of Flow Launcher seems to fix it. Sometimes it happens multiple times in a row requiring multiple restarts.

Win builds where i have experienced the issue:
- 26100.2605 (Regular availability)
- 10.0.22635.4655 (Beta)
- 10.0.26120.2705 (Dev)

Possibly related issue: [#2307](https://github.com/Flow-Launcher/Flow.Launcher/issues/2307)

Looked through logs and don't think errors are related, but they're included anwyay

### To Reproduce

1. Start Flow Launcher
2. Open launcher
3. Open Settings

### Screenshots

![Flow Launcher_vn8VrMHECT](https://github.com/user-attachments/assets/1c880492-6c60-4530-8924-7542a024e4bf)

### Flow Launcher Version

1.19.14

### Windows Build Number

26100.2605

### Error Log

[2024-12-28.txt](https://github.com/user-attachments/files/18307826/2024-12-28.txt)
[2024-12-29.txt](https://github.com/user-attachments/files/18307827/2024-12-29.txt)
[2024-12-31.txt](https://github.com/user-attachments/files/18307828/2024-12-31.txt)
[2025-01-01.txt](https://github.com/user-attachments/files/18307829/2025-01-01.txt)
[2025-01-02.txt](https://github.com/user-attachments/files/18307830/2025-01-02.txt)
[2025-01-03.txt](https://github.com/user-attachments/files/18307831/2025-01-03.txt)
[2025-01-04.txt](https://github.com/user-attachments/files/18307832/2025-01-04.txt)

d3faulter:
Don't know if i should open another issue, as i just encountered the following: 

Opened Settings on an existing instance of Flow, no menu pages on the left except "About" responded (others were selectable but only "About" showed.

When trying to resize the window, white border first appeared and then white box as on the image below.

Updated logs (from 12.48):
[...]

Issue #3139, opened 2024-12-19
BUG: Setting Window Freeze Issue after System Sleeping


Jack251970:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

When you change the system to sleep mode, and the settings window will possibly freeze.

In `SettingWindows.xaml.cs`,
`
private void OnLoaded(object sender, RoutedEventArgs e)
{
    RefreshMaximizeRestoreButton();
    // Fix (workaround) for  (Win+L)
    // https://stackoverflow.com/questions/4951058/software-rendering-mode-wpf
    HwndSource hwndSource = PresentationSource.FromVisual(this) as HwndSource;
    HwndTarget hwndTarget = hwndSource.CompositionTarget;
    hwndTarget.RenderMode = RenderMode.Default;

    InitializePosition();
}
`
these codes are used for `the window freezes after lock screen`, but it seems that these codes cannot fix the problem when the system goes to the sleep mode.

This is probably the correct reason of the issue of PR #3088.

### To Reproduce

1. Sleep Windows.
2. Start Windows.
3. Open settings window, and the window is frozen.

### Screenshots

https://github.com/user-attachments/assets/faaea9be-83a2-4a4d-8852-65002e242352

When the settings window is frozen, the main window can still work but the right context menu of the search will freeze.

And the `NavigationView_SelectionChanged(NavigationView sender, NavigationViewSelectionChangedEventArgs args)` event in `SettingWindows.xaml.cs` can still be triggered when you click the menu item but the interface will not refresh at all.

### Flow Launcher Version

1.19.4

### Windows Build Number

10.0.22631.4037

taooceros:
I think it's related to the hardware acceleration of WPF. There are some underlying issue of the framework.

Jack251970:
It seems that I cannot find any way to stably reproduce this.

PR #3204, opened 2025-01-27
Fix setting window freeze issue & Improve singleton window opener

Jack251970:
Fix setting window freeze issue after lock screen (Win+L) or sleep (Some commits seem to change the codes that try to fix this issue 😢)
Improve singleton window opener

Already tested on my computer and it works well now!

Related issue: #3139, #3169, #3187


jjw24 (approved):
👍

Issue #3212, opened 2025-01-29
Option to change priority at startup, so it launches sooner


MikeSesketcharba:
Flowlauncher starts up with my machine. But it takes 30 seconds or more to be ready. I'd like the ability within the settings, to be able to optionally set the startup priority to higher. This would require admin privileges.

Thanks! Love this app, fell in love instantly.
```
Edition	Windows 11 Pro
Version	23H2
Installed on	‎10/‎10/‎2024
OS build	22631.4751
Experience	Windows Feature Experience Pack 1000.22700.1055.0
```

deefrawley:
Hi. As far as I'm aware Windows controls startup order and this can't be changed from within apps themselves. You can Google possible ways to do this. Flow also takes some time to load at startup as it loads itself and plugins. If you go to the log folder you can read the main log where it shows exactly how long each plugin takes (so you can see if any one plugin is the culprit for a longer startup time).

MikeSesketcharba:
Oh well. I might check that out. Thank you for helping me!

29 Jan 2025 21:57:02 Damien ***@***.***>:

Jack251970:
@MikeSesketcharba Fully understand your request. I have similar experience when I start my system and need to launch other apps but FL has not launched yet. Now I create a PR for using logon task and this can make FL launch faster than apps using startup entries. Hope this can meet your need!

Robert-K:
I've also been looking to solve this problem.

[Wallpaper Engine handles this really well, they have a "high-priority" option which allows it to launch much earlier.](https://help.wallpaperengine.io/en/functionality/automaticstartup.html)

Looks like they are registering it as a Windows service. I think a similar option would fit well in FL.

chrisl8:
I literally quit using Flow Launcher because of this. If I have to use the normal Start menu when I start my computer, my muscle memory never locks onto the Flow keyboard shortcut.

I did some testing today and can confirm.

Time from Flow Launcher first log entry to last is from 4 to 6 seconds, very fast.
However, Flow Launcher's first log entry can be as much as 2 minutes after login!

Somehow literally every other thing on my computer gets priority, including Steam, Everything, Signal, Discord, etc.

Sure, I could have less crap on startup, but these things cause me no grief, they just do their thing while I work, but I need/want Flow to be first on the scene, so anything we could do to improve that would be amazing!

Let me know if I can help debug, test and/or experiment with anything for you to further this effort.

[...]

PR #3218, opened 2025-01-31
Support Logon Task for Faster Startup Experience

Jack251970:
# Add support for FL to use logon task instead of startup entry for faster strartup experience.

# Test

After using logon task for startup, FL launches as fast as PowerToys (the center app in the tray) which uses logon task for startup as well.

Also, it launchers faster than other apps using startup entry, like Microsoft Security Center (You cannot see it here because it has not launched yet.)
Screenshot 2025-01-31 215138

Related issue: #3212.

### DIFF
--- a/Flow.Launcher/Helper/AutoStartup.cs
+++ b/Flow.Launcher/Helper/AutoStartup.cs
@@ -1,18 +1,31 @@
 ﻿using System;
+using System.IO;
+using System.Linq;
+using System.Security.Principal;
 using Flow.Launcher.Infrastructure;
 using Flow.Launcher.Infrastructure.Logger;
 using Microsoft.Win32;
+using Microsoft.Win32.TaskScheduler;
 
 namespace Flow.Launcher.Helper;
 
 public class AutoStartup
 {
     private const string StartupPath = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Run";
+    private const string LogonTaskName = $"{Constant.FlowLauncher} Startup";
+    private const string LogonTaskDesc = $"{Constant.FlowLauncher} Auto Startup";
 
     public static bool IsEnabled
     {
         get
         {
+            // Check if logon task is enabled
+            if (CheckLogonTask())
+            {
+                return true;
+            }
+
+            // Check if registry is enabled
             try
             {
                 using var key = Registry.CurrentUser.OpenSubKey(StartupPath, true);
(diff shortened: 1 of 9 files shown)


jjw24:
@Jack251970 I noticed PT had issues removing schedule logon task when uninstalling. Could you test this part please- how does the schedule task gets handled by installed flow version?

If it is a manual process to remove the task when uninstalling flow, especially for portable version, I think we need to have this explained under the option so user is aware that manual removal is needed.

Jack251970:
@jjw24 Yeah, when I uninstall the FL, I find the logon task will not be deleted. I need to add tooltip for this.

Jack251970:
> @Jack251970 I noticed PT had issues removing schedule logon task when uninstalling. Could you test this part please- how does the schedule task gets handled by installed flow version?
> 
> If it is a manual process to remove the task when uninstalling flow, especially for portable version, I think we need to have this explained under the option so user is aware that manual removal is needed.

@jjw24 Please check.

rottakore:
Creating a string key on HKLM...\Run to launch FL is a tiny tiny bit faster than task scheduler

Jack251970:
@rottakore Thanks for your idea! I have decided to use this way.

Jack251970:
@jjw24 I attempted to use the HKLM\...\Run registry key, but it requires significant adjustments since FL is not designed to run with administrator privileges. For now, I’ll continue using the logon task to avoid complications.

Flow.Launcher/SettingPages/ViewModels/SettingsPaneGeneralViewModel.cs
```diff
@@ -42,9 +42,16 @@ public bool StartFlowLauncherOnSystemStartup
...
-                    AutoStartup.Enable();
+                {
+                    // Enable either registry or task scheduler
+                    AutoStartup.Enable(UseLogonTaskForStartup);
+                }
                 else
-                    AutoStartup.Disable();
+                {
+                    // Disable both registry and task scheduler
+                    AutoStartup.Disable(true);
```
jjw24, line 52: One nitpick, could we change this to DisableViaLogonTask & EnableViaLogonTask and same type of naming for registry.

jjw24: Keep these original methods as private and just add the additional wrapper functions for easier understanding.

Jack251970: Well, that is find. Please check the latest commit.

Flow.Launcher/Helper/AutoStartup.cs
```diff
@@ -42,17 +117,74 @@ public static void Disable()
...
+
+    private static bool UnscheduleLogonTask()
+    {
+        using var taskService = new TaskService();
+        try
+        {
+            taskService.RootFolder.DeleteTask(LogonTaskName);
+            return true;
+        }
+        catch (Exception)
```
jjw24, line 177: Could you catch the error message so can be logged.

Flow.Launcher/Helper/AutoStartup.cs
```diff
@@ -42,17 +117,74 @@ public static void Disable()
...
+        td.Settings.StopIfGoingOnBatteries = false;
+        td.Settings.DisallowStartIfOnBatteries = false;
+        td.Settings.ExecutionTimeLimit = TimeSpan.Zero;
+
+        try
+        {
+            TaskService.Instance.RootFolder.RegisterTaskDefinition(LogonTaskName, td);
+            return true;
+        }
+        catch (Exception)
```
jjw24, line 162: Log the exception error message too here

Flow.Launcher/Helper/AutoStartup.cs
```diff
@@ -28,12 +41,74 @@ public static bool IsEnabled
...
+                var action = task.Definition.Actions.FirstOrDefault()!.ToString().Trim();
+                if (!Constant.ExecutablePath.Equals(action, StringComparison.OrdinalIgnoreCase) && !File.Exists(action))
+                {
+                    UnscheduleLogonTask();
+                    ScheduleLogonTask();
+                }
+
+                return true;
+            }
+            catch (Exception)
```
jjw24, line 62: Log exception please

jjw24 (commented):
Let's log exception for easier troubleshooting later if needed.

Jack251970:
> Let's log exception for easier troubleshooting later if needed.

Please check

jjw24 (approved):

Susexe:
I believe the tooltip is no longer needed, since toggling the option off removed the task from Task Scheduler.

Jack251970:
> I believe the tooltip is no longer needed, since toggling the option off removed the task from Task Scheduler.

Here I mean if you uninstall Flow with this option on, the logon task will not be removed

PR #3230, opened 2025-02-07
add BackToQueryResults to JsonRPCPublicAPI

cibere:
#3087 created the `BackToQueryResults` API method, but didn't make it accessible to Jsonrpc-v2 plugins, so this pr lets jsonrpc v2 plugins use the api method


taooceros (approved):

Jack251970:
@cibere Hello, FYI, if you use this api function to make sure `ChangeQuery` and `ReQuery` work when results are from context menu or history, you can remove this function. See more in #3240.

Issue #3229, opened 2025-02-07
BUG: Explorer Plugin: Error when query ends with a forward slash


cibere:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I am using the latest version of Flow Launcher.

### Problem Description

When using the explorer plugin to look up a path, ending off your query in a forward slash results in an error

### To Reproduce

1. Type `C:\/`
2. See the error result from the explorer plugin

### Screenshots

![Image](https://github.com/user-attachments/assets/9dd3b0e3-a837-4754-81e6-b99fcee73ceb)

### Flow Launcher Version

1.19.5

### Windows Build Number

 Microsoft Windows [Version 10.0.22631.4751]

### Error Log

<details>

```shell

17:55:29.4035-08:00 - ERROR - DirectoryInfoSearch.DirectorySearch - Error occurred while searching path
EXCEPTION OCCURS: System.ArgumentException: Second path fragment must not be a drive or UNC name. (Parameter 'expression')
   at System.IO.Enumeration.FileSystemEnumerableFactory.NormalizeInputs(String& directory, String& expression, MatchType matchType)
   at System.IO.DirectoryInfo.InternalEnumerateInfos(String path, String searchPattern, SearchTarget searchTarget, EnumerationOptions options)
   at Flow.Launcher.Plugin.Explorer.Search.DirectoryInfo.DirectoryInfoSearch.DirectorySearch(EnumerationOptions enumerationOption, String search, String searchCriteria, CancellationToken token) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\DirectoryInfo\DirectoryInfoSearch.cs:line 81
   at Flow.Launcher.Plugin.Explorer.Search.SearchManager.PathSearchAsync(Query query, CancellationToken token) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\SearchManager.cs:line 242
   at Flow.Launcher.Plugin.Explorer.Search.SearchManager.SearchAsync(Query query, CancellationToken token) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Search\SearchManager.cs:line 136
   at Flow.Launcher.Plugin.Explorer.Main.QueryAsync(Query query, CancellationToken token) in C:\projects\flow-launcher\Plugins\Flow.Launcher.Plugin.Explorer\Main.cs:line 88
[...]

PR #3231, opened 2025-02-07
Fix unix directory seperator issue

Jack251970:
Fix issue when user uses forward slash as directory seperator, like C:\Test/Test.

Related issue in #3229.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/Constants.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/Constants.cs
@@ -22,7 +22,6 @@ namespace Flow.Launcher.Plugin.Explorer.Search
         internal const string WindowsIndexErrorImagePath = "Images\\index_error2.png";
         internal const string GeneralSearchErrorImagePath = "Images\\robot_error.png";
 
-
         internal const string ToolTipOpenDirectory = "Ctrl + Enter to open the directory";
 
         internal const string ToolTipOpenContainingFolder = "Ctrl + Enter to open the containing folder";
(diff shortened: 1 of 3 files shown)


Jack251970:
@cibere Does this work for you?

Jack251970:
> > @cibere Does this work for you?
> 
> It indeed works for me, however I'm still able to reproduce the [second error I mentioned in the comments](https://github.com/Flow-Launcher/Flow.Launcher/issues/3229#issuecomment-2641752711). However I'm still unsure how related the two errors are, and if it would be best to split them up

Sure this path `c:\test\test` is valid?

cibere:
> > > @cibere Does this work for you?
> > 
> > 
> > It indeed works for me, however I'm still able to reproduce the [second error I mentioned in the comments](https://github.com/Flow-Launcher/Flow.Launcher/issues/3229#issuecomment-2641752711). However I'm still unsure how related the two errors are, and if it would be best to split them up
> 
> Sure this path `c:\test\test` is valid?

It doesn't exist. However, I feel like flow should treat it the same way it does if you only use backslashes (which is to return no results instead of an error)

image
image

Jack251970:
> > > > @cibere Does this work for you?
> > > 
> > > 
> > > It indeed works for me, however I'm still able to reproduce the [second error I mentioned in the comments](https://github.com/Flow-Launcher/Flow.Launcher/issues/3229#issuecomment-2641752711). However I'm still unsure how related the two errors are, and if it would be best to split them up
> > 
> > 
> > Sure this path `c:\test\test` is valid?
> 
> It doesn't exist. However, I feel like flow should treat it the same way it does if you only use backslashes (which is to return no results instead of an error)
> 
> image image

I see. I should handle this in a more external function. Now it works fine for me.

cibere (approved):
I can't reproduce either of the errors on this

Jack251970:
@taooceros Ready to merge💪

Issue #3185, opened 2025-01-16
BUG: Turning off `Query window shadow effect` breaks scrollbar selection


Sneakpeakcss:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

https://github.com/user-attachments/assets/90ba66a7-9591-4adf-8997-1cc61fd931f3

### To Reproduce

1. Turn off `Query window shadow effect` (i think it's default since latest releases ?)
2. Try to drag scroll

### Flow Launcher Version

1.19.4

### Windows Build Number

Win10 22H2 19045.3693

PR #3232, opened 2025-02-07
Fix main window resize border thickness issue

Jack251970:
# Issue

Because adding drop shadow effect will change the margin of the window border, we need to update the resize thickness of the window chrome to correct set the resize border. Otherwise, scrollbar selection is unable to work because resize area override it.

* Related issue: #3185

# Test

Regardless of whether the shadow effect is enabled, the window border remains outside the resizing area.

https://github.com/user-attachments/assets/5619dccf-d8ae-464d-b036-a10d40a5bc34


taooceros:
@Jack251970 feel free to merge once you are ready.

PR #3240, opened 2025-02-14
Fix ChangeQuery & ReQuery Work Issue When Results Are Not From Query Results

Jack251970:
# Fix ChangeQuery & ReQuery Work Issue When Results Are Not From Query Results

The `ChangeQuery` and `ReQuery` function fail to work when results are from the context menu or history. Now we will back to query results when we call these functions.

# Test

* Call `ChangeQuery` when results are from the context menu or history.
* Call `ReQuery` when results are from the context menu or history.


taooceros:
What would be the current behavior of change query under context menu? Directly change the query text under the context menu search text?

Jack251970:
> What would be the current behavior of change query under context menu? Directly change the query text under the context menu search text?

As described in #3087, this is used to make sure FL will be back to query results so that `ChangeQuery` can work.

As for the current behavior, `ChangeQuery` will fail to work. (Around 0:00:10)

https://github.com/user-attachments/assets/598a606a-b46d-441e-a370-782d2e400484

taooceros:
> > What would be the current behavior of change query under context menu? Directly change the query text under the context menu search text?
> 
> As described in #3087, this is used to make sure FL will be back to query results so that `ChangeQuery` can work.
> 
> As for the current behavior, `ChangeQuery` will fail to work. (Around 0:00:10)
> 
>  2025-02-17.14-45-11.mp4

I feel like ChangeQuery now simply behave weird when not at result. Should we fix that behavior (like make sure it change to Results before changing query or something)?

Jack251970:
> > > What would be the current behavior of change query under context menu? Directly change the query text under the context menu search text?
> > 
> > 
> > As described in #3087, this is used to make sure FL will be back to query results so that `ChangeQuery` can work.
> > As for the current behavior, `ChangeQuery` will fail to work. (Around 0:00:10)
> > 2025-02-17.14-45-11.mp4
> 
> I feel like ChangeQuery now simply behave weird when not at result. Should we fix that behavior (like make sure it change to Results before changing query or something)?

From my perspective, I believe we do not need to fix that behavior because, in most scenarios, `ChangeQuery` or `Requery` operates on query results. This is why I created that PR without modifying the `ChangeQuery` and `Requery` functions... Perhaps you could ask @jjw24 about his opinion. (He approves that PR.)

taooceros:
I think if change query works fluently on context menu (i.e. change the query text for context menu) then we can skip that. If it will make the behavior of flow weird then we should fix the change query directly.

Backtoqueryresult can have other usage.

Jack251970:
> I think if change query works fluently on context menu (i.e. change the query text for context menu) then we can skip that. If it will make the behavior of flow weird then we should fix the change query directly.
> 
> Backtoqueryresult can have other usage.

Well, I will do that.

taooceros:
I will take a look on how changequery work. We can proceed this first.

Flow.Launcher/ViewModel/PluginViewModel.cs
```diff
@@ -146,6 +146,7 @@ private void OpenSourceCodeLink()
         [RelayCommand]
         private void OpenDeletePluginWindow()
         {
+            PluginManager.API.BackToQueryResults();
```
taooceros, line 149: Should we make it consistent to App.API? Similar to the following.

Jack251970: I use PluginManager.API here because OpenDeletePluginWindow use it to get the api instance.

jjw24:
> @jjw24 Hello, what are your thoughts on the API function `BackToQueryResults`? Should we require devs to call this method before `ChangeQuery` and `ReQuery`, or should we integrate it directly into `ChangeQuery` & `ReQuery`?

Both of you are correct. I would think though have change query and requery methods call it under the hood just makes it a little easier and more convenient for plugin devs when they use those two methods. Sorry I missed this part during the PR 3087 review.

Jack251970:
@taooceros Following jjw24's comments, now I remove #3087. When we call `ChangeQuery` & `ReQuery`, we will firstly back to query results when we call these functions. Already tested in desc.

jjw24:
Sorry, I don't mean remove the back to query result API though, just have requery and change query call it before the rest of the code.

Jack251970:
> Sorry, I don't mean remove the back to query result API though, just have requery and change query call it before the rest of the code.

Oh, sorry for misunderstanding.

Jack251970:
@jjw24 Ready for review.💪

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -17,7 +17,8 @@ namespace Flow.Launcher.Plugin
     public interface IPublicAPI
     {
         /// <summary>
-        /// Change Flow.Launcher query
+        /// Change Flow.Launcher query.
+        /// When current results are from context menu or history, we will back to query results before changing query.
```
jjw24, line 21:
```suggestion
        /// When current results are from context menu or history, it will back to query results before changing query.
```

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -299,7 +300,7 @@ public interface IPublicAPI
 
         /// <summary>
         /// Reloads the query.
-        /// This method should run when selected item is from query results.
+        /// When current results are from context menu or history, we will back to query results before requerying.
```
jjw24, line 303:
```suggestion
        /// When current results are from context menu or history, it will go back to query results before requerying.
```

jjw24 (approved):
👍

PR #3246, opened 2025-02-16
Fix Exception When Getting Thumbnails

Jack251970:
# Fix issue in getting thumbnails function

For some files in the recycle bin, we may encounter issue like this:
Screenshot 2025-02-16 134936

This can cause:
Screenshot 2025-02-16 223918

# Test
Screenshot 2025-02-16 224108

### DIFF
--- a/Flow.Launcher.Infrastructure/Image/ThumbnailReader.cs
+++ b/Flow.Launcher.Infrastructure/Image/ThumbnailReader.cs
@@ -31,6 +31,8 @@ namespace Flow.Launcher.Infrastructure.Image
 
         private static readonly Guid GUID_IShellItem = typeof(IShellItem).GUID;
 
+        private static readonly HRESULT S_ExtractionFailed = (HRESULT)0x8004B200;
+
         public static BitmapSource GetThumbnail(string fileName, int width, int height, ThumbnailOptions options)
         {
             HBITMAP hBitmap = GetHBitmap(Path.GetFullPath(fileName), width, height, options);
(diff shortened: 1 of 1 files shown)


Flow.Launcher.Infrastructure/Image/ThumbnailReader.cs
```diff
@@ -79,9 +77,14 @@ private static unsafe HBITMAP GetHBitmap(string fileName, int width, int height,
                 {
                     imageFactory.GetImage(size, (SIIGBF)options, &hBitmap);
                 }
-                catch (COMException ex) when (ex.HResult == S_ExtractionFailed && options == ThumbnailOptions.ThumbnailOnly)
+                catch (COMException)
```
taooceros, line 80: I somehow feel like you can merge this two branch with some pattern matching? like `catch (Exception ex) when (ex is COMException or FileNotFoundException)`

taooceros (changes requested):

Flow.Launcher.Infrastructure/Image/ThumbnailReader.cs
Jack251970: > I somehow feel like you can merge this two branch with some pattern matching? like `catch (Exception ex) when (ex is COMException or FileNotFoundException)`

I can also catch some ink files with COMException when getting thumbnail for them, so I also change COMException here. (I forget how to reproduce it...)

Jack251970: > I somehow feel like you can merge this two branch with some pattern matching? like `catch (Exception ex) when (ex is COMException or FileNotFoundException)`

Now I revert these codes.

taooceros (approved):

PR #3248, opened 2025-02-17
Fix Custom Query Hotkey / Shortcuts Preview Issue

Jack251970:
# Fix hotkey preview issue for custom query hotkeys and custom query shortcuts

In `CustomQueryHotkeySetting.cs` & `CustomShortcutSetting.cs`, original codes are like these:
```
Application.Current.MainWindow.Show();
Application.Current.MainWindow.Opacity = 1;
```

Here we will not call the correct `Show` function for main window.
So we need to change to:
```
_mainViewModel.Show();
```

And it can help us to address preview issue that main window will not close when it is unfocused.

# Test

Original:

https://github.com/user-attachments/assets/884b2d54-e33c-464b-a77e-7e4d09893383

After:

https://github.com/user-attachments/assets/9dc8491c-7b43-4a24-87bf-20c91c736164

### DIFF
--- a/Flow.Launcher/CustomQueryHotkeySetting.xaml.cs
+++ b/Flow.Launcher/CustomQueryHotkeySetting.xaml.cs
@@ -7,20 +7,23 @@ using System.Windows;
 using System.Windows.Input;
 using System.Windows.Controls;
 using Flow.Launcher.Core;
+using Flow.Launcher.ViewModel;
 
 namespace Flow.Launcher
 {
     public partial class CustomQueryHotkeySetting : Window
     {
         private SettingWindow _settingWidow;
+        private readonly Settings _settings;
+        private readonly MainViewModel _mainViewModel;
         private bool update;
         private CustomPluginHotkey updateCustomHotkey;
-        public Settings Settings { get; }
 
-        public CustomQueryHotkeySetting(SettingWindow settingWidow, Settings settings)
+        public CustomQueryHotkeySetting(SettingWindow settingWidow, Settings settings, MainViewModel mainVM)
         {
             _settingWidow = settingWidow;
-            Settings = settings;
+            _settings = settings;
+            _mainViewModel = mainVM;
             InitializeComponent();
         }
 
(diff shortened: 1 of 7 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1368,7 +1368,8 @@ public void Show()
                 MainWindowOpacity = 1;
 
                 MainWindowVisibilityStatus = true;
-                VisibilityChanged?.Invoke(this, new VisibilityChangedEventArgs { IsVisible = true });
+                if (invokeEvent)
```
taooceros, line 1371: why is this important?

Jack251970: The original codes does not invoke VisibliityChanged event, so I keep it here.

Jack251970: @taooceros If you think we should invoke this event here, feel free to tell me and I will do it.

taooceros: I am not sure. I don't recall much things registered in this event. I just want to check as you introduce an api change to the method, which is not yet sure important.

Jack251970: > I am not sure. I don't recall much things registered in this event. I just want to check as you introduce an api change to the method, which is not yet sure important.

Well, I think we should invoke visibility changed event here and I change the related codes.

Jack251970:
So sorry that I forget there is an api function for showing main window😢

taooceros:
> So sorry that I forget there is an api function for showing main window😢

Haha no worries 😝. You are already doing really good!

Flow.Launcher/CustomQueryHotkeySetting.xaml.cs
```diff
@@ -13,14 +13,14 @@ namespace Flow.Launcher
     public partial class CustomQueryHotkeySetting : Window
     {
         private SettingWindow _settingWidow;
+        private readonly Settings _settings;
```
taooceros, line 16: any reason to change this?

taooceros: Technically this is a field instead of property (maybe it worth to figure out the difference). For here, I think it's fine.

taooceros (approved):

PR #3250, opened 2025-02-18
Improve Shutdown & Restart & Hibernate in Sys Plugin

Jack251970:
# Improve Shutdown & Restart & Hibernate in Sys Plugin

Replace them with PInvoke function which can slightly improve execution speed.

# Test

* Shutdown & Restart & Restart With Advanced Boot Options & Hibernate action

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Sys/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Sys/Main.cs
@@ -120,10 +120,9 @@ namespace Flow.Launcher.Plugin.Sys
                             context.API.GetTranslation("flowlauncher_plugin_sys_dlgtext_shutdown_computer"),
                             context.API.GetTranslation("flowlauncher_plugin_sys_shutdown_computer"),
                             MessageBoxButton.YesNo, MessageBoxImage.Warning);
+
                         if (result == MessageBoxResult.Yes)
-                        {
                             Process.Start("shutdown", "/s /t 0");
-                        }
 
                         return true;
                     }
@@ -140,10 +139,9 @@ namespace Flow.Launcher.Plugin.Sys
                             context.API.GetTranslation("flowlauncher_plugin_sys_dlgtext_restart_computer"),
                             context.API.GetTranslation("flowlauncher_plugin_sys_restart_computer"),
                             MessageBoxButton.YesNo, MessageBoxImage.Warning);
+
                         if (result == MessageBoxResult.Yes)
-                        {
                             Process.Start("shutdown", "/r /t 0");
-                        }
 
                         return true;
                     }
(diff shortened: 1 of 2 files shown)


taooceros:
Do you know what would happen if you use `EWX_SHUTDOWN` or `EWX_HYBRID_SHUTDOWN`?

Jack251970:
> Do you know what would happen if you use `EWX_SHUTDOWN` or `EWX_HYBRID_SHUTDOWN` instead of poweroff?

If you use the former, the system in a state where other processes (e.g., firmware) can intervene before power-off and it may leave the system waiting for a manual power-off. Considering that user click to shutdown the computer, I think FL should not wait.

If you use the later, the system will performs a hybrid shutdown (Fast Startup). Instead of fully shutting down, Windows hibernates the kernel and driver state. I think FL should not use this mode.

taooceros:
> > Do you know what would happen if you use `EWX_SHUTDOWN` or `EWX_HYBRID_SHUTDOWN` instead of poweroff?
> 
> If you use the former, the system in a state where other processes (e.g., firmware) can intervene before power-off and it may leave the system waiting for a manual power-off. Considering that user click to shutdown the computer, I think FL should not wait.
> 
> If you use the later, the system will performs a hybrid shutdown (Fast Startup). Instead of fully shutting down, Windows hibernates the kernel and driver state. I think FL should not use this mode.

I am actually pretty curious whether we should have a way to use the hybrid shutdown. It is enabled by default for many computer I think, and for many people they wouldn't need a full shutdown but a fast startup would help. If they want to refresh some hardware state, most likely they will use restart instead of shutdown. I am not sure whether the current shutdown uses fast startup or the full shutdown. Maybe it worth to check.

Jack251970:
> > > Do you know what would happen if you use `EWX_SHUTDOWN` or `EWX_HYBRID_SHUTDOWN` instead of poweroff?
> > 
> > 
> > If you use the former, the system in a state where other processes (e.g., firmware) can intervene before power-off and it may leave the system waiting for a manual power-off. Considering that user click to shutdown the computer, I think FL should not wait.
> > If you use the later, the system will performs a hybrid shutdown (Fast Startup). Instead of fully shutting down, Windows hibernates the kernel and driver state. I think FL should not use this mode.
> 
> I am actually pretty curious whether we should have a way to use the hybrid shutdown. It is enabled by default for many computer I think, and for many people they wouldn't need a full shutdown but a fast startup would help. If they want to refresh some hardware state, most likely they will use restart instead of shutdown. I am not sure whether the current shutdown uses fast startup or the full shutdown. Maybe it worth to check.

The current codes `Process.Start("shutdown", "/s /t 0");` use the full shutdown so I use `EWX_SHUTDOWN | EWX_POWEROFF` to keep their behaviour consistent.

taooceros:
Sounds good, let's keep it for now. We can have the fast startup sometimes later.

taooceros (approved):

PR #3257, opened 2025-02-20
Fix Issue that Plugin Cannot Cache Records

Jack251970:
# Use null as default value for record key

* Because old data does not have this key, we should use null as the default value for consistency.

# Deep clone result list before updating results

* Because FL will change the score property of the results during results update, so before we update the results, we need to deep clone the result list, so that plugin can cache them for performance improvement (like Flow.Launcher.Plugin.Program).

This can fix issue in #3203. Tested.

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -231,8 +231,8 @@ namespace Flow.Launcher.ViewModel
 
                     var token = e.Token == default ? _updateToken : e.Token;
 
-                    // make a copy of results to avoid plugin change the result when updating view model
-                    var resultsCopy = e.Results.ToList();
+                    // make a clone to avoid possible issue that plugin will also change the list and items when updating view model
+                    var resultsCopy = DeepCloneResults(e.Results, token);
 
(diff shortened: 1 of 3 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1469,9 +1494,9 @@ public void UpdateResultView(ICollection<ResultsForUpdate> resultsForUpdates)
                 {
                     if (_topMostRecord.IsTopMost(result))
                     {
-                        result.Score = Result.MaxScore;
+                        result.Score = 100000; //Result.MaxScore;
```
jjw24, line 1497: Do we need to hardcode this or keep as variable?

Jack251970: Sorry, I forget to change this Debug code.

Jack251970: Thanks! I have reverted this.

Jack251970:
@jjw24 I have done my work, please check.

jjw24 (approved):

PR #3259, opened 2025-02-20
Support removing duplicated programs

Jack251970:
# Support removing duplicated Win32 apps in Program plugin

For Program plugin, we can get results with duplicated apps. (One is Win32 app, another is UWP app.)

Screenshot 2025-02-20 145447

So I add option to let us remove the duplicated apps.

I remove the Win32 apps instead of UWP apps because the name, description and some other properties of Win32 apps are always wrong if these apps are duplicated.

As you can see above, the top item is from Win32 app and its name is `olk` but actually this name should be `Outlook (Classic)` (same as bottom item which is from UWP app).

# Test

Settings Panel:

image

Search:

image

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Main.cs
@@ -72,6 +72,8 @@ namespace Flow.Launcher.Plugin.Program
         private const string ExeUninstallerSuffix = ".exe";
         private const string InkUninstallerSuffix = ".lnk";
 
+        private const string WindowsAppPath = "c:\\program files\\windowsapps";
+
         static Main()
         {
         }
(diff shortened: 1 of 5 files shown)


Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -152,6 +163,23 @@ private bool HideUninstallersFilter(IProgram program)
             return true;
         }
 
+        private static bool HideDuplicatedWindowsAppFilter(IProgram program, string[] uwpsDirectories)
```
jjw24, line 166: Can this be a different name? Maybe ShowWindowsApp?

Plugins/Flow.Launcher.Plugin.Program/Views/ProgramSetting.xaml.cs
```diff
@@ -57,6 +57,16 @@ public bool HideUninstallers
             }
         }
 
+        public bool HideDuplicatedWindowsApp
+        {
+            get => _settings.HideDuplicatedWindowsApp;
+            set
+            {
+                Main.ResetCache();
```
jjw24, line 65: why we are resetting cache?

Plugins/Flow.Launcher.Plugin.Program/Main.cs
Jack251970: Sorry, why we need other name? The option is Hide duplicated app..?

Plugins/Flow.Launcher.Plugin.Program/Views/ProgramSetting.xaml.cs
Jack251970: The same reason as Hide Uninstallers. If we search Outlook, the results contains duplicated results and we need to reset cache to make sure next time it will remove duplicated apps.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
Jack251970: I just follow the style as Hide Uninstallers option.

jjw24: hmm the naming is a bit hard to follow at first, yeah I kinda get it now. Nvm all good.

jjw24 (approved):

PR #3261, opened 2025-02-21
Support Multiple keywords for Plugins

Jack251970:
# Support multiple keywords for plugins

## Support user to set multiple keywords for plugins

I change the action keyword separator from `;` to ` ` because ` ` is the character we used to split terms so we should not let user include that character in one action keyword so we can use it as separator which can help us to fix one possible issue that action keyword with ` ` cannot work (I find FL does not notify users about this).

## Support plugin to not allow user change its action keyword

I find `WebSearch` and `Explorer` plugin need this because these plugins let user change its keywords from settings panel.

## Fix issue that plugin cannot update ActionKeyword after changing its keywords

Check https://github.com/Flow-Launcher/Flow.Launcher/commit/d9ba38b3221174996201f8df2b1a93a9000e00e5

# Test

https://github.com/user-attachments/assets/22fce4a7-3d6d-45b7-83f7-53397fa5ed73


taooceros:
Does this mean the explorer plugin (and the web search plugin) will demonstrate the actionkeyword assign panel? They should not display the control for user to change the actionkeyword.

Jack251970:
> Does this mean the explorer plugin (and the web search plugin) will demonstrate the actionkeyword assign panel? They should not display the control for user to change the actionkeyword.

Yes, if one plugin can change its action keyword in its settings panel, it can add `"AllowModifyActionKeywords": false,` in its `plugin.json` so that we will not show action keyword panel.

Jack251970:
To make sure one plugin can manage its action keywords fully, I add new api function `ReplaceActionKeyword`. Example: `WebSearch` plugin.

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -341,6 +340,24 @@ public static List<Result> GetContextMenusForPlugin(Result result)
             return results;
         }
 
+        public static bool ActionKeywordRegistered(IReadOnlyList<string> newActionKeywords, IReadOnlyList<string> oldActionKeywords)
```
taooceros, line 343: This method does not make sense to me. If user want to exclude the current action keywords, they could check it themselves, rather than passing it as a parameter in this method.

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -386,16 +404,63 @@ public static void RemoveActionKeyword(string id, string oldActionkeyword)
...
+                    plugin.Metadata.ActionKeyword = newActionKeywords[0];
+                }
+                else
+                {
+                    plugin.Metadata.ActionKeyword = string.Empty;
+                }
+            }
+        }
+
+        private static bool CheckActionKeywordChanged(IReadOnlyList<string> oldActionKeyword, IReadOnlyList<string> newActionKeyword)
```
taooceros, line 439: Also I don't really understand this method. Does it simply check whether the two list is equal?

Jack251970: Yes.

Flow.Launcher.Core/Plugin/PluginManager.cs
Jack251970: Check https://github.com/Flow-Launcher/Flow.Launcher/pull/3261/commits/34cb078602f3c04ff428123415ac55e97e9b9daf please.

Flow.Launcher/ActionKeywords.xaml.cs
```diff
@@ -53,5 +54,24 @@ private void btnDone_OnClick(object sender, RoutedEventArgs _)
                 MessageBoxEx.Show(msg);
             }
         }
+
+        private static bool ActionKeywordRegistered(IReadOnlyList<string> newActionKeywords, IReadOnlyList<string> oldActionKeywords)
```
taooceros, line 58: why not just exclude old action keywords when checking availability?

Jack251970: Sorry, you mean first exclude old action keywords from new action keywords? I think this function works fine.

Jack251970: I am a litle confused about your meaning? Could you please explain it further?

Flow.Launcher.Core/Plugin/PluginManager.cs
taooceros: well I would say we should check without order.

Flow.Launcher/ActionKeywords.xaml.cs
taooceros: > Sorry, you mean first exclude old action keywords from new action keywords? I think this function works fine.

yeah just curious. Although it's the same idea, but this makes me a little bit confusing when checking it. On the other hand, this method requires the order being the same for excluding, which I don't think should be what's happening.

taooceros: https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.except?view=net-9.0

Probably using this would be better than implement one ourselves.

taooceros: I will do this

```
if (!newActionKeywords.Except(oldActionKeywords).Any(PluginManager.ActionKeywordRegistered))
            {
                pluginViewModel.ChangeActionKeyword(newActionKeywords, oldActionKeywords);
                Close();
            }
```

taooceros:
As it introduces an api change, @Flow-Launcher/team how do you think about the new change?

Flow.Launcher/ActionKeywords.xaml.cs
Jack251970: Could you please suggest these changes? I can merge them now.

cibere:
> As it introduces an api change, @Flow-Launcher/team how do you think about the new change?

I think it's a fine api change, however this seems like an api method that can also be added to the json-rpc v2 api interface

Jack251970:
> As it introduces an api change, @Flow-Launcher/team how do you think about the new change?

@taooceros Sorry, I find that it seems to be a unnecessary API function since it can be implemented by `AddActionKeyword` & `RemoveActionKeyword`. Now I remove it and please check.

Jack251970:
> > As it introduces an api change, @Flow-Launcher/team how do you think about the new change?
> 
> I think it's a fine api change, however this seems like an api method that can also be added to the json-rpc v2 api interface

Yeah, this API function makes sense, but it seems to be a unnecessary API function since it can be implemented by `AddActionKeyword` & `RemoveActionKeyword`, so I would like to remove it.😔

taooceros:
I am also thinking about whether we should add a reference count for a specific action keyword, which could support multiple plugins shares the same action keyword? Probably a follow up for #3151. Though that's a different story.

Jack251970:
> I am also thinking about whether we should add a reference count for a specific action keyword, which could support multiple plugins shares the same action keyword? Probably a follow up for #3151. Though that's a different story.

Emm... Any issue requests that?

taooceros:
Don't think so. Just some thoughts.

jjw24 (commented):
The option to include plugin results in global search, is it possible to just have a toggle for that like PT Run? I feel like that's a better UI/UX implementation than to add a the global actionkeyword after a space.

Only thing I am not sure about is JSON-RPC plugins, do they have a default plugin panel initialised? If they do, can we also add the toggle in by default?

taooceros:
> Only thing I am not sure about is JSON-RPC plugins, do they have a default plugin panel initialised? If they do, can we also add the toggle in by default?

could you explained?

Jack251970:
> The option to include plugin results in global search, is it possible to just have a toggle for that like PT Run? I feel like that's a better UI/UX implementation than to add a the global actionkeyword after a space.
> 
> Only thing I am not sure about is JSON-RPC plugins, do they have a default plugin panel initialised? If they do, can we also add the toggle in by default?

@jjw24 Do you think we should change to something like Run's direct action keyword and a global search toggle button? I think it can cause conflicts with users' habits, since previously FL uses `*` to let user set a global search plugins, and users should prefer to use action keyword to set global search.

Jack251970:
> The option to include plugin results in global search, is it possible to just have a toggle for that like PT Run? I feel like that's a better UI/UX implementation than to add a the global actionkeyword after a space.
> 
> Only thing I am not sure about is JSON-RPC plugins, do they have a default plugin panel initialised? If they do, can we also add the toggle in by default?


Moreover, I believe FL does not have a strictly defined global search state, since plugins like WebSearch and Explorer may have multiple global search identifiers.

onesounds:
If there is no bug, let's just do it and make another PR for UI/UX and deal with it.

jjw24:
Yeah ok in this case let's keep how you have done it @Jack251970.

jjw24:
Also what's the reason for disallowing the WebSearch and Explorer action keywords been changed from settings panel?

Jack251970:
> Yeah ok in this case let's keep how you have done it @Jack251970.

Well, we can do this UI improvement in the future.

Jack251970:
> Also what's the reason for disallowing the WebSearch and Explorer action keywords been changed from settings panel?

The main reason is that those plugins can let users edit their action keywords from their settings controls, so action keyword panels seem to be redundant.

image

Also, I follow the original logic that plugins with multiple keywords do no show the action keyword panel and those plugins are all plugins with multiple action keywords by default.

Additionally, if we want to enable action keyword panels for those plugins, we need to add more like `ActionKeywordsChanged` callback to update their settings controls when we change their keywords from action keyword panels.

Flow.Launcher/PublicAPIInstance.cs
```diff
@@ -26,6 +26,7 @@
 using System.Diagnostics;
 using System.Collections.Specialized;
 using Flow.Launcher.Core;
+using Flow.Launcher.Infrastructure.UserSettings;
```
VictoriousRaptor, line 29: Is it accidental?

Jack251970: Yeah, have reverted

jjw24:
I don't think there is a need to allow users to assign more than one action keyword for a single plugin action unless the plugin like Explorer and WebSearches have different actions and specifically support it. It's just confusing.
image

If this is to allow results to also be included in global search, can we not achieve this via a plugin settings property `IncludeInGlobalSearch` and if the plugin has this true/false then add/remove from `GlobalPlugins`?

Jack251970:
> I don't think there is a need to allow users to assign more than one action keyword for a single plugin action unless the plugin like Explorer and WebSearches have different actions and specifically support it. It's just confusing. image
> 
> If this is to allow results to also be included in global search, can we not achieve this via a plugin settings property `IncludeInGlobalSearch` and if the plugin has this true/false then add/remove from `GlobalPlugins`?

I don't think this will cause confusion because it's all set by the user. (Users still can only set one action keyword)

Jack251970:
> If this is to allow results to also be included in global search, can we not achieve this via a plugin settings property `IncludeInGlobalSearch` and if the plugin has this true/false then add/remove from `GlobalPlugins`?

Yeah, we can do that and I want to do that in another PR. (I need onesounds's help with UI desgin)

jjw24:
> > If this is to allow results to also be included in global search, can we not achieve this via a plugin settings property `IncludeInGlobalSearch` and if the plugin has this true/false then add/remove from `GlobalPlugins`?
> 
> Yeah, we can do that and I want to do that in another PR. (I need onesounds's help with UI desgin)

Ok sounds good.

jjw24:
I don't think `AllowModifyActionKeywords` is working? As you can see in my screenshot Explorer plugin still shows the change action keyword box.

Also maybe call it `HideActionKeywordPanel` or something because that name suggests no longer allowing modifying action keywords.

Plugins/Flow.Launcher.Plugin.Calculator/Flow.Launcher.Plugin.Calculator.csproj
```diff
@@ -43,6 +43,8 @@
   
   <ItemGroup>
     <ProjectReference Include="..\..\Flow.Launcher.Core\Flow.Launcher.Core.csproj" />
+    <ProjectReference Include="..\..\Flow.Launcher.Infrastructure\Flow.Launcher.Infrastructure.csproj" />
```
jjw24, line 46: Don't include this, infra project is low level and should not be included in plugins. Expose anything needed via API.

We also want the default plugins to be treated like other plugins, so only the plugin project should be included, and eventually we should completely remove the default plugins' dependence on Infrastructure and Core projects.

Jack251970: I found Calculator plugin has used classes in Infrastructure namespace. So I add this project reference here to make consistent with other plugins.

Jack251970:
> I don't think `AllowModifyActionKeywords` is working? As you can see in my screenshot Explorer plugin still shows the change action keyword box.
> 
> Also maybe call it `HideActionKeywordPanel` or something because that name suggests no longer allowing modifying action keywords.

It works fine for me. I guess you have not changed `plugin.json` in plugin directories?

image

Plugins/Flow.Launcher.Plugin.Calculator/Flow.Launcher.Plugin.Calculator.csproj
jjw24: The other plugins will have infrastructure & core project reference removed. If it's not used, no need to add here.

Jack251970: Sure!

Jack251970:
@jjw24 Hi, is this working as expected?

PR #3265, opened 2025-02-21
Unify Settings Panel Design

Jack251970:
Unify desgin in `InstalledPluginDisplayKeyword.xaml.cs`, `JsonRPCPluginSettings.cs` and settings panel xaml files of all plugins.


Jack251970:
@onesounds Hi, I’d like to get your idea on the design for the settings panel. Which option do you think is better, or do you feel they both work well?

Left aligned version:

Screenshot 2025-02-21 222837

Right aligned version:

Screenshot 2025-02-21 222837 - Copy

Jack251970:
> @onesounds Hi, I’d like to get your idea on the design for the settings panel. Which option do you think is better, or do you feel they both work well?
> 
> Left aligned version:
> 
> Screenshot 2025-02-21 222837
> 
> Right aligned version:
> 
> Screenshot 2025-02-21 222837 - Copy

@Flow-Launcher/team Hello everyone, I’d like to get your feedback on this design. What are your thoughts?

cibere:
> @Flow-Launcher/team Hello everyone, I’d like to get your feedback on this design. What are your thoughts?

I definitely like the left align more.

With the right align, it seems unnecessary to add the separation and make it a little harder to match the label to the input. *for me at least, it takes me a little bit longer to match the label to the inputs on the right align vs the left align*

onesounds:
For the check box type, left alignment is recommended, and for others, right alignment is recommended. (I referred to powertoy run and Windows 11's settings window as a reference.) I've tried to unify this a few times, but it wasn't easy because there were a lot of plug-ins with complicated settings. Anyway, the basic design recommends the Power Toy layout.

Jack251970:
> > @Flow-Launcher/team Hello everyone, I’d like to get your feedback on this design. What are your thoughts?
> 
> I definitely like the left align more.
> 
> With the right align, it seems unnecessary to add the separation and make it a little harder to match the label to the input. _for me at least, it takes me a little bit longer to match the label to the inputs on the right align vs the left align_

Thanks for your reply!❤

Jack251970:
> For the check box type, left alignment is recommended, and for others, right alignment is recommended. (I referred to powertoy run and Windows 11's settings window as a reference.) I've tried to unify this a few times, but it wasn't easy because there were a lot of plug-ins with complicated settings. Anyway, the basic design recommends the Power Toy layout.

Thanks for your reply!❤ I wonder why for the check box type, left alignment is recommended?

image

As you can see in Windows Settings, it is also right-aligned. 😢

Jack251970:
> For the check box type, left alignment is recommended, and for others, right alignment is recommended. (I referred to powertoy run and Windows 11's settings window as a reference.) I've tried to unify this a few times, but it wasn't easy because there were a lot of plug-ins with complicated settings. Anyway, the basic design recommends the Power Toy layout.

NVM. I confused Toggle Button and Check Box.

VictoriousRaptor:
> Thanks for your reply!❤ I wonder why for the check box type, left alignment is recommended?

Just some personal idea:

Right alignment (or justified alignemnt if you consider labels and controls as a whole line) is harder to match label and control as cibere mentioned. but windows 11 in your example image has a box containing the label and control so it's not a big deal. As in our program we don't have boxes so left alignment is better IMO. 

Here's an example of windows 11 using left alignment.

image

Jack251970:
@Flow-Launcher/team Hi everyone! I have created a unified settings panel design for both preinstalled and JSON-RPC plugins. Could you spare some time to review this update?

You just need to install the installer [here](https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/51613775/artifacts). This installer is quick to install and fully compatible with existing Flow Launcher installations.

If you do not have JSON-RPC plugins, you can use [my demo plugin](https://github.com/user-attachments/files/19041646/Hello.World.Python-1.0.0.zip). You need to download and extract the files into `%APPDATA%\FlowLauncher\Plugins`. Then restart FL to see the changes.

Thank you for your time, and I would greatly appreciate your thoughts! 😊

cibere:
> If you do not have JSON-RPC plugins, you can use [my demo plugin](https://github.com/user-attachments/files/19041646/Hello.World.Python-1.0.0.zip). You need to download and extract the files into `%APPDATA%\FlowLauncher\Plugins`. Then restart FL to see the changes.

btw you can just run the following command with the package manager to install it:
```
pm install https://github.com/user-attachments/files/19041646/Hello.World.Python-1.0.0.zip
```

onesounds (dismissed):

Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
```diff
@@ -23,26 +19,44 @@ public class JsonRPCPluginSettings
...
+                    {
+                        Settings[key] = jsonElement.ValueKind switch
+                        {
+                            JsonValueKind.String => jsonElement.GetString() ?? value,
+                            JsonValueKind.True => jsonElement.GetBoolean(),
+                            JsonValueKind.False => jsonElement.GetBoolean(),
+                            JsonValueKind.Null => null,
+                            _ => value
+                        };
+                    }
```
Yusyuriv, line 57: I would really like to avoid as much nesting as possible, so could you please invert the conditional?

```suggestion
                    if (value is not JsonElement jsonElement) continue;

                    Settings[key] = jsonElement.ValueKind switch
                    {
                        JsonValueKind.String => jsonElement.GetString() ?? value,
                        JsonValueKind.True => jsonElement.GetBoolean(),
                        JsonValueKind.False => jsonElement.GetBoolean(),
                        JsonValueKind.Null => null,
                        _ => value
                    };
```

Yusyuriv (dismissed):

Yusyuriv (approved):

PR #3267, opened 2025-02-22
Improve Progress Box Design

Jack251970:
# Add minimize & background button for progress box & Set progress box topmost

* Minimize button is for users to temporarily hide progress box for this progress when they do not want to check the progress for now. They can view progress if they want from taskbar.

* Background button is for users to permanently hide progress for this progress when they do not care the progress. This will **not** cancel the progress. (It is different from cancel button & close button which is used to cancel the progress.)

* Set progress box topmost since we support to temporarily and permanently hide this window now.

# Test

Original:
Screenshot 2025-02-22 190132

New:
Screenshot 2025-02-22 190249


Jack251970:
@onesounds Sorry for the ping. What do you think of this design?

onesounds:
> @onesounds Sorry for the ping. What do you think of this design?

 Good.

jjw24 (approved):

PR #3272, opened 2025-02-24
Improve Plugin Metadata & Path Management

Jack251970:
# Improve Plugin Metadata & Path Management

### (1)`AssemblyName`

For .Net plugins, we do not need to use reflection to get its assembly name.

### (2)`PluginSettingsDirectoryPath` & (3)`PluginCacheDirectoryPath`

For csharp & JsonRPC v2 plugins, FL can provide them with these directories so that they can store their temporary and permeant data in that place.
   * PluginSettingsDirectoryPath is the path to the plugin settings directory which is used to store plugin settings files and data files. When plugin is deleted, FL will ask users whether to keep its settings. If users do not want to keep, this directory will be deleted.
   * PluginCacheDirectoryPath is the path to the plugin cache directory which is used to store cache files. When plugin is deleted, this directory will be deleted as well.

## 2. Improve path management

Add many fields in `DataLocation.cs` so that we do not need to use `Path.Combine` to construct the paths.

## 3. Improve `Flow.Launcher.Plugin` documents

Add docuements in `Flow.Launcher.Plugin` project

## 4. Fix Settings & Cache Directory Removing Issue

Dispose plugin when we need to delete plugin Settings or Cache Directory.

# Test

* Loaded plugin metadata is correct.
* Delete dotnet & non-dotnet plugins will clean their cache directory and if selected to remove their settings, settings directories will be removed.
* After updating to new version, `Program` can move old cache files to its `PluginCacheDirectoryPath`.
* For those plugins which locks files until disposing them, settings & cache directories can be deleted.

# Future
* Provide manual `Cache` clean feature.


Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -546,54 +562,41 @@ internal static void UninstallPlugin(PluginMetadata plugin, bool removePluginFro
 
             if (removePluginSettings)
             {
-                if (AllowedLanguage.IsDotNet(plugin.Language))  // for the plugin in .NET, we can use assembly loader
+                // For dotnet plugins, we need to remove their PluginJsonStorage instance
+                if (AllowedLanguage.IsDotNet(plugin.Language))
                 {
-                    var assemblyLoader = new PluginAssemblyLoader(plugin.ExecuteFilePath);
-                    var assembly = assemblyLoader.LoadAssemblyAndDependencies();
```
taooceros, line 552: why could these be deleted? Isn't this is where we load plugin?

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -188,9 +191,61 @@ public async Task InitAsync(PluginInitContext context)
...
+                        return;
+                    }
+
+                    if (File.Exists(destinationPath))
+                    {
+                        try
+                        {
+                            File.Delete(sourcePath);
+                        }
+                        catch (Exception)
```
taooceros, line 209: when will this happen?

Flow.Launcher.Core/Plugin/PluginManager.cs
Jack251970: oh, this is not for loading the plugin. this is for getting plugin assembly name to delete its settings directory

Jack251970: so now I use the AssemblyName initialized in metadata to avoid reflection

Plugins/Flow.Launcher.Plugin.Program/Main.cs
Jack251970: the old cache files are not cleaned. next time it wil retry it again

Flow.Launcher.Core/Plugin/PluginManager.cs
taooceros: ah I see, this is very similar to the load plugin code so I get confused

Flow.Launcher.Plugin/PluginMetadata.cs
```diff
@@ -30,28 +83,69 @@ internal set
...
         [JsonIgnore]
         public int QueryCount { get; set; }
+
+        /// <summary>
+        /// The path to the plugin settings directory which is not validated.
+        /// It is used to store plugin settings files and data files.
+        /// When plugin is deleted, FL will ask users whether to keep its settings.
+        /// If users do not want to keep, this directory will be deleted.
+        /// </summary>
+        [JsonIgnore]
```
taooceros, line 131: I think we can not ignore these, as that would help jsonrpc plugin (V2) to retrive these information.

However, I am not sure whether we save it to somewhere.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
taooceros: no I mean why the delete will fail. If it fail shouldn't next time it will also failed?

Jack251970: > no I mean why the delete will fail.

it will fail if user happens to open this file. 

>  If it fail shouldn't next time it will also failed?

not exactly. if last time user happens to open this file, it will success next time if user closes this file. no matter how many times it has failed before, if it succeeds once, the file will be deleted permanently since we will never create file in the old path.

Flow.Launcher.Plugin/PluginMetadata.cs
Jack251970: sorry, I do not think so. from my view, this data should be runtime only, and need to be initialized every time we start FL. perhaps in the future, we will change the plugin settings path to other place and we should not use the original paths

Jack251970: Oh, I know your meaning. I find that JsonRPC plugins cannot get metadata instance which means they cannot get these paths.

Could we pass this parameter to the plugins in function like `InitAsync` (Dotnet plugins has this but JsonRPC not, why?)

Additionally, I have changed the codes to make sure metadata of these JsonRPC plugins is correct before creating `JsonRPCPlugin` instance.

Jack251970: Similar to https://github.com/Flow-Launcher/Flow.Launcher/blob/dbf51a31a8d49743317c646dbe90c960a52cbb4b/Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs#L26-L27

I believe we should also expose `PluginMetadata` to JsonRPC plugins.

I guess we can add here: https://github.com/Flow-Launcher/Flow.Launcher/blob/dbf51a31a8d49743317c646dbe90c960a52cbb4b/Flow.Launcher.Core/Plugin/JsonRPCPluginV2.cs#L67

However, I’m not very familiar with this part of the codes and might not be able to contribute directly at the moment. 😢

taooceros: That's what I do to v2. For v1, it's kindly not in the protocol so I don't think we can do that.

Jack251970: So just give this feature only to v2 plugins?

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -577,7 +577,7 @@ internal static async Task UninstallPluginAsync(PluginMetadata plugin, bool remo
                 throw new ArgumentException($"Plugin {plugin.Name} has been modified");
             }
 
-            if (removePluginFromSettings)
+            if (removePluginSettings || removePluginFromSettings)
```
taooceros, line 588: what's the difference between these two?

Jack251970: `removePluginFromSettings`: When one plugin is deleted, this flag will be true and we will remove it from  `Settings.Plugins` and `AllPlugins` so that FL will not have this plugin. Additionally, we need to clean the cache folder of this plugin.

`removePluginSettings`: When one plugin is deleted, FL will ask users if they want to remove the settings of this plugin. Users can choose no and FL will keep its settings for next usage.

Because these two will all delete folders of plugins (one for cache folder, another for settings folder), so we need to dispose plugin if either of them is true.

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -154,7 +159,35 @@ public static void LoadPlugins(PluginsSettings settings)
...
             Settings = settings;
             Settings.UpdatePluginSettings(_metadatas);
+            // Update Json RPC plugin directory before loading plugins so that we can pass the correct plugin directory
+            UpdateJsonRPCPluginDirectory(_metadatas);
             AllPlugins = PluginsLoader.Plugins(_metadatas, Settings);
+            // Update dotnet plugin directory after loading plugins because we need to get assembly name first
+            UpdateNotNetPluginDirectory(_metadatas);
+        }
+
+        private static void UpdateJsonRPCPluginDirectory(List<PluginMetadata> metadatas)
```
taooceros, line 169: I think we can merge these two function.

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -228,10 +261,9 @@ public static ICollection<PluginPair> ValidPluginsForQuery(Query query)
             if (query is null)
                 return Array.Empty<PluginPair>();
 
-            if (!NonGlobalPlugins.ContainsKey(query.ActionKeyword))
+            if (!NonGlobalPlugins.TryGetValue(query.ActionKeyword, out var plugin))
```
taooceros, line 264: why?

Flow.Launcher.Core/Plugin/PluginManager.cs
taooceros: emmm I think the first one is pretty weird...but sure

Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -134,19 +137,35 @@ public static IEnumerable<PluginPair> ExecutablePlugins(IEnumerable<PluginMetada
...
+                .Select(metadata =>
                 {
-                    Plugin = new ExecutablePlugin(metadata.ExecuteFilePath), Metadata = metadata
+                    var plugin = new PluginPair
+                    {
+                        Plugin = new ExecutablePlugin(metadata.ExecuteFilePath),
+                        Metadata = metadata
+                    };
+
+                    plugin.Metadata.AssemblyName = string.Empty;
```
taooceros, line 148: why do we need to manually set this? isn't the default is either empty or null?

Flow.Launcher.Plugin/PluginMetadata.cs
taooceros: I think the simplest way to do this is by removing the JsonIgnore for both `PluginSettingsDirectoryPath` and `PluginCacheDirectoryPath`. It's not a big deal to save a little bit more information in the setting.

taooceros (changes requested):

Flow.Launcher.Core/Plugin/PluginManager.cs
Jack251970: Code quality improvement

Flow.Launcher.Core/Plugin/PluginManager.cs
Jack251970: Okay. Since only JsonRPC v2 & dotnet plugins both use InitAsync to get this path, we do not need to update directory before their constuctors

Flow.Launcher.Core/Plugin/PluginsLoader.cs
Jack251970: Yeah, we can remove that

Jack251970:
> I think the simplest way to do this is by removing the JsonIgnore for both PluginSettingsDirectoryPath and PluginCacheDirectoryPath. It's not a big deal to save a little bit more information in the setting.

@taooceros Sorry, I do not think we should save them in settings. The storage usage does not matter. From my view, this data should be runtime only, and need to be initialized every time we start FL. Similar to Windows `ApplicationData` folder, although it is the same for many Windows version, but in codes we still use `Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData)` to get it during runtime.

taooceros:
no i mean even if we save it during the setting we can still load them from runtime. its just for allow it to serialize when sending to jsonrpc plugins.

Jack251970:
> no i mean even if we save it during the setting we can still load them from runtime. its just for allow it to serialize when sending to jsonrpc plugins.

I really have no idea how to do that. How can we send it to jsonrpc plugins? Could you please help me with it?

taooceros:
> > no i mean even if we save it during the setting we can still load them from runtime. its just for allow it to serialize when sending to jsonrpc plugins.
> 
> I really have no idea how to do that. How can we send it to jsonrpc plugins? Could you please help me with it?

Sorry I think I confused you. I am just saying we should remove the JsonIgnore attribute so when jsonrpc v2 plugins are initialized, these property will be sent to them.

Jack251970:
@taooceros Oh, I get it. Now I changed my codes and please review.

taooceros (approved):

PR #3282, opened 2025-02-26
try some new version

taooceros:


taooceros:
well I guess let's just update to .net 9...

PR #3275, opened 2025-02-25
Fix Dependency Injection Issue

Jack251970:
* We should not create MainViewModel when creating api instance because it can cause strange issue which makes all functions related to `_mainVM` in api instance cannot work (Tested)
* Add exception message box in App constructor (Tested)
* Fix `System.PlatformNotSupportedException` (In progess: https://github.com/dahall/TaskScheduler/issues/1001)

PR #3283, opened 2025-02-26
Fix System.PlatformNotSupportedException & Add exception message box in App constructor

Jack251970:
From #3275.

### DIFF
--- a/Flow.Launcher/App.xaml.cs
+++ b/Flow.Launcher/App.xaml.cs
@@ -165,17 +165,17 @@ namespace Flow.Launcher
         {
             // we try to enable auto-startup on first launch, or reenable if it was removed
             // but the user still has the setting set
-            if (_settings.StartFlowLauncherOnSystemStartup && !Infrastructure.AutoStartup.IsEnabled)
+            if (_settings.StartFlowLauncherOnSystemStartup && !Helper.AutoStartup.IsEnabled)
             {
                 try
                 {
                     if (_settings.UseLogonTaskForStartup)
                     {
-                        Infrastructure.AutoStartup.EnableViaLogonTask();
+                        Helper.AutoStartup.EnableViaLogonTask();
                     }
                     else
                     {
-                        Infrastructure.AutoStartup.EnableViaRegistry();
+                        Helper.AutoStartup.EnableViaRegistry();
                     }
                 }
                 catch (Exception e)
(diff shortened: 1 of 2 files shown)


Jack251970:
@taooceros Hello, I am trying to fix System.PlatformNotSupportedException for installer here.

taooceros:
I guess probably just upgrade to .net 9 would fix this

Jack251970:
We should not upgrade the version, but instead we need to downgrade it.

taooceros:
Task scheduler seems to use 8.0 dependency injection.

taooceros:
> We should not upgrade the version, but instead we need to downgrade it.

yeah if it works then let's go with this first.

Jack251970:
> Task scheduler seems to use 8.0 dependency injection.

Really? Let us try this first.

Jack251970:
Oh, I forget to clean old codes.

Jack251970:
@taooceros Hello, can you download the installer? It seems that I have reached my daily limit...

taooceros:
> @taooceros Hello, can you download the installer? It seems that I have reached my daily limit...

That's shared among all of us...

taooceros:
that's appveyor limit for a certain account (which is the same regardless who is trying to download)

Jack251970:
> that's appveyor limit for a certain account (which is the same regardless who is trying to download)

I have targeted the issue is related the version of `System.Diagnostics.EventLog` package and have created PR for `TaskScheduler` (https://github.com/dahall/TaskScheduler/pull/1004). Now we just need to wait this fix!

taooceros:
Sure, although I think net9 would also fix this issue (I haven't tested).

taooceros:
Have you tested whether the fix work?

Jack251970:
> Have you tested whether the fix work?

Yeah, I have tested.

Jack251970:
You can also check it by installing installer here.

Jack251970:
> Sure, although I think net9 would also fix this issue (I haven't tested).

yeah, net9 can resolve this. and I just want to fix the usage of prereleased installers little quickly

Jack251970:
@taooceros Hello, could we merge it now? cibere need this fix for new api functions under dev branch.

taooceros (approved):
Sorry was taking lunch before. Looks cool!

Issue #3234, opened 2025-02-10
BUG: Flow.Launcher retarget to .NET 10 failed with exception about ambiguous reference between 'System.Windows.Controls.ContextMenu' and 'System.Windows.Forms.ContextMenu'


CancanTang:
### Checks

- [x] I have checked that this issue has not already been reported.

- [ ] I am using the latest version of Flow Launcher.

### Problem Description

When retargeting Flow.Launcher app to .NET 10, it is built failed with errors:

` 'MenuItem' is an ambiguous reference between 'System.Windows.Controls.MenuItem' and 'System.Windows.Forms.MenuItem'`

This is By Design for .Net 10
Please refer to https://github.com/dotnet/winforms/issues/12903 to get details  

### To Reproduce

1. Retarget Flow.Launcher\Flow.Launcher.csproj to net10.0-windows10.0.19041.0
2. run "dotnet publish Flow.Launcher" 

It will be show error : "` 'MenuItem' is an ambiguous reference between 'System.Windows.Controls.MenuItem' and 'System.Windows.Forms.MenuItem'`"

### Flow Launcher Version

1.0.0.0

### Windows Build Number

NA

### Error Log

<details>

```shell

Replace this line with the important log contents.

```
</details>

Jack251970:
Apologies for the confusion - I’m just curious why the target framework needs to be changed. I think this is our team's work to upgrade the target framework when need.

Jack251970:
And .net 9 is on the way. Please refer to #3286.

PR #3286, opened 2025-02-26
update to .net 9

taooceros:
We have experienced various issue with outdated framework, and also .Net 9 provides an excited theme for WPF, which we could later use to improve our UI.

Additional Change:
- Use Locked mode for dependency (this will help us caching nuget packages and make the behavior more consistent)
- [x] ~~Segoe Font Icons still have some problem (which is probably why .net 8 is paused)~~ (Remove the space in the file name makes it work).


Jack251970:
Hello, do you think we need to fix this build issue - ambiguous reference between 'System.Windows.Controls.ContextMenu' and 'System.Windows.Forms.ContextMenu' from #3234?

taooceros:
.net 10 seems to be pretty far away (at least another half year). I guess we can fix it in a separate pr.

jjw24:
Merge after v1.20.0

Jack251970:
After Flow Launcher 1.20.0 is released, I propose we bump the version to 2.x to signify a major upgrade, including a .NET version update and deprecation of Windows 7 and 8 support.

taooceros:
> After Flow Launcher 1.20.0 is released, I propose we bump the version to 2.x to signify a major upgrade, including a .NET version update and deprecation of Windows 7 and 8 support.

We don't do that for previous update to .net framework. I am not sure whether updating the .Net version is that much a big deal...?

Jack251970:
> > After Flow Launcher 1.20.0 is released, I propose we bump the version to 2.x to signify a major upgrade, including a .NET version update and deprecation of Windows 7 and 8 support.
> 
> We don't do that for previous update to .net framework. I am not sure whether updating the .Net version is that much a big deal...?

I am planning for deprecating support for Window 7, 8. And version 1.x and version 2.x can help users easily distinguish them.

jjw24:
Yes, we didn't bump because there weren't any breaking changes. With the stopping of win 7 support that's a breaking change from user side so best to bump to V2.

taooceros:
Sure I just feel like it's not that a big change. Support of Win 7 is already not really existing.

jjw24:
It's not a big technical change for us, it is a breaking change for any users that may for whatever reason still be on Win 7, so it's always best to clearly communicate this change. Additionally, not likely that we will do but an example scenario where we need to ship a version that fixes Win 7 bug, we can still ship as 1.20.1 while the production version continues 2.x.x onwards.

Jack251970:
@taooceros Is this ready?

taooceros:
> @taooceros Is this ready?

Yeah I think so. I download the executable and it seems working fine. Would you mind taking a try?

Jack251970 (approved):
LGTM

Jack251970:
@taooceros Please check if we need to upgrade target frameworks in workflow files as well.

image

Flow.Launcher/Properties/Net9.0-SelfContained.pubxml
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!--
+https://go.microsoft.com/fwlink/?LinkID=208121.
+-->
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <PropertyGroup>
+        <PublishProtocol>FileSystem</PublishProtocol>
+        <Configuration>Release</Configuration>
+        <Platform>Any CPU</Platform>
+        <TargetFramework>net9.0-windows10.0.19041.0</TargetFramework>
```
jjw24: Why is there an extra file here?

Jack251970: > Why is there an extra file here?

I think this file is added accidentally and I removed it as request.

Jack251970:
> Please update default plugin and dotnet workflows to use 9

Done

Jack251970 (dismissed):

taooceros:
Thanks!

Flow.Launcher.Infrastructure/Flow.Launcher.Infrastructure.csproj
```diff
@@ -71,6 +72,11 @@
     </PackageReference>
     <PackageReference Include="SharpVectors.Wpf" Version="1.8.4.2" />
     <PackageReference Include="System.Drawing.Common" Version="7.0.0" />
+    <PackageReference Include="PropertyChanged.Fody" Version="3.4.0">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="SharpVectors.Wpf" Version="1.8.4.2" />
+    <PackageReference Include="System.Drawing.Common" Version="9.0.2" />
```
jjw24: why are these new packages added?

Jack251970: > why are these new packages added?

I found they are duplicated in this project. So I removed them

jjw24:
Hmm on second thought we probably don't want to put through the readme changes, better to have the readme done in a separate PR because it might mislead plugin devs to think Flow is already out with .Net 9 since dev branch is the default landing page.

I will split out readme changes to a separate PR.

Jack251970:
> Hmm on second thought we probably don't want to put through the readme changes, better to have the readme done in a separate PR because it might mislead plugin devs to think Flow is already out with .Net 9 since dev branch is the default landing page.
> 
> I will split out readme changes to a separate PR.

Agree

jjw24 (approved):

PR #3295, opened 2025-02-28
New API for StringMatcherKey of Results

Jack251970:
StringMatcherKey is the key to be compared with the user input for fuzzy search.

This can be useful when the plugin wants to provide a different string for fuzzy search.

If the plugin does not specific this, FL just uses Title and SubTitle to compare with the user input.

For a plugin like Obsidian, the user would expect title and note content to match the string, we can set this property with title and note content.

### DIFF
--- a/Flow.Launcher.Infrastructure/StringMatcher.cs
+++ b/Flow.Launcher.Infrastructure/StringMatcher.cs
@@ -228,7 +228,7 @@ namespace Flow.Launcher.Infrastructure
             return new MatchResult(false, UserSettingSearchPrecision);
         }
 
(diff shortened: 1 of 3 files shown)


taooceros:
This is only used for context menu?

Jack251970:
> This is only used for context menu?

No, it is for general search score.

taooceros:
If you go through the code more carefully, only context menu will automatically do a fuzzy search based on title/subtitle. It could be helpful, but I think the usage needs to be discussed more (like null means plugin do its own fuzzy search while non-null means flow do it for general query).

PR #3296, opened 2025-02-28
Fix issue that mainVM fails to work

Jack251970:
# From #3275.

We should not create `MainViewModel` when creating API instance in constructor because it can cause unknown issue that makes all functions related to _mainVM in API instance cannot work.

# Test

Originally (Tested in VMWare, installed latest dev branch version 65f85cf)

https://github.com/user-attachments/assets/9fe90d8a-edef-45cb-9b71-3fafa12db3a2

Because installing & uninstalling plugins need _mainVM to `ChangeQuery` and `ShowWindow` so nothing happens when _mainVM fails to work.

After

### DIFF
--- a/Flow.Launcher/App.xaml.cs
+++ b/Flow.Launcher/App.xaml.cs
@@ -135,6 +135,7 @@ namespace Flow.Launcher
                 await imageLoadertask;
 
                 var mainVM = Ioc.Default.GetRequiredService<MainViewModel>();
+                ((PublicAPIInstance)API).Initialize(mainVM);
                 var window = new MainWindow(_settings, mainVM);
 
                 Log.Info($"|App.OnStartup|Dependencies Info:{ErrorReporting.DependenciesInfo()}");
--- a/Flow.Launcher/PublicAPIInstance.cs
+++ b/Flow.Launcher/PublicAPIInstance.cs
@@ -34,16 +34,21 @@ namespace Flow.Launcher
     public class PublicAPIInstance : IPublicAPI
     {
         private readonly Settings _settings;
-        private readonly MainViewModel _mainVM;
+        private MainViewModel _mainVM;
 
-        #region Constructor
+        #region Constructor & Initialization
 
-        public PublicAPIInstance(Settings settings, MainViewModel mainVM)
+        public PublicAPIInstance(Settings settings)
         {
             _settings = settings;
-            _mainVM = mainVM;
             GlobalHotkey.hookedKeyboardCallback = KListener_hookedKeyboardCallback;
             WebRequest.RegisterPrefix("data", new DataWebRequestFactory());
+        }
+
+        // We must initialize mainVM later to avoid unknown issue that _mainVM fails to work
+        public void Initialize(MainViewModel mainVM)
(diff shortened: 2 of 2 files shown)


Jack251970:
@taooceros Hello, after testing on a fresh Windows system, I have confirmed that the issue occurs consistently. Earlier, you mentioned you couldn’t reproduce the behavior, and I initially suspected it might be specific to my system. However, it appears to be a broader issue.

taooceros:
yeah I think I could reproduce now. Maybe previously the version is messed up. Sorry for the confusion.

taooceros:
How about we firstly create the mainviewmodel and pass that instance to the dependency injection. Would that also fix the issue?

Jack251970:
Let me check. IIRC, it cannot fix this issue.

Jack251970:
> How about we firstly create the mainviewmodel and pass that instance to the dependency injection. Would that also fix the issue?

No, we cannot do that. Because Updater will call API construction firstly.

```
.AddSingleton(sp => new Updater(sp.GetRequiredService<IPublicAPI>(), Launcher.Properties.Settings.Default.GithubRepo))
```

Jack251970:
I tried this but it cannot help.

image

Jack251970:
Yes, I find the MainViewModel is created twice.

Jack251970:
I am still checking.

Jack251970:
> I think it's important to understand why this is happening. Does this problem happen when build with VS?

Yes.

Jack251970:
> > > I think it's important to understand why this is happening. Does this problem happen when build with VS?
> > 
> > 
> > Yes.
> 
> That's good, so we don't have to test the appveyor build to understand why. I need to be afk for a while.

Certainly! For now I found that `PublicAPIInstance` is created twice, and I will keep you informed of any progress or updates.

taooceros:
> > > > I think it's important to understand why this is happening. Does this problem happen when build with VS?
> > > 
> > > 
> > > Yes.
> > 
> > 
> > That's good, so we don't have to test the appveyor build to understand why. I need to be afk for a while.
> 
> Certainly! For now I found that `PublicAPIInstance` is created twice, and I will keep you informed of any progress or updates.

Oh if that's the case then things can certainly goes wrong. It can link to the wrong publicapiinstance (which associate with a wrong mainviewmodel), which causes the problem. Let's try to figure out where's the problem.

Jack251970:
> > > > > I think it's important to understand why this is happening. Does this problem happen when build with VS?
> > > > 
> > > > 
> > > > Yes.
> > > 
> > > 
> > > That's good, so we don't have to test the appveyor build to understand why. I need to be afk for a while.
> > 
> > 
> > Certainly! For now I found that `PublicAPIInstance` is created twice, and I will keep you informed of any progress or updates.
> 
> Oh if that's the case then things can certainly goes wrong. It can link to the wrong publicapiinstance (which associate with a wrong mainviewmodel), which causes the problem. Let's try to figure out where's the problem.

Ok, I found the reason. `PluginManager` uses static property for API initialization which creates one more instance.

I will close this PR, and create a new one.

PR #3297, opened 2025-02-28
Fix API instance create twice issue & Make PluginManager.API private

Jack251970:
# From #3296 

We should not initialize API in static constructor or it will create another API instance.


Jack251970:
@taooceros I have targeted the source of this issue and now it works fine for me.

taooceros:
looks cool! wonderful

i will review it once i get to computer

taooceros:
Ideally we should also make PluginManager as a Singleton, but let's keep it like this for now as that might take quite a few work.

PR #3303, opened 2025-03-03
Fix Possible Plugin Initialization Issue

Jack251970:
# Issue

If using `Task.Run` to initialize my plugin Clipboard+, my plugin will fail to monitor the clipboard. (I do not know why, but my plugin cannot listen to system clipboard event if we do so)

The issue with StartAsSTA is that you registered a hook via the specific thread, but once it is created, the thread is finishing and die, which leaves Windows couldn't fire the event because it cannot find the thread that is registering the event.

So we should not use `Task` to run plugin init event. Instead, we would better to run `InitAsync` in main thread with `foreach`.

# Test

Use `foreach` will affect performance but just a little bit. I use 15 plugins and test with `Stopwatch.DebugAsync`.

With `Task.WhenAll` the plugins are fully initialized in about 1350ms, while with `foreach` the plugins are fully initialized in around 1250ms.

Considering majority of plugins do not need much time to initialize, this solution can serve the purpose.

Here is my plugin for test:
[Clipboard+-2.1.1 - Test InitAsync.zip](https://github.com/user-attachments/files/19053680/Clipboard%2B-2.1.1.-.Test.InitAsync.zip)

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -165,7 +165,8 @@ namespace Flow.Launcher.Core.Plugin
         {
             var failedPlugins = new ConcurrentQueue<PluginPair>();
 
-            var InitTasks = AllPlugins.Select(pair => Task.Run(async delegate
+            // Some plugins should not be initialized in task, so we cannot use Task.WhenAll here
+            foreach (var pair in AllPlugins)
             {
                 try
                 {
@@ -182,9 +183,7 @@ namespace Flow.Launcher.Core.Plugin
                     pair.Metadata.Disabled = true;
                     failedPlugins.Enqueue(pair);
                 }
-            }));
-
-            await Task.WhenAll(InitTasks);
+            }
 
             _contextMenuPlugins = GetPluginsForInterface<IContextMenu>();
             foreach (var plugin in AllPlugins)
--- a/Flow.Launcher.Plugin/Interfaces/IPlugin.cs
+++ b/Flow.Launcher.Plugin/Interfaces/IPlugin.cs
@@ -30,7 +30,12 @@ namespace Flow.Launcher.Plugin
         /// <param name="context"></param>
         void Init(PluginInitContext context);
 
-        Task IAsyncPlugin.InitAsync(PluginInitContext context) => Task.Run(() => Init(context));
+        async Task IAsyncPlugin.InitAsync(PluginInitContext context)
+        {
+            // Some plugins should not be initialized in task
+            Init(context);
+            await Task.CompletedTask;
+        }
 
         Task<List<Result>> IAsyncPlugin.QueryAsync(Query query, CancellationToken token) => Task.Run(() => Query(query));
     }


jjw24:
it's still async init so impact should be minimal. thoughts @taooceros ? If looks good then approve + merge please.

taooceros:
> it's still async init so impact should be minimal. thoughts @taooceros ? If looks good then approve + merge please.

well it should still be async, so plugins with network requirement can still fire those request concurrently (I am not sure), but not parallel. I think the assumption about plugin initialization is pretty short can be true? I didn't see much CPU intense plugin except Program plugin.

The clipboard issue is because it requires the thread being a STA rather than a MTA (which will be what `Task.Run` uses). I think we previously add a utility method to run a task in STA? Although that can be slower I think.

On the other hand, it will block the main thread, which means this will disallow us to show up flow's window before plugin initialization. Now it's fine as that's the current behavior, but I am not sure whether in the future we will want to do that?

Jack251970:
> > it's still async init so impact should be minimal. thoughts @taooceros ? If looks good then approve + merge please.
> 
> well it should still be async, so plugins with network requirement can still fire those request concurrently (I am not sure), but not parallel. I think the assumption about plugin initialization is pretty short can be true? I didn't see much CPU intense plugin except Program plugin.
> 
> The clipboard issue is because it requires the thread being a STA rather than a MTA (which will be what `Task.Run` uses). I think we previously add a utility method to run a task in STA? Although that can be slower I think.
> 
> On the other hand, it will block the main thread, which means this will disallow us to show up flow's window before plugin initialization. Now it's fine as that's the current behavior, but I am not sure whether in the future we will want to do that?

I do not think we need to show up FL window before plugin initialization. The plugin list is uninitialized and we cannot get anything from query, even may encounter issue when executing query methods.

And Devs can handle CPU intense task in their own plugin and this will not break anything.

I am not sure the clipboard issue is related to sta because  I just want to register clipboard monitor event and it does not need sta as I remember.

VictoriousRaptor:
> > > it's still async init so impact should be minimal. thoughts @taooceros ? If looks good then approve + merge please.
> > 
> > 
> > well it should still be async, so plugins with network requirement can still fire those request concurrently (I am not sure), but not parallel. I think the assumption about plugin initialization is pretty short can be true? I didn't see much CPU intense plugin except Program plugin.
> > The clipboard issue is because it requires the thread being a STA rather than a MTA (which will be what `Task.Run` uses). I think we previously add a utility method to run a task in STA? Although that can be slower I think.
> > On the other hand, it will block the main thread, which means this will disallow us to show up flow's window before plugin initialization. Now it's fine as that's the current behavior, but I am not sure whether in the future we will want to do that?
> 
> I do not think we need to show up FL window before plugin initialization. The plugins are uninitialized and we cannot get anything from query, and even may encounter issue when executing query methods.
> 
> And Devs can handle CPU intense task in their own plugin and this will not break anything.
> 
> I am not sure the clipboard issue is related to sta because I just want to register clipboard monitor event and it does not need sta as I remember.

May I propose one workaround? You can use windows api to access the built-in clipboard histroy (requires windows 10 1809+ iirc, and rerquires users to turn on the feature in windows settings), rather than creating your own monitor. It might need lots of changes though.

Jack251970:
> > > > it's still async init so impact should be minimal. thoughts @taooceros ? If looks good then approve + merge please.
> > > 
> > > 
> > > well it should still be async, so plugins with network requirement can still fire those request concurrently (I am not sure), but not parallel. I think the assumption about plugin initialization is pretty short can be true? I didn't see much CPU intense plugin except Program plugin.
> > > The clipboard issue is because it requires the thread being a STA rather than a MTA (which will be what `Task.Run` uses). I think we previously add a utility method to run a task in STA? Although that can be slower I think.
> > > On the other hand, it will block the main thread, which means this will disallow us to show up flow's window before plugin initialization. Now it's fine as that's the current behavior, but I am not sure whether in the future we will want to do that?
> > 
> > 
> > I do not think we need to show up FL window before plugin initialization. The plugins are uninitialized and we cannot get anything from query, and even may encounter issue when executing query methods.
> > And Devs can handle CPU intense task in their own plugin and this will not break anything.
> > I am not sure the clipboard issue is related to sta because I just want to register clipboard monitor event and it does not need sta as I remember.
> 
> May I propose one workaround? You can use windows api to access the built-in clipboard histroy (requires windows 10 1809+ iirc, and rerquires users to turn on the feature in windows settings), rather than creating your own monitor. It might need lots of changes though.

I have tried this. But this also fails to work....

Jack251970:
I believe using `Task` seems to break sth.

taooceros:
> I believe using `Task` will break sth.

yes because its thread pool

Jack251970:
> > I believe using `Task` will break sth.
> 
> yes because its thread pool

I learnt it. Are we happy to merge?

VictoriousRaptor:
> > > > > it's still async init so impact should be minimal. thoughts @taooceros ? If looks good then approve + merge please.
> > > > 
> > > > 
> > > > well it should still be async, so plugins with network requirement can still fire those request concurrently (I am not sure), but not parallel. I think the assumption about plugin initialization is pretty short can be true? I didn't see much CPU intense plugin except Program plugin.
> > > > The clipboard issue is because it requires the thread being a STA rather than a MTA (which will be what `Task.Run` uses). I think we previously add a utility method to run a task in STA? Although that can be slower I think.
> > > > On the other hand, it will block the main thread, which means this will disallow us to show up flow's window before plugin initialization. Now it's fine as that's the current behavior, but I am not sure whether in the future we will want to do that?
> > > 
> > > 
> > > I do not think we need to show up FL window before plugin initialization. The plugins are uninitialized and we cannot get anything from query, and even may encounter issue when executing query methods.
> > > And Devs can handle CPU intense task in their own plugin and this will not break anything.
> > > I am not sure the clipboard issue is related to sta because I just want to register clipboard monitor event and it does not need sta as I remember.
> > 
> > 
> > May I propose one workaround? You can use windows api to access the built-in clipboard histroy (requires windows 10 1809+ iirc, and rerquires users to turn on the feature in windows settings), rather than creating your own monitor. It might need lots of changes though.
> 
> I have tried this. But this also fails to work....
> 
> https://github.com/Jack251970/Flow.Launcher.Plugin.ClipboardPlus/blob/85f9b398cdf0d6db2efcec1b8c8bde7c4fa24815/src/Flow.Launcher.Plugin.ClipboardPlus.Core/Data/Models/Clipboard/ClipboardHandleWin.cs#L73-L87

Maybe we don't need to actually monitor clipboard, but only access it when user queries? I made a concept plugin long ago and it worked.

Jack251970:
@VictoriousRaptor Get it. But you solution fully relies on Windows Clipboard History (10.0.17763+) and every time you query, plugin will get results from Windows Clipboard History.

But my plugin is designed for win7+ and I do not want to rely on Windows Clipboard History feature.

By the way, my plugin also supports the solution you described above (Sync with Windows Clipboard History).

taooceros:
> I am not sure the clipboard issue is related to sta because I just want to register clipboard monitor event and it does not need sta as I remember.

Could you try STA first? I am more leaning to let the plugin handle STA themselves. If not then we can think about this.

If we provide this guarantee, then in the future we would have a hard time to revert it back as some plugin may depend on this behavior.

Jack251970:
> > I am not sure the clipboard issue is related to sta because I just want to register clipboard monitor event and it does not need sta as I remember.
> 
> Could you try STA first? I am more leaning to let the plugin handle STA themselves. If not then we can think about this.
> 
> If we provide this guarantee, then in the future we would have a hard time to revert it back as some plugin may depend on this behavior.

I have tested, but the clipboard monitor created in the STA thread cannot work as well.

Jack251970:
I really have no ideas about 'some plugin may depend on this behavior'? I do not think there are any functions which cannot work on main thread, but it can work with thread pool.

Jack251970:
> > > > I am not sure the clipboard issue is related to sta because I just want to register clipboard monitor event and it does not need sta as I remember.
> > > 
> > > 
> > > Could you try STA first? I am more leaning to let the plugin handle STA themselves. If not then we can think about this.
> > > If we provide this guarantee, then in the future we would have a hard time to revert it back as some plugin may depend on this behavior.
> > 
> > I have tested, but the clipboard monitor created in the STA thread cannot work as well.
> > 
> > image
> 
> No you can't create a monitor in STA thread and return that to the main thread. You should hold the monitor consistently in the STA thread (which will be used to handle the message from the operating system), and uses some message passing or whatever to pass the event to the place where you want to handle them.

Sounds like very complex. Could we add new interface called ISTAPlugin which uses the above solution?

Jack251970:
Or can I pass return value to let FL know this?

Jack251970:
Additionally, if I create this monitor in `constructor` but not `InitAsync`, it can work. I do not think we need to put all codes in STA thread.

taooceros:
> Additionally, if I create this monitor in `constructor` but not `InitAsync`, it can work.

Yeah because the constructor is not run in parallel and is run on the UI thread. Though it's not general a good idea as `disabled` plugin will still run its constructor.

I guess the issue with StartAsSTA is that you registered a hook via the specific thread, but once it is created, the thread is finishing and die, which leaves Windows couldn't fire the event because it cannot find the thread that is registering the event.

I think maybe we can also provide an API that exposes the `Dispatcher` of the main thread for this...? Though, if everyone else agree that it's ok to let `Init` run not in parallel then we can maybe proceed with this change...

Jack251970:
> Yeah because the constructor is not run in parallel and is run on the UI thread. Though it's not general a good idea as `disabled` plugin will still run its constructor.

That is why I want to move this to `InitAsync` function. I want to reduce the memory usage.

Jack251970:
@Flow-Launcher/team Hi, I would like to ask your ideas. Thank you!

Jack251970:
Keep as it was

PR #3344, opened 2025-03-13
Use Command Keyword to Search System Command

### LINKED ISSUE #3343: System plugin translation
Translations for the System plugin are not applied to the Flow Launcher. This is the Command column in the plugin settings.

Image

Jack251970:
# Use Command Keyword to Search System Command

Previously, FL uses the English title & translated title & subtitle to search the commands. If English title is matched, the result title will be in English which I think can cause confusion for users.

This PR Introduces command keyword to replace the original function of searching with English title. (Users can still search with translated title & subtitle) And users can change this command keyword.

By default, the command keywords are the same as the English title. And users can set whatever they like for faster query.

# Demo

https://github.com/user-attachments/assets/4ae1ee9d-5f97-47bb-a1a7-8d4adf048bf3

# Test

* Change command keyword & Search them.
* Name & Description column are now translated.
* Sys plugin can correctly save & load settings.

Close #3343

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Sys/Command.cs
+++ b/Plugins/Flow.Launcher.Plugin.Sys/Command.cs
@@ -6,12 +6,39 @@ namespace Flow.Launcher.Plugin.Sys
     {
         public string Key { get; set; }
 
+        private string name;
         [JsonIgnore]
-        public string Name { get; set; }
+        public string Name
+        {
+            get => name;
+            set
+            {
+                name = value;
+                OnPropertyChanged();
+            }
+        }
 
+        private string description;
         [JsonIgnore]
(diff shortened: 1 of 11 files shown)


Jack251970:
@onesounds I have completed my work and please review when you have a chance. Regarding the tips for changing the command keyword, I would like your feedback on whether they are clear enough for users.

onesounds (approved):

onesounds:
Good. there are no issues. (It works fine even when no command is entered or when there are duplicate inputs.) There was a request before to allow certain items to be disabled (since options like "Restart with Advanced" aren’t used often), but I’m not sure if that should be included. For now, this PR is good to go, so I'll approve it.

PR #3350, opened 2025-03-15
Input Smooth & Search delay

Jack251970:
# Delay time when typing to search

Delay plugin query when typing. We can set one default search delay time, and we can set search delay time for each plugin so that they will not follow default search delay time.

Close #1677.

# Demo

Delay:

https://github.com/user-attachments/assets/e090b475-dd90-46ba-85b4-2c589ac05521

No delay:

https://github.com/user-attachments/assets/b09ecec3-5418-4b0f-8a43-583dc5473678

# TODO

- [x] Support default delay speed for every plugin.
- [x] Specific delay time speed for some plugins (overwrite default value).
- [x] Add default value from `plugin.json`. (You can add `"SearchDelaySpeed": "Slow"` in `plugin.json` to set default search delay speed value)
- [x] Improved design in #3265.
- [x] Add documents for properties in `Plugin` project like #3272.
- [x] Give default search delay for preinstalled plugins. (WebSearch: 'Slow')
- [ ] Remove debug codes: `// TODO: Remove debug codes.`

# Test

- [x] Turning On and off Search delay can work.
- [x] Context menu / History still work.
- [x] Plugin search delay can work.
- [x] If changing query text and some tasks are still running, those task will be cancelled.\
- [x] Default search delay time in `plugin.json` can be applied.


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -631,6 +630,7 @@ public void ChangeQueryText(string queryText, bool isReQuery = false)
                 {
                     // re-query is done in QueryText's setter method
                     QueryText = queryText;
+                    Query();
```
taooceros, line 633: why do we need this?

Jack251970: From original codes, when QueryText is changed, it will call `Query()`.

Jack251970: And I remove `Query()` in `QueryText` setter, so we need to manually call it.

taooceros: Why do we do that?

Jack251970: `QueryText` is a property used by others functions like auto complete text. So we need to update it once users input.

`Query` is a function to load results and we can delay them.

onesounds (approved):

PR #3354, opened 2025-03-17
Fix custom query hotkey settings issue

Jack251970:
# Add new `enum` type to handle logic for custom query hotkey setting dialog

As you can see, the new version of `HotkeyControl` uses `Type` to specific the `Hotkey` getter and setter.

And previously, the default value of `Type` is `Hotkey` so the `HotkeyControl` in `CustomQueryHotkeySettings` will fallback to `case HotkeyType.Hotkey:`, resulting the change of the search hotkey.

So we need to add a new `enum` type so that `HotkeyControl` will cache the hotkey value and return it as result which is needed for `CustomQueryHotkeySettings` (it uses )

Follow on with #3310.

# Test

* Set / Update / Delete custom query hotkey.

### DIFF
--- a/Flow.Launcher/HotkeyControl.xaml.cs
+++ b/Flow.Launcher/HotkeyControl.xaml.cs
@@ -84,7 +84,7 @@ namespace Flow.Launcher
             nameof(Type),
             typeof(HotkeyType),
             typeof(HotkeyControl),
-            new FrameworkPropertyMetadata(HotkeyType.None, FrameworkPropertyMetadataOptions.BindsTwoWayByDefault, OnHotkeyChanged)
+            new FrameworkPropertyMetadata(HotkeyType.CustomQueryHotkey, FrameworkPropertyMetadataOptions.BindsTwoWayByDefault, OnHotkeyChanged)
         );
 
         public HotkeyType Type
@@ -95,8 +95,8 @@ namespace Flow.Launcher
 
         public enum HotkeyType
         {
-            // Custom hotkeys
-            None,  // Used for getting hotkey from dialog result
+            // Custom query hotkeys
+            CustomQueryHotkey,
             // Settings hotkeys
             Hotkey,
             PreviewHotkey,
@@ -125,8 +125,8 @@ namespace Flow.Launcher
             {
                 return Type switch
                 {
-                    // Custom hotkeys
-                    HotkeyType.None => hotkey,
+                    // Custom query hotkeys
+                    HotkeyType.CustomQueryHotkey => hotkey,
                     // Settings hotkeys
                     HotkeyType.Hotkey => _settings.Hotkey,
                     HotkeyType.PreviewHotkey => _settings.PreviewHotkey,
@@ -149,8 +149,10 @@ namespace Flow.Launcher
             {
                 switch (Type)
                 {
-                    // Custom hotkeys
-                    case HotkeyType.None:
+                    // Custom query hotkeys
+                    case HotkeyType.CustomQueryHotkey:
+                        // We just need to store it in a local field
+                        // because we will save to settings in other place
                         hotkey = value;
                         break;
                     // Settings hotkeys
--- a/Flow.Launcher/HotkeyControl.xaml.cs
+++ b/Flow.Launcher/HotkeyControl.xaml.cs
@@ -84,7 +84,7 @@ namespace Flow.Launcher
             nameof(Type),
             typeof(HotkeyType),
             typeof(HotkeyControl),
-            new FrameworkPropertyMetadata(HotkeyType.Hotkey, FrameworkPropertyMetadataOptions.BindsTwoWayByDefault, OnHotkeyChanged)
+            new FrameworkPropertyMetadata(HotkeyType.None, FrameworkPropertyMetadataOptions.BindsTwoWayByDefault, OnHotkeyChanged)
         );
 
         public HotkeyType Type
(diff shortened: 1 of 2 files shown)


taooceros:
do you mind briefly introduce what's the issue?

Jack251970:
> do you mind briefly introduce what's the issue?

As you can see, the new version of `HotkeyControl` uses `Type` to specific the `Hotkey`.

And previously, the default value of `Type` is `Hotkey` so it will fallback to `case HotkeyType.Hotkey:`, resulting the change of the search hotkey.

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -95,6 +95,9 @@ public HotkeyType Type
 
         public enum HotkeyType
         {
+            // Custom hotkeys
+            None,  // Used for getting hotkey from dialog result
```
jjw24, line 99: Is it set as None because custom hotkey doesnt directly open the hotkey dialogue yet? Isn't here a better representation than None?

Jack251970: here none means not directly changing the settings

Jack251970: > Is it set as None because custom hotkey doesnt directly open the hotkey dialogue yet? Is there a better representation than None?

If you have other better name, please let me know and I will use it.

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -185,7 +197,7 @@ public string Hotkey
                         _settings.SelectNextItemHotkey2 = value;
                         break;
                     default:
-                        return;
+                        throw new System.NotImplementedException("Hotkey type not setted");
```
jjw24, line 200:
```suggestion
                        throw new System.NotImplementedException("Hotkey type not set");
```

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -135,13 +142,18 @@ public string Hotkey
                     HotkeyType.SelectPrevItemHotkey2 => _settings.SelectPrevItemHotkey2,
                     HotkeyType.SelectNextItemHotkey => _settings.SelectNextItemHotkey,
                     HotkeyType.SelectNextItemHotkey2 => _settings.SelectNextItemHotkey2,
-                    _ => string.Empty
+                    _ => throw new System.NotImplementedException("Hotkey type not setted")
```
jjw24, line 145:
```suggestion
                    _ => throw new System.NotImplementedException("Hotkey type not set")
```

Flow.Launcher/HotkeyControl.xaml.cs
jjw24: `Prescreen` (before the actual hotkey dialog screen) or could it just be `CustomQueryHotkey` (maybe this will also confuse others in the future?)

Jack251970: What do you think of `Cache`? It means we do not store the data in the settings directly but we cache hotkey in local field.

jjw24: IIRC these hotkeys are still stored in settings file aren't they?

Jack251970: Oh yeah. Alright., I choose `CustomQueryHotkey`.

Jack251970: And I add comments so that it will confuse others too much

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -135,13 +142,20 @@ public string Hotkey
...
             }
             set
             {
                 switch (Type)
                 {
+                    // Custom query hotkeys
+                    case HotkeyType.CustomQueryHotkey:
+                        // We just need to store it in a local field
+                        // because we will save to settings in other place
+                        hotkey = value;
```
jjw24, line 156: Oh I see what you mean here. Yeah the design of custom query hokeys screens is slightly different to others to accommodate setting many custom hotkeys.

Jack251970: yes. that is why we need to add a new enum type for it.

Flow.Launcher/HotkeyControl.xaml.cs
```diff
@@ -84,7 +84,7 @@ public ICommand? ChangeHotkey
             nameof(Type),
             typeof(HotkeyType),
             typeof(HotkeyControl),
-            new FrameworkPropertyMetadata(HotkeyType.Hotkey, FrameworkPropertyMetadataOptions.BindsTwoWayByDefault, OnHotkeyChanged)
+            new FrameworkPropertyMetadata(HotkeyType.CustomQueryHotkey, FrameworkPropertyMetadataOptions.BindsTwoWayByDefault, OnHotkeyChanged)
```
jjw24, line 87: Is this correct? Doesn't seem to read well.

Jack251970: I would like to set this type as default value because it will not cause any strange operation to Settings.

Jack251970: Previously, the default value is `Hotkey` and it can cause the issue that we will change query hotkey if we do not set `Type`

Jack251970: nvm, I have added a `None` type so that we are forced to set this `Type`.

jjw24 (approved):

Flow.Launcher/HotkeyControl.xaml.cs
jjw24: Yeah seems better this way.

PR #3365, opened 2025-03-21
Fix Plugin Manager Issue

Jack251970:
When we call other functions that need `AllPlugins`, we should put them after `PluginManager.LoadPlugins(_settings.PluginSettings);`.

* Since FL introduces dependency injection #3175, the constructor of `MainViewModel` is brought forward because `PublicAPIInstance` needs its instance, so we cannot put `RegisterResultsUpdatedEvent()` in its constructor.
* Since `Ioc.Default.GetRequiredService<Internationalization>().ChangeLanguage(_settings.Language);` needs to check plugins with `IPluginI18n` interface, it should be put after `Ioc.Default.GetRequiredService<MainViewModel>().RegisterResultsUpdatedEvent();`

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -168,7 +168,6 @@ namespace Flow.Launcher.ViewModel
             };
 
             RegisterViewUpdate();
-            RegisterResultsUpdatedEvent();
             _ = RegisterClockAndDateUpdateAsync();
         }
 
@@ -213,7 +212,7 @@ namespace Flow.Launcher.ViewModel
             }
         }
 
-        private void RegisterResultsUpdatedEvent()
+        public void RegisterResultsUpdatedEvent()
         {
             foreach (var pair in PluginManager.GetPluginsForInterface<IResultUpdated>())
             {
--- a/Flow.Launcher/App.xaml.cs
+++ b/Flow.Launcher/App.xaml.cs
@@ -131,6 +131,7 @@ namespace Flow.Launcher
                 Ioc.Default.GetRequiredService<Internationalization>().ChangeLanguage(_settings.Language);
 
                 PluginManager.LoadPlugins(_settings.PluginSettings);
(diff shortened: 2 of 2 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -213,7 +212,7 @@ void continueAction(Task t)
             }
         }
 
-        private void RegisterResultsUpdatedEvent()
+        public void RegisterResultsUpdatedEvent()
```
taooceros, line 215: actually maybe we can move this to PluginsManager?

Jack251970: okay

Jack251970: oh we cannot do that. since `PluginManager` is in core project, it cannot access function in main project (MainViewModel.RegisterResultsUpdatedEvent)

taooceros: I see, sorry for that.

taooceros:
Now it looks like PluginManager should always be initialized first? (correct me if I am wrong). Is it possible to move that to DI and acts as a requirement for other components? (We can do that later).

taooceros (approved):

Jack251970:
> Now it looks like PluginManager should always be initialized first? (correct me if I am wrong). Is it possible to move that to DI and acts as a requirement for other components? (We can do that later).

Yeah, we should move plugin manager to DI. And we need to consider more about this since the relationship could be quite complex.

jjw24:
> Now it looks like PluginManager should always be initialized first? (correct me if I am wrong). Is it possible to move that to DI and acts as a requirement for other components? (We can do that later).

> Yeah, we should move plugin manager to DI. And we need to consider more about this since the relationship could be quite complex.

Yes agree.

joespinoza:
hi, i updated to de new 1.20 version then updated Plugins Manager and stop working
[2025-05-22.txt](https://github.com/user-attachments/files/20401573/2025-05-22.txt)

PR #3369, opened 2025-03-22
Enable acrylic effect on first launch

onesounds:
## What's the PR
- Enable acrylic effect on first launch if running on Windows 11
- Default backdrop is none.
- Change the backdrop setting to Acrylic on first launch if the conditions are met.
- The first impression improves when the program is launched for the first time 😉


Jack251970 (approved):
LGTM

PR #3370, opened 2025-03-22
Support Preview Panel for History Results

Jack251970:
# Support Preview Panel for History Results

image

# TODO

- [x] When selected items are from history, the preview panel height is 0.
- [x] History item preview image cannot be loaded.


taooceros:
Looks cool, but I am wondering why is this needed?

Jack251970:
> Looks cool, but I am wondering why is this needed?

Since context menu will not hide preview panel, we should also provide preview panel for history records

Jack251970:
Or you will get this if you enable preview panel: 
image

taooceros:
I see, better than empty

taooceros:
> > Looks cool, but I am wondering why is this needed?
> 
> Since context menu will not hide preview panel, we should also provide preview panel for history records

It's a bit different. ContextMenu's results are provided by plugins, which may include a special preview (I don't know whether anyone do that though).

taooceros (dismissed):

onesounds (approved):

Issue #3373, opened 2025-03-22
BUG: Context menu and arrow keys


deefrawley:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I am using the latest version of Flow Launcher.

### Problem Description

When using right arrow key to go to a search entry context menu, then left arrow key to get back to search results, the cursor is placed at the start of the search query not the end. If you then use the down arrow key to go to another search result, you have to hit the right arrow key enough times to skip through the query one character at a time to get to the end of the query before you can use the context menu again.
When returning from the context menu the cursor should be placed at the end of the search query.

### Flow Launcher Version

1.19.5

### Windows Build Number

10.0.19045.5608

### Error Log

No error produced

onesounds:
I saw this issue too, but whenever I tried to reproduce it, it would always disappear, so I thought I was just imagining it. I've confirmed the issue and made the necessary fixes.

PR #3374, opened 2025-03-23
Fix cursor doesn't move to the end on initial launch

onesounds:
## What's the PR ##
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/3373

## What's the Issue
- In the initial launch state, when navigating back from the context menu to the search results, the cursor is incorrectly positioned at the beginning.
- After Hide(), the cursor moves to the end without any issues.

## Cause of the issue and analysis
- QueryTextCursorMovedToEnd is triggered whenever the value changes, but it seems it wasn't triggered initially because there was no default value or it was already True. When the window is hidden for the first time, a default value of false is added, which allowed it to function correctly afterward.
 
## Solution
- A default value(false) has now been added during OnLoaded.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -143,7 +143,9 @@ namespace Flow.Launcher
 
             // Since the default main window visibility is visible, so we need set focus during startup
             QueryTextBox.Focus();
-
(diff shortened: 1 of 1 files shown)


Jack251970 (approved):
Work well for me

PR #3366, opened 2025-03-21
Implement auto-switching to English when the option is enabled

Yusyuriv:
> [!WARNING]
> **DO NOT MERGE YET**. Highly experimental, needs thorough testing with both IME and non-IME languages; also needs someone to confirm there are no memory leaks with using the native Windows APIs.

Fixes #2108, fixes #2569, fixes #3126.

# What it does
When the `Always Start Typing in English Mode` option is enabled, when the Flow Launcher window opens, it remembers the current keyboard layout and then switches to English. When the window hides, it restores the remembered keyboard layout.

# Testing I've done
- [x] It correctly switches to English when the window opens and restores the original keyboard layout when the window hides.

# Known issues
- [x] When it gets closed via Esc, Alt+Space, or Enter, it restores the keyboard layout correctly, but when the window loses focus, the layout does not get restored.

  ~~I haven't found a way to fix this, but I think it's still better than not having this functionality at all.~~
  Fixed.

# Things to consider
- [x] When used with a language that has an IME mode, don't switch the language, just disable the IME mode.

PR #3375, opened 2025-03-24
Force FL window foreground when switching keyboard layout

Jack251970:
Follow on with #3366.

When switching keyboard layout, we need to make sure foreground window is FL main window.


Flow.Launcher.Infrastructure/Win32Helper.cs
```diff
@@ -124,6 +124,16 @@ public static bool SetForegroundWindow(nint handle)
             return PInvoke.SetForegroundWindow(new(handle));
         }
 
+        public static bool IsForegroundWindow(Window window)
+        {
+            return GetWindowHandle(window).Equals(PInvoke.GetForegroundWindow());
```
taooceros, line 129: let call IsForegroundWindow with the handle from GetWindowHandle

Jack251970: Cannot get your idea. I already have a `internal` version which uses the handle from `GetWindowHandle`.
```
internal static bool IsForegroundWindow(HWND handle)
{
    return handle.Equals(PInvoke.GetForegroundWindow());
}
```

Jack251970: Since `HWND` is an internal class, we cannot make that `internal` function to be `public`

taooceros: huh? You can. Just do `IsForegroundWindow(GetWindowHandle(window))`

taooceros: I think you misunderstand how the modifier works. It won't restrict which method it can call inside the body, but only whether other can call it.

Jack251970: But it gives me error here:
image

Jack251970: sorry for my misunderstanding, now I use `GetWindowHandle` inside its function

Jack251970:
@Yusyuriv Are we happy to merge this?

PR #3377, opened 2025-03-24
Fix font select function

onesounds:
## What's the PR
- Fixed an issue where theme font changes were not being applied.
- Fixed an issue where resources could not be found at certain times during debugging.

## What was the problem
- There was an issue where, due to the timing of removing oldResource, there was a brief moment when no resource dictionaries were present.
- Font settings were not applied correctly for blur themes because of the timing of blur theme application.

## Test Cases
- [x] Font style changes should be applied in real time regardless of whether a blur theme is used.
- [x] On first launch, the previously saved font settings should be applied. 
- [x] There should be no "resource not found" messages in debug mode.
image

### DIFF
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -124,13 +124,15 @@ namespace Flow.Launcher.Core.Resource
         {
             try
             {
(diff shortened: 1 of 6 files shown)


Jack251970 (approved):
LGTM

PR #3384, opened 2025-03-26
Fix freezing with gpu

onesounds:
# Reference

* https://github.com/dotnet/wpf/issues/2158
* https://github.com/dotnet/wpf/issues/4276
* https://github.com/dotnet/wpf/issues/8919
----
* https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/issues/559
* https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/pull/3587
* https://forums.ni.com/t5/Measurement-Studio-for-NET/Window-with-WPF-Graph-freezes-when-windows-account-locked/td-p/2419516
* https://github.com/Kinnara/ModernWpf/issues/364
# TODO
- [ ] Check if #3404 is fixed.


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -236,6 +236,7 @@ private async void OnLoaded(object sender, RoutedEventArgs _)
             DependencyPropertyDescriptor
                 .FromProperty(VisibilityProperty, typeof(StackPanel)) // History는 StackPanel이라고 가정
                 .AddValueChanged(History, (s, e) => UpdateClockPanelVisibility());
+            _viewModel.ClearAllCacheModes(this);
```
lindexi, line 239: Emmmm, can I know all UIElement in all the window be clear cache? Thank you.

onesounds: Should the settings window (secondary window) also have its cachemode=null? For now, separate from this PR, I tried settingwindow the cache mode of all UI elements in both the main window and the settings window to Null in code, but it didn’t resolve the issue.

lindexi: @onesounds What worries me is that elements are added dynamically, and they are added after loading.

lindexi: @onesounds Can I ask any other try update?

onesounds: It didn't go well. Since we're using ModernWpf, I thought that instead of creating and building a new NuGet package to remove all the resources used by ModernWpf, I should find and remove them in the code — and I tried that, but it didn't work. I can’t tell if I’m doing it right, so I discussed it with Copilot and confirmed through debug code that all the caches were cleared, but even then, I’m not sure if it really worked correctly.

I also tested a few suggestions from cyclonring, but they all failed. The only thing that almost worked was an idea I came up with — showing the first window again at the moment of a "system change" (which removed the freezing of the settings window). But during testing, a focus loss situation. When the first window is hidden in this case, it causes the first window itself to freeze instead. I ended up giving up on this task because I discovered that the first window itself freezes as a result. https://github.com/Flow-Launcher/Flow.Launcher/pull/3378

I’ve asked my team members for help with this a few times, but no one seems interested. It feels like programmers tend to dislike this kind of work — the kind that doesn’t have a clear logic or defined answer and requires a lot of trial and error over a long period. Everyone seems to think it would be better and easier to just abandon WPF and move to something else. From my experience, I know that it's not that easy — which is why I want to find a solution within WPF.

In any case, I think it would be better if someone more capable than me took over this task and asked you questions. (The person you're talking to — me — is more of a designer and doesn’t really know much about programming.) 

@Flow-Launcher/team

lindexi: @onesounds I'm not familiar with the code of Flow.Launcher, so that I'm afraid I can't help you more. If you need any help, please Call me at any time.

onesounds: > @onesounds I'm not familiar with the code of Flow.Launcher, so that I'm afraid I can't help you more. If you need any help, please Call me at any time.

Yeah. Thank you. 👍

Jack251970: `ClearAllCacheModes` cannot bring any difference. I suddenly found that the freeze window issue disappeared when I opened main window with hotkey and I am not sure if this can help.

onesounds: @Jack251970 This issue basically comes down to the fact that the freeze gets lifted when the primary window is shown. So the behavior you're describing is actually expected. What’s happening is that when the primary window is hidden, something keeps trying to render to it, and that’s causing the second window (in our case, the settings window) to crash.

lindexi: > ClearAllCacheModes cannot bring any difference.

@Jack251970 I not sure this issus is same as https://github.com/dotnet/wpf/issues/2158 . Maybe other case cause the freeze render issues.

onesounds: > > @Jack251970 This issue basically comes down to the fact that the freeze gets lifted when the primary window is shown. So the behavior you're describing is actually expected. What’s happening is that when the primary window is hidden, something keeps trying to render to it, and that’s causing the second window (in our case, the settings window) to crash.
> 
> Sorry for my confusion. I mean the second window starts to work when I brought main window with hotkey.
> 
> Open FL -> Sleep -> Unlock -> Open settings window (Freeze) -> Open main window with hotkey (Settings window starts to render)

You're basically saying the same thing I described. After the lock screen, the settings window freezes, and then when the first window is shown, the settings window becomes active again. That's the basic structure of the changes I made in my initial PR. So you had told me, "There's an issue where the window is being shown," right?

lindexi: @Jack251970 and @onesounds  Can you make sure that no new elements with BitmapCache are added after OnLoaded?

Jack251970: > @Jack251970 and @onesounds Can you make sure that no new elements with BitmapCache are added after OnLoaded? Thank you.
> 
> As [#3384 (comment)](https://github.com/Flow-Launcher/Flow.Launcher/pull/3384#discussion_r2015192917) , what worries me is that elements are added dynamically, and they are added after loading.

For second window, I do think there are any elements added dynamically. For main window, I think there could be some dynamically added elements if we change query text. But even if we do not change query text, the freezing issue still exists.

Jack251970:
Continue in #3593.

PR #3401, opened 2025-03-30
Fix Application.Current null exception

Jack251970:
Fix this exception:

Screenshot 2025-03-30 145619


Jack251970:
Small patch and it is good to me.

onesounds (approved):

PR #3408, opened 2025-04-01
New API Function LogError

Jack251970:
# Add `LogError` API function from `Log.Error` for both csharp and jsonrpc plugins

Firstly, many preinstalled plugins use `Log.Error` function and I want to convert them into api usage.

Secondly, for JsonRPC plugins, they do not have `LogException` and providing them with this api can give them ability to log exception-level information.


Flow.Launcher.Infrastructure/Logger/Log.cs
```diff
@@ -135,12 +135,14 @@ private static string CheckClassAndMessageAndReturnFullClassWithMethod(string cl
             return className;
         }
 
+#if !DEBUG
```
VictoriousRaptor, line 138: Why would we need DEBUG macro here?

Jack251970: > Why would we need DEBUG macro here? Is `ExceptionInternal` only used in DEBUG mode?

yes, this function is only used in https://github.com/Flow-Launcher/Flow.Launcher/blob/a3c18a6e3b2c7827038a030364fe07dfe55515e9/Flow.Launcher.Infrastructure/Logger/Log.cs#L175

Jack251970: reverted

Flow.Launcher/PublicAPIInstance.cs
```diff
@@ -27,7 +27,7 @@
 using Flow.Launcher.Plugin.SharedCommands;
 using Flow.Launcher.ViewModel;
 using JetBrains.Annotations;
-using Flow.Launcher.Core.Resource;
+using Squirrel;
```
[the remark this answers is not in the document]
Jack251970: Reordering namespaces （s is after f)

Flow.Launcher/PublicAPIInstance.cs
```diff
@@ -11,9 +11,9 @@
 using System.Threading.Tasks;
 using System.Windows;
 using CommunityToolkit.Mvvm.DependencyInjection;
-using Squirrel;
 using Flow.Launcher.Core;
 using Flow.Launcher.Core.Plugin;
+using Flow.Launcher.Core.Resource;
```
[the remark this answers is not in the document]
Jack251970: I adjusted the sequence of the namespace here for code quality. Nothing changed in namespace

Flow.Launcher.Infrastructure/Logger/Log.cs
VictoriousRaptor: If we alter the usage in log.cs we need to also tweak it here.

Jack251970: > If we alter the usage in log.cs we need to also tweak it here.

Could you please explain it further?

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -227,6 +227,11 @@ public interface IPublicAPI
         /// </summary>
         void LogWarn(string className, string message, [CallerMemberName] string methodName = "");
 
+        /// <summary>
+        /// Log error message
+        /// </summary>
+        void LogError(string className, string message, [CallerMemberName] string methodName = "");
```
jjw24, line 233: Why do we need this, doesn't LogException already provides this?

Jack251970: > Why do we need this, doesn't LogException already provide this?

Firstly, many plugins use `Log.Error` function and I want to convert them into api usage.

Secondly, for JsonRPC plugins, they do not have `LogException` and providing them with this api can give them ability to log exception-level information.

Flow.Launcher.Core/Plugin/JsonRPCV2Models/JsonRPCPublicAPI.cs
```diff
@@ -156,6 +155,11 @@ public void LogWarn(string className, string message, [CallerMemberName] string
             _api.LogWarn(className, message, methodName);
         }
 
+        public void LogError(string className, string message, [CallerMemberName] string methodName = "")
```
cibere, line 158: I'm not familiar with the `[CallerMemberName]` syntax. Is that just apart of the typing of the `methodName` parameter?

How would this work for json-rpc plugins? Would they make a request like (this is a mockup, I don't remember the exact structure):
```json
{
"method": "LogError",
"params": ["plugin_name", "traceback/stacktrace", "name of method that the plugin was handling"]
}
```

Jack251970: > I'm not familiar with the `[CallerMemberName]` syntax. Is that just apart of the typing of the `methodName` parameter?
> 
> How would this work for json-rpc plugins? Would they make a request like (this is a mockup, I don't remember the exact structure):
> ```json
> {
> "method": "LogError",
> "params": ["plugin_name", "traceback/stacktrace", "name of method that the plugin was handling"]
> }
> ```

No sure how to use it, I think we need to manually pass methodName. Just forget CallMemberName.

And usage is the same as LogInfo etc.

Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -227,6 +227,11 @@ public interface IPublicAPI
         /// </summary>
         void LogWarn(string className, string message, [CallerMemberName] string methodName = "");
 
+        /// <summary>
+        /// Log error message
```
jjw24, line 231:
```suggestion
        /// Log error message. Preferred error logging method for plugins.
```

jjw24 (approved):

PR #3413, opened 2025-04-02
New API Function ValidateDirectory and ValidateDataDirectory & Cleanup Helper.cs in Flow.Launcher.Infrastructure project

Jack251970:
# Move format function location

Move `Flow.Launcher.Infrastructure.Format` function to `Updater.cs` since this function is only used by `Updater.cs`.

# Move `ValidateDirectory` functions location

Move `ValidateDataDirectory()` and `ValidateDirectory()` function in `Flow.Launcher.Infrastructure` project to `FilesFolders.cs` since they are functions related to directory and `Program` and `WebSearch` plugin use them so we would better to put them in `Flow.Launcher.Plugin` project for all plugin usage.

### DIFF
--- a/Flow.Launcher.Core/Updater.cs
+++ b/Flow.Launcher.Core/Updater.cs
@@ -17,6 +17,7 @@ using Flow.Launcher.Infrastructure.UserSettings;
 using Flow.Launcher.Plugin;
 using System.Text.Json.Serialization;
 using System.Threading;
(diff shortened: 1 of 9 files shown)


Jack251970:
I think it is good to merge since it only provide more useful static functions.

onesounds (approved):

PR #3367, opened 2025-03-22
Graceful shutdown

Jack251970:
# Graceful shutdown

Refer to PowerToys Run.

* Add diposable interface for `MainViewModel`
* Add diposable interface for `MainWindow.xaml.cs`
* Remove hook event for `MainWindow.xaml.cs`
* Improve code quality for `SingleInstance.cs`
* Add more logs for FL exit events
* Dispose  `MainWindow` and `MainViewModel` when disposing `App`

# TODO

- [x] Fix `PluginManager.DisposePluginsAsync` cannot be waited

PR #3417, opened 2025-04-03
Fix environment exit & Wait image cache save before restarting & Add settings saving lock & Do not crash when caught exception saving

Jack251970:
# Fix `Environment.Exit(0)` issue

Since #3367 has introduced `Application.Current.Shutdown` for app dispose, using `Environment.Exit(0)` will cause issue.

```
// If we call Environment.Exit(0), the application dispose will be called before _mainWindow.Close()
// Accessing _mainWindow?.Dispatcher will cause the application stuck
// So here we need to check it and just return so that we will not acees _mainWindow?.Dispatcher
```

# Wait image cache save

Before restarting app, we should wait `ImageLoader.SaveAsync`.

# Add settings save lock

Make multiple thread safe.

# Do not crash when caught exception in saving

Use try when calling `JsonStorage` and `BinaryStorage` save.

# Test

- [x] Install original dev version and install one plugin. When application is exited for restarting, it will stuck. Using installer in this PR, the issue is fixed.

### DIFF
--- a/Flow.Launcher.Core/Configuration/Portable.cs
+++ b/Flow.Launcher.Core/Configuration/Portable.cs
@@ -159,7 +159,7 @@ namespace Flow.Launcher.Core.Configuration
                 {
                     FilesFolders.OpenPath(Constant.RootDirectory, (s) => API.ShowMsgBox(s));
 
-                    Application.Current.Shutdown();
+                    Environment.Exit(0);
                 }
             }
             // Otherwise, if the portable data folder is marked for deletion,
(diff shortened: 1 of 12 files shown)


Jack251970:
Tested and good to go

onesounds (approved):

Jack251970:
Sorry, I forget to fix typos here and please review.

onesounds (approved):

Jack251970:
Sorry for asking your review again. I didn't handle conflict resolution well causing me to need an additional commit to fix it.

onesounds (approved):

PR #3429, opened 2025-04-08
New API Function from Stopwatch & Improve Program Plugin

Jack251970:
# New API Function from Stopwatch

Added new api functions to helper csharp plugin developers quickly log debug & info message with function executing time.

```
/// <summary>
/// Log debug message of the time taken to execute a method
/// Message will only be logged in Debug mode
/// </summary>
/// <returns>The time taken to execute the method in milliseconds</returns>
public long StopwatchLogDebug(string className, string message, Action action, [CallerMemberName] string methodName = "");

/// <summary>
/// Log debug message of the time taken to execute a method asynchronously
/// Message will only be logged in Debug mode
/// </summary>
/// <returns>The time taken to execute the method in milliseconds</returns>
public Task<long> StopwatchLogDebugAsync(string className, string message, Func<Task> action, [CallerMemberName] string methodName = "");

/// <summary>
/// Log info message of the time taken to execute a method
/// </summary>
/// <returns>The time taken to execute the method in milliseconds</returns>
public long StopwatchLogInfo(string className, string message, Action action, [CallerMemberName] string methodName = "");

/// <summary>
/// Log info message of the time taken to execute a method asynchronously
/// </summary>
/// <returns>The time taken to execute the method in milliseconds</returns>
public Task<long> StopwatchLogInfoAsync(string className, string message, Func<Task> action, [CallerMemberName] string methodName = "");
```

# Improve Program Plugin

Use api functions instead of project reference in Program plugin.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginsLoader.cs
+++ b/Flow.Launcher.Core/Plugin/PluginsLoader.cs
@@ -11,12 +11,17 @@ using Flow.Launcher.Infrastructure.Logger;
 #pragma warning restore IDE0005
 using Flow.Launcher.Infrastructure.UserSettings;
 using Flow.Launcher.Plugin;
-using Stopwatch = Flow.Launcher.Infrastructure.Stopwatch;
 
 namespace Flow.Launcher.Core.Plugin
 {
     public static class PluginsLoader
     {
+        private static readonly string ClassName = nameof(PluginsLoader);
+
+        // We should not initialize API in static constructor because it will create another API instance
+        private static IPublicAPI api = null;
+        private static IPublicAPI API => api ??= Ioc.Default.GetRequiredService<IPublicAPI>();
+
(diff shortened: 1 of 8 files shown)


Flow.Launcher.Core/Plugin/PluginsLoader.cs
```diff
@@ -11,12 +11,17 @@
...
 
 namespace Flow.Launcher.Core.Plugin
 {
     public static class PluginsLoader
     {
+        private static readonly string ClassName = nameof(PluginsLoader);
+
+        // We should not initialize API in static constructor because it will create another API instance
+        private static IPublicAPI api = null;
+        private static IPublicAPI API => api ??= Ioc.Default.GetRequiredService<IPublicAPI>();
```
jjw24, line 23: Not an issue for now but we should think about a better way to safeguard this, maybe a design pattern we can use or something.

jjw24 (approved):

PR #3435, opened 2025-04-09
Dependency Code Quality

Jack251970:
# Improve code quality

* Remove obsoleted `InternationalizationManager.Instance` and use api function `App.API.GetTranslation` instead.
* Remove `Notification.Show` and use api function `App.API.ShowMsg` instead.
* Small fixes and improments.


Jack251970:
I think it is good to go since it does not change any data models.

onesounds (approved):

PR #3452, opened 2025-04-11
Fix storage save directory issue

Jack251970:
Check directory before saving storage to fix `System.IO.DirectoryNotFoundException: Could not find a part of the path`.
Add old constructor for old `Program` plugin.

Fix #3451.

PR #3456, opened 2025-04-12
Improve FL Exiting & Exception Handler

### LINKED ISSUE #3453: BUG: Ignore errors on exit, so that Flow launcher can exit properly, and not keep a ghost process in background
So these issues #3428 and #3409 and more are the main reasons i am making this.

Solution:

I want that flow can exit and ignore the errors on exit if happened.

Alternatives:

~~Maybe add an toggle?~~

Add a watcher process so that this process watches for errors and then confirms if this is a startup error, a plugin error, a Flow Launcher error (like everything not installed-error), or a **Flow Launcher Exit-time** error.

The watcher process only becomes super active on the startup, an execution of a important feature, like installation of plugins, read and write on disk, **an exception** (That's a very important one), and on flow launcher exit.

The watcher process can also be made to startup earlier than the flow executable itself.

Jack251970:
# Use try-catch to call plugin Save & Dispose method so that Application can exit normally even plugins have exceptions

Tested. When plugin throw exception during saving & disposing, the application can exit normally.

# Fix storage DirectoryPath issue

Make `DirectoryPath` is not null. Follow on with #3452.

# Support TaskScheduler.UnobservedTaskException handler

Fix #3453.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -37,7 +37,7 @@ namespace Flow.Launcher.Core.Plugin
 
         private static PluginsSettings Settings;
         private static List<PluginMetadata> _metadatas;
-        private static List<string> _modifiedPlugins = new();
+        private static readonly List<string> _modifiedPlugins = new();
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


Jack251970:
I think it is good to go

onesounds (approved):

PR #3460, opened 2025-04-13
Improve log & Ignore plugin source http exception

Jack251970:
# Improve log

For all other projects, we use API to get log functions.

For Infrastructure project, we use Log static class to get log functions.

# Ignore plugin source http exception

Ignore exceptions like:

```

Please open new issue in https://github.com/Flow-Launcher/Flow.Launcher/issues
1. Upload log file: C:\Users\11602\AppData\Roaming\FlowLauncher\Logs\1.19.5\2025-04-13.txt
2. Copy below exception message

Flow Launcher version: 1.19.5
OS Version: 26100.3775
IntPtr Length: 8
x64: True

Python Path: C:\Users\11602\AppData\Roaming\FlowLauncher\Environments\Python\PythonEmbeddable-v3.11.4\pythonw.exe
Node Path: C:\Program Files\nodejs\node.exe

Date: 04/13/2025 15:45:04
Exception:
System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (The requested address is not valid in its context. (raw.githubusercontent.com:443))
 ---> System.Net.Http.HttpRequestException: The requested address is not valid in its context. (raw.githubusercontent.com:443)
 ---> System.Net.Sockets.SocketException (10049): The requested address is not valid in its context.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at System.Net.Sockets.Socket.<ConnectAsync>g__WaitForConnectWithCancellation|281_0(AwaitableSocketAsyncEventArgs saea, ValueTask connectTask, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
[...]

PR #3472, opened 2025-04-17
Fix Possible Window Popup During Startup & Fix Store Plugin List Refresh Issue

Jack251970:
# Fix Possible Window Popup During Startup

If FL enabled both logon task and registry,  `Hide Flow Launcher on startup` will not work since the later one will trigger main window show event. So we need to check both because if both of them are enabled.

Follow on with #3218.

# Fix Store Plugin List Refresh Issue

Use null as flag to return value so that we can handle plugin source issue.

Follow on with #3460.


Jack251970:
Tested and I think it is good to go.

onesounds (approved):

PR #3480, opened 2025-04-21
Remember selected position

onesounds:
# Improve SelectedIndex Retention on Query Updates and Context Menu Return

## Summary

This PR improves the behavior of `SelectedIndex` to prevent it from resetting to 0 in the following scenarios:

- When loading occurs during a query search and completes
- When returning from an opened context menu

By avoiding unnecessary index resets, the user experience becomes more consistent and intuitive.

---

### Main Changes

- **Retain SelectedIndex on context menu return:**  
  Previously, returning from a context menu caused the query to be re-applied, which reset the index to 0.  
  A new logic now updates the `QueryText` without refreshing the result list, preventing index reset in this case.

- **Handle null QueryText:**  
  When `QueryText` becomes `null`, the stored index is now properly cleared.

- **Changed behavior for users with "retain query" mode enabled:**  
  Previously, the index would reset even if the query remained the same.  
  Now, the index is preserved as long as the query is not cleared.  
  This is considered more intuitive, so no setting toggle was added.

---

## Additional Fixes

- **Prevent simultaneous display of context menu and result list during loading:**  
  A bug was discovered where opening the context menu before loading completed could result in both the context menu and results being shown at the same time.  
  The update logic now checks for open context menus and skips result updates if one is active.

---

## Test Cases

- On initial launch, typing a query should place the index at 0.
- There should never be a case where no result is selected (i.e., index must always be valid).
- Opening and closing the context menu should not reset the selection.
- Loading during search should no longer reset the index on completion.

During testing, there were a few instances where no index was selected, so further testing is needed.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -1125,6 +1125,13 @@ namespace Flow.Launcher
         {
             var textBox = (TextBox)sender;
             _viewModel.QueryText = textBox.Text;
+    
+            // Reset Index when query null
+            if (string.IsNullOrEmpty(textBox.Text))
+            {
+                _viewModel.Results.ResetSelectedIndex();
+            }
+    
             _viewModel.Query(_settings.SearchQueryResultsWithDelay);
         }
 
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -34,6 +34,7 @@ namespace Flow.Launcher.ViewModel
         private bool _isQueryRunning;
         private Query _lastQuery;
         private string _queryTextBeforeLeaveResults;
+        private bool _blockQueryExecution = false;
 
         private readonly FlowLauncherJsonStorage<History> _historyItemsStorage;
         private readonly FlowLauncherJsonStorage<UserSelectedRecord> _userSelectedRecordStorage;
@@ -638,20 +639,19 @@ namespace Flow.Launcher.ViewModel
         /// </summary>
         /// <param name="queryText"></param>
         /// <param name="isReQuery">Force query even when Query Text doesn't change</param>
-        public void ChangeQueryText(string queryText, bool isReQuery = false)
+        public void ChangeQueryText(string queryText, bool isReQuery = false, bool suppressQueryExecution = false)
         {
-            _ = ChangeQueryTextAsync(queryText, isReQuery);
+            _ = ChangeQueryTextAsync(queryText, isReQuery, suppressQueryExecution);
         }
-
         /// <summary>
         /// Async version of <see cref="ChangeQueryText"/>
         /// </summary>
-        private async Task ChangeQueryTextAsync(string queryText, bool isReQuery = false)
+        private async Task ChangeQueryTextAsync(string queryText, bool isReQuery = false, bool suppressQueryExecution = false)
         {
             // Must check access so that we will not block the UI thread which cause window visibility issue
             if (!Application.Current.Dispatcher.CheckAccess())
             {
-                await Application.Current.Dispatcher.InvokeAsync(() => ChangeQueryText(queryText, isReQuery));
+                await Application.Current.Dispatcher.InvokeAsync(() => ChangeQueryText(queryText, isReQuery, suppressQueryExecution));
                 return;
             }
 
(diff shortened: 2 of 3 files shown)


onesounds:
@coderabbitai Review it please.

onesounds:
@copilot Review it

Flow.Launcher/ViewModel/ResultsViewModel.cs
```diff
@@ -184,36 +200,98 @@ public void AddResults(List<Result> newRawResults, string resultId)
...
+            {
+                if (SelectedItem != null)
+                {
+                    lastSelectedItem = SelectedItem;
+                    lastSelectedIndex = SelectedIndex;
+                }
+            });
+
+            // 캡처한 값 저장
+            _lastSelectedItem = lastSelectedItem;
```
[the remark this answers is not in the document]
onesounds: Fixed

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -1125,6 +1125,13 @@ private void QueryTextBox_TextChanged1(object sender, TextChangedEventArgs e)
         {
             var textBox = (TextBox)sender;
             _viewModel.QueryText = textBox.Text;
+    
+            // Reset Index when query null
```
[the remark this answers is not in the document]
onesounds: Rabbit made a similar review, but in my opinion, these are used in different contexts, so I think they should remain separated

Flow.Launcher/ViewModel/ResultsViewModel.cs
```diff
@@ -184,36 +200,97 @@ public void AddResults(List<Result> newRawResults, string resultId)
...
-                    break;
-                case Visibility.Visible when Results.Count == 0:
-                    Visibility = Visibility.Collapsed;
-                    break;
+                }
+            }
+            else if (Visibility == Visibility.Visible && Results.Count == 0)
+            {
+                Visibility = Visibility.Collapsed;
             }
```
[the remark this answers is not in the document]
onesounds: Applied.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -701,11 +711,31 @@ private ResultsViewModel SelectedResults
...
+                    if (isReturningFromContextMenu && _lastSelectedResultItem != null)
                     {
-                        _queryText = _queryTextBeforeLeaveResults;
-                        OnPropertyChanged(nameof(QueryText));
-                        QueryTextCursorMovedToEnd = true;
+                        try
+                        {
+                            _blockQueryExecution = true;
+                            // Set querytext without running the query for keep index. if not use this, index will be 0.
+                            _queryText = _queryTextBeforeLeaveResults;
```
taooceros, line 720: why not just change QueryText

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -701,11 +711,31 @@ private ResultsViewModel SelectedResults
...
+                            OnPropertyChanged(nameof(QueryText));
+                            QueryTextCursorMovedToEnd = true;
+                        }
+                        finally
+                        {
+                            _blockQueryExecution = false;
+                        }
+                        // restore selected item
+                        if (Results.Results.Contains(_lastSelectedResultItem))
```
taooceros, line 729: Would this be slow for large results? And I am curious when will Results not contains the last selected result item? Why not just use _lastSelectedIndex

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1620,67 +1660,27 @@ public void Save()
         /// </summary>
         public void UpdateResultView(ICollection<ResultsForUpdate> resultsForUpdates)
         {
-            if (!resultsForUpdates.Any())
+            if (resultsForUpdates == null || !resultsForUpdates.Any())
                 return;
-            CancellationToken token;
 
+            // Block the query execution when Open ContextMenu is called
+            if (!QueryResultsSelected())
```
taooceros, line 1667: I don't understand here. Query shouldn't be executed when context menu is called? It will only call QueryContextMenu?

Flow.Launcher/ViewModel/MainViewModel.cs
onesounds: I noted this in the PR description:

If you simply update the `queryText`, the index resets to 0 when the results are refreshed. The flow goes like this:

1. Remember the previously selected item  
2. Open the context menu (which clears the query box text)  
3. Close the context menu (the previous query text is restored)

Even if the selected index is correctly remembered and restored at step 3, it becomes pointless because restoring the `queryText` causes the index to reset to 0 again.

Flow.Launcher/ViewModel/MainViewModel.cs
onesounds: I already mentioned this in the PR description. If you open the context menu while results are still loading, both the context menu and the results panel end up being shown once the loading finishes. Please read the PR description. Why do you always ignore what others have already said? I always write it down, whether it's on Discord or here, but you keep asking the same questions. You’re making me explain the same thing over and over again.

Flow.Launcher/ViewModel/MainViewModel.cs
onesounds: OK. Modified it to work using only the index.

onesounds:
@coderabbitai Review it.

onesounds:
Fixed in #3503

PR #3482, opened 2025-04-22
Fix clock not displaying in certain situations

onesounds:
### 🛠 Fix clock not displaying on first launch (when animation is enabled)

- When `HideOnStartup = off` and `Animation = on`, there was a bug where the clock would **not appear at all on the initial launch**.
- After the first launch, the clock behaves normally.
- This PR addresses the issue by **forcibly triggering the animation on `OnLoaded`**, ensuring the clock becomes visible on the first load.

### ⚠️ Notes

- Even after this fix, **the animation might appear to be missing** due to rendering timing.  
  If you slow down the animation duration, you’ll notice it’s actually running.
- Adjusting the position of the animation trigger within `OnLoaded` can make the animation look better.  
  However, doing so caused a **slight delay in visibility when animation was disabled**.
- While a more refined fix is possible, it would **increase complexity**.  
  Since the clock **does show up correctly**, and this is the key UX issue, no further changes were made.

### 💡 UX Consideration

- The lack of visible animation on the first load is likely due to **rendering timing** and is not perceived as unnatural by most users.
- If users ask, just say:  
  _"The animation was too fast because your PC is so fast."_ 😄

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -138,8 +138,12 @@ namespace Flow.Launcher
             else
             {
                 _viewModel.Show();
+                if (_settings.UseAnimation)
+                {
+                    WindowAnimation();
+                }
             }
-
+            
             // Show notify icon when flowlauncher is hidden
             InitializeNotifyIcon();
 
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -138,6 +138,9 @@ namespace Flow.Launcher
             else
             {
                 _viewModel.Show();
+                // When HideOnStartup is off and UseAnimation is on,
+                // there was a bug where the clock would not appear at all on the initial launch
+                // So we need to forcibly trigger animation here to ensure the clock visible
                 if (_settings.UseAnimation)
                 {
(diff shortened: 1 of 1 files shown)


taooceros:
Could we just set the location of the clock on load rather than running the animation?

onesounds:
> Could we just set the location of the clock on load rather than running the animation?

You want to just show it without any animation? This involves changing the alpha value and making an exception for the initial launch timing, so it could potentially affect the existing show/hide logic. Are you sure that’s okay? Later on, during maintenance, we’ll probably have to come back and work on this part separately. If you’re fine with that, I’ll go ahead and do it.

onesounds:
@coderabbitai Review it. Please also review taooceros's suggestion and include that in the discussion.

Jack251970:
> > Could we just set the location of the clock on load rather than running the animation?
> 
> You want to just show it without any animation? This involves changing the alpha value and making an exception for the initial launch timing, so it could potentially affect the existing show/hide logic. Are you sure that’s okay? Later on, during maintenance, we’ll probably have to come back and work on this part separately. If you’re fine with that, I’ll go ahead and do it.

I tried other methods but none of them can help. If you think this can fix all things, I agree with you to use this.

Jack251970 (approved):

PR #3486, opened 2025-04-23
Let more controls follow setting window font & Add Use Logon Task option in welcome page 5

Jack251970:
# Let more controls follow setting window font

Follow on with #3470.

This PR adds more controls like button, combo box, plugin settings control (preinstalled Csharp plugins & JsonRPC plugins), etc. And this PR also exposes setting window font in `{DynamicResource SettingWindowFont}` so that plugin developers can use them in their settings control.

# Allow users to enable Use Logon Task in welcome page 5

This PR adds Use Logon Task check box in welcome page 5.

### DIFF
--- a/Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
+++ b/Flow.Launcher.Core/Plugin/JsonRPCPluginSettings.cs
@@ -215,7 +215,7 @@ namespace Flow.Launcher.Core.Plugin
                             TextWrapping = TextWrapping.WrapWithOverflow
                         };
 
-                        desc.SetResourceReference(TextBlock.StyleProperty, "SettingPanelTextBlockDescriptionStyle"); // for theme change
+                        desc.SetResourceReference(FrameworkElement.StyleProperty, "SettingPanelTextBlockDescriptionStyle"); // for theme change
                     }
 
                     // Add the name and description to the panel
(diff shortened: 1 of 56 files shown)


onesounds:
Wow, I can’t believe you added all that. If this feature was necessary, you should’ve just told me. I’m not entirely sure, but if SettingFont can be used as a DynamicResource, then instead of setting FontFamily on every control, it might be possible to simply override it in the styles by adding a FontFamily setter for each control type. (Like how controls inside a CustomControlTemplate can be styled to reflect font changes elsewhere.)

The current approach, after merging this, effectively forces developers to set FontFamily in every XAML control, doesn’t it? That’s not something devs are going to be happy about. Anyway, it does work.

onesounds:
@coderabbitai Review it. This PR currently adds FontFamily settings to every control, but instead of doing that, isn’t there a simpler way to override the styles of specific controls (like ComboBox, Button, etc.) to apply a certain font?

What I want is: if a control doesn’t have a FontFamily explicitly set in XAML, it should use font “A” by default. But if the FontFamily is set in XAML, then that should take precedence. I don’t want to manually set FontFamily every time I add a new UI element in the future.

Jack251970:
> Wow, I can’t believe you added all that. If this feature was necessary, you should’ve just told me. I’m not entirely sure, but if SettingFont can be used as a DynamicResource, then instead of setting FontFamily on every control, it might be possible to simply override it in the styles by adding a FontFamily setter for each control type. (Like how controls inside a CustomControlTemplate can be styled to reflect font changes elsewhere.)
> 
> The current approach, after merging this, effectively forces developers to set FontFamily in every XAML control, doesn’t it? That’s not something devs are going to be happy about. Anyway, it does work.

Oh, I have not noticed you have set fonts here...

Jack251970:
But you still need to explicitly set style by its key here.

Jack251970:
Well, I get it. I will check it then.

Jack251970:
Implemented in #3487

Issue #3498, opened 2025-04-30
Dev branch- results jump around when return from context menu


jjw24:
Results jump around when return from context menu on a dev build fresh install. Notice in the screenshot the WebSearch result '1' appears on search and first time return from context menu, but disappears (drops down the list) after the second return from context menu. 

This is actually hard to reproduce consistently like in the screenshot below.

![Image](https://github.com/user-attachments/assets/8686c028-7b98-4d1d-8b7d-8aa3f8f76a99)

PR #3503, opened 2025-05-02
Do not query again when setting private field of QueryText

### LINKED ISSUE #3497: Dev branch- return to result list from context menu jump back to top result
Return to result list from context menu jump back to top result in the dev branch build:

Production v1.19.5

Image

Dev build:

Image

Jack251970:
# Do not query again when setting private field of QueryText

From #3350, we use TextChanged event to invoke Query, which can cause duplicated call of Query if we use OnPropertyChange(nameof(QueryText)).

In this PR, we use 

```
// When executing OnPropertyChanged, QueryTextBox_TextChanged1 and Query will be called
// So we need to ignore it so that we will not call Query again
_ignoredQueryText = _queryText;
```

to resolve that.

Fix #3497, #3498.

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -34,6 +34,7 @@ namespace Flow.Launcher.ViewModel
         private bool _isQueryRunning;
         private Query _lastQuery;
         private string _queryTextBeforeLeaveResults;
+        private string _ignoredQueryText = null;
 
         private readonly FlowLauncherJsonStorage<History> _historyItemsStorage;
         private readonly FlowLauncherJsonStorage<UserSelectedRecord> _userSelectedRecordStorage;
@@ -730,6 +731,9 @@ namespace Flow.Launcher.ViewModel
                     if (isReturningFromContextMenu)
                     {
                         _queryText = _queryTextBeforeLeaveResults;
+                        // When executing OnPropertyChanged, QueryTextBox_TextChanged1 and Query will be called
+                        // So we need to ignore it so that we will not call Query again
+                        _ignoredQueryText = _queryText;
                         OnPropertyChanged(nameof(QueryText));
                         QueryTextCursorMovedToEnd = true;
                     }
@@ -1076,6 +1080,20 @@ namespace Flow.Launcher.ViewModel
 
         public void Query(bool searchDelay, bool isReQuery = false)
         {
+            if (_ignoredQueryText != null)
+            {
+                if (_ignoredQueryText == QueryText)
+                {
+                    _ignoredQueryText = null;
+                    return;
+                }
+                else
+                {
+                    // If _ignoredQueryText does not match current QueryText, we should still execute Query
+                    _ignoredQueryText = null;
+                }
+            }
+
             if (QueryResultsSelected())
             {
                 _ = QueryResultsAsync(searchDelay, isReQuery);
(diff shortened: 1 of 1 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1076,6 +1080,20 @@ private bool QueryResultsPreviewed()
 
         public void Query(bool searchDelay, bool isReQuery = false)
         {
+            if (_ignoredQueryText != null)
```
jjw24, line 1083: Do we need to check for empty string as well to consider in the future when we display saved/favourited results on start- up which will have empty query text?

Jack251970: > Do we need to check for empty string as well to consider in the future when we display saved/favourited results on start-up which will have empty query text?

Startup will not trigger this event and we can handle that in future.

Jack251970: And here I use null to check, so empty string will be checked here.

jjw24: Ok but fyi empty != null on a string though.

jjw24 (approved):

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: yes, so if the ignored string is empty, it will  be check here?

Jack251970: I get it. Here I use null as an invalid flag so that I can check all strings including empty string

jjw24: From reading it just two thoughts come to mind:
1. _queryTextBeforeLeaveResults sounds like it's the query text before entering context menu? That's why I thought if query ever be `""` when user enters context menu and back, the query async will execute again, which would cause the jumping back to first result issue?
2. It's assigning an empty string on class instantiation, would this empty string (`""`) ever get to this part of the code?
https://github.com/Flow-Launcher/Flow.Launcher/blob/247272c9aa7a986395a14beab3d6a5b2e22e4b75/Flow.Launcher/ViewModel/MainViewModel.cs#L60

Basically, I am just wondering if it is necessary to use string.IsNullOrEmpty?

Jack251970: > From reading it just two thoughts come to mind:
> 
> 1. _queryTextBeforeLeaveResults sounds like it's the query text before entering context menu? That's why I thought if query ever be `""` when user enters context menu and back, the query async will execute again, which would cause the jumping back to first result issue?
> 2. It's assigning an empty string on class instantiation, would this empty string (`""`) ever get to this part of the code?
>    https://github.com/Flow-Launcher/Flow.Launcher/blob/247272c9aa7a986395a14beab3d6a5b2e22e4b75/Flow.Launcher/ViewModel/MainViewModel.cs#L60
> 
> Basically, I am just wondering if it is necessary to use string.IsNullOrEmpty?

I think we should still consider empty as valid query text. (In future, I am planning to provider query feature even if query is empty: see #3399) So I think we should use IsNull instead of IsNullOrEmpty.

jjw24: Ok no worries.

PR #3525, opened 2025-05-10
Fix history results clear logic & Make sure results cleared

Jack251970:
Follow on with #3524 to fix two issues in that PR.

### 1. Clear existing results for history results query function

We also need to check `shouldClearExistingResults` when querying history results.

### 2. Make sure results cleared when update tasks are not called

When update tasks are not called, we need to clear results manually afterwards. This can fix an issue that results are not cleared when Home Page feature is off, and `QueryTaskAsync` and `QueryHistoryTask` are not called.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -105,7 +105,7 @@ namespace Flow.Launcher
             Win32Helper.DisableControlBox(this);
         }
 
-        private async void OnLoaded(object sender, RoutedEventArgs _)
+        private void OnLoaded(object sender, RoutedEventArgs _)
         {
             // Check first launch
             if (_settings.FirstLaunch)
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -1284,8 +1284,6 @@ namespace Flow.Launcher.ViewModel
             // Update the query's IsReQuery property to true if this is a re-query
             query.IsReQuery = isReQuery;
 
-            
-
             ICollection<PluginPair> plugins = Array.Empty<PluginPair>();
             if (currentIsHomeQuery)
             {
@@ -1458,8 +1456,13 @@ namespace Flow.Launcher.ViewModel
 
                 App.API.LogDebug(ClassName, $"Update results for history");
 
+                // Indicate if to clear existing results so to show only ones from plugins with action keywords
+                var shouldClearExistingResults = ShouldClearExistingResults(query, currentIsHomeQuery);
+                _lastQuery = query;
+                _previousIsHomeQuery = currentIsHomeQuery;
+
                 if (!_resultsUpdateChannelWriter.TryWrite(new ResultsForUpdate(results, _historyMetadata, query,
-                    token)))
+                    token, reSelect, shouldClearExistingResults)))
                 {
                     App.API.LogError(ClassName, "Unable to add item to Result Update Queue");
                 }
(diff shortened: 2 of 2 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -105,7 +105,7 @@ private void OnSourceInitialized(object sender, EventArgs e)
             Win32Helper.DisableControlBox(this);
         }
 
-        private async void OnLoaded(object sender, RoutedEventArgs _)
+        private void OnLoaded(object sender, RoutedEventArgs _)
```
jjw24, line 108: Why we changing this?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1445,6 +1447,11 @@ await PluginManager.QueryHomeForPluginAsync(plugin, query, token) :
                 {
                     App.API.LogError(ClassName, "Unable to add item to Result Update Queue");
                 }
+                else
+                {
+                    // Only update the clear flag when we successfully write to the channel
```
jjw24, line 1452: Do we actually get failure writes?

Jack251970: I think we should still take that into consideration.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1376,6 +1375,9 @@ private async Task QueryResultsAsync(bool searchDelay, bool isReQuery = false, b
                 // nothing to do here
             }
 
+            // If results are not cleared, we need to clear the results
+            if (!resultsCleared) Results.Clear();
+
```
jjw24, line 1380: Is this really necessary? This could cause unexplained/unexpected flicker

Flow.Launcher/MainWindow.xaml.cs
Jack251970: Previous PR removed async function in OnLoaded, so we can remove async modifier here to resolve one more warning.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: Noooo. This is used to make sure results cleared when there are not any update tasks.

Jack251970: Think about one situation. Home page feature is off and you just change query text from `e` to empty. You will find the results will not be cleared because there are not any update tasks. So we should force call `Results.Clear()` here when it is not be called in `QueryResult` or `QueryHistory` functions.

jjw24: When you say not any update task are you referring to one example where you toggle off home page in settings and results are not cleared?

Jack251970: Let me give you one demo video

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: Hmm we probably should take into consideration but setting to true when results could very well not be cleared is misleading. If it failed to add to queue for example.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: @jjw24 It seems that I have issues when uploading video here. Please see more in Discord.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: But that will cause issue as below

Jack251970: Reverted

Jack251970:
Fixed flickering issue when typing very fast

jjw24:
I haven't had time to sit down and think about this. Will find some in the upcoming days.

Jack251970:
I do not know why, but sometimes if I input whitespace, the results are not cleared.

Jack251970:
Continue in #3553.

PR #3529, opened 2025-05-11
Fix Homepage with triggers history results on arrow up

### LINKED ISSUE #3519: Dev branch - Homepage with triggers history results on arrow up
When there is results history, using homepage and press arrow up triggers results history instead of just navigating up:

Image

Jack251970:
Resolve #3519.

Tested:
- When home page feature is on and input is empty, if there are items in result list box, arrow up will trigger result index change instead of history results change.

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -520,7 +520,7 @@ namespace Flow.Launcher.ViewModel
         {
             if (QueryResultsSelected() // Results selected
                 && string.IsNullOrEmpty(QueryText) // No input
-                && Results.Visibility != Visibility.Visible // Results closed which means no items in Results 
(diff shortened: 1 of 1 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -518,9 +518,10 @@ private void SelectNextPage()
         [RelayCommand]
         private void SelectPrevItem()
         {
-            if (_history.Items.Count > 0
-                && QueryText == string.Empty
-                && QueryResultsSelected())
+            if (QueryResultsSelected() // Results selected
+                && string.IsNullOrEmpty(QueryText) // No input
+                && Results.Visibility != Visibility.Visible // Results closed which means no items in Results
```
jjw24, line 523:
```suggestion
                && Results.Visibility != Visibility.Visible // No items in result list, e.g. when home page is off and no query text is entered, therefore the view is collapsed.
```

jjw24 (approved):

PR #3541, opened 2025-05-13
Support Showing Search Window at Topmost

Jack251970:
# Support Showing Search Window At Topmost

Use this option to show window above other windows. Inspired by Fluent Search.

image

# Test

* On/Off the option can set window Topmost state.

https://github.com/user-attachments/assets/c6cbe1cd-4c77-4bab-8659-cdf5039b042c


jjw24:
What's the difference between this and setting hide when lost focus to off?

Jack251970:
> What's the difference between this and setting hide when lost focus to off?

Setting Hide when lost focus off and this on will make this window always on top of all other windows. After showing the search window, it will pin to the top.(z-order)

Inspired by Fluent Search.

Jack251970:
@jjw24 Here is the demo video.

https://github.com/user-attachments/assets/c6cbe1cd-4c77-4bab-8659-cdf5039b042c

onesounds (approved):

PR #3545, opened 2025-05-15
Add dynamic font family to SettingWindow

onesounds:
## Summary

image

- Fixed an issue where the font was not applied to certain elements in the settings window.

## Reproduce
- In the Dev environment, the font in the settings window is changed. You can observe that the font is not applied to titles and card elements within the settings window, while it is correctly applied to regular controls.

## Details

- The root cause is unclear; the font application was working correctly at the time of the previous merge.

- To resolve the issue, the style was explicitly applied to the settings window itself, following the initial implementation approach.

- Verified that other windows still apply fonts correctly without needing explicit style settings.


Jack251970 (approved):
Good!

Issue #1945, opened 2023-03-02
BUG: Run Plugin - No focus when using win+r replacement


LinguaOccultus:
Hello!

When using the setting to replace the run dialogue (Win + R) it shows the command dialogue correctly instead of the normal Flow Launcher dialogue, but focus is lost. I cannot select one of the shown entries or type or anythin. Only hitting enter and starting the default entry does work. What is wrong here?

Greetings!

deefrawley:
Hi, can confirm this. Win hitting win+r the Flow search window doesn't have focus, I have to click in the window to be able to typoe commands and use.
@taooceros could you have a look when you get time? Thanks

LinguaOccultus:
As far as I can remember, this worked several main versions back and then stopped at some point since I have the Win + R replacement enabled from the very beginning of using Flow Launcher. However, this issues is standing for me for quite some version back now. Interestingly, this is not happening with the normal Flow Launcher dialogue. Also, this is not happening every time for me. As I am writing this, I can use the normal dialogue, the normal dialogue with command prefix (>) entered manually  and the Win + R replacement just fine. Bust mostly I am forced in the third case to use the mouse to select one of the autofill suggestions since I cannot type in anything. So I guess, this might be in context with something else.

If I can help at some point with anything, just let me know ...

jjw24:
Should have been resolved with the latest update.

LinguaOccultus:
I cannot confirm that. I am using the most recent stable 1.15.0 and for me, nothing changed with this issue.

However, maybe since I am using the Win + R replacement and trigger it also by using Win + R since I am used to that key to open the run dialogue and then enter RegEdit and stuff before using Flow Launcher.

There might also be another setting which affects this; I cannot tell for sure. I could send in my configuration in our Discord conversation if that helps?

Jack251970:
Fixed in #3568

Jack251970:
Issue still exists in 2.0.0

Jack251970:
@LinguaOccultus Could you please check if this can help? I made some adjustments and I am not sure if this can resolve this issue.

https://ci.appveyor.com/project/JohnTheGr8/flow-launcher/builds/52759143/artifacts

Btw, I enlarged delay time before focusing to 1s, so after clicking win+r, please wait for a while to check if focus works correctly.

LinguaOccultus:
[...]

PR #3568, opened 2025-05-21
Fix: Ensure QueryTextBox receives focus after Win+R

onesounds:
## What's the PR
- Fixes an issue where QueryTextBox did not receive focus when the launcher was shown via Win+R shortcut. (Shell Plugin)
- This was due to OS-level focus interference. Resolved by introducing a short delay before applying focus.

### Reproduce
- In dev build, enable the Win+R setting in the Shell plugin and run the launcher
- Verify that the input box does not receive focus.

### Problem
When launching Flow Launcher via the Win+R shortcut (with the Shell plugin's setting enabled), the main window appears, but the input box (QueryTextBox) does not receive focus, preventing immediate typing.

### Cause
This happens because Win+R is a system-reserved shortcut. Even though the plugin intercepts the keyboard event and shows the main window, Windows continues to process the Win key, and briefly reclaims focus, typically sending it to the Start menu or background. As a result, any Focus() call issued immediately after ShowMainWindow() is overridden.

### Solution 
To work around the OS-level focus interference:

- We introduced a slight delay (Task.Delay) after showing the main window before explicitly calling FocusQueryTextBox().

- This ensures that Windows has completed handling the Win key and the plugin regains control to correctly apply focus.

### DIFF
--- a/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
+++ b/Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
@@ -88,6 +88,11 @@ namespace Flow.Launcher.Plugin
         /// Show the MainWindow when hiding
         /// </summary>
         void ShowMainWindow();
+        
+        /// <summary>
+        /// Focus the query text box in the main window
+        /// </summary>
+        void FocusQueryTextBox();
 
         /// <summary>
         /// Hide MainWindow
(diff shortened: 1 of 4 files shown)


Jack251970 (approved):

LinguaOccultus:
As I reported this initially in #1945 (now closed due to this solution) I can confirm that this happens on my end until today in version 2.0.0 and if you need information or testing later, just let me know.

Jack251970:
> As I reported this initially in #1945 (now closed due to this solution) I can confirm that this happens on my end until today in version 2.0.0 and if you need information or testing later, just let me know.

It seems that #1945 is still not fixed in the latest version. I will open that issue.

PR #3577, opened 2025-05-23
Do not cancel tasks which is for clearing existing results

### LINKED ISSUE #3576: Dev Branch: Results Clear Issue
Typing from empty to one word, I found the home query results are not cleared.

[2025-05-22.txt](https://github.com/user-attachments/files/20403516/2025-05-22.txt)

Jack251970:
Resolve #3576

### DIFF
--- a/Flow.Launcher/ViewModel/MainViewModel.cs
+++ b/Flow.Launcher/ViewModel/MainViewModel.cs
@@ -212,7 +212,8 @@ namespace Flow.Launcher.ViewModel
                     await Task.Delay(20);
                     while (channelReader.TryRead(out var item))
                     {
-                        if (!item.Token.IsCancellationRequested)
+                        // If this update task is for clearing existing results, we must need to add it to the queue
+                        if (!item.Token.IsCancellationRequested || item.shouldClearExistingResults)
                             queue[item.ID] = item;
                     }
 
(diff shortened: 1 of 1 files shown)


Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1460,8 +1462,9 @@ void QueryHistoryTask(CancellationToken token)
                 _lastQuery = query;
                 _previousIsHomeQuery = currentIsHomeQuery;
 
+                // If this update task is for clearing existing results, do not pass token to make sure it is handled
```
jjw24, line 1465: What scenario could have this occur?

Jack251970: If one task which needs to clear existing results, we should make sure it is handled in `Results`. So we need to ignore its origin token here.

jjw24: If a task is cancelled, it wouldn't get to this part of the code right

Jack251970: Please see the update flow above.

Jack251970: This is still in the process to update the results though.

jjw24: OK #3580 does help demonstrate the potential issue. One thing I am not sure on is if we do this, wouldn't the results that were supposed to be cancelled be shown because they are marked to be have results cleared? Even if it's for a moment. For example, if I type 'b' and before results are shown I change to typing 's', instead of the results from 'b' be cancelled, theoretically they would be shown before the results from 's' are updated.

jjw24: I suppose in reality because everything is happening so fast this is probably a non-issue. Let's move forward with this PR.

Jack251970: Since `public void Update(List<ResultViewModel> newItems, CancellationToken token = default)` should be the last step to update the results, so can I set a flag to check if the results are finally updated into results.

jjw24: Actually in this PR we are just trying to handle the case where token is cancelled but require clearing results still, so we can just capture this where token is cancelled and require clearing are true then return empty result set in that NewResults method.

Jack251970: @jjw24 Please go ahead with this. I find I am unable to do that

Jack251970: I am little unfamiliar with this clear logic and the only thing I can change is about handling with token cancellation.

Jack251970:
Continue in #3588.

PR #3588, opened 2025-05-26
Use new model to clear results & Fix clear existing results when using IResultUpdate

Jack251970:
Follow on with #3524.

# CHANGE

- Add new property in Query class `IsHomeQuery` to indicate home query state
- Handle `ShouldClearExistingResultsForQuery` in `UpdateActionAsync`
    - We will skip if token is cancelled there
    - We will also check clear existing results in `IResultUpdate` event to fix the issue that one query consumes much time, the results will be cleared once the results updated (as below)
- Remove token cancellation check between `var newResults = NewResults(resultsForUpdates);` & `if (Count != 0) RemoveAll(newItems.Count);` or `Clear();`
    - All clear actions will be executed.
- Check token cancellation before `AddAll(newItems, token);` & `BulkAddAll(newItems, token);` 
    - We will not add new items from the cancelled queries. It can make sure we only need to clear existing results for one time.
- Use local cancellation in `ResultCollection` for code quality.
- Pass shouldClearExistingResults flag when overriding results in `queue`.

# Issue Example

image

The first item is from new query while others are from old one. If we do not add clear existing results when using IResultUpdate and this query happens to consume long time, the list will display old items for a long time until the query is completed and updated.

# TEST

- Testing


jjw24:
I hit the result not updating issue with this PR again after several times typing in random jibberish:
result not updating
result not updating2

Should clear and clear results are not being called.

Jack251970:
@jjw24 Do you think this PR has any issues in clear logic?

Jack251970:
@jjw24 Please see changes in description above. I removed global flag.

Jack251970:
> Yeah I will test drive it for a couple of days. Could you do the same please.

Okay, will do

jjw24:
Yeah ok. Seems good so far, I will test drive it one more day to make sure.

jjw24:
This iteration of the change has the code really clear now, I am happy with this change. Thank you for taking the time to do this.

Jack251970:
Yeah, good

jjw24:
@coderabbit review

jjw24:
@coderabbitai review

jjw24 (approved):

PR #3596, opened 2025-05-30
Update delete confirmation message

onesounds:
## What's the PR
**Before**
image

**After**
image

- Added a "Title" to the delete confirmation MessageBox in the Explorer plugin.
- The title uses the same key as the "Delete" option in the context menu.
- Fixed an issue where the dialog could not be closed using the ESC key.

### DIFF
--- a/Flow.Launcher/MessageBoxEx.xaml.cs
+++ b/Flow.Launcher/MessageBoxEx.xaml.cs
@@ -159,12 +159,19 @@ namespace Flow.Launcher
 
         private void KeyEsc_OnPress(object sender, ExecutedRoutedEventArgs e)
         {
-            if (_button == MessageBoxButton.YesNo)
-                return;
-            else if (_button == MessageBoxButton.OK)
-                _result = MessageBoxResult.OK;
-            else
-                _result = MessageBoxResult.Cancel;
+            switch (_button)
+            {
+                case MessageBoxButton.OK:
+                    _result = MessageBoxResult.None; 
+                    break;
+                case MessageBoxButton.OKCancel:
+                case MessageBoxButton.YesNoCancel:
+                    _result = MessageBoxResult.Cancel;
+                    break;
+                case MessageBoxButton.YesNo:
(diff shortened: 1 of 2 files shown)


Flow.Launcher/MessageBoxEx.xaml.cs
```diff
@@ -159,12 +159,19 @@ private async Task SetImageAsync(string imageName)
 
         private void KeyEsc_OnPress(object sender, ExecutedRoutedEventArgs e)
         {
-            if (_button == MessageBoxButton.YesNo)
-                return;
-            else if (_button == MessageBoxButton.OK)
-                _result = MessageBoxResult.OK;
-            else
-                _result = MessageBoxResult.Cancel;
+            switch (_button)
```
Jack251970, line 162: Could you please tell me why you changed this?

Note: This follows the behaviour of `System.Windows.MessageBox`: `return` if MessageBoxButton is YesNo

onesounds: I added it because pressing ESC didn't close the dialog — I felt it should.
If the behavior is meant to match System.Windows.MessageBox, I'm fine with rolling this change back.

Jack251970: > I added it because pressing ESC didn't close the dialog — I felt it should. I could have just closed the dialog when no matching condition was found, but Copilot warned that doing so would return None. I'm not entirely sure, but I assumed that would require all functions calling the message box to handle exceptions. So instead, I included all possible conditions and made pressing ESC default to the option that's closest to "Cancel."
> 
> If the behavior is meant to match System.Windows.MessageBox, I'm fine with rolling this change back.

yeah, iirc, esc behaviour of `System.Windows.MessageBox` should also like this

Jack251970: tested

Jack251970 (dismissed):

Issue #2806, opened 2024-06-27
BUG: FL Window Height Stuck After Double-Click with Resize Cursor


AminSallah:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I am using the latest version of Flow Launcher.

### Problem Description

Issue Description:
When double-clicking on window margins instead of dragging with the resize cursor, the window height becomes fixed even if the query is empty.

Why use double-click?
Double-clicking the top bezel of a window to make it fit the height of the screen is a popular behavior and useful particularly when many results are displayed, allowing users to maximize the window height.
Additionally, users can return to the previous height by double-clicking again.

### To Reproduce

1. Double-click using the resize cursor.
2. The FL window height gets stuck at the double-click height.

To revert manually:
- Restart FL
- Resize the window manually using the resize cursor

### Flow Launcher Version

v1.19.0 pre-release cb77f8d 

### Windows Build Number

10.0.19045.2965

jjw24:
I believe this is fixed in the official 1.19.0 release.

AminSallah:
It seems like you're referring to a different bug. This video using FL 19.1 (official latest version) shows what i mentioned in this bug report:

https://github.com/user-attachments/assets/c7330775-7cb0-4368-bc2f-88c08882d0ce

AtaraxyState:
I seem to have encountered the same issue here is my use case and screenshots for more clarity:

So if I hold left click and drag to expend the window it correctly shows more items:

![Image](https://github.com/user-attachments/assets/e98d1466-f9e7-4e84-9f11-5427b206ab20)


But if I use the double click to make it fit my screen it doesnt seem to update to size of the window or at least prevents me from prompting more items:

![Image](https://github.com/user-attachments/assets/5780a534-fb54-484b-986e-5b84e1de03c4)

Do note that this might be a problem from my side but I think its probably an issue with Flow Launcher himself

PR #3606, opened 2025-06-02
Fix issues related to window maximization and Snap behavior

onesounds:
## What's the PR
- Fix https://github.com/Flow-Launcher/Flow.Launcher/issues/2806
- Prevent window from being maximized or resized by double-clicking the title bar or window borders.
- Improve behavior when the window is maximized via the Snap feature.
- Improve behavior when the window is snapped to the left or right or restored from a snapped state.
- Fix window border issues when the window is snapped to the screen edges.
- Disable the ability to minimize the window.

## Details
Issue #2806 was resolved by preventing double-clicking on the window border. But when using the Snap feature with Win+Arrow keys (automatically resizing the window by dragging it to the top or sides of the monitor), several issues were identified with Flow’s behavior. The explanation is a bit complicated, but trying out various Snap scenarios easily reveals many problematic cases. No one will probably read all of this, but here’s a rough breakdown anyway:

**Issue:** When snapping, the window border is forcibly turned into a rectangle. This results in visible rounded corners and shadows being removed. The behavior differs between Blur and non-Blur themes.
**Solution:** Adjusted the border rendering logic to display correctly. This was not something that could be handled by application code alone. The window still appears as a rectangle, but we've ensured there are no remaining rounded corners or visual artifacts. Since becoming a rectangle is standard Windows behavior, we are keeping this behavior as is.

**Issue:** Pressing Win+Down Arrow minimizes the window, but there is no way to restore it except with Win+Up Arrow.
**Solution**: Prevented the window from being minimized.

**Issue**: While normal maximizing is blocked, OS-level forced maximization via Snap cannot be blocked.
Solution: The only way to exit this state is by dragging the title bar, but in Flow, the title area is hidden, making it non-intuitive for users. To address this, dragging from the title area while maximized now automatically restores the window to normal and allows a smooth drag transition.

**Issue**: When a window is unsnapped from left/right Snap, the OS automatically reduces its height. Flow interpreted this as a SizeChanged event and responded by updating the result count, often reducing it to 2.
**Solution**: If the window height shrinks to below the size needed to show even a single result, the result count is now preserved and not recalculated.

## Test Cases
- [x] Flow should not be minimized using the Windows + Down Arrow key.
- [x] When Flow's window is snapped using Win + Left/Right Arrow or by dragging, the rounded border should not be visible inside the rectangular window.
- [x] When a Flow window snapped to the left or right is restored to normal state by dragging, the MaxResult count should remain unchanged.
- [x] When Flow's window is maximized by dragging it to the top of the monitor, the transition back to the normal window state via dragging should behave naturally.
- [x] Even if the window border is double-clicked, the result panel height should still display correctly on the next query. 

## ETC
- Since the maximized state can now be exited more naturally (though it might feel slightly off if the result window is already visible, it's not too distracting), allowing the WIN+Up shortcut seems acceptable. However, just in case, I’ve decided to block all maximize behaviors except those triggered by Snap.

- I remember it was possible to adjust the slightly oversized window that appears when maximized, but I didn’t bother implementing that due to the hassle.

- There doesn’t seem to be a reliable way to detect snap events. I initially considered estimating snap status by calculating window size, but dropped that idea because Windows 11 supports various snap layouts (such as 1:1:1 splits instead of just halves), and there may also be interactions with tools like PowerToys FancyZones.

### DIFF
--- a/Flow.Launcher.Core/Resource/Theme.cs
+++ b/Flow.Launcher.Core/Resource/Theme.cs
@@ -671,7 +671,15 @@ namespace Flow.Launcher.Core.Resource
                     windowBorderStyle.Setters.Remove(windowBorderStyle.Setters.OfType<Setter>().FirstOrDefault(x => x.Property.Name == "Background"));
                     windowBorderStyle.Setters.Add(new Setter(Border.BackgroundProperty, new SolidColorBrush(Colors.Transparent)));
                 }
-
+                
+                // For themes with blur enabled, the window border is rendered by the system, so it's treated as a simple rectangle regardless of thickness.
+                //(This is to avoid issues when the window is forcibly changed to a rectangular shape during snap scenarios.)
+                var cornerRadiusSetter = windowBorderStyle.Setters.OfType<Setter>().FirstOrDefault(x => x.Property == Border.CornerRadiusProperty);
+                if (cornerRadiusSetter != null)
+                    cornerRadiusSetter.Value = new CornerRadius(0);
+                else
+                    windowBorderStyle.Setters.Add(new Setter(Border.CornerRadiusProperty, new CornerRadius(0)));
+                
                 // Apply the blur effect
                 Win32Helper.DWMSetBackdropForWindow(mainWindow, backdropType);
                 ColorizeWindow(theme, backdropType);
--- a/Flow.Launcher.Infrastructure/Win32Helper.cs
+++ b/Flow.Launcher.Infrastructure/Win32Helper.cs
@@ -162,13 +162,6 @@ namespace Flow.Launcher.Infrastructure
 
             SetWindowStyle(hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, (int)newExStyle);
         }
-
-        public static void BlockWindowMaximize(Window window, HwndSourceHook hook)
-        {
-            var handle = GetWindowHandle(window, true);
-            var hwndSource = HwndSource.FromHwnd(handle);
-            hwndSource.AddHook(hook);
-        }
         
         /// <summary>
         /// Restore window display in the Alt+Tab window list.
--- a/Flow.Launcher.Infrastructure/Win32Helper.cs
+++ b/Flow.Launcher.Infrastructure/Win32Helper.cs
@@ -163,6 +163,13 @@ namespace Flow.Launcher.Infrastructure
             SetWindowStyle(hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, (int)newExStyle);
         }
 
+        public static void BlockWindowMaximize(Window window, HwndSourceHook hook)
+        {
+            var handle = GetWindowHandle(window, true);
+            var hwndSource = HwndSource.FromHwnd(handle);
+            hwndSource.AddHook(hook);
+        }
+        
         /// <summary>
         /// Restore window display in the Alt+Tab window list.
         /// </summary>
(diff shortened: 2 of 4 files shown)


Flow.Launcher.Infrastructure/Win32Helper.cs
```diff
@@ -163,6 +163,13 @@
             SetWindowStyle(hwnd, WINDOW_LONG_PTR_INDEX.GWL_EXSTYLE, (int)newExStyle);
         }
 
+        public static void BlockWindowMaximize(Window window, HwndSourceHook hook)
+        {
+            var handle = GetWindowHandle(window, true);
+            var hwndSource = HwndSource.FromHwnd(handle);
+            hwndSource.AddHook(hook);
+        }
+
```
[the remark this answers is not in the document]
onesounds: Removed. Unused Function.

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -541,7 +598,20 @@ private IntPtr WndProc(IntPtr hwnd, int msg, IntPtr wParam, IntPtr lParam, ref b
 
                 handled = true;
             }
-
+            if (msg == WM_NCLBUTTONDBLCLK)
+            {
+                SizeToContent = SizeToContent.Height;
+                handled = true;
+            }
+            else if (msg == WM_SYSCOMMAND)
```
onesounds, line 606: Block Maximize/Minimize by Win+Up and Win+Down Arrow

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -541,7 +598,20 @@ private IntPtr WndProc(IntPtr hwnd, int msg, IntPtr wParam, IntPtr lParam, ref b
 
                 handled = true;
             }
-
+            if (msg == WM_NCLBUTTONDBLCLK)
```
onesounds, line 601: Block the double click in frame

Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -465,7 +465,52 @@ private void OnPreviewMouseMove(object sender, MouseEventArgs e)
 
         private void OnMouseDown(object sender, MouseButtonEventArgs e)
         {
-            if (e.ChangedButton == MouseButton.Left) DragMove();
+            if (e.ChangedButton == MouseButton.Left)
```
onesounds, line 468: When the window is maximized via Snap, dragging attempts will first switch the window from Maximized to Normal state, and adjust the drag position accordingly.

Flow.Launcher.Core/Resource/Theme.cs
```diff
@@ -671,7 +671,15 @@ private void SetBlurForWindow(string theme, BackdropTypes backdropType)
                     windowBorderStyle.Setters.Remove(windowBorderStyle.Setters.OfType<Setter>().FirstOrDefault(x => x.Property.Name == "Background"));
                     windowBorderStyle.Setters.Add(new Setter(Border.BackgroundProperty, new SolidColorBrush(Colors.Transparent)));
                 }
-
+
```
onesounds, line 674: Eliminated the issue where internal border values remained visible when the window border was forcibly set to a rectangle via Snap. This behavior only applies when blur is enabled; non-blur themes are unaffected.

onesounds:
@coderabbitai Review again please, lady.

onesounds:
Note: When the fixed size option is enabled, the snap feature doesn't work at all. Might be worth trying to enable it only while dragging.

jjw24 (approved):

Issue #3624, opened 2025-06-03
BUG: Recent Update Broke Some Acrylic/Blur Themes


skiIItissue:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I am using the latest version of Flow Launcher.

### Problem Description

I've been using the OnsetGlaze (Acrylic/Glassmorphism) theme for a couple of months now, and after a recent update its background has become pitch black. I tried modifying the .xaml file, however, given I'm not qualified enough to fix the code myself, I didn't succeed. 

So in an attempt to fix this I tried using another theme called Acrylic for Flow Launcher. That did fix the background, but now the Flow window displays double corners, which to my understanding is not intended. I believe there's a good chance it's tied to the recent semi-transparent corners bug fix.

---

On a side note, Acrylic/Glassmorphism themes seem to look less transparent. However, I'm too used to the transparency of the OnsetGlaze theme which is now fully opaque, so I can't reasonably compare. Also, please let me know if it's impossible to tweak the transparency in general (without modyfing the .xaml) or if I'm missing something. 

### To Reproduce

1. Download and install either the "OnsetGlaze" or the "Acrylic for Flow Launcher" theme
2. Set either as a theme in Flow Launcher settings
3. Open Flow Launcher

### Screenshots

OnsetGlaze:

![Image](https://github.com/user-attachments/assets/679691b0-65a5-45be-a243-766bfdf634ba)

What it should look like:

![Image](https://github.com/user-attachments/assets/824fe4ce-9dd6-4bfe-9453-48c3b3dd627e)

---

Acrylic for Flow Launcher (notice the double corners):

![Image](https://github.com/user-attachments/assets/fd281386-9039-4a7f-9845-88dddb04d16e)

What it should look like:

![Image](https://github.com/user-attachments/assets/eb0f87ae-c9f0-4fa6-bc10-df385b5a5aaf)

### Flow Launcher Version

1.20.0

### Windows Build Number

10.0.26100.4061

### Error Log

N/A

indigofairyx:
Broken Blur++ on Win 10

this is the same theme on v1.19.5 vs v1.20.0 on win 10

with the blur being ignore the transparency of the background make flow unreadable when\if there's a lot going on in the background.

my win 11 install apply's a different blur affect altogether. i'll have to follow up with a screen shot for that later. 

[...]

PR #3630, opened 2025-06-04
Change window color tint handling method

onesounds:
## What's the PR 
- Resolve https://github.com/Flow-Launcher/Flow.Launcher/issues/3624
- In version 1.20.0, tint for acrylic is currently applied through a separate layer, but it has been changed to let DWM handle the color and opacity directly.
- **This is a PR that should not be applied due to existing issues.**
- Ref : https://github.com/Maplespe/ExplorerBlurMica

## Problem
- In actual use, there are occasional issues where DWM doesn't update properly during show/hide situations. Calling SetBlur every time the window becomes active can fix it, but this consumes a lot of resources. The root cause is unknown. If this issue can be resolved, the change could be applied.


onesounds:
This bug is a bit tricky to fix. In some cases, previous rendering artifacts remain, so we need to flush them when showing the window, but it's not working well. Even if it does work, if the transparency is too high, it's hard to design the UI to look like a proper application (text color becomes an issue). The property helps prevent rendering blinks without transparency, but it doesn't seem necessary at the moment. I'll keep it in mind for now but won't implement it yet.

PR #3694, opened 2025-06-10
Update Button width for responsive design

stefanroelofs:
Modified Button to use auto width with a minimum width of 100. This change allows the button to grow when keywords do not fit.
image


Jack251970 (approved):

Jack251970:
Thank you for the fix! It looks good to me.

Jack251970:
@onesounds Should we add maximize width for it?

onesounds:
We probably don't need to set it. If it looks weird, people will likely adjust it themselves.

stefanroelofs:
Yeah you don't need maxWidth really. Because settings window itself has variable width.
Very long keywords correctly fit in available space:
image

The action keyword dialog that you open with this button could handle longer keywords better though. I might make a separate PR for that.

PR #3703, opened 2025-06-11
Skip release note notification for dev builds

onesounds:
## What's the PR
- Skip release note notification for dev version (1.0.0)
- When we add the release‐channel switch feature later, we’ll need to adjust this part too. We should suppress the notification if the version number goes down (for example, when moving from the beta channel back to stable). 
- I totally forgot to mention that during the code review.


Jack251970 (approved):

Jack251970:
Let us keep it until 1.20.1 release

PR #3712, opened 2025-06-12
Add preview information back for external preview

Jack251970:
Follow on with #3631.

We need to add preview information back for external preview plugin like `Quick Look`.


Jack251970:
Already tested

onesounds (approved):

PR #3728, opened 2025-06-14
Improve Topmost Feature

Jack251970:
# Changes

- Set topmost default value to false

Follow on with #3541.

# TODO

- [x] Improve topmost feature description
- [x] Reorganize topmost setting option position


Jack251970:
@onesounds If you have a chance, could you please implement the TODOs as described above?

onesounds:
image

Done. I’m not entirely happy with using the converter (since I’m not sure if it will ever be reused in the future, or if it might need to handle reverse logic at some point — in which case, we might consider renaming it). Anyway, I’ll leave the wording and the converter details up to you.

Jack251970:
> image
> 
> Done. I’m not entirely happy with using the converter (since I’m not sure if it will ever be reused in the future, or if it might need to handle reverse logic at some point — in which case, we might consider renaming it). Anyway, I’ll leave the wording and the converter details up to you.

why do we need this disable relationship? even if hiding option is on, this option can help make sure Flow window is on topper of other Topmost windows.

onesounds:
As I understand it, this feature only works when "lost focus" is turned off, right? I'm genuinely asking because I’m not sure. From what I know, this feature operates only when "lost focus" is off. If it's on, the window hides anyway when focus is lost, so this feature wouldn't make sense. If this isn't the functionality I’m thinking of, then I guess I don’t really understand what this feature is. I don't think I'm the right person to write or organize the description for it.

Wait a second — is this feature actually meant to ignore other programs' "Always on Top" setting?
If that's the case, then it shouldn’t be linked (to the lost focus setting).

onesounds:
image
I made some changes. This seems like the most accurate description.

Jack251970:
@onesounds I think it is good to go👍

onesounds (approved):

PR #3754, opened 2025-06-19
Add rad2deg and deg2rad to default calculator plugin.

Koisu-unavailable:
This PR adds two functions to the default calculator plugin: rad2deg which converts from radians to degrees, and deg2rad which converts from degrees to radians. I believe calculators that can do the trignometric functions are expected to be able to convert between degrees and radians.
# Examples:
image


VictoriousRaptor (approved):

Koisu-unavailable:
Sorry, I forgot how pull requests worked and accidentally committed other changes to the same branch, so I had to revert them.

PR #3779, opened 2025-06-27
Fix Card Control

onesounds:
## What's the PR
- 
image
- Fixes a XAML error caused by a card property that used to be set automatically within a card group.
- Changed the 'First', 'Middle', and 'Last' properties, which were previously set automatically within the Card control, to be added manually upon registration.
- Added the property to each pages

### Details
- This property was introduced to adjust the borders when there are multiple cards inside a CardGroup. The previous method, which seems to have operated automatically, would display an error message when the card was not part of a CardGroup. In other words, the issue was caused by a dependency on whether or not the control was within a CardGroup. Even with a fallback value, the error persisted, so we've modified the logic to function regardless of whether it's in a CardGroup. Now, developers are required to explicitly set the value if needed.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -1,3 +1,5 @@
+# This file should contain names of products, companies, or individuals that aren't in a standard dictionary (e.g., GitHub, Keptn, VSCode).
+
 crowdin
 DWM
 workflows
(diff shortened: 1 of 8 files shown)


Jack251970 (approved):
LGTM! The performance of the settings window under debug configuration is improved largely!

onesounds:
I turned off this debug message, so I was completely aware of it.

PR #3791, opened 2025-07-01
Check if plugin is modified before querying

Jack251970:
If one plugin is modified, do not use it when querying results. 

From #3572

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginManager.cs
+++ b/Flow.Launcher.Core/Plugin/PluginManager.cs
@@ -290,7 +290,14 @@ namespace Flow.Launcher.Core.Plugin
                 return Array.Empty<PluginPair>();
 
             if (!NonGlobalPlugins.TryGetValue(query.ActionKeyword, out var plugin))
-                return GlobalPlugins;
+            {
+                return GlobalPlugins.Where(p => !PluginModified(p.Metadata.ID)).ToList();
+            }
+
+            if (API.PluginModified(plugin.Metadata.ID))
+            {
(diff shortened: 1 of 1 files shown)


Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -290,7 +290,14 @@
                 return Array.Empty<PluginPair>();
 
             if (!NonGlobalPlugins.TryGetValue(query.ActionKeyword, out var plugin))
-                return GlobalPlugins;
+            {
+                return GlobalPlugins.Where(p => !PluginModified(p.Metadata.ID)).ToList();
+            }
```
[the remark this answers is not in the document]
Jack251970: We can use internal function here

Flow.Launcher.Core/Plugin/PluginManager.cs
```diff
@@ -300,7 +307,7 @@ public static ICollection<PluginPair> ValidPluginsForQuery(Query query)
 
         public static ICollection<PluginPair> ValidPluginsForHomeQuery()
         {
-            return _homePlugins.ToList();
+            return _homePlugins.Where(p => !PluginModified(p.Metadata.ID)).ToList();
```
jjw24, line 310: Not related to this PR but just thought, does this then also need to take into consideration of Disabled plugins? Will they show if disabled?

Jack251970: No need since it will be checked in MainViewModel

jjw24 (approved):

PR #3796, opened 2025-07-04
Return results when query is empty for Program

Jack251970:
Check `string.IsNullOrEmpty(query.Search)` for Program plugin so that Flow can return results when query is empty for program


jjw24:
How do I reproduce the issue this PR is resolving?

Jack251970:
> How do I reproduce the issue this PR is resolving?

set an action keyword for program plugin and search it

jjw24:
So this can then be improved to return a list of saved program results and also via the home page?

Jack251970:
> So this can then be improved to return a list of saved program results and also via the home page?

Hmm, I have thought about this before. And I think we can open another PR for it.

My thinking is to add a new context menu item - Pin to home - so that some applications will be shown in home page.

PR #3811, opened 2025-07-09
Update home page when language changes

Jack251970:
# Changes

- When search is empty and Home Page feature is enabled, if the language is changed, the home page will not be updated by its translation until next non-empty query. So we should check if home page is shown and update the translations

# Test

- Enable Home Page and change language, the search window results will be updated

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -284,6 +284,10 @@ namespace Flow.Launcher
                         break;
                     case nameof(Settings.Language):
                         UpdateNotifyIconText();
+                        if (_settings.ShowHomePage && _viewModel.QueryResultsSelected() && string.IsNullOrEmpty(_viewModel.QueryText))
+                        {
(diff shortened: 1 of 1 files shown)


Flow.Launcher/MainWindow.xaml.cs
```diff
@@ -284,6 +284,10 @@ private void OnLoaded(object sender, RoutedEventArgs _)
                         break;
                     case nameof(Settings.Language):
                         UpdateNotifyIconText();
+                        if (_settings.ShowHomePage && _viewModel.QueryResultsSelected() && string.IsNullOrEmpty(_viewModel.QueryText))
```
jjw24, line 287: Why do we need to check query result selected?

Jack251970: > Why do we need to check query result selected?

If query results are not selected, users may disable `Hide Flow when focus is lost` and they are using context menu or history menu perhaps. So we should not refresh search window which can break their work.

jjw24 (approved):

PR #3827, opened 2025-07-14
Add Support for Automatically Checking Plugin Updates

Jack251970:
# Changes

* Add Auto Update description

image

* Add Auto Update Plugins option which can automatically check plugin updates in the background every 5 hours

Screenshot 2025-07-14 163214

* Add an option to check plugin updates in plugin store page

Screenshot 2025-07-14 163239

# Test

Screenshot 2025-07-14 163145

Screenshot 2025-07-14 163424

# One more thing

Since the plugin update message is implemented with message instead of message box which I do not think it will annoy users, I use default True value for Auto Update Plugins.

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginInstaller.cs
+++ b/Flow.Launcher.Core/Plugin/PluginInstaller.cs
@@ -284,7 +284,7 @@ public static class PluginInstaller
     /// <param name="usePrimaryUrlOnly">If true, only use the primary URL for updates.</param>
     /// <param name="token">Cancellation token to cancel the update operation.</param>
     /// <returns></returns>
-    public static async Task UpdatePluginAsync(bool silentUpdate = true, bool usePrimaryUrlOnly = false, CancellationToken token = default)
+    public static async Task CheckForPluginUpdatesAsync(bool silentUpdate = true, bool usePrimaryUrlOnly = false, CancellationToken token = default)
     {
         // Update the plugin manifest
         await API.UpdatePluginManifestAsync(usePrimaryUrlOnly, token);
--- a/Flow.Launcher.Core/Plugin/PluginInstaller.cs
+++ b/Flow.Launcher.Core/Plugin/PluginInstaller.cs
@@ -278,6 +278,97 @@ public static class PluginInstaller
     }
 
     /// <summary>
+    /// Updates the plugin to the latest version available from its source.
+    /// </summary>
+    /// <param name="silentUpdate">If true, do not show any messages when there is no udpate available.</param>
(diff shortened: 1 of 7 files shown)


Flow.Launcher/Languages/en.xaml
```diff
@@ -93,6 +93,7 @@
     <system:String x:Key="typingStartEn">Always Start Typing in English Mode</system:String>
     <system:String x:Key="typingStartEnTooltip">Temporarily change your input method to English mode when activating Flow.</system:String>
     <system:String x:Key="autoUpdates">Auto Update</system:String>
+    <system:String x:Key="autoUpdatesTooltip">Automatically check app updates and notify if there are any updates available</system:String>
```
jjw24: This auto update for the app actually updates so the description needs a bit more specific.

```suggestion
    <system:String x:Key="autoUpdatesTooltip">Automatically check and update the app when available</system:String>
```

Flow.Launcher/Languages/en.xaml
```diff
@@ -150,6 +151,8 @@
     <system:String x:Key="autoRestartAfterChangingToolTip">Restart Flow Launcher automatically after installing/uninstalling/updating plugin via Plugin Store</system:String>
     <system:String x:Key="showUnknownSourceWarning">Show unknown source warning</system:String>
     <system:String x:Key="showUnknownSourceWarningToolTip">Show warning when installing plugins from unknown sources</system:String>
+    <system:String x:Key="autoUpdatePlugins">Auto update plugins</system:String>
+    <system:String x:Key="autoUpdatePluginsToolTip">Automatically check plugin updates and notify if there are any updates available</system:String>
```
jjw24, line 155: Does it auto updates or just notify with a button to click on to update all?

Jack251970: > Does it auto updates or just notify with a button to click on to update all?

Just notify the user with the message as shown in the PR description, and if users click that, the plugins will be updated without restarting

Flow.Launcher.Core/Plugin/PluginInstaller.cs
```diff
@@ -277,6 +278,100 @@ await DownloadFileAsync(
         }
     }
 
+    /// <summary>
+    /// Updates the plugin to the latest version available from its source.
+    /// </summary>
+    /// <param name="silentUpdate">If true, do not show any messages when there is no udpate available.</param>
```
jjw24:
```suggestion
    /// <param name="silentUpdate">If true, do not show any messages when there is no update available.</param>
```

jjw24 (approved):
👍

PR #2427, opened 2023-11-18
Double pinyin query

VictoriousRaptor:
This pull request introduces a comprehensive set of changes to enhance Pinyin support in the application, including the addition of Double Pinyin (“双拼”) functionality, improved translation mappings, and related settings. The most significant updates include the implementation of the `IAlphabet` interface, the refactoring of the `PinyinAlphabet` class, the introduction of Double Pinyin schemas, and updates to the user settings and localization files.

## UI in General Panel

Clip_20250713_194052

### Pinyin and Double Pinyin Enhancements:
* Added the `IAlphabet` interface to define methods for language translation and determining translatability. (`Flow.Launcher.Infrastructure/IAlphabet.cs`)
* **Refactored the `PinyinAlphabet` class to support Double Pinyin**, utilize a cache for translations, and dynamically reload schemas based on user settings. (`Flow.Launcher.Infrastructure/PinyinAlphabet.cs`)
* Introduced Double Pinyin schemas as an enum and added settings for enabling/disabling Double Pinyin and selecting schemas. (`Flow.Launcher.Infrastructure/UserSettings/Settings.cs`) [[1]](diffhunk://#diff-7128cb802625c6dba2865b74f04d4ad61c7cdc2d19a727235636e2b46354ca99R302-R331) [[2]](diffhunk://#diff-7128cb802625c6dba2865b74f04d4ad61c7cdc2d19a727235636e2b46354ca99R605-R617)

### Translation Mapping Improvements:
* **Reimplemented the `TranslationMapping` class to simplify and optimize the mapping of original to translated indices.** (`Flow.Launcher.Infrastructure/TranslationMapping.cs`)
* Added unit tests for `TranslationMapping` to validate its functionality. (`Flow.Launcher.Test/TranslationMappingTest.cs`)

### Localization and Resource Updates:
* Updated localization files to include strings for Double Pinyin settings and schemas. (`Flow.Launcher/Languages/en.xaml`)
* **Added `double_pinyin.json` to project resources for schema definitions**. (`Flow.Launcher/Flow.Launcher.csproj`)

### Codebase Simplifications:
* Updated methods in `StringMatcher.cs` to use the new `ShouldTranslate` method from `IAlphabet`. (`Flow.Launcher.Infrastructure/StringMatcher.cs`)
[...]

PR #3829, opened 2025-07-14
Change double pinyin panel design

Jack251970:
# Change double pinyin panel design

image

Reason for this changing:
* Originally, we use visibility to organize the relationship of these options, but it looks little strange since all elements in setting window do not use this in its design.
* Now I change it to expand card & card design to make it look more unified with others.

Follows on from https://github.com/Flow-Launcher/Flow.Launcher/pull/2427


onesounds (approved):

VictoriousRaptor:
Do we need some rework on this design since if `double pinyin` is true then `use pinyin` must be true as well?

PR #3843, opened 2025-07-19
Add plugin update dialog for selective plugin updating

Jack251970:
# Changes

Follow on with #3827. Pop up a plugin update dialog to let users select the plugins they want to update and whether to restart application after updating plugins.

# Test

Screenshot 2025-07-18 153108

Screenshot 2025-07-18 153204

### DIFF
--- a/Flow.Launcher.Core/Plugin/PluginInstaller.cs
+++ b/Flow.Launcher.Core/Plugin/PluginInstaller.cs
@@ -281,11 +281,12 @@ public static class PluginInstaller
     /// <summary>
     /// Updates the plugin to the latest version available from its source.
     /// </summary>
+    /// <param name="updateAllPlugins">Action to execute when the user chooses to update all plugins.</param>
     /// <param name="silentUpdate">If true, do not show any messages when there is no update available.</param>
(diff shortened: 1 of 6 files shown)


Flow.Launcher.Core/Plugin/PluginInstaller.cs
```diff
@@ -363,13 +370,28 @@ await DownloadFileAsync(
...
+                anyPluginSuccess = true;
             }
             catch (Exception e)
             {
                 API.LogException(ClassName, "Failed to update plugin", e);
                 API.ShowMsgError(API.GetTranslation("ErrorUpdatingPlugin"));
             }
         }));
+
+        if (!anyPluginSuccess) return;
```
jjw24, line 383: Do we need to notify failed plugin updates if any?

Jack251970: > Do we need to notify failed plugin updates if any?

I think we do not need to do that since the try-catch sentences above will show the error messages

jjw24 (approved):

PR #3886, opened 2025-08-07
Remove unnecessary runtime files

Jack251970:
Remove unnecessary runtime files for Flow.Launcher

### DIFF
--- a/Flow.Launcher/Flow.Launcher.csproj
+++ b/Flow.Launcher/Flow.Launcher.csproj
@@ -39,6 +39,48 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
 
+  <Target Name="RemoveUnnecessaryRuntimesAfterBuild" AfterTargets="Build">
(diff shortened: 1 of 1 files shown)


jjw24:
Are these unnecessary files introduced by the .Net upgrade?

Jack251970:
> Are these unnecessary files introduced by the .Net upgrade?

Not sure. I noticed this when I was checking another package size issue.

Jack251970:
> Is there a better way to have the project not build these files than to remove them post-build?

As far as I know, one another way is to set specific `TargetIdentifier` to `win64` or `win86` for Flow and it will remove the runtime folder. But this needs to build Flow with different architecture.

Jack251970:
Setting `<RuntimeIdentifiers>win-x64;win-x86</RuntimeIdentifiers>` makes no difference

taooceros (approved):

taooceros:
> Setting `<RuntimeIdentifiers>win-x64;win-x86</RuntimeIdentifiers>` makes no difference

I remember that this is problematic

PR #3889, opened 2025-08-09
Fix property binding in WebSearch plugin

zoltanvi:
The `OnPropertyChanged` was not triggered, that is why there was a workaround for binding Settings.EnableSuggestion in the WebSearch plugin.


Jack251970 (approved):
Good work! Thanks for your contribution!

PR #3898, opened 2025-08-13
Load images into ImageCache & Use non-async method in ImageCache

Jack251970:
# Load images into ImageCache

Load `Constant.ImageIcon` & `Constant.LoadingImgIcon` so that they can be freeze resources which can improve performance.

# Use non-async method in ImageCache

Add non-async method in `BinaryStorage` so that `ImageCache` can use it to remove async methods.

And we can remove `ImageLoader.WaitSaveAsync()` to improve code quality

# Test

- Loading missing icon can work


taooceros:
why do we need waitsave before?

Jack251970:
> why do we need waitsave before?

Before Flow exiting, I think we need to make sure all data including image cache are saved.

jjw24 (approved):

PR #3921, opened 2025-08-25
Check plugin updates only for Release

Jack251970:
# Check plugin updates only for Release

Follow on with #3827.

When we start Flow on Debug configuration, Flow will pop up this message every time it starts since all pre-released plugins are in version 1.0.0. So let us check plugin updates only for release to remove that annoying message to improve development experience.

image

### DIFF
--- a/Flow.Launcher/App.xaml.cs
+++ b/Flow.Launcher/App.xaml.cs
@@ -297,6 +297,7 @@ namespace Flow.Launcher
             });
         }
 
+        [Conditional("RELEASE")]
         private static void AutoPluginUpdates()
         {
             _ = Task.Run(async () =>


VictoriousRaptor:
I've seen the update tip when using local release build on dev. Should we check whether running dev plugin to solve this issue?

Jack251970:
> I've seen the update tip when using local release build on dev. Should we check whether running dev plugin to solve this issue?

I think this solution is simple and it will not break anything (We can still check plugin updates manually in plugin update page). If we want to check dev plugins, there should be more logic which I think is unnecessary.

jjw24:
I think what Raptor is saying is that when you do a 'Release' build locally this issue still exists. Problem is I am assuming everyone uses different IDEs for flow e.g. vscode, rider, vs studio etc, we don't have a uniform way of upping the plugin versions during the build process, unless we run a dedicated build script.

We can still solve this by creating:
- An GitHub action that will create a PR to bump plugin versions when a PR is tagged with 'release' label. Or
- introduce an option in settings to turn off plugin updates so you only need to do this once in your dev environment and not be bothered by this message again.

I personally think it's a minor annoyance update message that can be ignored, so up to you guys if you think it's worth implementing.

For now this change is good as well that solves the debug environment issue.

PR #3927, opened 2025-08-31
Test Remove unused runtime files

Jack251970:


jjw24:
let's see the build size with x86 removed.

Jack251970:
> let's see the build size with x86 removed.

I wonder if `arm` is supported. IIRC, it is not supported?

Jack251970:
> let's see the build size with x86 removed.

If we only keep runtime files of ×64, the installer size is 90 MB

jjw24:
>I wonder if arm is supported. IIRC, it is not supported?

It's requested, but we have had issues with Everything plugin and the fact that the rest of the plugins in the community are not publish arm versions, so for now we should remove ARM runtime files until we figure out a proper path forward for ARM builds.

Jack251970:
> > I wonder if arm is supported. IIRC, it is not supported?
> 
> It's requested, but we have had issues with Everything plugin and the fact that the rest of the plugins in the community are not publish arm versions, so for now we should remove ARM runtime files until we figure out a proper path forward for ARM builds.

Should we keep runtime files for ×86? If not, this PR should be mergable.

Jack251970:
Anyway, we should remove runtimes for arm to decrease size.

jjw24:
Yeah, let's keep x86 for now.

Jack251970:
> keep x86

Done

jjw24 (approved):
Thank you for this 👍

PR #3944, opened 2025-09-04
Enhancement: Support Custom Browser Path & Open in window / tab & In private for URL Plugin

Jack251970:
# Support Custom Browser Path & Open in window / tab & In private for URL Plugin

When open one url, we can set custom browser path, choose whether to open in new window or tab, and choose whether to open in private mode.

# Test

image

* Setting a custom browser path can work.
* Opening in new window can work.
* Opening in private can work.

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Url/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Url/Main.cs
@@ -117,7 +117,7 @@ namespace Flow.Launcher.Plugin.Url
         public void Init(PluginInitContext context)
         {
             Context = context;
-            
+
             Settings = context.API.LoadSettingJsonStorage<Settings>();
         }
 
--- a/Plugins/Flow.Launcher.Plugin.Url/SettingsControl.xaml.cs
+++ b/Plugins/Flow.Launcher.Plugin.Url/SettingsControl.xaml.cs
@@ -16,7 +16,7 @@ public partial class SettingsControl : UserControl
     {
         var dlg = new Microsoft.Win32.OpenFileDialog
         {
-            Filter = Main.Context.API.GetTranslation("flowlauncher_plugin_url_plugin_filter")
+            Filter = Localize.flowlauncher_plugin_url_plugin_filter()
         };
 
         if (dlg.ShowDialog() == true && !string.IsNullOrEmpty(dlg.FileName))


taooceros:
Actually I remember this is implemented as the default browser option here? 
image

Jack251970:
> Actually I remember this is implemented as the default browser option here? 
> image
> 

Yes, but this needs to use api function OpenUrl. Here I use another api in ShareCommands so we can set custom browser only for URL plugin

Plugins/Flow.Launcher.Plugin.Url/Main.cs
```diff
@@ -39,52 +41,62 @@
...
+                        SubTitle = string.Format(Context.API.GetTranslation("flowlauncher_plugin_url_open_url"),raw),
                         IcoPath = "Images/url.png",
                         Score = 8,
                         Action = _ =>
                         {
-                            if (!raw.ToLower().StartsWith("http"))
+                            if (!raw.StartsWith("http", StringComparison.OrdinalIgnoreCase))
                             {
                                 raw = "http://" + raw;
                             }
```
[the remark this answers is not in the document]
Jack251970: No need

Plugins/Flow.Launcher.Plugin.Url/Main.cs
```diff
@@ -39,52 +41,62 @@
...
-                                
+                                var browserPath = Settings.UseCustomBrowser ? Settings.BrowserPath : string.Empty;
+                                if (Settings.OpenInNewBrowserWindow)
+                                {
+                                    SearchWeb.OpenInBrowserWindow(raw, browserPath, Settings.OpenInPrivateMode);
+                                }
+                                else
+                                {
+                                    SearchWeb.OpenInBrowserTab(raw, browserPath, Settings.OpenInPrivateMode);
+                                }
```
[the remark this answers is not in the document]
Jack251970: No need

Plugins/Flow.Launcher.Plugin.Url/Settings.cs
```diff
@@ -1,9 +1,25 @@
...
+                {
+                    _browserPath = value;
+                    OnPropertyChanged();
+                }
+            }
+        }
+
+        public bool OpenInNewBrowserWindow { get; set; } = false;
+
+        public bool OpenInPrivateMode { get; set; } = false;
```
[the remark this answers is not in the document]
Jack251970: Will do

Plugins/Flow.Launcher.Plugin.Url/Settings.cs
```diff
@@ -1,9 +1,25 @@
...
+            {
+                if (_browserPath != value)
+                {
+                    _browserPath = value;
+                    OnPropertyChanged();
+                }
+            }
+        }
+
+        public bool OpenInNewBrowserWindow { get; set; } = false;
```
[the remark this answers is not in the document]
Jack251970: Will do

Plugins/Flow.Launcher.Plugin.Url/Settings.cs
```diff
@@ -1,9 +1,25 @@
 ﻿namespace Flow.Launcher.Plugin.Url
 {
-    public class Settings
+    public class Settings : BaseModel
     {
-        public string BrowserPath { get; set; }
+        public bool UseCustomBrowser { get; set; } = false;
```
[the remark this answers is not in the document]
Jack251970: Will do

Plugins/Flow.Launcher.Plugin.Url/SettingsControl.xaml.cs
```diff
@@ -0,0 +1,27 @@
...
+    public SettingsControl()
+    {
+        InitializeComponent();
+    }
+
+    private void SelectBrowserPath(object sender, RoutedEventArgs e)
+    {
+        var dlg = new Microsoft.Win32.OpenFileDialog
+        {
+            Filter = Main.Context.API.GetTranslation("flowlauncher_plugin_url_plugin_filter")
```
jjw24: Don't think it's good to use a translatable string as filter, that filter shouldn't even need translation right?

Plugins/Flow.Launcher.Plugin.Url/Languages/en.xaml
```diff
@@ -1,18 +1,22 @@
...
     <system:String x:Key="flowlauncher_plugin_url_cannot_open_url">Can't open url:{0}</system:String>
 
     <system:String x:Key="flowlauncher_plugin_url_plugin_name">URL</system:String>
     <system:String x:Key="flowlauncher_plugin_url_plugin_description">Open the typed URL from Flow Launcher</system:String>
 
-    <system:String x:Key="flowlauncher_plugin_url_plugin_set_tip">Please set your browser path:</system:String>
     <system:String x:Key="flowlauncher_plugin_url_plugin_choose">Choose</system:String>
     <system:String x:Key="flowlauncher_plugin_url_plugin_filter">Application(*.exe)|*.exe|All files|*.*</system:String>
+
+    <system:String x:Key="flowlauncher_plugin_url_use_custom_browser">Use custom web browser instead of Flow default web browser</system:String>
```
jjw24:
```suggestion
    <system:String x:Key="flowlauncher_plugin_url_use_custom_browser">Use custom instead of Flow's default web browser</system:String>
```

jjw24 (commented):
Everything else LGTM other than the use of translation string as filter.

Plugins/Flow.Launcher.Plugin.Url/SettingsControl.xaml.cs
Jack251970: > Don't think it's good to use a translatable string as filter, that filter shouldn't even need translation right?

I do not think so. In `Application(*.exe)|*.exe|All files|*.*`, I think `Application` & `All files` should be translated

Jack251970: Take `Notepad` filter for example, those filters are translated:

image

Plugins/Flow.Launcher.Plugin.Url/Converters/BoolToVisibilityConverter.cs
VictoriousRaptor, line 1: IIRC we've got a same converter. Do we need this?

Jack251970: > IIRC we've got a same converter. Do we need this?

This is used in url project. We must need to keep it.

PR #3987, opened 2025-09-17
Fix shell plugin focus issue

### LINKED ISSUE #1945: BUG: Run Plugin - No focus when using win+r replacement

Jack251970:
Fix #1945


Jack251970:
Cannot work

PR #4000, opened 2025-09-21
Build script fixed + enhanced

dcog989:
Main issues: redundant 95MB file removed from portable zip; fixed plugin build logic so that we always get the latest version and not an obsolete DLL.

---

1.  **Fixed Portable Package Creation**

    * The portable build process was including the temporary `packages` directory in the final archive. This added `FlowLauncher-x.x.x-full.nupkg` file, increasing zip size by ~95MB. We now explicitly remove this directory before compression.
    * We now use a staging directory to assemble the portable version by running a silent install which ensures package structure is correct and contains no unnecessary build artifacts.

2.  **Enhanced Build Process Reliability**

    * **Dynamic Squirrel Path:** The script no longer uses a hardcoded path to `squirrel.windows`.
    * **Improved Dependency Handling:** By changing `<Private>` from `false` to `true` in `Directory.Build.targets`, project-to-project dependencies are now correctly copied to the output directory. This prevents "missing assembly" errors, especially for plugins.
    * **Safer File Deletion:** The `Remove-UnusedFiles` function now includes a safety check to prevent accidentally deleting a plugin's main DLL if it shares a name with a common dependency.
    * **Structured Build Flow:** The main build logic has been restructured to follow a sequence of building the solution, publishing to a temporary directory, and then merging into the final release folder, which prevents interference from old build artifacts.
    * **More Reliable Version Retrieval:** The `Build-Version` function was updated to use `(Get-Item).VersionInfo.ProductVersion` instead of `Get-Command`. This is more reliable for retrieving file version in PowerShell and aligns with `AssemblyInformationalVersion`.

3.  **Modernized NuGet Package**

    * The `.nuspec` file has been updated to use the modern `<icon>` and `<readme>` tags (note - I failed to get rid of the deprecated `<iconUrl>` due to Squirrel).

---

> [!NOTE]
> I've been playing with https://github.com/velopack/velopack for https://github.com/dcog989/Cliptoo - would there be interest in converting Flow to use this given that Squirrel was abandoned long ago? (Velopack originates from a fork of Squirrel).

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -5,10 +5,10 @@
 			Condition="'$(OutputType)' == 'Library' and '$(CopyLocalLockFileAssemblies)' == 'true' and $(AssemblyName.EndsWith('Tests')) == 'false' ">
 	  <ItemGroup>
 		<ProjectReferenceWithConfiguration Update="@(ProjectReferenceWithConfiguration)" >
-		  <Private>false</Private>
+		  <Private>true</Private>
 		</ProjectReferenceWithConfiguration>
 		<ProjectReference Update="@(ProjectReference)" >
-		  <Private>false</Private>
+		  <Private>true</Private>
 		</ProjectReference>
 	  </ItemGroup>
 	</Target>
--- a/Scripts/flowlauncher.nuspec
+++ b/Scripts/flowlauncher.nuspec
@@ -7,6 +7,7 @@
     <authors>Flow-Launcher Team</authors>
     <projectUrl>https://github.com/Flow-Launcher/Flow.Launcher</projectUrl>
     <icon>images\app.png</icon>
+    <iconUrl>https://raw.githubusercontent.com/Flow-Launcher/Flow.Launcher/master/Flow.Launcher/Resources/app.ico</iconUrl>
     <readme>README.md</readme>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <description>Flow Launcher - Quick file search and app launcher for Windows with community-made plugins</description>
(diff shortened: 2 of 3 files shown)


Jack251970:
We have tried to move to Velopack + Github Action in #2616

jjw24:
Before we get too deep into enhancing squirrel.windows I actually would like to move to Velopack if possible, it provides build versioning and ability to use pre-release channel.

jjw24:
>The portable build process was including the temporary packages directory in the final archive. This added FlowLauncher-x.x.x-full.nupkg file, increasing zip size by ~95MB. We now explicitly remove this directory before compression.

Is this safe to do? Been a long while I looked at squirrel, is this file needed for updating purposes?

dcog989:
> We have tried to move to Velopack + Github Action in #2616

Thanks. It's not clear from that why it was closed?? 

> Before we get too deep into enhancing squirrel.windows I actually would like to move to Velopack

That makes the most sense for sure. Lots of advantages.

> ...is this file needed for updating purposes?

I don't think it is but have done minimal testing (installer + portable works), I'm not familiar with Squirrel and was going to rely on you guys to review the PR / help test.

But if Velopack is the preferred solution then I'm happy to have a go, and let this PR sit for now?

Jack251970:
I have no idea about why #2616 is closed. And I would prefer Velopack. If you have similar ideas, I think we can put this PR and let us go with Velopack.

dcog989:
> ...I think we can put this PR and let us go with Velopack.

OK, I will make a start after I finish the new BrowserBookmarks plugin, which I'm looking at now.

PR #4003, opened 2025-09-22
New browser bookmarks plugin

dcog989:
# Replacement for `BrowserBookmark` Plugin

This is a new plugin to provide browser bookmarks and favicon functionality. It replaces the existing 'BrowserBookmark` plugin. It improves performance, extensibility, and UX over the original.

Favicons are now available without the need to restart the app, with instant updates for Chrome/ium browsers and a three-hour refresh cycle for Firefox.

## Features & Enhancements

* **Favicon Handling:**
  * **Local Extraction:** The plugin first extracts favicons from the browser's local database. This is a fast / low cost function so the previous warning that this might be slow has been removed.
  * **Web Fetching (Optional):** An optional setting fetches missing favicons from the web. We use a smart fetch favicon.ico / parse the HTML function, processed on a background thread.
  * **Format Support:** ICO / PNG / WEBP / SVG are supported.

* **Real-time & Periodic Updates:**
  * **Chromium:** Changes to bookmarks in Chromium-based browsers are detected real-time using a `FileSystemWatcher` with debouncing to prevent excessive reloads.
  * **Firefox:** To avoid performance issues from watching the frequently-updated `places.sqlite` file, Firefox bookmarks are now reloaded on a **periodic 3-hour timer**.

## Architectural & Performance

* **Fully Asynchronous Loading:** The entire bookmark and favicon loading pipeline is asynchronous (`async/await`).

* **Parallel Processing:** Favicon processing is parallelized to significantly speed up the initial caching process.

* **Separation of Concerns:** The logic is cleanly divided:
  * `BookmarkLoaderService`: Handles discovering and loading bookmarks from different browser types.
  * `BookmarkWatcherService`: Manages real-time file system monitoring for Chromium browsers.
  * `FaviconService`: Responsible for all favicon extraction, fetching, and caching.
  * `BrowserDetector`: Contains the logic to automatically identify a browser's engine from its directory path.

## UX & UI Improvements

* **Automatic Engine Detection:** The user no longer needs to specify a custom browser engine, we auto-detect.


Jack251970:
@dcog989 Why did you change the directory of the plugin from `Flow.Launcher.Plugin.Bookmarks` to `Flow.Launcher.Plugin.Bookmark`? I just think it is unnecessary and I do not recommend to do that.

dcog989:
> Why did you change the directory of the plugin

It's a new plugin. All code is new apart from some reused XAML. Is this critical?

Jack251970:
> > Why did you change the directory of the plugin
> 
> It's a new plugin. All code is new apart from some reused XAML. Is this critical?

A very noticeable problem is that users will lose their settings and caches of old Bookmark plugin since you change the namespace. And I am not sure if Crowdin can work here. Additionally, we need to check related Github Actions &  CI/workflows.

Plugins/Flow.Launcher.Plugin.BrowserBookmarks/Models/BaseModel.cs
```diff
@@ -0,0 +1,16 @@
+#nullable enable
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+
+namespace Flow.Launcher.Plugin.BrowserBookmarks.Models;
+
+// A base class that implements INotifyPropertyChanged for view models.
+public abstract class BaseModel : INotifyPropertyChanged
```
Jack251970, line 8: Why we do not use this?

https://github.com/Flow-Launcher/Flow.Launcher/blob/1e8819af0c41d2ed23ac34bbd789d88ce9ffbbcd/Flow.Launcher.Plugin/BaseModel.cs#L10

dcog989:
> A very noticeable problem is that users will lose their settings and caches of old Bookmark plugin since you change the namespace.

The only issue I see here is that users would lose manually entered custom browsers. Caches would be rebuilt so that would be transparent to the user. Plus, they have a new option to choose either way.

I can easily switch this back to `BrowserBookmark` if you think that is the best solution?? Which is the path of least pain??

Jack251970:
@jjw24 Any ideas about this?

jjw24:
- Keep the namespace, plugin name, plugin ID and project filename as is please (as Jack mentioned, there are indeed automations built around them)
- Cater for any transition that is needed for user's data to move from old to new plugin (haven't had time to look at PR, this is an assumption on my part). Transition should be seamless, for example if users don't look they wouldn't even notice they have been moved to this new one.
- how much different is this to the old plugin, is it a complete rewrite?

dcog989:
@jjw24 - 👍 

>  is it a complete rewrite?

Yes, completely refactored, rewritten. The only thing I have reused / updated is XAML/.cs

dcog989:
I'm closing this and opening a new PR to replace the existing `BrowserBookmark` plugin.

Note: I have incorporated all feedback from this PR in to the new code.

PR #4019, opened 2025-09-26
Center the window vertically when settings = Center

### LINKED ISSUE #3917: BUG: 'Center' for window position is not vertically centered
### Checks

### Problem Description

Settings / 'Search Window Position on Monitor'* > 'Center' positions Flow centred horizontally but not vertically.

DPI / Scaling issue or...???

Monitor = 3840x2160 @ 150% scaling

NOTE - some overlap with https://github.com/Flow-Launcher/Flow.Launcher/issues/1140

### Screenshots

Image

---

Image

### Flow Launcher Version

Dev

### Windows Build Number

Microsoft Windows [Version 10.0.26100.4946]

dcog989:
Closes #3917

### DIFF
--- a/Flow.Launcher/MainWindow.xaml.cs
+++ b/Flow.Launcher/MainWindow.xaml.cs
@@ -991,7 +991,7 @@ namespace Flow.Launcher
         {
             var dip1 = Win32Helper.TransformPixelsToDIP(this, 0, screen.WorkingArea.Y);
             var dip2 = Win32Helper.TransformPixelsToDIP(this, 0, screen.WorkingArea.Height);
-            var top = (dip2.Y - QueryTextBox.ActualHeight) / 4 + dip1.Y;
+            var top = (dip2.Y - ActualHeight) / 2 + dip1.Y;
             return top;
         }
 


Jack251970 (changes requested):
There are two bugs here:

1. When window is shown at the first time. Its position is weird.
2. When ActualHeight is changed, its position is not updated. And the next time you open the main window, it will stay in the same position as origin. (In demo video, I disabled Home Page feature to change the ActualHeight of main window when query is empty)

Demo video:

https://github.com/user-attachments/assets/984fc808-f008-42c7-9d85-f7277428496c

VictoriousRaptor:
Please do not change it this way. If we really vertically center it when we have 9 results, the bottom ones may be out of the screen.

We can create a new option that really centers, but please do not change this one.

dcog989:
> we have 9 results, the bottom ones may be out of the screen.

That is not the result - it centers on `ActualHeight` so whether we display 1 line or 20, that 'block' is vertically centered on the screen. So that provides what the user would expect when they choose 'Center', as opposed to 'somewhere near the top of the page' with current behavior?

Please clarify then I can address issues that @Jack251970 raises.

VictoriousRaptor:
> > we have 9 results, the bottom ones may be out of the screen.
> 
> That is not the result - it centers on `ActualHeight` so whether we display 1 line or 20, that 'block' is vertically centered on the screen. So that provides what the user would expect when they choose 'Center', as opposed to 'somewhere near the top of the page' with current behavior?
> 
> Please clarify then I can address issues that @Jack251970 raises.

Well my suggestion is don't change the behavior of the current Center option. Created a new one maybe better for user experience.

Jack251970:
I agree. Please do not change the behaviour of this option.

jjw24:
Marked as draft. Please introduce as new option if needed.

PR #4023, opened 2025-09-29
Refactor ShellSettings with Binding & Fix IsEnabled logic

Jack251970:
# CHANGES

* Use `Binding` instead of event handler to handle UI logic.
* Use `Flow.Launcher.Localization` to improve code quality.
* Fix `IsEnabled` logic: When command type is `RunCommand`, these two options cannot take effect - `Close Command Prompt after pressing any key` & `Do not close Command Prompt after command execution` - so we should disabled these two options.
* Fix `ShowOnlyMostUsedCMDsNumber` default value issue: Since the old Settings class set default value of `ShowOnlyMostUsedCMDsNumber` to 0 which is a wrong value, we need to fix it here to make sure the default value is 5

# TEST

Setting panel can work well


Plugins/Flow.Launcher.Plugin.Shell/Views/ShellSetting.xaml
```diff
@@ -23,50 +29,72 @@
             Grid.Row="0"
             Margin="{StaticResource SettingPanelItemRightTopBottomMargin}"
             HorizontalAlignment="Left"
-            Content="{DynamicResource flowlauncher_plugin_cmd_relace_winr}" />
+            Content="{DynamicResource flowlauncher_plugin_cmd_relace_winr}"
+            IsChecked="{Binding Settings.ReplaceWinR, Mode=TwoWay}" />
         <CheckBox
```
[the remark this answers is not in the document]
Jack251970: No need

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -383,9 +384,15 @@ public void Init(PluginInitContext context)
             Context = context;
             _settings = context.API.LoadSettingJsonStorage<Settings>();
             context.API.RegisterGlobalKeyboardCallback(API_GlobalKeyboardEvent);
+            // Since the old Settings class set default value of ShowOnlyMostUsedCMDsNumber to 0 which is a wrong value,
+            // we need to fix it here to make sure the default value is 5
+            if (_settings.ShowOnlyMostUsedCMDsNumber == 0)
```
jjw24, line 390: What happens if the default is 0? Does it just not show any or is it causing an error?

Jack251970: > What happens if the default is 0? Does it just not show any or is it causing an error?

I think nothing will be shown in the history list

jjw24: But that's still fine though because some users may not want to use this option. Can we remove this extra code (so we don't need to come back and clean it up) and just let new installations use the default 5?

Jack251970: > But that's still fine though because some users may not want to use this option. Can we remove this extra code (so we don't need to come back and clean it up) and just let new installations use the default 5?

I am afraid this will cause some UI problems since 0 is not the member of sources which lead to blank ui in that ComboBox.

jjw24: Oh ok. Let me add a comment.

Plugins/Flow.Launcher.Plugin.Shell/Main.cs
```diff
@@ -383,9 +384,15 @@ public void Init(PluginInitContext context)
             Context = context;
             _settings = context.API.LoadSettingJsonStorage<Settings>();
             context.API.RegisterGlobalKeyboardCallback(API_GlobalKeyboardEvent);
+            // Since the old Settings class set default value of ShowOnlyMostUsedCMDsNumber to 0 which is a wrong value,
+            // we need to fix it here to make sure the default value is 5
```
jjw24, line 388:
```suggestion
            // we need to fix it here to make sure the default value is 5
            // todo: remove this code block after release v2.2.0
```

PR #4037, opened 2025-10-09
Prefer https over http setting for url

### LINKED ISSUE #3773: Option to prefer https:// urls
There’s no option to choose https:// URLs by default.
If you type any URL without https://, it opens with http://.
It could be a good option to have, since it’s also useful that it opens http:// URLs — in my case, it works fine to open localhost.
Or maybe there could be a function to detect whether it should open with http:// or https://.

I would like to contribute in this too :)

AWAS666:
Closes #3773

This is still missing some translations though and it also adds a settings panel to the url plugin.

There are some other unused properties in this plugin, not sure if these can be deleted?

Also default behaviour is http but I think https should be the state of the art at this point

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Url/SettingsViewModel.cs
+++ b/Plugins/Flow.Launcher.Plugin.Url/SettingsViewModel.cs
@@ -1,9 +1,0 @@
-﻿using System.Collections.Generic;
-using System.Linq;
-
-namespace Flow.Launcher.Plugin.Url;
-
-public class SettingsViewModel(Settings settings) : BaseModel
-{
-    public Settings Settings { get; } = settings;   
-}


Plugins/Flow.Launcher.Plugin.Url/Main.cs
```diff
@@ -58,13 +59,13 @@ public List<Result> Query(Query query)
                         Score = 8,
                         Action = _ =>
                         {
-                            if (!raw.ToLower().StartsWith("http"))
+                            if (!raw.StartsWith("http://", StringComparison.OrdinalIgnoreCase) && !raw.StartsWith("https://", StringComparison.OrdinalIgnoreCase))
                             {
-                                raw = "http://" + raw;
+                                raw = GetHttpPreference() + "://" + raw;
```
VictoriousRaptor, line 65: If https is not reachable is there a fallback? Or it's just browser's job to handle that?

jjw24: Don't think it needs a fallback, if it's specified to use https and that fails then it's fine and vice versa.

AWAS666: I mean before it always used http and it relied on the browser/website to redirect.

As most websites are https nowadays I think this is fine or rather should be defaulted to even

Jack251970: Agree with jjw24. We just need to let users make the choice

Jack251970:
LGTM but I would like to merge it after #3944 since #3944 brings more changes and I would prefer the resolve the conflicts in this PR.

Btw,  @jjw24 can #3944 be merged?

jjw24:
Merged 3944.

PR #4053, opened 2025-10-15
BUG: Explorer plugin navigate path should only show results in current path

01Dri:
# CHANGES

- Union results with Quick Access Links only when it’s not a path search action.


jjw24:
@01Dri the change is good, just needed to handle one more scenario where users may search through Quick Access list using path query. I have added the change, please have a look and let me know if it's ok.

01Dri:
> @01Dri the change is good, just needed to handle one more scenario where users may search through Quick Access list using path query. I have added the change, please have a look and let me know if it's ok.

I tested the new scenario and it worked correctly. Everything looks good

jjw24 (approved):

PR #4057, opened 2025-10-17
History results display actual result icon for Last Opened history style via IcoPath persistence

jjw24:
**Context:**
Display icon on last history results

Follows on with #4042

**Changes:**
- Fixed result highlighting and scrolling when switched to history mode
- Changed the property flow uses to display icons, from IcoPath to new IcoPathAbsolute. This helps IcoPath be serializable while flow uses the constructed full path from IcoPathAbsolute.
- Changed LastOpenedHistoryItem class to LastOpenedHistoryResult.
- Introduced inheritance, LastOpenedHistoryResult is now a child of Result. This helps retaining properties from the Result class and not needing to new up the Result class every time to return results for display. LastOpenedHistoryResult and Result are now serializable, useful for future when we need to provide similar special Results that need to be saved to disk.
- Fixed existing production preview issues where icons missing preview after history mode
 
**Before**
{7AD844BE-A3D1-412F-8F99-5BF9B6B3CCA5}

**After**
{080F9018-460F-4CEC-9893-8C38F1FF3947}

**Tested**
1. Results with the same query do not duplicate when saved
2. IcoPathAbsolute contains the full path
3. IcoPath updated from user selection when different
4. Moved flow to a different location and relative path images still work
5. History mode navigation behaves correctly


jjw24:
@Jack251970 @01Dri thoughts on showing the last opened with the actual result icon? Should it just show the icon instead of badge icon as well?

jjw24:
Also, there is a bug in the comparison logic when saving history results because as seen in the screenshot 'Save Settings' result shouldn't be duplicated.

01Dri:
> Also, there is a bug in the comparison logic when saving history results because as seen in the screenshot 'Save Settings' result shouldn't be duplicated.

It’s not really a bug. The equality comparison also uses the query, so in scenarios where I type “Ter” and open Terminal, and then type “Term” and open Terminal again, I end up with two history items “Ter” and “Term” both referring to the same application (Terminal), which causes duplicate results.

In Query style, that makes sense because the two queries are different.
However, in Last Opened style, it doesn’t really make sense.
A suggestion would be to filter the history items, removing duplicates by name and sorting them by the most recent execution date/time.
**Here’s an example in MainViewModels.cs -> GetHistoryItems:**

```
    foreach (var h in historyItems
               .OrderByDescending(x => x.ExecutedDateTime) 
               .DistinctBy(x => x.GetCompositeKey()).ToList())
```
 

**CompositeKey (This name is just an example; it could be something else)**

```
   public string GetCompositeKey()
    {
        var parts = new[] { Title, SubTitle, RecordKey }
            .Where(s => !string.IsNullOrEmpty(s));

        return string.Join("-", parts);
    }

```

01Dri:
> @Jack251970 @01Dri thoughts on showing the last opened with the actual result icon? Should it just show the icon instead of badge icon as well?

I think the icon with the badge is more intuitive for the user, as it indicates that the item is a history entry.

Jack251970:
> Also, there is a bug in the comparison logic when saving history results because as seen in the screenshot 'Save Settings' result shouldn't be duplicated.

As 01Dri said, it is not a bug. Duplicated items will occur if you click A for many times.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1349,7 +1349,9 @@ private List<Result> GetHistoryItems(IEnumerable<LastOpenedHistoryItem> historyI
                             Localize.executeQuery(h.Query) :
                             h.Title,
                         SubTitle = Localize.lastExecuteTime(h.ExecutedDateTime),
-                        IcoPath = Constant.HistoryIcon,
+                        IcoPath = Settings.ShowBadges ? h.IcoPath : Constant.HistoryIcon,
+                        BadgeIcoPath = Settings.ShowBadges ? Constant.HistoryIcon : h.IcoPath,
```
Jack251970: Please check if `h.IcoPath` exists to prevent possible blank icon

Flow.Launcher/Storage/QueryHistory.cs
```diff
@@ -59,6 +59,8 @@ public void Add(Result result)
                     PluginID = result.PluginID,
                     Query = result.OriginQuery.RawQuery,
                     RecordKey = result.RecordKey,
+                    IcoPath = result.IcoPath,
```
Jack251970: `result.IcoPath` should be absolute path which depends on Flow installed place, plugin version, Flow portable mode, etc. I think here we should use the relative path:

If this icon is inside the directory of the preinstalled plugins, let us store a relative path based on preinstalled plugin directory.

Else if the icon is inside the directory of the Flow data directory, let us store a relative path based on Flow data directory.

Else let us just use the absolute path.

Jack251970 (changes requested):
Showing icons enhance users' experience and I agree with this change. But there is still something we need to resolve

jjw24:
@01Dri @Jack251970

>I think the icon with the badge is more intuitive for the user, as it indicates that the item is a history entry.

What about just use the actual results icons, no badge?

>However, in Last Opened style, it doesn’t really make sense.
A suggestion would be to filter the history items, removing duplicates by name and sorting them by the most recent execution date/time.

I think we should apply filtering of duplicates when last opened style is selected. What do you think?

Jack251970:
>I think we should apply filtering of duplicates when last opened style is selected. What do you think?

@jjw24 I think we can do the filtering here as long as we keep the history in the order of execute time so that we can utilize the history list more effectively.

Flow.Launcher/Storage/QueryHistory.cs
jjw24: >It’s not really a bug. The equality comparison also uses the query, so in scenarios where I type “Ter” and open Terminal, and then type “Term” and open Terminal again, I end up with two history items “Ter” and “Term” both referring to the same application (Terminal), which causes duplicate results.

>As 01Dri said, it is not a bug. Duplicated items will occur if you click A for many times.

@01Dri @Jack251970 I still don't quite see why the duplicates should exist if the same query e.g. 'paint' is pressed several times, shouldn't you just find the exact record from the history list and update it? If I typed 'pain' then selected 'paint' then I can see that the result should be duplicated due to query difference.

The code here just adds the record to history if the last item is not the same as the user selected. So I can essentially end up with duplicate of the exact query if I selected something else in between:

Image

Jack251970: @jjw24 This is the logic Flow used before. And I just do not want to change that in that PR. If you think duplication really annoys your, feel free to change it

jjw24: Ok understand, I will change it in this PR, I don't think it's necessary/correct to duplicate on the exact same query.

01Dri:
@Jack251970 @jjw24 Sorry for the question, I’m new to the open source community. My question is: am I allowed to make the requested changes, or is this PR only for jjw? Thanks!

Jack251970:
> @Jack251970 @jjw24 Sorry for the question, I’m new to the open source community. My question is: am I allowed to make the requested changes, or is this PR only for jjw? Thanks!

You can create PR to last_history_show_result_icon branch

01Dri:
> > @Jack251970 @jjw24 Sorry for the question, I’m new to the open source community. My question is: am I allowed to make the requested changes, or is this PR only for jjw? Thanks!
> 
> You can create PR to last_history_show_result_icon branch

Ok thanks!

jjw24:
@01Dri just to add, of course you can, just let the PR author know what/which area you are changing to avoid overlapping. If very minor like comments, typos or small corrections to the logic that you are sure won't overlap with what the author is doing then go ahead, otherwise its good to let the PR author know you intend to help out with certain areas of the PR.

Having said that, I am still making changes to this draft but help is very welcomed still.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1320,78 @@ private void QueryHistory()
...
+        /// <c>PluginDirectory</c> (which in turn resolves <c>IcoPathAbsolute</c>).
+        ///
+        /// Important:
+        /// - Plugins must be initialized (their metadata and <c>PluginDirectory</c> set)
+        ///   before calling this method; otherwise icon resolution cannot be performed.
+        /// </remarks>
+        internal void RefreshLastOpenedHistoryResults()
+        {
+            // TODO: remove after release v2.3.0
+            _history.PopulateHistoryFromLegacyHistory();
```
Jack251970, line 1387: Could we mark it as obsolete?

Jack251970 (changes requested):

Flow.Launcher/Storage/QueryHistory.cs
jjw24: Done. I have changed the way Flow uses IcoPath, which now allows this property to be serializable with relative path when provided. Flow now uses IcoPathAbsolute which is the full path constructed from IcoPath. This makes the plugin icon images portable.

jjw24:
Change done. @01Dri Please have look at the change and take it for a spin if you can, see if everything works as expected.

Flow.Launcher/Storage/LastOpenedHistoryResult.cs
```diff
@@ -0,0 +1,121 @@
...
+                && PluginID == r.PluginID
+                && Query == r.OriginQuery.TrimmedQuery;
+        }
+        else
+        {
+            return RecordKey == r.RecordKey
+                && PluginID == r.PluginID
+                && Query == r.OriginQuery.TrimmedQuery;
+        }
+    }
```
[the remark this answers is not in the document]
jjw24: Let it throw if null so can be fixed.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1323,77 @@ private void QueryHistory()
...
-                        },
-                        Glyph = new GlyphInfo(FontFamily: "/Resources/#Segoe Fluent Icons", Glyph: "\uE81C")
-                    };
-                    results.Add(result);
-                }
+                // Items saved to disk are differentiated by Query also, but LastOpened style only cares about unique results
+                historyItems = historyItems
+                                 .GroupBy(r => new { r.Title, r.SubTitle, r.PluginID, r.RecordKey })
+                                 .Select(g => g.First());
             }
```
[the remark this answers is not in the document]
Jack251970: @jjw24 Do you think we need to apply this change?

Flow.Launcher/App.xaml.cs
```diff
@@ -259,6 +259,8 @@ await API.StopwatchLogInfoAsync(ClassName, "Startup cost", async () =>
 
                     await PluginManager.InitializePluginsAsync(_mainVM);
 
+                    _mainVM.RefreshLastOpenedHistoryResults();
```
Jack251970, line 263: Why do we call `RefreshLastOpenedHistoryResults` in `App.xaml.cs`? Can we call it in the constructor of the `MainViewModel` in the original position?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -152,11 +153,10 @@ public MainViewModel()
             };
 
             _historyItemsStorage = new FlowLauncherJsonStorage<History>();
-            _userSelectedRecordStorage = new FlowLauncherJsonStorage<UserSelectedRecord>();
-            _topMostRecord = new FlowLauncherJsonStorageTopMostRecord();
             _history = _historyItemsStorage.Load();
-            _history.PopulateHistoryFromLegacyHistory();
+            _userSelectedRecordStorage = new FlowLauncherJsonStorage<UserSelectedRecord>();
```
Jack251970, line 156: Why do we call `RefreshLastOpenedHistoryResults` in `App.xaml.cs`? Can we call it in this position?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -355,7 +355,11 @@ private void LoadHistory()
             if (QueryResultsSelected())
             {
                 SelectedResults = History;
-                History.SelectedIndex = _history.LastOpenedHistoryItems.Count - 1;
+                if (History.Results.Count > 0)
+                {
+                    History.SelectedIndex = 0;
```
Jack251970: Why do you use the first item instead of the last item as origin?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1322,82 @@ private void QueryHistory()
...
-                            App.API.BackToQueryResults();
-                            App.API.ChangeQuery(h.Query);
-                            return false;
-                        },
-                        Glyph = new GlyphInfo(FontFamily: "/Resources/#Segoe Fluent Icons", Glyph: "\uE81C")
-                    };
-                    results.Add(result);
-                }
+                // Items saved to disk are differentiated by Query also, but LastOpened style only cares about unique results
+                historyItems = historyItems
```
Jack251970, line 1331: Do you think we need to consider ordering by ExecutedDateTime before deduplication?

For Last Opened style, the current grouping takes the first occurrence of each unique result, but the most recent execution (highest ExecutedDateTime) should be preferred. According to the PR discussion, contributor 01Dri suggested ordering by ExecutedDateTime descending before applying DistinctBy (or in this case, GroupBy).

```
.OrderByDescending(r => r.ExecutedDateTime)
```

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1322,82 @@ private void QueryHistory()
...
-            else
+
+            foreach (var item in historyItems)
             {
-                foreach (var h in historyItems)
+                var copiedItem = item.DeepCopy();
+
+                if (Settings.HistoryStyle == HistoryStyle.Query)
+                {
+                    copiedItem.Title = Localize.executeQuery(copiedItem.Query);
```
Jack251970: Will the old migrated history items have no title here if query history style is LastOpened? (You can check the previous version)

Jack251970 (changes requested):
Please check the comments

Flow.Launcher/Storage/LastOpenedHistoryResult.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                App.API.BackToQueryResults();
+                App.API.ChangeQuery(queryValue);
+                return false;
+            },
+            // Used for Last Opened History style reopening, currently need to be assigned at MainViewModel.cs
+            AsyncAction = null,
+            Glyph = glyphValue != null 
+                        ? new GlyphInfo(Glyph.FontFamily, Glyph.Glyph)
+                        : null,
```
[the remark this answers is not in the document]
Jack251970: @jjw24 I think we can just refer to the same instance here?

Flow.Launcher/App.xaml.cs
jjw24: Because refresh needs to happen after plugins are initialised which doesnt happen during the MainViewModel constructor phase.

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: hmm, I thought it felt odd to be the last result, when you go to a new page would you expect to be on the first result?

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: The old migrated items will have the title already set in PopulateHistoryFromLegacyHistory call.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > hmm, I thought it felt odd to be the last result, when you go to a new page would you expect to be on the first result?

image

For now, our historical records are sorted from the earliest to the latest. If you want to select the first item, it means we select the earliest record. This should be wrong. Normally, we want to check the latest history record since we remember it clearliest.

I think we should select the latest history record at least.

Jack251970: IMO We have two options: either sort the list from new to old and then select the first one, or sort the list from old to new and then select the last one

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: I don't think it's necessary to do this because GroupBy preserves the order and results are added sequentially so the one taken after grouping always has the latest result execution time. Unless user changes the order directly in the json file content which I prefer they don't do, order shouldn't be a problem.

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: I will change it to sort new to old.

jjw24: @Jack251970 you changed the following from SelectedResults
History.SelectedIndex = 0;
History.SelectedItem = History.Results[0];

While functionally no difference but I think we should be adjusting the SelectedResults' index (the target) not the original object Hisotry (source) right? Reads better at least. Ok for me to change back?

Jack251970: Okay. I just hope to remain the same as before. Feel free to change it

jjw24:
@coderabbitai full review

Flow.Launcher/Storage/LastOpenedHistoryResult.cs
```diff
@@ -0,0 +1,117 @@
...
+        // Used for Query History style reopening
+        Action = _ =>
+        {
+            App.API.BackToQueryResults();
+            App.API.ChangeQuery(result.OriginQuery.TrimmedQuery);
+            return false;
+        };
+        // Used for Last Opened History style reopening, currently need to be assigned at MainViewModel.cs
+        AsyncAction = null;
+    }
```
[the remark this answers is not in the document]
jjw24: Let it fail.

Flow.Launcher/Storage/LastOpenedHistoryResult.cs
```diff
@@ -0,0 +1,117 @@
...
+                && PluginID == r.PluginID
+                && Query == r.OriginQuery.TrimmedQuery;
+        }
+        else
+        {
+            return RecordKey == r.RecordKey
+                && PluginID == r.PluginID
+                && Query == r.OriginQuery.TrimmedQuery;
+        }
+    }
```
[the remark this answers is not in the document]
jjw24: Let it fail.

Jack251970:
@jjw24 I found one bug. If we enable history items in home page and we switch from history page to home page. The preview panel of the first item will be blank.

image

And the item order is not correct either.

image

Here is my settings.

image

Flow.Launcher/Storage/QueryHistory.cs
```diff
@@ -44,23 +88,50 @@ public void Add(Result result)
...
+            // If the last item is the same as the current result, just update the timestamp and the icon path
+            if (LastOpenedHistoryItems.Count > 0 &&
+                TryGetLastOpenedHistoryResult(result, out var existingHistoryItem))
             {
-                LastOpenedHistoryItems.Last().ExecutedDateTime = DateTime.Now;
+                existingHistoryItem.ExecutedDateTime = DateTime.Now;
+                if (existingHistoryItem.IcoPath != result.IcoPath)
+                {
+                    existingHistoryItem.IcoPath = result.IcoPath;
+                }
```
Jack251970: Do you think we should also copy Glyph here?

Jack251970 (changes requested):

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1352,6 +1352,18 @@ private List<Result> GetHistoryItems(IEnumerable<LastOpenedHistoryResult> histor
                 if (Settings.HistoryStyle == HistoryStyle.Query)
                 {
                     copiedItem.Title = Localize.executeQuery(copiedItem.Query);
+                    copiedItem.IcoPath = Constant.HistoryIcon;
+                    // TODO: Add Glyph here
+                    // copiedItem.Glyph = new GlyphInfo(FontFamily: "/Resources/#Segoe Fluent Icons", Glyph: "\uE81C");
```
Jack251970: I think we should also copy Glyph here.

VictoriousRaptor:
Don't have much time review these days:(

Flow.Launcher/Storage/QueryHistory.cs
```diff
@@ -44,23 +70,72 @@ public void Add(Result result)
...
                 {
-                    Title = result.Title,
-                    SubTitle = result.SubTitle,
-                    PluginID = result.PluginID,
-                    Query = result.OriginQuery.TrimmedQuery,
-                    RecordKey = result.RecordKey,
-                    ExecutedDateTime = DateTime.Now
-                });
+                    existingHistoryItem.IcoPath = result.IcoPath;
+                }
```
Jack251970: Do you think we should also copy Glyph here?

Flow.Launcher/Helper/ResultHelper.cs
```diff
@@ -24,7 +24,7 @@ public static class ResultHelper
         if (query == null) return null;
         try
         {
-            var freshResults = await plugin.Plugin.QueryAsync(query, CancellationToken.None);
+            var freshResults = await PluginManager.QueryForPluginAsync(plugin, query, CancellationToken.None);
```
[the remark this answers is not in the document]
Jack251970: We need to use this function so that we can handle plugin initialization correctly

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1317,106 @@ private void QueryHistory()
...
-                        Action = _ =>
-                        {
-                            App.API.BackToQueryResults();
-                            App.API.ChangeQuery(h.Query);
-                            return false;
-                        },
-                        Glyph = new GlyphInfo(FontFamily: "/Resources/#Segoe Fluent Icons", Glyph: "\uE81C")
-                    };
-                    results.Add(result);
+                        copiedItem.IcoPath = Constant.MissingImgIcon;
```
jjw24: IcoPath shouldn't be null, this should not be necessary. If you just add the default IcoPath and Glyph as history icon when calling the PopulateHistoryFromLegacyHistory then you don't need to assign them here. Why are you setting the default IcoPath in PopulateHistoryFromLegacyHistory as null?

jjw24:
> @jjw24 The item order is not correct under Last Opened History option. (I have openned `Save Settings` item for two times. One before `Wechat` and another after `WeChat`)
> 
> image
> Here are my settings.
> 
> image

Homepage display takes into consideration of result selection ranking, the behavior is, when on fresh boot the result is ordered descending by date, and then they move up the more you select them. Is this not the case?

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: I think you are getting the missing image icon during testing like in one of your screenshots is because you are testing with History.json from dev branch after it has been migrated by the previous PR. This means when you try this PR PopulateHistoryFromLegacyHistory is not called and consequently IcoPath is not set (because the code doesn't need to set it for normal results).

You need to test with History.json from prod build or as a fresh build where History.json doesn't exist.

Flow.Launcher/Storage/QueryHistory.cs
jjw24: Yes, I forgot about Glyph. If Glyph is different then reassign it.

Jack251970:
> > @jjw24 The item order is not correct under Last Opened History option. (I have openned `Save Settings` item for two times. One before `Wechat` and another after `WeChat`)
> > image
> > Here are my settings.
> > image
> 
> Homepage display takes into consideration of result selection ranking, the behavior is, when on fresh boot the result is ordered descending by date, and then they move up the more you select them. Is this not the case?

Oh yeah. I forgot the user selection score

Flow.Launcher/Storage/QueryHistory.cs
```diff
@@ -44,23 +70,73 @@ public void Add(Result result)
...
             {
-                LastOpenedHistoryItems.Add(new LastOpenedHistoryItem
+                LastOpenedHistoryItems.Add(new LastOpenedHistoryResult(result));
+            }
+        }
+
+        /// <summary>
+        /// Update a result into the last-opened history list (<see cref="LastOpenedHistoryItems"/>).
+        /// </summary>
+        public void Update(Result result)
```
jjw24: What scenario do we need this for? This is pretty much the same as the add method.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > IcoPath shouldn't be null, this should not be necessary. If you just add the default IcoPath and Glyph as history icon when calling the PopulateHistoryFromLegacyHistory then you don't need to assign them here. Why are you setting the default IcoPath in PopulateHistoryFromLegacyHistory as null?

I think setting IcoPath to history icon path is unnecessary since this path is accessible in codes and we can use a `if` sentence to set the history icon

Jack251970: > I think you are getting the missing image icon during testing like in one of your screenshots is because you are testing with History.json from dev branch after it has been migrated by the previous PR. This means when you try this PR PopulateHistoryFromLegacyHistory is not called and consequently IcoPath is not set (because the code doesn't need to set it for normal results).
> 
> You need to test with History.json from prod build or as a fresh build where History.json doesn't exist.

I think if we do not know the actual icon for some history items, we should use this missing icon instead of the history icon as the default icon under Last Opened Mode.

Flow.Launcher/Storage/QueryHistory.cs
Jack251970: > What scenario do we need this for? This is pretty much the same as the add method.

When we are executing one command from history item itself.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: Since the badge icon is already the history icon...

jjw24: >I think setting IcoPath to history icon path is unnecessary since this path is accessible in codes and we can use a if sentence to use history icon

Setting by default where it is loaded is better than putting in if statement, I think the correct use of if here is set a different value 'if the required property (such as SubTitle) needs to be different than the source'. 

>I think if we do not know the actual icon for some history items, we should use this missing icon instead of the history icon as the icon under Last Opened Mode.

We should know the actual icon though, because LastOpenHistoryResult is a child class of Result, so unless Result doesn't for some reason have IcoPath assigned, in this case if I remember correctly if IcoPath is null or pointing to a non-existent image then code later on will assign missing icon image by default anyway.

Flow.Launcher/Storage/QueryHistory.cs
jjw24: hm, I am aware that when you select a result from the history page or I think also home page it doesn't call `Add()`, I thought this was by design as selecting history results don't count towards user selection like normal acutal results do. However, changing this looks like introducing a little bit of unnecessary complexity in the code.

Jack251970: > hm, I am aware that when you select a result from the history page or I think also home page it doesn't call `Add()`, I thought this was by design as selecting history results don't count towards user selection like normal acutal results do. However, changing this looks like introducing a little bit of unnecessary complexity in the code.

I think this logic is necessary since it keeps the icon & execute date up-to-date. And in codes, we also handle the user select record, right?

```
if (reflectResult != null)
{
    // Record the user selected record for result ranking
    _userSelectedRecord.Add(reflectResult);

    // Update the history with the reflected result (I think it is necessary)
    _history.Update(reflectResult);

    // Since some actions may need to hide the Flow window to execute
    // So let us populate the results of them
    return await reflectResult.ExecuteAsync(c);
}
```

Jack251970: If you think it is unnecessary, let us remove them all.

```
// Record the user selected record for result ranking
_userSelectedRecord.Add(reflectResult);

// Update the history with the reflected result (I think it is necessary)
_history.Update(reflectResult);
```

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > Setting by default where it is loaded is better than putting in if statement, I think the correct use of if here is set a different value 'if the required property (such as SubTitle) needs to be different than the source'. 

Sorry, I do not agree with you. The path saved locally is static, while the path in the code is dynamic and runtime. If we change this path from `Images\history.png` to `Images\history_icon.png`, will the old path of that history icon still work? But the path from codes should always be right.

> We should know the actual icon though, because LastOpenHistoryResult is a child class of Result, so unless Result doesn't for some reason have IcoPath assigned, in this case if I remember correctly if IcoPath is null or pointing to a non-existent image then code later on will assign missing icon image by default anyway.

Yes, LastOpenHistoryResult should know the actual icons. But if we set a static history icon as the icon paths of those old data, the badge icons will be the same as the icons.

Flow.Launcher/Storage/QueryHistory.cs
Jack251970: Reverted as you said

Flow.Launcher/Storage/QueryHistory.cs
Jack251970: > Yes, I forgot about Glyph. If Glyph is different then reassign it.

However, the Glyph is a init-only property. How can we edit it?

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: I think `IcoPath` should store its actual icons

Flow.Launcher/Storage/QueryHistory.cs
```diff
@@ -14,28 +16,52 @@ public class History
...
         {
             if (Items.Count == 0) return;
             // Migrate old history items to new LastOpenedHistoryItems
             foreach (var item in Items)
             {
-                LastOpenedHistoryItems.Add(new LastOpenedHistoryItem
+                LastOpenedHistoryItems.Add(new LastOpenedHistoryResult
                 {
+                    Title = Localize.executeQuery(item.Query),
+                    IcoPath = null,
```
Jack251970: Set it to null and parse it during runtime

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -431,14 +435,17 @@ private void LoadContextMenu()
             {
                 // When switch to ContextMenu from QueryResults, but no item being chosen, should do nothing
                 // i.e. Shift+Enter/Ctrl+O right after Alt + Space should do nothing
-                if (SelectedResults.SelectedItem != null)
+                if (SelectedResults.SelectedItem?.Result != null &&
+                    !string.IsNullOrEmpty(SelectedResults.SelectedItem.Result.PluginID))  // Do not show context menu for history items
```
Jack251970: Do not show context menu for history items since their PluginID is empty which can cause issues

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -431,14 +435,17 @@ private void LoadContextMenu()
...
                 {
                     SelectedResults = ContextMenu;
                 }
             }
             else
             {
                 SelectedResults = Results;
+                PreviewSelectedItem = Results.SelectedItem;
```
Jack251970, line 448: Fix preview panel blank issue

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -642,6 +649,8 @@ private void Esc()
             if (!QueryResultsSelected())
             {
                 SelectedResults = Results;
+                PreviewSelectedItem = Results.SelectedItem;
```
Jack251970, line 653: Fix preview panel blank issue

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1252,22 +1261,12 @@ private void QueryContextMenu()
 
             var selected = Results.SelectedItem?.Result;
 
-            if (selected != null) // SelectedItem returns null if selection is empty.
+            if (selected != null && // SelectedItem returns null if selection is empty.
+                !string.IsNullOrEmpty(selected.PluginID))  // SelectedItem must have a valid PluginID
```
Jack251970: Check item plugin ID

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1317,100 @@ private void QueryHistory()
             }
         }
 
-        private List<Result> GetHistoryItems(IEnumerable<LastOpenedHistoryItem> historyItems)
+        private List<Result> GetHistoryItems(IEnumerable<LastOpenedHistoryResult> historyItems, int? selectLast = null)
```
Jack251970: Add selectLast parameter for querying history results in home page

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1317,100 @@ private void QueryHistory()
...
+
+            foreach (var item in historyItems)
+            {
+                var copiedItem = item.DeepCopy();
+
+                if (Settings.HistoryStyle == HistoryStyle.Query)
                 {
-                    var result = new Result
+                    copiedItem.Title = Localize.executeQuery(copiedItem.Query);
+                    copiedItem.IcoPath = Constant.HistoryIcon;
```
Jack251970: Use history icon under Query History Mode

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1317,100 @@ private void QueryHistory()
...
+                    // TODO: Add Glyph here
+                    // copiedItem.Glyph = new GlyphInfo(FontFamily: "/Resources/#Segoe Fluent Icons", Glyph: "\uE81C");
+                }
+                else
+                {
+                    if (string.IsNullOrEmpty(copiedItem.IcoPath)) // Must manually set missing image icon here
```
Jack251970: Use actual icon + history icon badge icon under Last Opened Mode

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1318,68 +1317,100 @@ private void QueryHistory()
...
-                                // Record the user selected record for result ranking
-                                _userSelectedRecord.Add(reflectResult);
-
-                                // Since some actions may need to hide the Flow window to execute
-                                // So let us populate the results of them
-                                return await reflectResult.ExecuteAsync(c);
-                            }
+                            // Since some actions may need to hide the Flow window to execute
+                            // So let us populate the results of them
+                            return await reflectResult.ExecuteAsync(c);
```
Jack251970, line 1356: Do not record the user selected record for result ranking

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1617,10 +1648,8 @@ await PluginManager.QueryHomeForPluginAsync(plugin, query, token) :
 
             void QueryHistoryTask(CancellationToken token)
             {
-                // Select last history results and revert its order to make sure last history results are on top
-                var historyItems = _history.LastOpenedHistoryItems.TakeLast(Settings.MaxHistoryResultsToShowForHomePage).Reverse();
-
-                var results = GetHistoryItems(historyItems);
+                // Select last history results
+                var results = GetHistoryItems(_history.LastOpenedHistoryItems, Settings.MaxHistoryResultsToShowForHomePage);
```
Jack251970, line 1630: Combine get history item for home page logic into `GetHistoryItems`

Jack251970 (commented):
Changes

Jack251970:
Final effect under my changes:

Query history Mode: Use history icon.
image

Last opened history mode: Use actual icon + history icon badge. Use missing icon as default value for old data.
image

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: >Sorry, I do not agree with you. The path saved locally is static, while the path in the code is dynamic and runtime. If we change this path from Images\history.png to Images\history_icon.png, will the old path of that history icon still work? But the path from codes should always be right.

Well, it will still work because in my code with PopulateHistoryFromLegacyHistory  we set `IcoPath = Constant.HistoryIcon,`, dynamically assigned on startup.

You are setting null for IcoPath in PopulateHistoryFromLegacyHistory then add a specific if statement to handle this null with `if (string.IsNullOrEmpty(copiedItem.IcoPath))` (Line 1354), I think is not required especially for a throw away backwards compatibility method that only gets run once. Legacy history results that we migrated for the user will always have the history icon and new history results will always contain a non-null IcoPath, so why do we need to handle IcoPath null? Have I missed a scenario?

Jack251970: > >Sorry, I do not agree with you. The path saved locally is static, while the path in the code is dynamic and runtime. If we change this path from Images\history.png to Images\history_icon.png, will the old path of that history icon still work? But the path from codes should always be right.
> 
> Well, it will still work because in my code with PopulateHistoryFromLegacyHistory  we set `IcoPath = Constant.HistoryIcon,`, dynamically assigned on startup.
> 
> You are setting null for IcoPath in PopulateHistoryFromLegacyHistory then add a specific if statement to handle this null with `if (string.IsNullOrEmpty(copiedItem.IcoPath))` (Line 1354), I think is not required especially for a throw away backwards compatibility method that only gets run once. Legacy history results that we migrated for the user will always have the history icon and new history results will always contain a non-null IcoPath, so why do we need to handle IcoPath null? Have I missed a scenario?

Well, feel free to make any changes for this

Jack251970: > >Sorry, I do not agree with you. The path saved locally is static, while the path in the code is dynamic and runtime. If we change this path from Images\history.png to Images\history_icon.png, will the old path of that history icon still work? But the path from codes should always be right.
> 
> Well, it will still work because in my code with PopulateHistoryFromLegacyHistory  we set `IcoPath = Constant.HistoryIcon,`, dynamically assigned on startup.
> 
> You are setting null for IcoPath in PopulateHistoryFromLegacyHistory then add a specific if statement to handle this null with `if (string.IsNullOrEmpty(copiedItem.IcoPath))` (Line 1354), I think is not required especially for a throw away backwards compatibility method that only gets run once. Legacy history results that we migrated for the user will always have the history icon and new history results will always contain a non-null IcoPath, so why do we need to handle IcoPath null? Have I missed a scenario?

Well, feel free to make any changes for this

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: How do I reproduce the blank preview panel issue? My last commit at 58d910d seemed fine:
{D633CE19-D645-4819-8A2B-B293CEBE2A31}

Jack251970: First press Ctrl+H to enter history page and then press Esc to return to result page. The preview panel of the selected item in result page is blank

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: What is this for?

Jack251970: For fixing issues when querying history items in home page. I unified the its querying logic

Flow.Launcher/ViewModel/MainViewModel.cs
jjw24: Ah ok, this has been a production bug actually.

Flow.Launcher/Storage/LastOpenedHistoryResult.cs
```diff
@@ -0,0 +1,146 @@
...
```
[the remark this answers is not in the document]
Jack251970: It will not be null here...

Flow.Launcher/Storage/QueryHistory.cs
jjw24: I changed it to set, don't think init is very necessary.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -426,19 +429,23 @@ private void LoadContextMenu()
...
+                if (SelectedResults.SelectedItem?.Result != null)
                     SelectedResults = ContextMenu;
-                }
             }
             else
             {
                 SelectedResults = Results;
+                PreviewSelectedItem = Results.SelectedItem;
+                _ = UpdatePreviewAsync();
             }
```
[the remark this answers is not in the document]
jjw24: Good catch thank you 👍

jjw24:
@Jack251970 should be good to go, please review.

Jack251970 (changes requested):
~~When we upgrade Flow from old version to new one, the icons of those old results are blank.~~

Image

The updated `history.json` contains null icon path, but the missing icon is not shown.

```
{
      "Query": "steam",
      "ExecutedDateTime": "2026-01-15T16:16:30.9760574+08:00",
      "Title": "Steam",
      "SubTitle": "",
      "ActionKeywordAssigned": null,
      "CopyText": "",
      "AutoCompleteText": null,
      "IcoPath": null,
      "IcoPathAbsolute": null,
      "BadgeIcoPath": null,
      "RoundedIcon": false,
      ...
    },
```

Flow.Launcher.Plugin/Result.cs
```diff
@@ -131,17 +155,19 @@ public string BadgeIcoPath
...
         /// Delegate to load an icon for the badge of this result.
         /// </summary>
+        [JsonIgnore]
         public IconDelegate BadgeIcon = null;
 
         /// <summary>
         /// Information for Glyph Icon (Prioritized than IcoPath/Icon if user enable Glyph Icons)
         /// </summary>
-        public GlyphInfo Glyph { get; init; }
+        public GlyphInfo Glyph { get; set; }
```
Jack251970: This change breaks combability with `Glyph` setters. Take Clipboard+ plugin for example, after upgrade to this version, it cannot work.

Image

jjw24:
I can't reproduce the issue, the missing image icon shows up for me. Maybe:
- Is it the theme you are using, what is it?
- Can you try do the upgrade on a fresh build
- Do you still have the history.json file you upgraded from? Maybe share it with me.

Jack251970:
> I can't reproduce the issue, the missing image icon shows up for me. Maybe:
> 
> * Is it the theme you are using, what is it?
> * Can you try do the upgrade on a fresh build
> * Do you still have the history.json file you upgraded from? Maybe share it with me.

It seems that this problem no longer appears in the latest commit. Let us forget it

Flow.Launcher.Plugin/Result.cs
Jack251970: @jjw24 I use a workaround to handle this combability in https://github.com/Flow-Launcher/Flow.Launcher/pull/4057/commits/935d41dc265972c433f599a5fbfdff0acd798542. Please check if this works for you

Jack251970 (dismissed):
Overall, it works well for me. And please do the final check if necessary

Flow.Launcher.Plugin/Result.cs
VictoriousRaptor: > This change breaks combability with `Glyph` setters. Take Clipboard+ plugin for example, after upgrade to this version, it cannot work.
> 
> Image

is it because the referred plugin project not upated yet in cb+ plugin?

Jack251970: > > This change breaks combability with `Glyph` setters. Take Clipboard+ plugin for example, after upgrade to this version, it cannot work.
> > 
> > Image
> 
> is it because the referred plugin project not upated yet in cb+ plugin?

But this can cause all old plugins cannot work if they use the Glyph. Forcing all of them to use the new Plugin nuget package is impossible

jjw24: > @jjw24 I use a workaround to handle this combability in [935d41d](https://github.com/Flow-Launcher/Flow.Launcher/commit/935d41dc265972c433f599a5fbfdff0acd798542). Please check if this works for you

Good catch, I overlooked the compatibility with plugins.

jjw24: I am finding an old issue, the preview no longer works after going to history and back- `ctrl-h` then `ctrl+h`. Can you see if you can reproduce please.

Jack251970: > I am finding an old issue, the preview no longer works after going to history and back- `ctrl-h` then `ctrl+h`. Can you see if you can reproduce please.

This is the preview panel issue I have described before. And I have resolved that already

Jack251970: > I am finding an old issue, the preview no longer works after going to history and back- `ctrl-h` then `ctrl+h`. Can you see if you can reproduce please.

Check https://github.com/Flow-Launcher/Flow.Launcher/pull/4057/changes/938a87ceca9973ea8dceeb95edc39ac434c4c3eb & https://github.com/Flow-Launcher/Flow.Launcher/pull/4057/changes/cf4268d1c32ba70860d6e5a1a762b8910bd45ac4 please

jjw24: Nah not fully, this issue is from a different code path than the two commits. I found the problem and will push a fix soon.

Jack251970 (approved):

PR #4093, opened 2025-11-05
[FEATURE] Folder and File Action Keywords

01Dri:
## **CHANGES**

- Added new **File** and **Folder** action keywords  
- Implemented **action keyword–based result filtering**  
- **Refactored** the Search Manager for better maintainability 
---

### Note
The **Search Manager** in Explorer was refactored to simplify the addition of new action keywords.  
Previously, each new keyword required code changes in multiple places.  
Now, the logic for action keywords and filters is much easier to understand and extend.

The quick access results is filtered action keyword based

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
+++ b/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -23,6 +23,7 @@ namespace Flow.Launcher.Plugin.Explorer.Search
         {
             { ActionKeyword.FileSearchActionKeyword, [ResultType.File] },
             { ActionKeyword.FolderSearchActionKeyword, [ResultType.Folder, ResultType.Volume] },
+            { ActionKeyword.SearchActionKeyword, [ResultType.File, ResultType.Folder, ResultType.Volume] },
         };
 
 


01Dri:
image

Jack251970 (dismissed):
Thanks for your contribution! LGTM.

Jack251970:
@jjw24 It works well for me generally. But I am not particularly sure about the refactoring within the `SearchAsync` function, please review this for me, thanks!

jjw24:
Yeah no worries I will take a look too.

jjw24 (changes requested):
Just doing some initial testing, I am getting this with the file/folder action keyword when no query is inputted:
{A193A1F7-1DAA-4B5C-8AA0-F31C40811B46}

jjw24 (changes requested):
Searching with actual path should still be triggered with a general search:
{E74CA1F1-A0EC-4C48-9E7D-5B9E02F1D940}

jjw24:
> Just doing some initial testing, I am getting this with the file/folder action keyword when no query is inputted: {A193A1F7-1DAA-4B5C-8AA0-F31C40811B46}

Still getting this for file/folder action keyword.

jjw24:
>For example, the Folder Search action should only display folders, but if there’s a .txt file in Quick Access, it still appears in the results.
My idea introduces a new option to include or exclude Quick Access items when an action keyword is used.

Just re-read your description, with regards to this, folder search should not bring up the .txt file in Quick Access unless it's a folder. I don't think there is a need for an option to include/exclude Quick Access items. If file search action keyword-> .txt in Quick Access should be returned. If folder search action keyword -> a folder called txt in Quick Access should be returned.

An example use case is if I have a folder 'blah' that is not in the Windows Search index but added to Quick Access, using folder action keyword should still bring up my 'blah' folder. This is so I don't need to rely on the index for getting folders I may need every once in a while.

jjw24:
I also don't think we should list any files or folders if query contains only the file/folder action keywords. In this instance we should just return empty result.

01Dri:
> I also don't think we should list any files or folders if query contains only the file/folder action keywords. In this instance we should just return empty result.

I agree

01Dri:
> > For example, the Folder Search action should only display folders, but if there’s a .txt file in Quick Access, it still appears in the results.
> > My idea introduces a new option to include or exclude Quick Access items when an action keyword is used.
> 
> Just re-read your description, with regards to this, folder search should not bring up the .txt file in Quick Access unless it's a folder. I don't think there is a need for an option to include/exclude Quick Access items. If file search action keyword-> .txt in Quick Access should be returned. If folder search action keyword -> a folder called txt in Quick Access should be returned.
> 
> An example use case is if I have a folder 'blah' that is not in the Windows Search index but added to Quick Access, using folder action keyword should still bring up my 'blah' folder. This is so I don't need to rely on the index for getting folders I may need every once in a while.

That makes sense. So essentially, Quick Access results should respect the selected filter, right? When using a folder action, only Quick Access items that are folders should be included in the merged results.

01Dri:
I'm working on corrections.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -47,45 +56,57 @@
...
+
+            var activeActionKeywords = Settings.GetActiveActionKeywords(keyword);
+
+            // No action keyword matched - plugin should not handle this query, return empty results.
+            if (activeActionKeywords.Count == 0)
             {
-                if (string.IsNullOrEmpty(query.Search) && ActionKeywordMatch(query, Settings.ActionKeyword.QuickAccessActionKeyword))
-                    return QuickAccess.AccessLinkListAll(query, Settings.QuickAccessLinks);
+                return [];
             }
```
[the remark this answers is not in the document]
01Dri: GetActiveActionKeywords never returns a null value

01Dri:
@jjw24 Can you test again? Thanks!

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -47,83 +56,95 @@ public int GetHashCode(Result obj)
...
-            // This allows the user to type the below action keywords and see/search the list of quick folder links
-            if (ActionKeywordMatch(query, Settings.ActionKeyword.SearchActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.QuickAccessActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.PathSearchActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.IndexSearchActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.FileContentSearchActionKeyword))
+            var activeActionKeywords = Settings.GetActiveActionKeywords(keyword);
+
+            // No action keyword matched - plugin should not handle this query, return empty results.
+            if (activeActionKeywords.Count == 0)
```
jjw24, line 87: Is it possible to do this at the top before isPathSearch check so can return earlier?

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -231,7 +232,7 @@ private async Task<List<Result>> PathSearchAsync(Query query, CancellationToken
             }
 
             if (token.IsCancellationRequested)
-                return new List<Result>();
+                return [];
```
jjw24: `[.. results]`, neat 👍

Plugins/Flow.Launcher.Plugin.Explorer/Settings.cs
```diff
@@ -154,13 +164,15 @@ public enum ContentIndexSearchEngineOption
 
         #endregion
 
-        internal enum ActionKeyword
+        public enum ActionKeyword
```
jjw24: Why is this changed to public?

01Dri: It was changed in some tests, I fixed it now

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
01Dri: Yes, but if no action keyword is active and the user types 'c:\'\, the path search is not executed (empty result because early return). It's okay ?

Example: 
Screenshot 2025-12-04 231149

VictoriousRaptor: > Yes, but if no action keyword is active and the user types 'c:', the path search is not executed (empty result because early return). It's okay ?
> 

Well that's expected imo. No action keyword means you can't start a search.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -47,83 +56,92 @@ public int GetHashCode(Result obj)
         internal async Task<List<Result>> SearchAsync(Query query, CancellationToken token)
         {
             var results = new HashSet<Result>(PathEqualityComparator.Instance);
+            var keyword = query.ActionKeyword.Length == 0 ? Query.GlobalPluginWildcardSign : query.ActionKeyword;
```
jjw24, line 83:
```suggestion

            var keyword = query.ActionKeyword.Length == 0 ? Query.GlobalPluginWildcardSign : query.ActionKeyword;
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -48,82 +57,92 @@ internal async Task<List<Result>> SearchAsync(Query query, CancellationToken tok
...
+            // No action keyword matched - plugin should not handle this query, return empty results.
+            var activeActionKeywords = Settings.GetActiveActionKeywords(keyword);
+            if (activeActionKeywords.Count == 0)
             {
-                if (string.IsNullOrEmpty(query.Search) && ActionKeywordMatch(query, Settings.ActionKeyword.QuickAccessActionKeyword))
-                    return QuickAccess.AccessLinkListAll(query, Settings.QuickAccessLinks);
+                return [];
             }
-            else
+            var isPathSearch = query.Search.IsLocationPathString()
```
jjw24:
```suggestion

            var isPathSearch = query.Search.IsLocationPathString()
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -48,82 +57,93 @@ internal async Task<List<Result>> SearchAsync(Query query, CancellationToken tok
...
             var results = new HashSet<Result>(PathEqualityComparator.Instance);
 
-            // This allows the user to type the below action keywords and see/search the list of quick folder links
-            if (ActionKeywordMatch(query, Settings.ActionKeyword.SearchActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.QuickAccessActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.PathSearchActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.IndexSearchActionKeyword)
-                || ActionKeywordMatch(query, Settings.ActionKeyword.FileContentSearchActionKeyword))
+            var keyword = query.ActionKeyword.Length == 0 ? Query.GlobalPluginWildcardSign : query.ActionKeyword;
+            // No action keyword matched - plugin should not handle this query, return empty results.
```
jjw24, line 85:
```suggestion

            // No action keyword matched - plugin should not handle this query, return empty results.
```

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -48,82 +57,94 @@ internal async Task<List<Result>> SearchAsync(Query query, CancellationToken tok
...
             }
-            else
+
+            var isPathSearch = query.Search.IsLocationPathString()
+                || EnvironmentVariables.IsEnvironmentVariableSearch(query.Search)
+                || EnvironmentVariables.HasEnvironmentVar(query.Search);
+
+            var queryIsEmpty = string.IsNullOrEmpty(query.Search);
+
+            if (queryIsEmpty && activeActionKeywords.ContainsKey(ActionKeyword.QuickAccessActionKeyword))
```
jjw24, line 93: Almost there. Index search is giving the error without any query:

Image

01Dri: Hmm, strange, it works fine for me.

Screenshot 2025-12-07 133406

01Dri:
> I think we also need to make the folder search behaviour consistent- it should either allow path search for when there is a hotkey or global, or not allow at all:
> 
> With global hotkey:
> 
> Image
> With action keyword:
> 
> Image

I think Folder and Search should not allow global hotkey assignment, similar to Quick Access.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
jjw24: I believe it's still running the query and the error is specific to my environment, but we should have it returned empty list if no query input with the action keyword.

jjw24:
Pretty sure Quick Access allows global action keyword, only one doesn't is content search which is due to significant search time/load. 

I don't see why we shouldn't allow global action keyword for Folder and Search. What's the reasoning?

01Dri:
> Pretty sure Quick Access allows global action keyword, only one doesn't is content search which is due to significant search time/load.
> 
> I don't see why we shouldn't allow global action keyword for Folder and Search. What's the reasoning?

When both file and folder use the global keyword, which one should the search manager prioritize?
The results are filtered based on the action keyword, but for example, if I type “User” and both file and folder actions are active with the global keyword, which results should the search manager show, folders or files?

I think allowing global keywords for both actions can make the results confusing. What do you think? 

Either way, I’ve applied the requested fixes.

jjw24:
>When both file and folder use the global keyword, which one should the search manager prioritize?
The results are filtered based on the action keyword, but for example, if I type “User” and both file and folder actions are active with the global keyword, which results should the search manager show, folders or files?

Results from both file & folder should be returned. User can adjust search or change them to dedicated action keyword, and have Search as global if they want results from both file and folder.

Think of 'Search' as one that does everything (except content), user can disable or and just use Search, and they can further assign action keyword to File & Folder to filter types of search.

On the flip side, user can disable 'Search' and use 'File', 'Folder' & 'Path' as global to achieve the same.

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -44,86 +52,111 @@ public int GetHashCode(Result obj)
...
                     searchResults = Settings.ContentIndexProvider.ContentSearchAsync("", query.Search, token);
                     engineName = Enum.GetName(Settings.ContentSearchEngine);
                     break;
 
-                case false
-                    when ActionKeywordMatch(query, Settings.ActionKeyword.IndexSearchActionKeyword)
-                         || ActionKeywordMatch(query, Settings.ActionKeyword.SearchActionKeyword):
+                case true or false
+                    when activeActionKeywords.ContainsKey(ActionKeyword.QuickAccessActionKeyword):
+                    return QuickAccess.AccessLinkListMatched(query, Settings.QuickAccessLinks);
```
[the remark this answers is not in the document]
jjw24: This scenario won't be the case as QuickAccessActionKeyword is not allowed to be `*`, but you have noted an issue which actually exists where no results returned when both FileSearchActionKeyword and FolderSearchActionKeyword are set to '*'. Investigate the cause. I suspect line 143-155 of SearchManager.CS

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -308,18 +310,22 @@ private List<Result> GetQuickAccessResultsFilteredByActionKeyword(Query query, D
                        && !IsResultTypeFilteredByActionKeyword(result.Type, actions))
                 .ToList();
         }
-        private bool IsResultTypeFilteredByActionKeyword(ResultType type, Dictionary<ActionKeyword, string> actions)
+        private bool IsResultTypeFilteredByActionKeyword(ResultType type, List<ActionKeyword> actions)
         {
-            foreach (var action in actions.Keys)
+            var actionsWithWhitelist = actions.Intersect(_allowedTypesByActionKeyword.Keys).ToList();
+            if (actionsWithWhitelist.Count <= 0) return false;
```
jjw24: This shouldn't need '<', you can't have a negative list.

jjw24 (approved):
@01Dri Good to go, thank you for this amazing work! Explorer plugin due to its ability to handle different search types, has the most complicated code paths out of all plugins. Glad you were able to improve upon this 👍

01Dri:
> @01Dri Good to go, thank you for this amazing work! Explorer plugin due to its ability to handle different search types, has the most complicated code paths out of all plugins. Glad you were able to improve upon this 👍

Thanks a lot! Flow is my favorite app, and I’m glad I can help

Jack251970:
@01Dri I think something does not work right here. In #4287, quick access results seems to be missing under global query.

Jack251970:
@jjw24 Any ideas about #4287? Quick access results seems to be missing under global query.

01Dri:
> @01Dri I think something does not work right here. In #4287, quick access results seems to be missing under global query.

Sorry for the delay. I’m currently traveling and away from home, so I haven’t been able to work on the Flow.

It looks like the issue has already been resolved. I’ll be back home next week and will be able to help if anything else is needed.

Jack251970:
> > @01Dri I think something does not work right here. In #4287, quick access results seems to be missing under global query.
> 
> Sorry for the delay. I’m currently traveling and away from home, so I haven’t been able to work on the Flow.
> 
> It looks like the issue has already been resolved. I’ll be back home next week and will be able to help if anything else is needed.

Thanks! And it is already fixed in #4294

Issue #4016, opened 2025-09-25
BUG: COMException- Desktop composition is disabled


RonnyAllen:
### Checks

- [ ] I have checked that this issue has not already been reported.

- [ ] I am using the latest version of Flow Launcher.

- [ ] I am using the prerelease version of Flow Launcher.

### Problem Description

From [ndahirwaSosthene](https://github.com/ndahirwaSosthene]: This error froze my laptop that the whole screen went black. I don't know if this should be happening.

[2025-09-25.txt](https://github.com/user-attachments/files/2252981Flow Launcher version: 2.0.1
OS Version: 27950.1
IntPtr Length: 8
x64: True

Python Path: C:\Users\rohan\AppData\Roaming\FlowLauncher\Environments\Python\PythonEmbeddable-v3.11.4\pythonw.exe
Node Path: 

Date: 09/25/2025 12:06:23
Exception:
System.Runtime.InteropServices.COMException (0x80263001): {Desktop composition is disabled} The operation could not be completed because desktop composition is disabled. (0x80263001)
   at Standard.NativeMethods.DwmExtendFrameIntoClientArea(IntPtr hwnd, MARGINS& pMarInset)
   at System.Windows.Shell.WindowChromeWorker._ExtendGlassFrame()
   at System.Windows.Shell.WindowChromeWorker._UpdateFrameState(Boolean force)
   at System.Windows.Shell.WindowChromeWorker._HandleDwmCompositionChanged(WM uMsg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Shell.WindowChromeWorker._WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Interop.HwndSource.PublicHooksFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
5/2025-09-25.txt)

### To Reproduce

1. ...
2. ...
3. ...

### Flow Launcher Version

2.0.1

### Windows Build Number

27950.1

### Error Log

<details>

```shell

Replace this line with the important log contents.

```
</details>

[...]

PR #4113, opened 2025-11-16
Handle recoverable DWM composition exceptions gracefully

Jack251970:
Updated the `ErrorReporting` class to log recoverable DWM composition exceptions instead of displaying the "Report problem UI," addressing issues with crashes in
`PresentationFramework.dll`. Introduced a new `ExceptionHelper` class to encapsulate logic for identifying recoverable DWM composition exceptions based on specific HRESULT values and stack trace patterns. Added detailed comments for context and referenced a related GitHub issue. Improved modularity and maintainability by isolating exception handling logic in the `ExceptionHelper` class.

From: https://github.com/microsoft/PowerToys/pull/42588.
Resolve: #4016

### DIFF
--- a/Flow.Launcher/Helper/ExceptionHelper.cs
+++ b/Flow.Launcher/Helper/ExceptionHelper.cs
@@ -1,6 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation
-// The Microsoft Corporation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
+﻿// This is a direct copy of the file at https://github.com/microsoft/PowerToys/blob/main/src/modules/launcher/PowerLauncher/Helper/ExceptionHelper.cs and adapted for flow. 
 
 using System;
 using System.Runtime.InteropServices;


Flow.Launcher/Helper/ErrorReporting.cs
```diff
@@ -16,6 +16,14 @@ private static void Report(Exception e, bool silent = false, [CallerMemberName]
         var logger = LogManager.GetLogger(methodName);
         logger.Fatal(ExceptionFormatter.FormatExcpetion(e));
         if (silent) return;
+
+        // The crash occurs in PresentationFramework.dll, not necessarily when the Runner UI is visible, originating from this line:
+        // https://github.com/dotnet/wpf/blob/3439f20fb8c685af6d9247e8fd2978cac42e74ac/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Shell/WindowChromeWorker.cs#L1005
+        // Many bug reports because users see the "Report problem UI" after "the" crash with System.Runtime.InteropServices.COMException 0xD0000701 or 0x80263001.
+        // However, displaying this "Report problem UI" during WPF crashes, especially when DWM composition is changing, is not ideal; some users reported it hangs for up to a minute before the "Report problem UI" appears.
+        // This change modifies the behavior to log the exception instead of showing the "Report problem UI".
```
jjw24:
```suggestion
        // Workaround for issue https://github.com/Flow-Launcher/Flow.Launcher/issues/4016
        // The crash occurs in PresentationFramework.dll, not necessarily when the Runner UI is visible, originating from this line:
        // https://github.com/dotnet/wpf/blob/3439f20fb8c685af6d9247e8fd2978cac42e74ac/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Shell/WindowChromeWorker.cs#L1005
        // Many bug reports because users see the "Error report UI" after the crash with System.Runtime.InteropServices.COMException 0xD0000701 or 0x80263001.
        // However, displaying this "Error report UI" during WPF crashes, especially when DWM composition is changing, is not ideal; some users reported it hangs for up to a minute before the it appears.
        // This change modifies the behavior to log the exception instead of showing the "Error report UI".
```

Flow.Launcher/Helper/ExceptionHelper.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation
+// The Microsoft Corporation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
jjw24:
```suggestion
﻿// This is a direct copy of the file at https://github.com/microsoft/PowerToys/blob/main/src/modules/launcher/PowerLauncher/Helper/ExceptionHelper.cs and adapted for flow. 
```

jjw24 (approved):

PR #4131, opened 2025-11-26
Check debug level before logging debug & Expose LOGLEVEL API for performance optimization

Jack251970:
Enhanced logging with conditional debug checks based on log level, ensuring logs are only generated in DEBUG mode. Integrated `CommunityToolkit.Mvvm.DependencyInjection` for dependency injection, improving modularity and testability.

- Replaced direct settings access with `FlowSettings` via DI.
- Updated logging in `PluginManager.cs`, `PluginsLoader.cs`, `MainViewModel.cs`, and `ResultsViewModel.cs`.
- Improved plugin initialization error handling with conditional logging.
- Refactored query and result logging for better performance.
- Added necessary namespace imports for DI and logging.
- Cleaned up redundant logging and improved code readability.

### DIFF
--- a/Flow.Launcher.Infrastructure/Logger/Log.cs
+++ b/Flow.Launcher.Infrastructure/Logger/Log.cs
@@ -155,6 +155,9 @@ namespace Flow.Launcher.Infrastructure.Logger
 
         public static void Debug(string className, string message, [CallerMemberName] string methodName = "")
         {
+            // We check log level for Debug to optimize performance as Debug is the most verbose level
+            if (LogLevel != LOGLEVEL.DEBUG) return;
+
             LogInternal(NLog.LogLevel.Debug, className, message, methodName);
         }
 


Flow.Launcher.Plugin/Interfaces/IPublicAPI.cs
```diff
@@ -637,5 +637,11 @@ public interface IPublicAPI
         /// </summary>
         /// <returns></returns>
         string GetLogDirectory();
+
+        /// <summary>
+        /// Get the current log level of Flow Launcher.
```
jjw24:
```suggestion
        /// Get the current log level for Flow Launcher.
```

Flow.Launcher.Infrastructure/Http/Http.cs
```diff
@@ -158,7 +160,8 @@ public static Task<string> GetAsync([NotNull] string url, CancellationToken toke
         /// <returns>The Http result as string. Null if cancellation requested</returns>
         public static async Task<string> GetAsync([NotNull] Uri url, CancellationToken token = default)
         {
-            Log.Debug(ClassName, $"Url <{url}>");
+            if (Log.LogLevel == LOGLEVEL.DEBUG)
```
jjw24, line 163: I think instead of putting these if statements all over, should just embed it in the Log.Debug call, so when debug is only logged if the current level is debug.

Jack251970: > I think instead of putting these if statements all over, should just embed it in the Log.Debug call, so when debug is only logged if the current level is debug.

Since we add new API function GetLogLevel, I think we should also keep the same style in other places.

jjw24: What I mean is it's unnecessary to require check log level before logging debug, it is explicit enough that when log debug, it will only log when the log level is set to it. Shouldn't have the need to check log level first, otherwise it might as well just be any output method with a check log level to only output if log is debug.

jjw24: This looks like already the case as we are passing the log level to the logger, so the log level check is redundant. Can you test please:
https://github.com/Flow-Launcher/Flow.Launcher/blob/a49eb464c8af1b286c7a22d2502093a324657f91/Flow.Launcher.Infrastructure/Logger/Log.cs#L141-L148

Jack251970: > This looks like already the case as we are passing the log level to the logger, so the log level check is redundant. Can you test please:
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/a49eb464c8af1b286c7a22d2502093a324657f91/Flow.Launcher.Infrastructure/Logger/Log.cs#L141-L148

Understood. We can check the log level in Debug function but I think that is not enough.

Take `App.API.LogDebug(ClassName, $"Start query with ActionKeyword <{query.ActionKeyword}> and TrimmedQuery <{query.TrimmedQuery}>");` for example, if we check log level in Debug function, the strings of `$"Start query with ActionKeyword <{query.ActionKeyword}> and TrimmedQuery <{query.TrimmedQuery}>"` is constructed which can consume plenty of memory since this strings is different during each time. So I think we should check the log level before calling `LogDebug` function

jjw24: Yes we can have the check for those two plus this one `App.API.LogDebug(ClassName, $"Valid <{plugins.Count}> plugins: {string.Join(" ", plugins.Select(x => $"<{x.Metadata.Name}>"))}");`, the others are simple string constructions.

Jack251970: > Yes we can have the check for those two plus this one `App.API.LogDebug(ClassName, $"Valid <{plugins.Count}> plugins: {string.Join(" ", plugins.Select(x => $"<{x.Metadata.Name}>"))}");`, the others are simple string constructions.

Actually, all checks are for strings with interpolation.

Jack251970: > This looks like already the case as we are passing the log level to the logger, so the log level check is redundant. Can you test please:
> 
> https://github.com/Flow-Launcher/Flow.Launcher/blob/a49eb464c8af1b286c7a22d2502093a324657f91/Flow.Launcher.Infrastructure/Logger/Log.cs#L141-L148

I think the check in `Debug` function is still necessary since `Debug` is the most verbose level

jjw24: >Yes we can have the check for those two plus this one App.API.LogDebug(ClassName, $"Valid <{plugins.Count}> plugins: {string.Join(" ", plugins.Select(x => $"<{x.Metadata.Name}>"))}");, the others are simple string constructions.

>Actually, all checks are for strings with interpolation.

Those except for the above are simple string interpolations there is little impact/cost to speed and performance though.

Jack251970: > > Yes we can have the check for those two plus this one App.API.LogDebug(ClassName, $"Valid <{plugins.Count}> plugins: {string.Join(" ", plugins.Select(x => $"<{x.Metadata.Name}>"))}");, the others are simple string constructions.
> 
> > Actually, all checks are for strings with interpolation.
> 
> Those except for the above are simple string interpolations there is little impact/cost to speed and performance though.

In that case, let us close this PR

Issue #4139, opened 2025-11-30
BUG: Hide item option does not work


silverwings15:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I am using the latest version of Flow Launcher.

- [ ] I am using the prerelease version of Flow Launcher.

### Problem Description

After hiding an item from the dropdown list, the item remains visible afterwards.

### To Reproduce

1. Open Flow, search for an item, right click on it and press Hide
2. The popup notification appears, saying: "Success! Successfully disabled this program from displaying in your query"
3. The item remains in the query
4. Clear the query and search for the item again afterwards, it still remains visible.

### Flow Launcher Version

2.0.3

### Windows Build Number

10.0.26200.7171

### Error Log

N/A

PR #4141, opened 2025-11-30
Fix hide item option not working in Program plugin

Jack251970:
The original problem is the call of First() method. I fix this with null check.

Additionally, this pull request refactors and improves the program indexing and disabling logic in the `Flow.Launcher.Plugin.Program` plugin. The main changes introduce more robust cache management, better handling of program disabling, and improved asynchrony and error handling.

**Improvements to indexing and cache management:**

* Added a `resetCache` parameter (defaulting to `true`) to both `IndexWin32ProgramsAsync` and `IndexUwpProgramsAsync`, allowing more control over when the cache is reset. The cache is only reset if this parameter is `true`. [[1]](diffhunk://#diff-5ae7212ef856d329f8b9696d8009b568e45fb6ac33d84af7dd38cb2a6936d785L337-R337) [[2]](diffhunk://#diff-5ae7212ef856d329f8b9696d8009b568e45fb6ac33d84af7dd38cb2a6936d785R348-R351) [[3]](diffhunk://#diff-5ae7212ef856d329f8b9696d8009b568e45fb6ac33d84af7dd38cb2a6936d785L365-R368) [[4]](diffhunk://#diff-5ae7212ef856d329f8b9696d8009b568e45fb6ac33d84af7dd38cb2a6936d785R379-R382)
* Updated all places where program indexing is triggered (including directory and package change watchers) to explicitly call the indexing methods with `resetCache: true` for correct cache invalidation. [[1]](diffhunk://#diff-5ae7212ef856d329f8b9696d8009b568e45fb6ac33d84af7dd38cb2a6936d785L397-R408) [[2]](diffhunk://#diff-4ffcaa43a3ea3e64cbd4a1251a29b720a7c5c257727ed529797017e380ae7f16L320-R320) [[3]](diffhunk://#diff-deda9860fe7701be4c8e3c06744f5a5966459e8eb58b78a851317dd99e5acb9bL799-R799)

**Improvements to disabling programs:**

* Refactored `DisableProgramAsync` to return a `bool` indicating whether a program was actually disabled, and to only trigger reindexing for the relevant program type (UWP or Win32).
* Updated the context menu action for disabling a program to handle errors gracefully, reset the cache only if a program was actually disabled, and run the disabling logic asynchronously to avoid blocking the UI.

**Code quality and safety:**

* Changed the `catalog` field in `UWPPackage.cs` from nullable to non-nullable, reflecting that it is always initialized before use.
* Add null check for First() methods.

These changes improve the efficiency, reliability, and user experience of program management in the plugin.

Fixes Flow-Launcher/Flow.Launcher#4139

Tested

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Main.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Main.cs
@@ -413,9 +413,21 @@ namespace Flow.Launcher.Plugin.Program
 
         internal static void ResetCache()
         {
-            var oldCache = cache;
-            cache = new MemoryCache(cacheOptions);
-            oldCache.Dispose();
+            var newCache = new MemoryCache(cacheOptions);
+
+            // Atomically swap and get the previous cache instance, avoids double-dispose/lost-assignment race
+            // where each caller receives a distinct prior instance to dispose.
+            var oldCache = Interlocked.Exchange(ref cache, newCache);
+
+            // Dispose the previous instance (if any)- each caller gets a unique prior instance from above
+            try
+            {
+                oldCache?.Dispose();
+            }
+            catch (Exception e)
+            {
+                Context.API.LogException(ClassName, "Failed to dispose old program cache", e);
+            }
         }
 
         public Control CreateSettingPanel()


Plugins/Flow.Launcher.Plugin.Program/Programs/UWPPackage.cs
```diff
@@ -317,7 +317,7 @@ public static async Task WatchPackageChangeAsync()
                 {
                     await Task.Delay(3000).ConfigureAwait(false);
                     PackageChangeChannel.Reader.TryRead(out _);
-                    await Task.Run(Main.IndexUwpProgramsAsync);
+                    await Main.IndexUwpProgramsAsync(true).ConfigureAwait(false);
```
jjw24: @Jack251970 let's in the future with parameters that are not clear like bools, add the actual parameter name in so we know what we are setting true/false for.

Could you please make the changes to this and the win32 method calls too please.

jjw24:
@Jack251970 could you also include in the description what is the actual problem causing this issue.

Plugins/Flow.Launcher.Plugin.Program/Programs/UWPPackage.cs
Jack251970: > @Jack251970 let's in the future with parameters that are not clear like bools, add the actual parameter name in so we know what we are setting true/false for.
> 
> Could you please make the changes to this and the win32 method calls too please.

I think it is clear enough? resetCache

Jack251970:
> @Jack251970 could you also include in the description what is the actual problem causing this issue.

The root problem is the call of First() method. I fix this with null check

Jack251970:
> @Jack251970 could you also include in the description what is the actual problem causing this issue.

Already added

Plugins/Flow.Launcher.Plugin.Program/Programs/UWPPackage.cs
jjw24: I mean like this:
`await Main.IndexUwpProgramsAsync(reset cache: true).ConfigureAwait(false);`

Jack251970: > .IndexUwpProgramsAsync(

Added. Please review.

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -394,12 +400,12 @@ public static async Task IndexProgramsAsync()
         {
             var win32Task = Task.Run(async () =>
             {
-                await Context.API.StopwatchLogInfoAsync(ClassName, "Win32Program index cost", IndexWin32ProgramsAsync);
+                await Context.API.StopwatchLogInfoAsync(ClassName, "Win32Program index cost", () => IndexWin32ProgramsAsync(true));
```
jjw24:
```suggestion
                await Context.API.StopwatchLogInfoAsync(ClassName, "Win32Program index cost", () => IndexWin32ProgramsAsync(resetCache: true));
```

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -394,12 +400,12 @@ public static async Task IndexProgramsAsync()
...
             });
 
             var uwpTask = Task.Run(async () =>
             {
-                await Context.API.StopwatchLogInfoAsync(ClassName, "UWPProgram index cost", IndexUwpProgramsAsync);
+                await Context.API.StopwatchLogInfoAsync(ClassName, "UWPProgram index cost", () => IndexUwpProgramsAsync(true));
```
jjw24:
```suggestion
                await Context.API.StopwatchLogInfoAsync(ClassName, "UWPProgram index cost", () => IndexUwpProgramsAsync(resetCache: true));
```

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -458,52 +478,50 @@ public List<Result> LoadContextMenus(Result selectedResult)
...
-                var program = _uwps.First(x => x.UniqueIdentifier == programToDelete.UniqueIdentifier);
-                program.Enabled = false;
-                _settings.DisabledProgramSources.Add(new ProgramSource(program));
+                var program = _uwps.FirstOrDefault(x => x.UniqueIdentifier == programToDelete.UniqueIdentifier);
+                if (program != null)
+                {
+                    program.Enabled = false;
+                    _settings.DisabledProgramSources.Add(new ProgramSource(program));
+                    // Reindex UWP programs
+                    _ = Task.Run(() => IndexUwpProgramsAsync(false));
```
jjw24:
```suggestion
                    _ = Task.Run(() => IndexUwpProgramsAsync(resetCache: false));
```

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -458,52 +478,50 @@ public List<Result> LoadContextMenus(Result selectedResult)
...
-                var program = _win32s.First(x => x.UniqueIdentifier == programToDelete.UniqueIdentifier);
-                program.Enabled = false;
-                _settings.DisabledProgramSources.Add(new ProgramSource(program));
+                var program = _win32s.FirstOrDefault(x => x.UniqueIdentifier == programToDelete.UniqueIdentifier);
+                if (program != null)
+                {
+                    program.Enabled = false;
+                    _settings.DisabledProgramSources.Add(new ProgramSource(program));
+                    // Reindex Win32 programs
+                    _ = Task.Run(() => IndexWin32ProgramsAsync(false));
```
jjw24:
```suggestion
                    _ = Task.Run(() => IndexWin32ProgramsAsync(resetCache: false));
```

Plugins/Flow.Launcher.Plugin.Program/Main.cs
```diff
@@ -442,12 +448,26 @@ public List<Result> LoadContextMenus(Result selectedResult)
...
-                                "flowlauncher_plugin_program_disable_dlgtitle_success_message"));
-                        Context.API.ReQuery();
+                        _ = Task.Run(async () =>
+                        {
+                            try
+                            {
+                                var disabled = await DisableProgramAsync(program);
+                                if (disabled)
+                                {
+                                    ResetCache();
```
jjw24, line 470: How does the ResetCache call work when we are resetting multiple times ? For example user disables one program after another.

Jack251970: > How does the ResetCache call work when we are resetting multiple times ? For example user disables one program after another.

No idea about what you said. Every time one program is disabled, it will invoke a task to add the program in disable list and reindex.

VictoriousRaptor: > How does the ResetCache call work when we are resetting multiple times ? For example user disables one program after another.

It will reset the cache every time you disable a program.

The one elegant way is to only remove the cache entry of the disabled one. But yeah calling resetting cache one by one is acceptable imo.

jjw24: What I am mean is you have IndexProgramsAsync making two concurrent calls to ResetCache() via IndexWin32ProgramsAsync and IndexUwpProgramsAsync. This could lead to:
- Double disposal
- Final cache ends up as the last writer and the other cache is lost and leaked or disposed according to timing.
- Disposing the old MemoryCache while another is concurrently reading/writing to it can cause error behavior.

jjw24 (approved):

Issue #3984, opened 2025-09-16
Set as topmost automatically when selecting a result


tomasklaen:
When I select a result for a query, that result should be set as topmost for that query automatically. All other launchers I've used until now behaved that way. I don't want to do this via context menu every time :(

BTW, it took me quite a while to find out this feature even exists, and was about to uninstall Flow because of it missing.

Also, why does setting an item as topmost trigger a system level notification? Seems highly unnecessary, and kinda triggers my OCD of not wanting to have unread notifications so now I have to go there and clean them. Very annoying.

Adding an option to make this automatic would solve both problems:
- it not being automatic
- giving people a chance to discover it by stumbling upon an option for it

But imho it should be automatic by default.

01Dri:
I think it's a great feature, and I'm currently looking into how to implement it.

01Dri:
About the notifications, I'm not sure if we should remove them or not. @Jack251970 what do you think?

Jack251970:
> About the notifications, I'm not sure if we should remove them or not. @Jack251970 what do you think?

I would prefer to keep the notification. But if automatical mode is on, let us remove the notification since every action will trigger it

PR #4159, opened 2025-12-14
[FEATURE] Topmost automatically 

01Dri:
Issue: https://github.com/Flow-Launcher/Flow.Launcher/issues/3984

Example: 

https://github.com/user-attachments/assets/47a48aff-6f70-4982-980b-a111b298eeba


Flow.Launcher/Languages/en.xaml
```diff
@@ -177,6 +177,8 @@
     <system:String x:Key="homeToggleBoxToolTip">This can only be edited if plugin supports Home feature and Home Page is enabled.</system:String>
     <system:String x:Key="showAtTopmost">Show Search Window at Foremost</system:String>
     <system:String x:Key="showAtTopmostToolTip">Overrides other programs' 'Always on Top' setting and displays Flow in the foremost position.</system:String>
+    <system:String x:Key="autoTopmostOnOpen">Auto Topmost on Open</system:String>
+    <system:String x:Key="autoTopmostOnOpenToolTip">Set Flow Launcher to be topmost automatically when it is opened.</system:String>
```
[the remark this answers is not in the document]
VictoriousRaptor: Yeah please consider code rabbit's suggestion. I was staring at Flow window the whole time when watching your video.

Flow.Launcher.Infrastructure/UserSettings/Settings.cs
```diff
@@ -496,6 +496,21 @@ public bool ShowAtTopmost
             }
         }
 
+        private bool _autoTopmostResults = false;
+        public bool AutoTopmostResults
```
VictoriousRaptor:
```suggestion
        public bool AutoTopmostLastOpenedResult
```

What about it?

01Dri: Good!

Jack251970 (dismissed):

jjw24 (commented):
This leads to the MultipleTopMostRecord.json file growing indefinitely with unique queries when the option is on.

If the goal is to allow faster access to results by pushing frequently selected results to the top, I think the ideal change is to rebalance the 'selected record count' bump. The mechanism is there to allow frequently selected results to show up higher and seems like it is not as effective. We can rebalance it to have results show up better like this:

Image

01Dri:
> This leads to the MultipleTopMostRecord.json file growing indefinitely with unique queries when the option is on.
> 
> If the goal is to allow faster access to results by pushing frequently selected results to the top, I think the ideal change is to rebalance the 'selected record count' bump. The mechanism is there to allow frequently selected results to show up higher and seems like it is not as effective. We can rebalance it to have results show up better like this:
> 
> Image [ ](https://private-user-images.githubusercontent.com/26427004/530153435-7978849c-aa19-4ca6-9a93-5dc7a9403d02.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjY3MTgxMzAsIm5iZiI6MTc2NjcxNzgzMCwicGF0aCI6Ii8yNjQyNzAwNC81MzAxNTM0MzUtNzk3ODg0OWMtYWExOS00Y2E2LTlhOTMtNWRjN2E5NDAzZDAyLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTEyMjYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUxMjI2VDAyNTcxMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWIyZmExYWIxYjQzN2Y4NDY1OWM4NmQ0YTc5ZTMzZTcyNWIzOWRjYjA0MmExN2Q0MDBmY2ZkZDBjZDkyZGNjOTImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.JroL-XkS6T_9TViDFDBxNw9y0IAZAE5fgVyBCnAHXfk)

Hmm, it works, but changing the current result balance may affect other users who are already satisfied with the existing behavior.

What do you think about using the new option I added in my PR? When enabled, it would prioritize the last selected result by placing it at the top of the ranking, rebalancing the selected result with a higher score.

The new option could be something like "Prioritize the last query as the top result". When this option is enabled, the query score could be set to 100, guaranteeing priority in the results. Otherwise, we can keep using the default value (5).

jjw24:
@01Dri thank you for considering this option as well. I think currently I have seen some issues and comments that the current bump is not significant or obvious enough, users have had to select the result several times before they are satisfied with the placement. So I think the rebalancing is not going to be too negative on user experience.

What I propose, since this option is significantly less effort, how about we bump it appropriately and once we get some feedback then we can consider adding the toggle option if it is negatively impacting the users. What do you guys think?

jjw24:
One thing I would recommend is also we need to be mindful of introducing new features unless they are highly requested, because the more we add the more complex flow becomes to maintain. This is not to say this feature is not good, just that we can solve the root cause with something significantly simpler, it would be better for the long term.

01Dri:
> @01Dri thank you for considering this option as well. I think currently I have seen some issues and comments that the current bump is not significant or obvious enough, users have had to select the result several times before they are satisfied with the placement. So I think the rebalancing is not going to be too negative on user experience.
> 
> What I propose, since this option is significantly less effort, how about we bump it appropriately and once we get some feedback then we can consider adding the toggle option if it is negatively impacting the users. What do you guys think?

Ok for me.

My friend will take care of this feature because I’m traveling at the moment.

4yinn:
> > @01Dri thank you for considering this option as well. I think currently I have seen some issues and comments that the current bump is not significant or obvious enough, users have had to select the result several times before they are satisfied with the placement. So I think the rebalancing is not going to be too negative on user experience.
> > What I propose, since this option is significantly less effort, how about we bump it appropriately and once we get some feedback then we can consider adding the toggle option if it is negatively impacting the users. What do you guys think?
> 
> Ok for me.
> 
> My friend will take care of this feature because I’m traveling at the moment.

Sorry for the delay, I’ll take care of it.

4yinn:
> This leads to the MultipleTopMostRecord.json file growing indefinitely with unique queries when the option is on.
> 
> If the goal is to allow faster access to results by pushing frequently selected results to the top, I think the ideal change is to rebalance the 'selected record count' bump. The mechanism is there to allow frequently selected results to show up higher and seems like it is not as effective. We can rebalance it to have results show up better like this:
> 
> Image [ ](https://private-user-images.githubusercontent.com/26427004/530153435-7978849c-aa19-4ca6-9a93-5dc7a9403d02.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Njc3NDYzMzEsIm5iZiI6MTc2Nzc0NjAzMSwicGF0aCI6Ii8yNjQyNzAwNC81MzAxNTM0MzUtNzk3ODg0OWMtYWExOS00Y2E2LTlhOTMtNWRjN2E5NDAzZDAyLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAxMDclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMTA3VDAwMzM1MVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTM1YjE2M2NiYjM3MGNjMWI3NWI4NTU3OTUzNjYxMmU1MWFlNzE4MjgzOTkwY2Y3N2IyNDFlZDU1OWI5NDgwMzImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.7I67v7m6uOvK6bJEvGKpbh-X3doLO19CLuqKNwvhEFQ) Image [ ](https://private-user-images.githubusercontent.com/26427004/530153435-7978849c-aa19-4ca6-9a93-5dc7a9403d02.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Njc3NDYzMzEsIm5iZiI6MTc2Nzc0NjAzMSwicGF0aCI6Ii8yNjQyNzAwNC81MzAxNTM0MzUtNzk3ODg0OWMtYWExOS00Y2E2LTlhOTMtNWRjN2E5NDAzZDAyLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAxMDclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMTA3VDAwMzM1MVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTM1YjE2M2NiYjM3MGNjMWI3NWI4NTU3OTUzNjYxMmU1MWFlNzE4MjgzOTkwY2Y3N2IyNDFlZDU1OWI5NDgwMzImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.7I67v7m6uOvK6bJEvGKpbh-X3doLO19CLuqKNwvhEFQ)

I’ve reviewed the code, and the actual value doesn’t matter. The logic is that if a user clicks, for example, Notepad four times and then clicks Sublime once, Sublime should not move up because it hasn’t surpassed the user’s favorite item. However, I think a better approach would be to always move an item to the top whenever the user clicks it. 

https://github.com/user-attachments/assets/62498d2b-dadb-4da8-b490-8c87db14d9dd

jjw24:
Hi there, how high it will place the result will also depend on the result score constructed from matching against the query user typed. For example Sublime may not score as high as as Limelight if user searched 'lim'.

I don't think having the result placed at the top from one selection is feasible because, for example if I typed 'lime' and selected Sublime the first time, I might not want it again the second time I type it. I might choose Limelight instead the second time, so having Sublime jump immediately above Limelight means I need to press down arrow then enter instead of just enter to select it. However I think it is reasonable that if I selected Sublime the second time as well then it should be placed at the top, or maybe at the top the third time I select it.

I will draft up a new PR this weekend with an adjusted 'selected record count' bump and we can review, test and discuss there.

PR #4167, opened 2025-12-17
Fix: Result.Preview.Description Does not Display in Preview Panel

Heck-R:
According to the `Result.Preview.Description` ([`PreviewInfo.Description`](https://github.com/Flow-Launcher/Flow.Launcher/blob/20930617bb714638cbdfdfae57c4d16e66906176/Flow.Launcher.Plugin/Result.cs#L350)) documentation
```
/// <summary>
/// Result description text that is shown at the bottom of the preview panel.
/// </summary>
/// <remarks>
/// When a value is not set, the <see cref="SubTitle"/> will be used.
/// </remarks>
```

However, it actually doesn't do anything.
This PR fixes this, so the `Description` does what it is documented to do.
I left the related `TextBlock` control name as `PreviewSubTitle`, because although this is confusing, changing it could break user themes.

### DIFF
--- a/Flow.Launcher/MainWindow.xaml
+++ b/Flow.Launcher/MainWindow.xaml
@@ -526,7 +526,7 @@
                                     <TextBlock
                                         x:Name="PreviewSubTitle"
                                         Style="{DynamicResource PreviewItemSubTitleStyle}"
-                                        Text="{Binding DefaultPreviewCustomDescription}" />
+                                        Text="{Binding PreviewDescription}" />
                                 </StackPanel>
                             </Grid>
                         </Border>
--- a/Flow.Launcher/ViewModel/ResultViewModel.cs
+++ b/Flow.Launcher/ViewModel/ResultViewModel.cs
@@ -288,7 +288,7 @@ namespace Flow.Launcher.ViewModel
             }
         }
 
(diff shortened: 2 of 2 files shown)


VictoriousRaptor:
So it seems like when description is not set subtitle is not used. Did you test it out?

Heck-R:
If the description is not set, then the subtitle **is** used (as it was before, like the Description's documentation suggests).

Yes, I tested it. Here is a quick dummy example

With the Description:
image
image

Without the description (same with or without the PreviewInfo)
image
image
image

Jack251970:
> So it seems like when description is not set subtitle is not used. Did you test it out?

Could you please help me test it? It works well for me and I think it is ready to merge

Jack251970:
> So it seems like when description is not set subtitle is not used. Did you test it out?

Tested for uwp package preview panel:

After:
image

Before:
image

VictoriousRaptor:
> > So it seems like when description is not set subtitle is not used. Did you test it out?
> 
> Could you please help me test it? It works well for me and I think it is ready to merge

I was busy when I posted this comment and didn't notice how `Subtitle` and `Description` are connected. LGTM.

PR #4170, opened 2025-12-21
Replace old clipboard plugin with a new one in README

VictoriousRaptor:
Replace the old clipboard plugin in our readme with jack's new one.


VictoriousRaptor:
> Is there a specific reason for switching from Clipboard History to Clipboard+?

not

> Is there a specific reason for switching from Clipboard History to Clipboard+?

Hi, it's mainly because Clipboard History has not been updated for a long time (but doesn't mean it's not functioning) so we choose to feature the new clipboard+ on our readme.

PR #4196, opened 2026-01-02
Use Highlight matches from Everything 

VictoriousRaptor:
To solve inaccurate highlight in some cases.
Clip_20260322_162229

Our FuzzySearch() doesn't match the same highlight as Everything so sometimes the highlight looks weird.


Jack251970 (commented):
LGTM! Need jjw24 to do tests with Everything

jjw24:
@VictoriousRaptor can you show me a screenshot of the inaccurate highlighting please.
So highlighting is only inaccurate with Pinyin searches? Should we only use the conversion code when searching in Pinyin?

VictoriousRaptor:
> @VictoriousRaptor can you show me a screenshot of the inaccurate highlighting please. So highlighting is only inaccurate with Pinyin searches? Should we only use the conversion code when searching in Pinyin?

I can't give a screenshot now because I don't remember a specific query that has in accurate highlight. I'll post it if I have one. Sometimes it happens when I'm using double pinyin or with pinyin acronym. Not sure if it's only with pinyin, most of my english named documents works good. 

The last question should be discussed.

VictoriousRaptor:
> @VictoriousRaptor can you show me a screenshot of the inaccurate highlighting please. So highlighting is only inaccurate with Pinyin searches? Should we only use the conversion code when searching in Pinyin?

image
image

This is an example of highlighting.

VictoriousRaptor:
Clip_20260321_132125

Here's an incorrect highlight in English.

Jack251970 (changes requested):
Resolve the build issue please

Plugins/Flow.Launcher.Plugin.Explorer/Search/ResultManager.cs
```diff
@@ -92,15 +93,15 @@ internal static void ShowNativeContextMenu(string path, ResultType type)
...
+        internal static Result CreateFolderResult(string title, string subtitle, string path, Query query, int score = 0, bool windowsIndexed = false, List<int> highlightData = null)
         {
             return new Result
             {
                 Title = title,
                 IcoPath = path,
                 SubTitle = subtitle,
                 AutoCompleteText = GetAutoCompleteText(title, query, path, ResultType.Folder),
-                TitleHighlightData = Context.API.FuzzySearch(query.Search, title).MatchData,
+                TitleHighlightData = highlightData ?? Context.API.FuzzySearch(query.Search, title).MatchData,
```
jjw24, line 148: Is it a simple change to have the other search engine pass highlightData instead of calling it here? I think when FuzzySearch is used it will also contain the MatchData which, you can the assign to the SearchResult object.

VictoriousRaptor: > Is it a simple change to have the other search engine pass highlightData instead of calling it here? I think when FuzzySearch is used it will also contain the MatchData which, you can the assign to the SearchResult object.

But sadly windows index doesn't provide this feature so it has to fall back to FuzzySearch.

jjw24: Yeah but is it an easy change to have Windows Indexer to pass the highlight data the same way Everything is now doing?

VictoriousRaptor: > Yeah but is it an easy change to have Windows Indexer to pass the highlight data the same way Everything is now doing?

CreateFileResult() and CreateFolderResult() is not only invoked by indexers and is called in about 10 different locations. I think it changes too much to pass the highlight data by default.

jjw24: Yeah ok let's not worry about it.

Jack251970 (approved):

PR #4294, opened 2026-02-26
Fix Explorer plugin Quick Access items missing from general search results

### LINKED ISSUE #4287: BUG: quick access not on the first of the list
### Checks

### Problem Description

The latest version (2.1.0) doesn't put quick access items on the first one in the list. It was working fine in version 2.0.3.

### To Reproduce

1. Add items to quick access, in my case it's a .lnk file, file name starting with h and s
2. Query s
3. Items not found

### Screenshots

Version 2.1.0:
Version 2.1.0

Version 2.0.3:
Version 2.0.3

### Flow Launcher Version

2.1.0

### Windows Build Number

10.0.19044.6937

### Error Log

<details>

```shell

```
</details>

Jack251970:
Follow on with #4093

Resolve #4287

---
## Summary by cubic
Fixes a 2.1.0 regression where Quick Access items were missing from Explorer plugin general search. Matching Quick Access links are now always merged into results, regardless of the action keyword. Closes #4287.

- **Bug Fixes**
  - Removed the incorrect QuickAccessKeywordEnabled check in GetQuickAccessResultsFilteredByActionKeyword.
  - The action keyword now only controls Quick Access–only searches; general searches include matching Quick Access items.

<sup>Written for commit ae73b481d53f52dae6d98d05b647357ad4bce6c6. Summary will update on new commits.</sup>

### DIFF
--- a/Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
+++ b/Flow.Launcher.Plugin/Flow.Launcher.Plugin.csproj
@@ -15,10 +15,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>5.1.0</Version>
-    <PackageVersion>5.1.0</PackageVersion>
-    <AssemblyVersion>5.1.0</AssemblyVersion>
-    <FileVersion>5.1.0</FileVersion>
+    <Version>5.2.0</Version>
+    <PackageVersion>5.2.0</PackageVersion>
+    <AssemblyVersion>5.2.0</AssemblyVersion>
+    <FileVersion>5.2.0</FileVersion>
     <PackageId>Flow.Launcher.Plugin</PackageId>
     <Authors>Flow-Launcher</Authors>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
--- a/Flow.Launcher/Storage/QueryHistory.cs
+++ b/Flow.Launcher/Storage/QueryHistory.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json.Serialization;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: '2.0.3.{build}'
+version: '2.1.0.{build}'
 
 # Do not build on tags because we create a release on merge to master. Otherwise will upload artifacts twice changing the hash, as well as triggering duplicate GitHub release action & NuGet deployments.
 skip_tags: true


Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -299,9 +299,6 @@ private bool IsExcludedFile(SearchResult result)
 
         private List<Result> GetQuickAccessResultsFilteredByActionKeyword(Query query, List<ActionKeyword> actions)
         {
```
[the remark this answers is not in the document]
Jack251970: No need

Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs
```diff
@@ -299,9 +299,6 @@ private bool IsExcludedFile(SearchResult result)
...
-            if (!Settings.QuickAccessKeywordEnabled)
-                return [];
-
```
jjw24, line 304: If Quick Access Keyword is not enabled, why allow the results to come through?

Jack251970: > If Quick Access Keyword is not enabled, why allow the results to come through?

Check the original logic please.

Image

VictoriousRaptor: I did some search and I think it's related to #4053. #4053 introduced the idea that Quick Access are removed from global search. 

The change in this PR should be accepted IMO but I think Quick Access should be removed from global search.

jjw24: Ok let me test drive this part and figure out the behaviour.

Jack251970: > I did some search and I think it's related to #4053. #4053 introduced the idea that Quick Access are removed from global search. 
> 
> The change in this PR should be accepted IMO but I think Quick Access should be removed from global search.

I do not agree. I prefer to maintain the original logic.

Jack251970: > Well what should be the correct behaviour? I feel the action keyword to trigger quick access should behave the same as other action keywords in Explorer- triggered when enabled, if non-global then should trigger only with the dedicated keyword, if global then triggered with global results, and disabled if set to disabled. This is the behaviour standardised for the other action keywords e.g. files, folder or path search. This makes it easier to maintain.

So could you please tell me how to edit it? Could you please do me a favor?

jjw24: You are right, this change is the correct behaviour. Quick Access results should still be added to files/folder/general search even if the action keyword disabled. Disabling action keyword simply just doesnt trigger the Quick Access list.

Also the behaviour is outlined here: https://github.com/Flow-Launcher/Flow.Launcher/blob/dda900041acb1a7632daa68544f5c9e20b09a18d/Plugins/Flow.Launcher.Plugin.Explorer/Search/SearchManager.cs#L56-L78
My mistake.

jjw24 (approved):

PR #4353, opened 2026-03-13
Handle null plugin manifests with null-coalescing operator

Jack251970:
Updated usages of GetPluginManifest() to use the null-coalescing operator (?? []) for safer handling of potential null values since if Flow cannot get manifest data, this could be null.

```
/// <summary>
/// Get the plugin manifest.
/// </summary>
/// <remarks>
/// If Flow cannot get manifest data, this could be null
/// </remarks>
/// <returns></returns>
public IReadOnlyList<UserPlugin> GetPluginManifest();
```

---
## Summary by cubic
Safely handle null plugin manifests by using (GetPluginManifest() ?? []) across update/install and store flows. Prevents null reference crashes and shows empty lists when the manifest is missing.

**Summary of changes**
- Changed: Replaced direct GetPluginManifest() calls with (GetPluginManifest() ?? []) in joins/queries in PluginInstaller, SettingsPanePluginStoreViewModel, and PluginsManager.
- Added: Fallback that treats a missing manifest as “no available plugins,” keeping update/install and store UI stable.
- Removed: The assumption that the manifest is always non-null; no features or code paths removed.
- Memory: Negligible overhead from creating an empty collection; no sustained allocations.
- Security: No new risks; safer null handling reduces crash surface only.
- Tests: No new unit tests; existing update/install paths exercise this flow.

<sup>Written for commit 86d7c1e06f742afda12e83808b25b104e9b2214c. Summary will update on new commits.</sup>

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
+++ b/Plugins/Flow.Launcher.Plugin.BrowserBookmark/Flow.Launcher.Plugin.BrowserBookmark.csproj
@@ -109,7 +109,7 @@
   <ItemGroup>
     <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.0" />
     <PackageReference Include="Flow.Launcher.Localization" Version="0.0.6" />
-    <PackageReference Include="Microsoft.Data.Sqlite" Version="10.0.4" />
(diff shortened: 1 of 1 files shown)


VictoriousRaptor:
Why not making GetPluginManifest() not null?

Jack251970:
> Why not making GetPluginManifest() not null? 

Hmm. Yeah. Will change from nullable to non-nullable. It seems that this change will not cause issues in compability.

PR #4360, opened 2026-03-15
feat(search): add unicode character removal for fuzzy matching

### LINKED ISSUE #4149: [Describe Your Feature] Allow for searching without diacritics match (e.g., “camera” = “câmera”)
Windows Start Menu search uses a diacritics agnostic approach to searches. For example, "câmera" and "camera" are both acceptable in portuguese, and the Windows Start menu will show the same results for both, while Flow Launcher does not. This is a common approach for ease of use, as adding accents requires more buttons presses, and I imagine it to be common in other languages.

Windows:
Image
Image

Flow Launcher:
Image
Image

Allow us to toggle diacritics matching in searches, just as the Everything app allows for. That way, a query for "camera" should also include apps named "câmera" or "cámara", if that's what the user wants.

### LINKED ISSUE #4387: BUG: Turkish Localization Search Issue: Inconsistent handling of 'i' vs 'ı' (Case-folding/Normalization)
### Checks

### Problem Description

When using Flow Launcher on a Windows OS with Turkish language/region settings, there is a significant search inconsistency regarding the letters "i" and "ı".

In the Turkish alphabet, the lowercase of "I" is "ı" (dotless), and the uppercase of "i" (dotted) is "İ". Because of this, when searching for an application like "Internet Download Manager (IDM)" or "IntelliJ":

    Searching with "idm" (with lowercase 'i') yields no results.

    Searching with "ıdm" (with lowercase 'ı') works correctly.

This indicates that the search engine is using locale-sensitive string comparison instead of a culture-invariant or normalized comparison. Users should be able to find their apps using the standard English/Latin "i" regardless of their system language.

### To Reproduce

1. Use a Windows OS with Turkish language/region settings.
2. Install an application with a name starting with "I" (e.g., Internet Download Manager or IntelliJ).
3. Open Flow Launcher and type the search query using the standard Latin "i" (e.g., "idm" or "intellij").
4. Observe: The application does not appear in the results.
5. Now, type the same query using the Turkish-specific "ı" (e.g., "ıdm" or "ıntellij").
6. Observe: The application appears correctly.

### Screenshots

Image

Image

Image

Image

### Flow Launcher Version

2.1.1

### Windows Build Number

Microsoft Windows [Version 10.0.26200.8037]

4yinn:
### Description

Added a preprocessing step to `FuzzySearch` to normalize Unicode characters and remove diacritics before matching.

This allows searches to be accent-insensitive, improving usability. For example, searching for `camera` will also match `câmera`.

### Related Issue
Closes #4149

---
## Summary by cubic
Make fuzzy search accent-insensitive behind a user toggle. This improves matches across languages (e.g., “camera” matches “câmera”) without changing default behavior.

- **Summary of changes**
  - Changed: `StringMatcher` now normalizes characters on-the-fly when `Settings.IgnoreAccents` is enabled; preserves existing scoring, acronym, and substring matching; refines case handling to avoid unnecessary `.ToLower()` when accents are ignored.
  - Added: `DiacriticsNormalizer` with fast `NormalizeChar`/`Normalize` using a static map, a small lookup table, `stackalloc` (<=512 chars), and `ArrayPool<char>` for longer strings; early-exit path to skip work when no accents are present.
  - Added: `Settings.IgnoreAccents` option, UI toggle, and localized strings; prompts to enable when switching to languages that commonly use diacritics; integrated into Welcome and Language flows.
  - Added: `Settings.StringMatcherBehaviorChanged` event and `IPublicAPI.StringMatcherBehaviorChanged`; `Program` plugin subscribes to reset its cache when matcher behavior changes.
  - Removed: Obsolete `StringMatcher(IAlphabet)` test-only constructor and a redundant precision update path in favor of centralized change notifications.
  - Memory: Minimal steady-state overhead (one small static dictionary + lookup array). Normalization avoids full-string copies and uses pooling; no long-lived allocations added.
  - Security: No new I/O, permissions, or external surface area. Changes are limited to in-process string processing.
  - Tests: Updated existing tests to pass `Settings`; added a test to verify accent-insensitive matching when `IgnoreAccents` is on/off.

<sup>Written for commit b2344402111ca351a0e5eba9f5cdb9e67b8e5e42. Summary will update on new commits.</sup>


Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -67,6 +69,8 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
                 return new MatchResult(false, UserSettingSearchPrecision);
 
             query = query.Trim();
+            query = RemoveAccents(query);
+            stringToCompare = RemoveAccents(stringToCompare);
```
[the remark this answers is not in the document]
Jack251970: @4yinn Do you think we need to map back for correct highlight?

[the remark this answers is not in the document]
Jack251970: @4yinn Do you think we need to handle this?

Jack251970 (changes requested):
Thanks for your contribution! But from AI's review, this removal may change the length of the string which can cause some issues like incorrect highlights.

VictoriousRaptor:
Please don't involve too much code format changes so we can trace changes in the future.

jjw24 (changes requested):
Thank you for your PR.

StringMatcher is performance-critical, so we must minimize additional overhead. Please address the following:

1. Consolidate Loops: The current implementation performs an initial loop to identify and remove Unicode characters, followed by a second loop for the main matching logic. Please merge it into the main loop to avoid redundant iteration.

2. Conditional Unicode Processing: We should make Unicode removal optional—for instance, by adding a toggle for non-English languages. This prevents unnecessary processing when the feature isn't required.

Flow.Launcher.Infrastructure/StringMatcher.cs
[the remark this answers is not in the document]
4yinn: > @4yinn Do you think we need to handle this?

From my perspective as a Flow Launcher user, I think this feature is quite necessary. Many times when I search for programs that contain accents, such as “Câmera”, I tend to type the word without accents, like “camera”, because that’s how I usually type when searching in Windows.

Because of this, I often have difficulty finding applications in Flow when their names contain accents. The same issue was mentioned in the GitHub issue opened by another Brazilian user, which is why I decided to implement this improvement.

Sorry for the late response.

4yinn:
> Thank you for your PR.
> 
> StringMatcher is performance-critical, so we must minimize additional overhead. Please address the following:
> 
> 1. Consolidate Loops: The current implementation performs an initial loop to identify and remove Unicode characters, followed by a second loop for the main matching logic. Please merge it into the main loop to avoid redundant iteration.
> 2. Conditional Unicode Processing: We should make Unicode removal optional—for instance, by adding a toggle for non-English languages. This prevents unnecessary processing when the feature isn't required.

Yes, I can see the problem. I'll think about a way to implement it with better optimizations.

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -218,19 +227,39 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
...
-                var resultList = indexList.Select(x => translationMapping?.MapToOriginalIndex(x) ?? x).Distinct().ToList();
+                var resultList = indexList.Select(x => translationMapping?.MapToOriginalIndex(x) ?? x).Distinct()
+                    .ToList();
                 return new MatchResult(true, UserSettingSearchPrecision, resultList, score);
             }
 
             return new MatchResult(false, UserSettingSearchPrecision);
         }
 
+        private static string RemoveAccents(string value)
```
VictoriousRaptor: If this functions changes the length of `query` we should use `TranslationMapping` to avoid highlight mismatches.

Flow.Launcher.Infrastructure/StringMatcher.cs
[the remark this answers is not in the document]
Jack251970: > > @4yinn Do you think we need to handle this?
> 
> From my perspective as a Flow Launcher user, I think this feature is quite necessary. Many times when I search for programs that contain accents, such as “Câmera”, I tend to type the word without accents, like “camera”, because that’s how I usually type when searching in Windows.
> 
> Because of this, I often have difficulty finding applications in Flow when their names contain accents. The same issue was mentioned in the GitHub issue opened by another Brazilian user, which is why I decided to implement this improvement.
> 
> Sorry for the late response.

I'm sorry, it seems you misunderstood me. What I meant was that from AI's review, this removal may change the length of the string which can cause some issues like incorrect highlights.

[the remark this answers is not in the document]
Jack251970: Duplicated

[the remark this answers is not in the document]
4yinn: > > > @4yinn Do you think we need to handle this?
> > 
> > 
> > From my perspective as a Flow Launcher user, I think this feature is quite necessary. Many times when I search for programs that contain accents, such as “Câmera”, I tend to type the word without accents, like “camera”, because that’s how I usually type when searching in Windows.
> > Because of this, I often have difficulty finding applications in Flow when their names contain accents. The same issue was mentioned in the GitHub issue opened by another Brazilian user, which is why I decided to implement this improvement.
> > Sorry for the late response.
> 
> I'm sorry, it seems you misunderstood me. What I meant was that from AI's review, this removal may change the length of the string which can cause some issues like incorrect highlights.

Sorry for the late reply. hmm ok. Do u think this implementation isn’t necessary, or does it just need further optimization?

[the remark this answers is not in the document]
Jack251970: @4yinn I'm not sure if this Removal will bring about any changes to the string. Since this kind of character does not exist in my language, you need to confirm this issue. If so, improvements are needed here, that is, we need to map the string back and handle the content related to highlighting.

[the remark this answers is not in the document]
4yinn: > @4yinn I'm not sure if this Removal will bring about any changes to the string. Since this kind of character does not exist in my language, you need to confirm this issue. If so, improvements are needed here, that is, we need to map the string back and handle the content related to highlighting.

Yeah, my language is pretty boring when it comes to accents.  Sorry for the late reply again — I was working. I’ll check this now and look for improvements.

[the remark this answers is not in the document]
Jack251970: > > @4yinn I'm not sure if this Removal will bring about any changes to the string. Since this kind of character does not exist in my language, you need to confirm this issue. If so, improvements are needed here, that is, we need to map the string back and handle the content related to highlighting.
> 
> Yeah, my language is pretty boring when it comes to accents. Sorry for the late reply again — I was working. I’ll check this now and look for improvements.

Thanks for your reply!

4yinn:
@jjw24 @Jack251970  

Sorry for the delay — I’ve been thinking about ways to implement this and reached a possible solution, but I got a bit stuck in one situation and wanted to ask for your help if possible.

I implemented this approach to handle the issue:
```c#
private static readonly Dictionary<char, char> AccentMap = new()
{
    ['á'] = 'a', ['à'] = 'a', ['ã'] = 'a', ['â'] = 'a', ['ä'] = 'a', ['å'] = 'a',
    ['é'] = 'e', ['è'] = 'e', ['ê'] = 'e', ['ë'] = 'e',
    ['í'] = 'i', ['ì'] = 'i', ['î'] = 'i', ['ï'] = 'i',
    ['ó'] = 'o', ['ò'] = 'o', ['õ'] = 'o', ['ô'] = 'o', ['ö'] = 'o',
    ['ú'] = 'u', ['ù'] = 'u', ['û'] = 'u', ['ü'] = 'u',
    ['ç'] = 'c',
    ['ñ'] = 'n',
    ['ý'] = 'y', ['ÿ'] = 'y'
};

public static string Normalize(string value)
{
    Span<char> buffer = stackalloc char[value.Length];
    for (int i = 0; i < value.Length; i++)
    {
        var c = char.ToLowerInvariant(value[i]);
        buffer[i] = AccentMap.TryGetValue(c, out var mapped) ? mapped : c;
    }

    return new string(buffer);
}
```
And I’m using it like this inside method FuzzyMatch: 
```c#
var fullStringToCompareAndNormalize = opt.IgnoreCase ? Normalize(stringToCompare) : stringToCompare;
var queryWithoutCaseAndNormalize = opt.IgnoreCase ? Normalize(query) : query;

var fullStringToCompareWithoutCase = opt.IgnoreCase ? stringToCompare.ToLower() : stringToCompare;
var queryWithoutCase = opt.IgnoreCase ? query.ToLower() : query;

var fullStringToCompare = _settings.SensitiveAccents ? fullStringToCompareAndNormalize : fullStringToCompareWithoutCase;
var queryToCompare = _settings.SensitiveAccents ? queryWithoutCaseAndNormalize : queryWithoutCase;
```
This logic is only triggered when the user enables the option via a button in the general settings.

However, I realized that since the apps/results are cached in memory during queries, I would need to reset the cache whenever the user enables or disables this option so that the behavior updates correctly.

I got a bit stuck on how to handle this situation properly and would appreciate any guidance on the best approach.

VictoriousRaptor (dismissed):
Please remove all translations changes. Translation is done on crowdin after merging this PR.

Jack251970 (changes requested):
Could you please resolve the build issue first?

Flow.Launcher/SettingPages/Views/SettingsPaneGeneral.xaml
```diff
@@ -91,7 +91,26 @@
                     OffContent="{DynamicResource disable}"
                     OnContent="{DynamicResource enable}" />
             </ui:SettingsCard>
-
+            <ui:SettingsCard
+                Description="{DynamicResource ignoreAccentsToolTip}"
```
Jack251970: I think you forget to add these string resources in en.xaml.

Flow.Launcher/SettingPages/ViewModels/SettingsPaneGeneralViewModel.cs
```diff
@@ -22,13 +22,16 @@ public partial class SettingsPaneGeneralViewModel : BaseModel
     private readonly Updater _updater;
     private readonly Portable _portable;
     private readonly Internationalization _translater;
-
-    public SettingsPaneGeneralViewModel(Settings settings, Updater updater, Portable portable, Internationalization translater)
+    private readonly IPublicAPI _publicAPIInstance;
```
Jack251970: Could you do not create api here and use App.API instead to keep aligned with other parts?

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -80,10 +86,16 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
             int acronymsTotalCount = 0;
             int acronymsMatched = 0;
 
+            var fullStringToCompareAndNormalize = opt.IgnoreCase ? Normalize(stringToCompare) : stringToCompare;
```
Jack251970: Is the contents between 89 and 91 duplicated with the contents between 90 to 93?

If I disable `_settings._IgnoreAccents`, `Normalize` function seems to invoke here which I think could affect the performance. I think we should only invoke `Normalize` function when `_settings.IgnoreAccents` is enabled

Jack251970 (changes requested):
Please resolve my comments

Flow.Launcher/SettingPages/Views/SettingsPaneGeneral.xaml
4yinn: srry, I forgot, lmao.

Jack251970 (changes requested):
Can you tell me why we need to restart Flow when change accent option?

I think `_settings.IgnoreAccents` in `StringMatcher.cs` can be changed during runtime and we do not need to restart Flow for taking this into effect.

4yinn:
> Can you tell me why we need to restart Flow when change accent option?
> 
> I think `_settings.IgnoreAccents` in `StringMatcher.cs` can be changed during runtime and we do not need to restart Flow for taking this into effect.

I also wanted it to work that way, but I couldn’t find a way to do it without needing to restart. I had thought about using the method to reset the memory cache that exists in Main in the PluginProgram, but I wasn’t able to use it or call it in any way.

Jack251970:
> > Can you tell me why we need to restart Flow when change accent option?
> > I think `_settings.IgnoreAccents` in `StringMatcher.cs` can be changed during runtime and we do not need to restart Flow for taking this into effect.
> 
> I also wanted it to work that way, but I couldn’t find a way to do it without needing to restart. I had thought about using the method to reset the memory cache that exists in Main in the PluginProgram, but I wasn’t able to use it or call it in any way.

This is because Program plugin constructs memory cache for program results.

https://github.com/Flow-Launcher/Flow.Launcher/blob/6d8c35c43c97ad6f335948d0d192c2b4ac7fb9e2/Plugins/Flow.Launcher.Plugin.Program/Main.cs#L38-L39

Jack251970:
One possible solution is to reload plugins when this option changes

Jack251970:
@jjw24 ~~How do you think about this issue?~~

4yinn:
> One possible solution is to reload plugins when this option changes

I had noticed that the StringMatcher method is implemented as a singleton, so it ends up being a bit complicated. That’s why I was thinking about restarting the flow for this option, since it involves the query mechanism.

Jack251970:
@4yinn I added a new API event to handle this. Since I do not have any programs with accent-contained names, could you please help me check if this option can take effect for Program plugin without restarting? Note that you need to wait for a moment after you changed this option and test.

VictoriousRaptor:
> @4yinn I added a new API event to handle this. Since I do not have any programs with accent-contained names, could you please help me check if this option can take effect for Program plugin without restarting? Note that you need to wait for a moment after you changed this option and test.

Why not use the same architecture of PinyinAlphabet? It's doesn't require a restart.

Jack251970:
> > @4yinn I added a new API event to handle this. Since I do not have any programs with accent-contained names, could you please help me check if this option can take effect for Program plugin without restarting? Note that you need to wait for a moment after you changed this option and test.
> 
> Why not use the same architecture of PinyinAlphabet? It's doesn't require a restart.

I think 5yinn is talking about the cache behavior of Program plugin.

Jack251970:
Program plugin will cache the query and results in memory. And we meed to reset it when these settings change.

VictoriousRaptor:
> Program plugin will cache the query and results in memory. And we meed to reset it when these settings change.

Oh sorry I forget there's a memory cache.

4yinn:
> @4yinn I added a new API event to handle this. Since I do not have any programs with accent-contained names, could you please help me check if this option can take effect for Program plugin without restarting? Note that you need to wait for a moment after you changed this option and test.

That works.

Jack251970 (commented):
LGTM! Thanks for your contribution! Please wait for jjw24's review

Jack251970:
@jjw24 Hi, this PR looks good to me! Could you please finally review it? Thanks!

4yinn:
@Jack251970  I added more characters for testing after reviewing the issues you mentioned. Do you think this change is necessary?

Jack251970:
> @Jack251970 I added more characters for testing after reviewing the issues you mentioned. Do you think this change is necessary?

I cannot find any reasons to reject this. Good work!

Jack251970 (dismissed):
LGTM!

jjw24:
I will review soon.

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -218,19 +244,134 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
...
+            Span<char> buffer = value.Length <= 512
+                ? stackalloc char[value.Length]
+                : (arrayFromPool = ArrayPool<char>.Shared.Rent(value.Length));
+            try
+            {
+                for (int i = 0; i < value.Length; i++)
+                {
+                    var c = char.ToLowerInvariant(value[i]);
+                    buffer[i] = AccentMap.TryGetValue(c, out var mapped) ? mapped : c;
+                }
```
jjw24: Is there a reason this can't be out inside the existing loop? Normalize is still called to loop through both compare strings before the main loop, which also loops through the new strings.

4yinn: In a way, I placed this loop outside to keep it O(n). If I left this loop inside the existing loop, it would have to traverse the same string more than once, resulting in O(n²) complexity. This is my understanding—I could be wrong, but this is the conclusion I reached.

4yinn: @jjw24  I took a closer look to give a clearer answer: I didn’t put the normalization inside the loop because it needs to already be normalized before entering the loop to perform the processing.

jjw24: Yes, mostly true- the queryToCompare string will need to be normalized before entering because of the splitting into substrings, but I still think there are benefits to move the stringToCompare normalization inside the main loop so it's not doubling up.

jjw24: I will create a PR against this one to show and review, leave it with me.

jjw24: @4yinn please have a look at this PR and let me know what you think, I try to explain my reasoning in the description and hopefully it makes sense. @Jack251970 @VictoriousRaptor tagging you since you guys are across the changes in this PR and can also review my proposed changes.
https://github.com/4yinn/Flow.Launcher/pull/1

4yinn: I understand what you meant. I was analyzing it yesterday with the help of my friend @01Dri , and we tried to find a way to avoid doing two loops and instead reuse at least the one from fullStringToCompare, but so far, everything we’ve come up with leads us back to the same point where it needs to be normalized before entering.

jjw24: Hmm, do the changes in my PR work?

4yinn: > Hmm, do the changes in my PR work?

Sorry for the delay, I tested it now and it didn’t work.

jjw24: It doesn't normalize the character? Can you show me how to reproduce please.

4yinn: > It doesn't normalize the character? Can you show me how to reproduce please.

First test I did was the usual one: I enabled `IgnoreAccents` and searched for "camera", and the result was:

test1



Second, I tried searching with the accent "câmera", and that didn’t work either:

test2


When I disable it, the query works normally.

jjw24: I will take another look soon.

jjw24: @4yinn issue fixed with the PR, please pull latest and test again.

{DA7BEF58-7736-4E3F-A1A5-6849B47093B3}

4yinn: > @4yinn issue fixed with the PR, please pull latest and test again.
> {DA7BEF58-7736-4E3F-A1A5-6849B47093B3}

Works perfectly!! Thanks — I really like the logic you used. I’m going to study it because I found it genuinely impressive. Appreciate the modification!

jjw24: Could you please merge the PR in :)

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -218,19 +253,164 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
...
+        private static char[] BuildAccentLookup()
+        {
+            var lookup = new char[AccentRangeEnd - AccentRangeStart + 1];
+            foreach (var (key, value) in AccentMap)
+                lookup[key - AccentRangeStart] = value;
+            return lookup;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static char NormalizeChar(char c)
```
jjw24: To keep cleaner and less noisy, could we move everything related to normalization out of StringMatcher and into its own file please.

4yinn: > To keep cleaner and less noisy, could we move everything related to normalization out of StringMatcher and into its own file please.

Where do you think this belongs? Should it live in the Infrastructure layer as well?

jjw24: For now, the code of normalization is focused on diacritics, so I changed the file name to make it more specific and meaningful.

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -1,30 +1,42 @@
...
+
+            _settings.PropertyChanged += (sender, e) =>
+            {
+                switch (e.PropertyName)
+                {
+                    case nameof(Settings.QuerySearchPrecision):
+                        UserSettingSearchPrecision = _settings.QuerySearchPrecision;
+                        break;
+                }
+            };
```
jjw24: @Jack251970 @4yinn this is not needed because anytime the QuerySearchPrecision is set StringMatcher's copy is updated: https://github.com/Flow-Launcher/Flow.Launcher/blob/4e947273914eadfe9f4861fe1ca7218e8f86f59c/Flow.Launcher.Infrastructure/UserSettings/Settings.cs#L401-L408

Jack251970: Let's remove it

jjw24:
@cubic-dev-ai review

jjw24:
@coderabbitai full review

jjw24:
@4yinn I added a prompt asking users whether to ignore accents when switching languages.

4yinn:
> @4yinn I added a prompt asking users whether to ignore accents when switching languages.

great idea!!

4yinn:
> > @4yinn I added a prompt asking users whether to ignore accents when switching languages.
> 
> great idea!!

I tested it, and it works well for me.

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -80,10 +75,20 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
...
+            else if (ignoreCase)
+            {
+                queryToCompare = queryToCompare.ToLower();
+            }
 
-            var querySubstrings = queryWithoutCase.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
+            var querySubstrings = queryToCompare.Split([' '], StringSplitOptions.RemoveEmptyEntries);
             int currentQuerySubstringIndex = 0;
             var currentQuerySubstring = querySubstrings[currentQuerySubstringIndex];
             var currentQuerySubstringCharacterIndex = 0;
```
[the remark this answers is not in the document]
jjw24: Good pickup but not applicable at this stage, white spaces only query are ignored by high level callers.

Flow.Launcher.Infrastructure/StringMatcher.cs
```diff
@@ -80,10 +75,20 @@ public MatchResult FuzzyMatch(string query, string stringToCompare, MatchOption
...
-            var queryWithoutCase = opt.IgnoreCase ? query.ToLower() : query;
+            var queryToCompare = query;
+            bool ignoreAccents = _settings.IgnoreAccents;
+            bool ignoreCase = opt.IgnoreCase;
+
+            if (ignoreAccents)
+            {
+                queryToCompare = DiacriticsNormalizer.Normalize(queryToCompare);
+            }
+            else if (ignoreCase)
```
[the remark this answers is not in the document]
jjw24: DiacriticsNormalizer.Normalize() call already does the lowercasing of the string so that's why it's else if not a blanket if.

jjw24 (approved):
Good work. This is not an easy change. Thank you for taking this on. 🥇

4yinn:
> Good work. This is not an easy change. Thank you for taking this on. 🥇

I’m the one who’s grateful—and this won’t be the last!!!

PR #4370, opened 2026-03-21
feat: Everything 1.5beta support

### LINKED ISSUE #1716: Add support for Everything v1.5a
Hello, I'm using [Everthing 1.5a]( https://www.voidtools.com/forum/viewtopic.php?t=9787) . It seems that this plugin can't detect it and gives a "Warning: Everything service is not running". Clicking on it doesn't seem to solve the issue and I get no feedback from doing that. 
Initially I assumed that the problem was the service name, but following @void's instructions to [disable the alpha version's instance naming ](https://www.voidtools.com/forum/viewtopic.php?f=12&t=9799#isntance) didn't solve the issue. Unfortunately I didn't have time to check if that indeed changed the name back to 1.4's.

Although this is a niche problem, I wonder if it makes sense to provision for cases like that on the plugin's side, for futureproofing as well. 

At any rate, thank you for your sharing your code.

### LINKED ISSUE #1769: BUG: Incompatibility of plugin `Flow.Launcher.Plugin.Explorer` with Everything (1.5a)
### Checks

#### If [Everything 1.5a](https://www.voidtools.com/forum/viewtopic.php?f=12&t=9787) is used and properly installed as a service instead of Everything 1.4, the current latest version, `IsEverythingRunningAsync` will return `false`:

https://github.com/Flow-Launcher/Flow.Launcher/blob/c0f46f24de3ad1bf80f329f46be375fb18115596/Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingAPI.cs#L92-L106

#### More context on this might be revealed through the comment here:

https://github.com/Flow-Launcher/Flow.Launcher/blob/c0f46f24de3ad1bf80f329f46be375fb18115596/Plugins/Flow.Launcher.Plugin.Explorer/ViewModels/SettingsViewModel.cs#L405-L425

### To Reproduce

1. Install Flow Launcher and Everything 1.4, and observe that the Flow Launcher Explorer plugin works
2. Install Everything 1.5a (It is stated, that it seamlessly migrates from 1.4 to 1.5a which it does indeed), and observe the Flow Launcher Explorer plugin does not work

### Screenshots

image

image

image

### Flow Launcher Version

v1.11.0

### Windows Build Number

10.0.22621.963

### LINKED ISSUE #4603: Implement Everything 1.5 SDK (Named Pipes) for massive performance improvements
Hello Flow Launcher team!

First, thank you for the amazing work on this project.

I would like to request an update to the Everything plugin to utilize the new Everything 1.5 SDK.

Recently, Voidtools introduced a new SDK specifically designed for Everything 1.5. In older versions, the IPC communication relied on `WM_COPYDATA`. The new SDK transitions to using Named pipes, which allows keeping the connection open and results in significant performance gains.

According to the Everything developer, using Named pipes is **roughly 5 times faster** at sending all results compared to the older method, and it effectively eliminates IPC connection delays. Other tools (like EverythingToolbar in their recent 3.0.0 release) have already implemented this, and the speed improvement when listing search results is highly noticeable.

Here is the relevant documentation and discussions from the Voidtools forum:

New SDK Documentation: https://www.voidtools.com/forum/viewtopic.php?t=15853

Developer explanation on performance differences (5x faster): https://www.voidtools.com/forum/viewtopic.php?t=15770
"Everything 1.5 IPC will use pipes instead of WM_COPYDATA which is much faster (~5x times faster to send all results) There shouldn't be any delay if you keep the pipe connection alive."

Would it be possible to integrate this new SDK into Flow Launcher's Everything plugin? This would make searching even more instantaneous for users who have Everything 1.5 installed.

Thank you!

VictoriousRaptor:
## Description
This PR adds opt-in support for `Everything 1.5` in the Explorer plugin and refactors the `Everything` integration so Flow Launcher can work with both the legacy SDK and the newer `Everything3` API.

The main goal is to make ~~`Everything 1.5a`~~ `Everything 1.5beta` usable without requiring users to manually edit the `.ini` file or rely on compatibility workarounds.

## What changed
- Added a new `Everything 1.5` integration path based on `Everything3.dll`.
- Introduced an `IEverythingApi` abstraction to separate the legacy `Everything 1.4` implementation from the new `Everything 1.5` implementation.
- Refactored API factory / loader / availability handling so the Explorer plugin can initialize the correct SDK at startup.
- Moved shared query preparation and highlight parsing into helper code to reduce duplication.
- Removed support on x86.

## Settings and UX changes
- Added a setting to enable `Everything 1.5` support.
- Added a setting for the `Everything 1.5` instance name, defaulting to empty for 1.5 beta which just rolled out.
  - Everything 1.5a needs to type `1.5a`.
- Updated settings warnings and availability messages to better distinguish between:
  - legacy `Everything` not running,
  - `Everything 1.5` not available,
  - SDK loading / architecture issues.
- Added UI text and validation around the new configuration options.

## Refactoring
- Renamed the old static `EverythingApi` usage into a legacy implementation and routed calls through `EverythingSearchManager`.
- Centralized availability checks in `EverythingAvailabilityService`.
- Simplified `Main` initialization by letting the search manager own `Everything` API initialization.
- Added small safety improvements around empty paths and semaphore-based API access.

## Functional impact
With this change, users can explicitly enable `Everything 1.5` support and configure the instance name used by the Explorer plugin. This allows Flow Launcher to query `Everything 1.5a` through its dedicated API while preserving the existing legacy path for earlier `Everything` versions.

## Related issues
- Resolves #1716 - Add support for Everything v1.5a
- Resolves #1769 - BUG: Incompatibility of plugin `Flow.Launcher.Plugin.Explorer` with `Everything (1.5a)`
- Resolves #4603 - Implement Everything 1.5 SDK (Named Pipes) for massive performance improvements


Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/Everything3ApiDllImport.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (code == 0)
+            {
+                int err = Marshal.GetLastPInvokeError();
+                Marshal.ThrowExceptionForHR(err);
+            }
+        }
+
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+        private static extern int LoadLibrary(string name);
+        private const string Dll = "Everything3.dll";
```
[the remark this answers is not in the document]
VictoriousRaptor: @copilot apply changes based on [this feedback](https://github.com/Flow-Launcher/Flow.Launcher/pull/4370#discussion_r3004414484)

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingSearchManager.cs
```diff
@@ -12,19 +12,24 @@ namespace Flow.Launcher.Plugin.Explorer.Search.Everything
...
             Settings = settings;
+            _api = CreateApi(Settings.EnableEverything15Support, GetNormalizedInstanceName(Settings.Everything15InstanceName));
         }
 
         private async ValueTask ThrowIfEverythingNotAvailableAsync(CancellationToken token = default)
         {
             try
             {
-                if (!await EverythingApi.IsEverythingRunningAsync(token))
+                if (!await _api.IsEverythingRunningAsync(token))
```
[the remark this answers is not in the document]
VictoriousRaptor: @jjw24 @Jack251970 So is it acceptable that ticking/unticking the "support Everything 1.5" option requires a restart? The new Everything SDK for 1.5 alpha is not compatible with Everything 1.4. And I don't want to load both the new and the old SDK dlls, just to save memory. Loading/unloading dll in the flight causes potential issues.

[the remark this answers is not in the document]
Jack251970: > @jjw24 @Jack251970 So is it acceptable that ticking/unticking the "support Everything 1.5" option requires a restart? The new Everything SDK for 1.5 alpha is not compatible with Everything 1.4. And I don't want to load both the new and the old SDK dlls, just to save memory. Loading/unloading dll in the flight causes potential issues.

Hi, I think it is reasonable. And I strongly recommend to work in that way to prevent some possible strange issues!

jjw24:
@cubic-dev-ai review

jjw24 (commented):
@VictoriousRaptor If cubic doesn't update the PR description after the review, could you do so please.

Plugins/Flow.Launcher.Plugin.Explorer/EverythingSDK/x86/Everything3.dll
jjw24, line 1: Do we still use x86? Maybe these can be removed?

VictoriousRaptor: > Do we still use x86? Maybe these can be removed?

maybe.

VictoriousRaptor: Removed x86 support. If many people are using it we may revert later.

VictoriousRaptor:
@jjw24 It's ready. Please further review.

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingApiV3.cs
```diff
@@ -0,0 +1,460 @@
...
+                    ascending = true;
+                    return true;
+                case EverythingSortOption.FILE_LIST_FILENAME_DESCENDING:
+                    propertyId = EVERYTHING3_PROPERTY_ID_FILE_LIST_NAME;
+                    ascending = false;
+                    return true;
+                case EverythingSortOption.RUN_COUNT_DESCENDING:
+                    propertyId = EVERYTHING3_PROPERTY_ID_RUN_COUNT;
+                    ascending = false;
+                    return true;
```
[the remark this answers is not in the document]
VictoriousRaptor: @copilot apply changes based on [this feedback](https://github.com/Flow-Launcher/Flow.Launcher/pull/4370#discussion_r3178176415)

Plugins/Flow.Launcher.Plugin.Explorer/Views/ExplorerSettings.xaml
```diff
@@ -614,11 +621,46 @@
...
+                    IsChecked="{Binding EnableEverything15Support}" />
+
+                <TextBlock
+                    Grid.Row="1"
+                    Grid.Column="0"
+                    Margin="{StaticResource SettingPanelItemTopBottomMargin}"
+                    VerticalAlignment="Center"
+                    Foreground="{DynamicResource Color05B}"
+                    Text="{DynamicResource plugin_explorer_everything_15_instance_name}"
+                    Visibility="{Binding EnableEverything15Support, Converter={StaticResource BooleanToVisibilityConverter}}" />
```
DavidGBrett, line 634: I was a bit confused when this popped up,
I wasn't familiar with how instances work in everything but I believe for alpha it used 1.5a but now for beta it is back to the default empty name?

I think a tooltip is required to clarify this, e.g. that for most users they can leave that blank.

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingApiV3.cs
```diff
@@ -0,0 +1,436 @@
...
+                searchState = Everything3ApiDllImport.Everything3_CreateSearchState();
+                if (searchState == IntPtr.Zero)
+                {
+                    CheckAndThrowExceptionOnErrorFromEverything3();
+                    yield break;
+                }
+
+                _ = Everything3ApiDllImport.Everything3_SetSearchRegex(searchState, option.UseRegex);
+                _ = Everything3ApiDllImport.Everything3_SetSearchMatchPath(searchState, option.IsFullPathSearch);
+                _ = Everything3ApiDllImport.Everything3_SetSearchTextW(searchState, query.SearchText);
```
DavidGBrett, line 180: Is it fine to ignore the results of these calls?
Some are less important but Everything3_SetSearchTextW for example seems important no?
perhaps we could log this if we don't want to throw here

DavidGBrett (commented):
I'm not really that familiar with "everything" but I've tested it with the portable 1.5 beta version and everything I tried seems to be working
regex, highlighting, folder navigation, sorting

Legacy everything doesn't seem broken by this update either, can still use it when I turn off 1.5

I made a few comments regarding minor things worth fixing though

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingApiV3.cs
```diff
@@ -0,0 +1,436 @@
...
+                    if (!Everything3ApiDllImport.Everything3_AddSearchSort(searchState, sortPropertyId, ascending))
+                    {
+                        CheckAndThrowExceptionOnErrorFromEverything3();
+                        yield break;
+                    }
+                }
+
+                _ = Everything3ApiDllImport.Everything3_ClearSearchPropertyRequests(searchState);
+                _ = Everything3ApiDllImport.Everything3_AddSearchPropertyRequestHighlighted(searchState, EVERYTHING3_PROPERTY_ID_NAME);
+                //_ = Everything3ApiDllImport.Everything3_AddSearchPropertyRequestHighlighted(searchState, EVERYTHING3_PROPERTY_ID_PATH);
```
DavidGBrett, line 196: should remove this commented out line here

Plugins/Flow.Launcher.Plugin.Explorer/Views/ExplorerSettings.xaml
DavidGBrett: Ideally instance name should be something that you can set regardless of if 1.5 is chosen (unless that wasn't available in legacy)

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingApiV3.cs
VictoriousRaptor: They simply returns a bool to indicate success or not. As we check for errors later so it's fine to discard them.

DavidGBrett: It looks like we only call `CheckAndThrowExceptionOnErrorFromEverything3` if one of
- `Everything3_CreateSearchState`
- `Everything3_AddSearchSort`
- `Everything3_Search`
- `Everything3_GetResultFullPathNameW` (in `TryGetResultFullPath`)

fails, right?

Could one of those unchecked calls fail but all those other checked cases still pass, silently failing?
E.g. Everything3_SetSearchRegex not being set so characters are interpreted as raw instead? 

Probably only a very rare edge case that something like that call could fail I imagine though...
Might still be nice to have logging of this in case it comes up as an issue later.

VictoriousRaptor: Maybe I can  check the code this weekend. It's been a while so I'm not very sure now.

gerroon:
Hi will this be merged? it would be amazing since Everythign 1.5 is much better

DavidGBrett:
VictoriousRaptor has been working hard on this for quite a while, there are only a few minor things left so it won't be long before its merged

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingApiV3.cs
VictoriousRaptor: Added `LogDebug` calls for each of the previously-discarded `Everything3_SetSearch*`/`Everything3_SetSearchViewport*` calls, so if one of them fails silently it's now recorded for diagnosis without changing the existing throw-on-error behavior (which still happens later via `CheckAndThrowExceptionOnErrorFromEverything3`).

Plugins/Flow.Launcher.Plugin.Explorer/Search/Everything/EverythingApiV3.cs
VictoriousRaptor: Removed the commented-out line.

DavidGBrett (approved):
Thank you for your hard work on this!
All my feedback has been addressed.

PR #4454, opened 2026-05-10
FEATURE: Calculator Plugin History

01Dri:
## Changes
- Added `History` and `HistoryItem` objects to persist calculator results
- Added a plugin setting to enable or disable history
- Implemented a debounce strategy to reduce unnecessary history entries

## Debounce Strategy
Currently, the calculator plugin returns results even when the Enter key is not pressed. Because of this behavior, every partial expression can be added to the history while the user is typing.
For example, while the user is building the expression:
```txt
1+2+3+4+5
```
Without a debounce mechanism, each intermediate expression is stored in history, generating multiple unnecessary entries.
With the debounce strategy, the plugin waits for a short delay before saving the result. This ensures that only the final or more complete expression is added to history.

### Without debounce
- `1`
- `1+2`
- `1+2+3`
- `1+2+3+4`
- `1+2+3+4+5`

### With debounce
- `1+2+3+4+5`

This keeps the history cleaner and improves the overall user experience.

---

> [!NOTE]
> I'm not fully satisfied with the result title and subtitles yet, so suggestions are welcome!

### Time-Ago Formatting
We implemented custom, localized relative time strings (e.g., "just now", "5 minutes ago"). This was needed because there is no built-in datetime formatting helper in the Flow Launcher codebase. The implementation uses the plugin's own local translation keys (in `en.xaml`) to keep everything localized correctly.

### Debounce vs Immediate Modes
- **On Query Mode (Debounced)**: Uses a `PendingHistoryItem` together with an 800ms debounce timer to prevent saving incomplete expressions while the user is typing.
- **On Enter Mode (Immediate)**: When execution occurs (e.g., when the user presses Enter), the item is added or updated directly using a standard `HistoryItem` without any debounce.

### Code Quality & Refactorings
- **Constructor Chaining**: Chained the constructors in `HistoryItem.cs` so that the `PendingHistoryItem` constructor delegates to the primary one, avoiding initialization duplication.
- **Unified Action Delegation**: Structured `CreateClipboardActionWithHistory` in `Main.cs` to delegate directly to `CreateClipboardAction`, avoiding duplicated copy and error handling logic.
- **Unified Update Logic**: Created `AddOrUpdateInternal(HistoryItem)` in `History.cs` to standardize the update path for both modes.
- **Main Thread Execution**: Removed unnecessary locks (`lock (_syncRoot)`) on the normal `HistoryItem` addition path since it is triggered from the main thread.

---
Screenshot 2026-07-04 201306

https://github.com/user-attachments/assets/02af3d1f-3156-4692-afa2-c41a541512cf

---
## Summary by cubic
Adds an optional calculation history to the calculator plugin, showing up to 5 recent calculations beneath the current result. Also updates the history badge icon to a clearer design.

- **Summary of changes**
  - Changed: `Query` now returns the main result plus recent history (when `EnableHistory` is on), excluding the active expression. Main result shows a clear “copy to clipboard” subtitle. Clipboard is handled via shared actions; in `OnEnter` mode history saves only after a successful copy. History items render with a badge and a relative “time ago.” Updated the history badge icon for better clarity.
  - Added: Settings `EnableHistory` (default off) and `HistoryCreationMode` (`OnQuery` with 800ms debounce, `OnEnter`). Storage-backed `History` with a 5-item cap (drops oldest). `HistoryItem`, `PendingHistoryItem`, and `HistoryHelper` for pending items and localized time-ago strings. History icon, English strings, and settings UI with mode selector and an `OnQuery` warning. Unit tests for enabled/disabled behavior, stored fields, and debounce; tests reset history and flush timers.
  - Removed: Inline copy-to-clipboard logic from `Query` in favor of centralized helpers.
  - Memory impact: Low. At most 5 stored items, one pending item, and one debounce timer in settings storage.
  - Security risks: Low. History is local and off by default. `OnQuery` may capture partial expressions; UI warns; nothing is sent externally.
  - Unit tests: Added coverage for history enable/disable behavior, saved data, and debounce flush; includes per-test history reset.

- **Release Note**
  - You can now enable a calculator history that remembers up to 5 recent calculations, saved as you type or only when you press Enter.

<sup>Written for commit 285859d205bf14924a651102cdb6b563568061bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Flow-Launcher/Flow.Launcher/pull/4454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Plugins/Flow.Launcher.Plugin.Calculator/Storage/History.cs
```diff
@@ -0,0 +1,85 @@
...
+                currentItem.Refresh(_pendingItem.Result, _pendingItem.Action);
+            }
+            else
+            {
+                var item = new HistoryItem(_pendingItem.Result, _pendingItem.Expression, _pendingItem.Action);
+
+                if (Items.Count >= MaxItems)
+                {
+                    Items.RemoveAt(0);
+                }
```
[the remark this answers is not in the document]
DavidGBrett: If refreshing doesn't change the order then it definitely makes sense to consider the CalculatedAt time @01Dri

DavidGBrett (commented):
~~Currently my result icons aren't getting that clock (history.png) layered on top of them like yours are in the screenshots provded - not sure why~~
My mistake I didn't realise I needed the results badges to be enabled in the settings

One design concern is that if the user is searching for a numerical query, e.g. a date,
then there would be a lot of history results in the way of what they actually want.
Its a difficult problem to solve since as a query a date and a calculation can look identical

It's a little confusing that the title is not whats being copied to the clipboard
Perhaps it could be `query = result`?

Another approach that you could consider is switching the current query to that historical query, making it work more like a rewind button (that's what I was used to on my old physical calculator).
They can still use the existing result to copy it after switching - a context menu copy is also an option but perhaps unnecessary.

Instead of a timestamp, I think a time delta would be more helpful.
No need for the `Calculated at` part either I think, too verbose.
e.g. `2 minutes ago` 

image

Plugins/Flow.Launcher.Plugin.Calculator/Images/history.png
DavidGBrett, line 1: I can confirm this is from icons8 which we already give attribution to so is safe to use
https://icons8.com/icon/QDgOnr6UAOmg/time-machine

My concern is that at a small scale its not too readable

I found another alternative, which might be slightly more clear - though still not ideal
https://icons8.com/icon/--bXx-DYszmx/time-machine

Image

01Dri:
@DavidGBrett 
@Jack251970 

I made some adjustments. Could you test it again and let me know if everything looks good? Thanks!

Plugins/Flow.Launcher.Plugin.Calculator/Languages/pt-br.xaml
Jack251970, line 1: Hi, could you please revert changes in this file?

Jack251970 (changes requested):
Please revert changes in pt-br.xaml.

Jack251970:
@codex review

Plugins/Flow.Launcher.Plugin.Calculator/Main.cs
```diff
@@ -119,43 +123,68 @@ public List<Result> Query(Query query)
...
-                                    return true;
-                                }
-                                catch (ExternalException)
-                                {
-                                    Context.API.ShowMsgBox(Localize.flowlauncher_plugin_calculator_failed_to_copy());
-                                    return false;
-                                }
-                            }
+                                Title = item.Query,
+                                SubTitle = $"{item.CopyText} - {historySubtitle}",
```
DavidGBrett, line 176: Image

Using the subtraction sign `-` to separate the answer from the time ago part is definitely confusing ( it equals 3 minus 2?)
You could put these on separate lines,
but I think the best solution might be to just put it after an `=` sign in the title

E.g.
Title | Subtitle
-- | --
1 + 2 = 3 | 2 min ago

For consistency you could also put an `=` before the current query's answer

E.g.
Title | Subtitle
-- | --
= 8 | Copy this number to the clipboard

Plugins/Flow.Launcher.Plugin.Calculator/Main.cs
```diff
@@ -119,43 +123,68 @@ public List<Result> Query(Query query)
...
+                                IcoPath = item.IcoPath,
+                                BadgeIcoPath = item.BadgeIcoPath,
+                                ShowBadge = item.ShowBadge,
+                                Score = item.Score,
+                                CopyText = item.CopyText,
+                                Action = item.Action
```
DavidGBrett, line 182: I do think that instead of copying the answer to the clipboard history results should change the query to that old one

This lets you more easily switch between queries and if you do want to copy it then the pre-existing answer with "copy to clipboard" result is still there

DavidGBrett (changes requested):
Thank you for your continued work on this!

Good work on the time delta string system, a nice improvement.

Do have a look at the different history icons I suggested as well.

Plugins/Flow.Launcher.Plugin.Calculator/Languages/pt-br.xaml
DavidGBrett: The recent commit a7efe5d seems intended to revert that but there are still changes showing up for me

Plugins/Flow.Launcher.Plugin.Calculator/Images/history.png
DavidGBrett: Looking at the icons we can use again I think this one might be the best - clearer than the green one and less aggressive than the red one

image


Here's the link to the icon
https://icons8.com/icon/x1d6P0XJa15s/clock

I can commit the change myself though if you are happy with this one.

01Dri: Could commit please? Thanks!

DavidGBrett: Done!

01Dri:
@DavidGBrett 

Hi! I've been away these last few weeks because I've been busy with some personal matters, but I'll be back soon.

DavidGBrett:
No worries at all,
Best of luck with that!

PR #4529, opened 2026-06-12
Add markdown preview type and per result preview visibility control

TrueCrimeDev:
## What

The preview pane can now render formatted markdown with syntax-highlighted code blocks. Plugins opt in by setting `Result.Preview.ContentType` to `Markdown` and providing a markdown description.

A new `Result.PreviewVisibility` lets plugins control whether the preview pane shows:

- **Optional** (default) - respects the user's Always Preview setting and preference from the preview toggle
- **Always** - forces the pane open so users see the preview without manual toggling
- **Never** - hides the pane for results with nothing useful to preview

And `PreviewContentType` controls how the preview is rendered:

- **ImageWithText** (default) - the classic preview: image above, plain text below
- **Markdown** - formatted markdown with headings, links, and syntax-highlighted code blocks

Existing plugins are unaffected as all defaults match current behavior.

## Why

Plugins that produce rich text (AI assistants, documentation/snippet/note search, dictionary-style lookups) currently have to cram everything into plain `SubTitle`/`Description` text. This lets them opt into a proper rendered preview per result without affecting any existing plugin: every current result keeps the default `text` behavior.

The visibility controls exist so plugins can make sure users see the preview when it matters, without relying on them to manually toggle it or change settings. Results that exist primarily for their rendered content (like markdown) always show their preview; results with nothing useful to preview can hide it.

## Details

**Markdown rendering** — Built on MdXaml (already a dependency) with code blocks highlighted through AvalonEdit (the one new package). A `CodeHighlightTheme` setting under Settings > Theme lets users pick a syntax theme or use Auto to match the app colour scheme. Bundled themes include VS Code Dark+, VS Code Light, One Dark, and Catppuccin Macchiato.

**JSON-RPC** — Both `contentType` and `previewVisibility` are added as keys on the result object (with `contentType` nested in `preview`):
```json
{ "preview": { "contentType": "markdown", "description": "**hello**" }, "previewVisibility": "always" }
```

**Themed horizontal scrollbars** — Code blocks need horizontal scrolling, so horizontal scrollbar styles were added to Base.xaml and every built-in theme. Minimum thickness is 5px since horizontal bars are harder to click and have no scrollwheel alternative.

**New dependency** — `AvalonEdit` 6.3.0.90.

## Tests

Full suite passes (256/256), including new coverage: pane gating rules (`MainViewModelPreviewTest`), markdown control behavior (`PreviewMarkdownScrollViewerTest`, `PreviewMarkdownStyleTest`), and JSON-RPC `contentType` deserialization (`JsonRPCPluginTest`).

## Demo
Slightly outdated but still a good demonstration
📹 [markdown-preview-demo.mp4](https://github.com/TrueCrimeDev/Flow.Launcher/blob/markdown-preview-pane/docs/assets/markdown-preview-demo.mp4)

## TODO
1. Update https://github.com/Flow-Launcher/flow-launcher.github.io theme builder with the new changes
2. Update documentation on markdown usage

---
## Summary by cubic
Adds a markdown preview pane with syntax-highlighted code, per-result preview visibility (optional/never/always), and a `Code Highlight Theme` setting with Auto that follows the app theme. The pane won’t open without a selected result; existing results default to `imageWithText`.

**Summary of changes**
- Changed
  - Preview visibility: single toggle preference `_previewPreferenceFromToggle`; results can force show/hide via `PreviewVisibility`; forced previews restore on reopen; internal preview is no longer auto-hidden on window hide; toggle always affects the current result; no selection keeps the pane hidden; fixed startup so preview doesn’t show before `ResetPreviewAsync` runs.
  - Preview flow: `UpdatePreviewAsync` delegates to `ShowPreviewAsync`; all preview calls awaited with try/catch; `Reset` closes external before opening internal; navigation still respects plugin visibility.
  - Input/UX: only code‑block navigation keys are captured; while a code block is focused the window key handler is skipped; hyperlinks open with improved error handling; preview height bound to the result list.
  - Styling: themed horizontal scrollbars with overrideable `HorizontalScrollBarStyle`/`HorizontalThumbStyle`; explicit code‑block scrollbar template avoids last‑line overlap; subtler thumbs; hyperlink underlines persist on hover/focus; preview and code‑block scrollbars themed across built‑in themes.
  - Naming/strings: default preview type renamed to `imageWithText`; `PreviewVisibility.Default` to `Optional`; tooltips and README updated; `Result.Preview` now a new instance per result and cloning deep‑copies to avoid shared state.
- Added
  - `Result.Preview.ContentType` (`imageWithText`, `markdown`) and `Result.PreviewVisibility` (`optional`, `never`, `always`) with JSON‑RPC wire values.
  - `PreviewMarkdownScrollViewer` using `MdXaml` + `AvalonEdit`, per‑editor themed colorizer, language aliases, themed code‑block scrollbars, and `IsCodeBlockFocused`.
  - `Settings.CodeHighlightTheme` (default "Auto") with a Theme dropdown; Auto tracks light/dark and updates on theme change and startup. Dependency: `AvalonEdit` `6.3.0.90`.
- Removed
  - Old hidden preview content type (use `PreviewVisibility.Never`), `HidePreviewPane`/`ForcePreviewPane` wrappers, and the `IsSafeScheme` helper (allow only http/https).
  - Automatic hide of the internal preview on window hide.
- Memory
  - Small increase from `AvalonEdit`, per‑editor colorizers, and style resources; editors are created on demand.
- Security
  - Low risk: markdown renders to a FlowDocument (no scripts/autonavigation); links restricted to http/https; code blocks are read‑only.
- Unit tests
  - Added: `MainViewModelPreviewTest`, `PreviewMarkdownScrollViewerTest`, `PreviewMarkdownStyleTest`, `CodeHighlightThemeTest`, and `JsonRPCPluginTest`. Tests cover toggle behavior, startup/reset edge cases, no‑selection guard, restore on reopen, JSON‑RPC content types, and theme auto‑resolve.

**Release Note**
You can preview markdown with clickable links and highlighted code, choose a code theme, and plugins can auto‑show or hide previews for their results.

<sup>Written for commit 25754b196bdf8b76dd6fbadfdddbcd5fa1fc3941. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Flow-Launcher/Flow.Launcher/pull/4529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
DavidGBrett, line 1: The horizontal scrollbar on code blocks covers the last line.

Ideally this should properly be fixed in the UI,
but if thats not possible then perhaps its possible to insert a new line as a bandaid fix

Image

Flow.Launcher/MainWindow.xaml
DavidGBrett, line 1: When you close and re-open the main window (e.g. double tap the hotkey), the preview window doesn't show up again until you hover over a result.

This is only if you have "Always Preview" disabled

This is a problem if you are, for example, trying to "tab" between the response and another editor.

Image

Flow.Launcher/Resources/Controls/CodeHighlightTheme.cs
DavidGBrett, line 1: The code highlighting doesn't work well with light themes
None of the provided themes seem to suit a light theme anyways

Image

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
DavidGBrett, line 1: If you set flow to light mode but use a dark theme then you end up with an odd mix
Light coloured scrollbar and code background - but dark everything else

Admittedly this is an odd situation but specifically at least it would be better if the preview scrollbar used the same styling as the existing scrollbar

Image

With the same configuration the main window still uses the theme's scrollbar

Image

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
```diff
@@ -0,0 +1,245 @@
...
+using MdXaml;
+
+namespace Flow.Launcher.Resources.Controls;
+
+public class PreviewMarkdownScrollViewer : MarkdownScrollViewer
+{
+    private static readonly string[] EmphasisMarkers = ["**", "__", "*", "_"];
+    private const double MinimumDocumentPageWidth = 64;
+
+    internal static CodeHighlightTheme ActiveTheme { get; set; } = CodeHighlightTheme.VSCodeDarkPlus;
```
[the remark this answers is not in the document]
DavidGBrett: Definitely important - no point having different code highlight themes if they aren't set anywhere and should ideally be tied to the existing theme system somehow
At least there should be a setting to select the code theme in the settings window under appearance

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1167,6 +1216,42 @@ when PluginManager.AllowAlwaysPreview() && CanExternalPreviewSelectedResult(out
...
         {
+            // Respect per-result HidePreviewPane: collapse the pane on selections that
+            // explicitly opt out (e.g., Shorty's "Ask" placeholder), restore on others.
+            if (ShouldSuppressPreview(PreviewSelectedItem))
+            {
+                await SuppressPreviewAsync();
+                return;
+            }
+
+            if (_previewSuppressedBySelectedResult)
```
DavidGBrett: Why is there this double check for preview suppressed?
Does the previous call to ShouldSuppressPreview not handle this?
Or is this doing something else?

This probably needs a comment added to explain why

DavidGBrett (changes requested):

Flow.Launcher.Plugin/Result.cs
```diff
@@ -416,7 +422,32 @@ public record PreviewInfo
...
+        /// Render preview descriptions as markdown.
+        /// </summary>
+        [JsonStringEnumMemberName("markdown")]
+        Markdown,
+
+        /// <summary>
+        /// Suppress the preview pane for this result, even when the global "Always Preview" setting is on.
+        /// </summary>
+        [JsonStringEnumMemberName("hidden")]
+        Hidden,
```
DavidGBrett: I'm not sure if the content type should be used to hide the preview

It isn't really a content type and it requires the plugin author to create add PreviewInfo to the result just to hide it

Perhaps it would be better to add a new property outside PreviewInfo and on the Result directly such as `AllowPreview`
Should be more easily discoverable also.

DavidGBrett:
Incase anyone else is reviewing this, heres the link to the plugin @TrueCrimeDev made using this feature:
https://github.com/TrueCrimeDev/Flow.Launcher.Plugin.Shorty

Flow.Launcher.Plugin/Result.cs
DavidGBrett: Given that you want to force the markdown previews, perhaps we could instead use a visibility enum here? 

Something like:
```
public enum PreviewVisibility
{
    Default,
    Never,
    Always
}
```

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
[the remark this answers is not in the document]
DavidGBrett: Perhaps the default option for the code highlight theme could be something like `auto` or `match color scheme` - which uses a default light or dark theme depending on the existing color scheme setting

Still letting the user pick a specific theme if they want it though

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
TrueCrimeDev: Addressed in 1ade524 — the code-block editor now uses an explicit `ScrollViewer` template where the horizontal scrollbar sits in its own grid row, so it can no longer overlap the last line (no trailing-newline hack needed). Mind giving it a quick look on your end to confirm it renders as expected?

Flow.Launcher/MainWindow.xaml
TrueCrimeDev: Fixed in 1504177. `ResetPreview()` runs on every window show and, with Always Preview off, was unconditionally hiding the pane. It now honours a result's `PreviewVisibility.Always`, so a forced preview returns immediately on reopen instead of waiting for a hover. Added regression tests for both the forced (reopens) and default (stays hidden) cases.

Flow.Launcher/Resources/Controls/CodeHighlightTheme.cs
TrueCrimeDev: Added a **VS Code Light** code theme in 38c27fd, and the new default `Auto` setting selects it automatically when the app is rendering with a light colour scheme.

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
TrueCrimeDev: The preview pane and the code blocks now use Flow's thin themed scrollbar instead of the default WPF one (7092ad2, 1ade524), so they match the rest of the UI rather than standing out. Worth a quick visual check on your side.

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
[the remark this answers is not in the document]
TrueCrimeDev: Added a **Code Highlight Theme** dropdown under Settings → Theme (38c27fd). It persists to `Settings.CodeHighlightTheme` and is applied at startup and whenever the colour scheme changes.

Flow.Launcher/ViewModel/MainViewModel.cs
TrueCrimeDev: They're two different things — sorry that wasn't clear. The first is `ShouldSuppressPreview` (does the *current* result opt out, i.e. `PreviewVisibility.Never`). The second was a one-shot restore flag for when a *previous* Never result hid a pane we'd promised to bring back. Renamed it `_restorePreviewAfterNeverResult` and added comments spelling out why both checks exist (ea78b53).

Flow.Launcher.Plugin/Result.cs
TrueCrimeDev: Agreed — removed `PreviewContentType.Hidden` and added a `PreviewVisibility` property directly on `Result` (not nested in `PreviewInfo`) so it's discoverable and independent of the content type (ea78b53).

TrueCrimeDev: Implemented exactly as suggested (ea78b53): `public enum PreviewVisibility { Default, Never, Always }` on `Result`. `Never` replaces the old `Hidden`; `Always` forces the pane open even when Always Preview is off, which is how forced markdown previews now opt in.

One behaviour change worth flagging: content type no longer forces the pane open on its own, so markdown results that should auto-open now set `PreviewVisibility.Always` explicitly. I've migrated the Shorty plugin to match.

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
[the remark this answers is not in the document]
TrueCrimeDev: Done — the default is **Auto (match color scheme)**, which resolves to a light or dark code theme from the current app theme, while still letting users pick a specific theme (38c27fd).

TrueCrimeDev:
Thanks for the thorough review @DavidGBrett — all points addressed, with replies in each thread. Quick map:

- `PreviewVisibility { Default, Never, Always }` on `Result` replaces `PreviewContentType.Hidden`, decoupling pane visibility from content type (ea78b53)
- Code-highlight theme is now a **Settings → Theme** dropdown, with a new VS Code Light theme and an **Auto (match color scheme)** default (38c27fd)
- Preview pane reappears on reopen for forced previews, with regression tests (1504177)
- Preview + code-block scrollbars now use Flow's thin themed style; the code-block horizontal scrollbar sits in its own grid row so it can't cover the last line (7092ad2, 1ade524)
- Clarified the double preview-suppress check (renamed `_restorePreviewAfterNeverResult` + comments)

Two notes:
1. The scrollbar / last-line fixes are visual — I'd appreciate a quick check that they render as expected on your setup.
2. **Behaviour change**: content type no longer forces the pane open on its own; forced (e.g. markdown) previews now opt in via `PreviewVisibility.Always`. Migrated the Shorty plugin to match: TrueCrimeDev/Flow.Launcher.Plugin.Shorty#4

TrueCrimeDev:
@Flow-Launcher — this one's ready for another look whenever a maintainer has time.

All of @DavidGBrett's review feedback has been addressed (responses in each thread, summary above) and the branch is green. The only items worth a second pair of eyes are the two scrollbar visuals, which I've called out in-thread. Happy to make any further changes — thanks!

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
DavidGBrett: Unfortunately the previous changes did not do this, 
but I have pushed changes that now do use the base themes for scrollbars in the preview

DavidGBrett: Yes now it no longer covers the last line of code!

DavidGBrett:
As the shorty plugin is out of date with the latest changes, and to verify this works with JsonRPC, I built my own python plugin to test this:

https://github.com/DavidGBrett/Flow.Launcher.Plugin.MarkdownPreviewTest

Not in the plugin store yet - might not be a need to add it
So just manually install from the releases if anyone wants to use this

Flow.Launcher.Plugin/Result.cs
DavidGBrett: Thanks this is better and working correctly

DavidGBrett:
Marking this as draft as I plan to push some more commits later.
There are a few more things to look into and clean up.

Thanks again for your contribution @TrueCrimeDev

TrueCrimeDev:
Feel free to send me more fixes, I am
happy to continue to working on it to make it easier on your end.

DavidGBrett:
Ok then, thank you

Don't worry about the theming for the horizontal scrollbar I'll add that to the built in themes soon

So far the biggest issue I see right now is:
- Arrow keys eventually move the curser in the code block but still change selected query or go to context menu first - need to disable those actions while focused on the code block

API Design wise:
- I think its better to rename the default preview content type from Text to ImageWithText since it actually primarily shows a preview image with text below

Beyond that there's some cleanup to do such as:
- Removing the doc/assets folder and images - if you can re-render them after the final changes that would be great to include in the PR description, but there's no need to include them in the repo directly, only adds maintenance

Beyond that its checking over the issues raised by the bots to see if they are valid
For example the one about hyperlink underlines seems important

Flow.Launcher.Plugin/Result.cs
```diff
@@ -365,6 +377,7 @@ public Result Clone()
                 ProgressBar = ProgressBar,
                 ProgressBarColor = ProgressBarColor,
                 Preview = Preview,
+                PreviewVisibility = PreviewVisibility,
```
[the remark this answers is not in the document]
DavidGBrett: Moved to separate issue #4561 as this is a pre-existing issue, can be dealt with outside this PR

Flow.Launcher/Themes/Base.xaml
```diff
@@ -578,6 +578,180 @@
...
+                        <Setter Property="FontFamily" Value="Cascadia Code, Consolas" />
+                        <Setter Property="FontSize" Value="12" />
+                    </Trigger>
+                </Style.Triggers>
+            </Style>
+
+            <Style TargetType="{x:Type Hyperlink}">
+                <Setter Property="Foreground" Value="{DynamicResource BasicSystemAccentColor}" />
+                <Setter Property="TextDecorations" Value="None" />
+            </Style>
```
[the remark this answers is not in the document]
DavidGBrett: This does seem relevant - probably best to leave the underlines on the hyperlinks for accessibility and to match average user expectations

TrueCrimeDev:
@DavidGBrett thanks for the detailed pass — pushed `8636fb2c0` addressing the follow-up list:

- **Code-block key handling** — while a code block owns keyboard focus, the window's key handler now bows out, so arrows move the caret inside the block instead of changing the selected result or opening the context menu. Fixes the reported arrow-key behaviour.
- **API rename** — `PreviewContentType.Text` → `ImageWithText` (wire name `imageWithText`), since the default preview is the image with the text description below it. Defaults, docs and tests updated.
- **docs/assets** — removed the two preview PNGs from the repo; I'll re-render them for the PR description after the final changes rather than keep them tracked.
- **Hyperlink underlines** — kept the underline (dropped the `TextDecorations="None"`) so links retain a non-colour cue.

While I was in there I also folded in a still-valid bot finding: `RetintEditor` no longer mutates the shared AvalonEdit highlighting definition — colours now apply through a per-editor `ThemedHighlightingColorizer`, so the tint can't leak into other editors. It's inserted at index 0 so the selection colorizer still wins and selected code stays legible.

One item I deliberately left for you: window-level `KeyBinding`s that the read-only editor doesn't consume (Alt+Home/End, the context-menu hotkey, Ctrl/Alt+Enter, modifier+digit) still fire while a code block is focused — they're matched during bubbling, separate from the key handler above. A correct fix needs per-gesture decisions (which of Escape/F5/Ctrl+O/Alt+Home/Enter/PageUp to block vs keep) and blanket suppression would break code-block scrolling, so that one felt like yours to shape. Happy to take it on if you have a behaviour in mind.

Flow.Launcher/Themes/Base.xaml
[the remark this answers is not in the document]
TrueCrimeDev: Done in `8636fb2c0` — dropped the `TextDecorations="None"` so links keep their underline as a non-colour cue. Agreed on both counts (accessibility and matching user expectations).

Flow.Launcher.Plugin/Result.cs
```diff
@@ -406,6 +406,12 @@ public record PreviewInfo
             /// </summary>
             public string FilePath { get; set; } = null;
 
+            /// <summary>
+            /// Determines how <see cref="Description"/> should be rendered in Flow Launcher's preview panel.
+            /// </summary>
+            [JsonConverter(typeof(JsonStringEnumConverter<PreviewContentType>))]
+            public PreviewContentType ContentType { get; set; } = PreviewContentType.Text;
```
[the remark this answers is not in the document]
DavidGBrett: Moved to separate issue as this is a preexisting problem outside scope of this pr for now

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
```diff
@@ -0,0 +1,245 @@
...
+                color.Foreground = brush;
+                changed = true;
+            }
+        }
+
+        if (changed)
+        {
+            editor.TextArea.TextView.Redraw();
+        }
+    }
```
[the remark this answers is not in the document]
DavidGBrett: Should be fixed I believe

DavidGBrett:
Thanks for sorting those out
I will look further into the key binding concerns you mentioned - the current behavior seems fine to me so far anyways.

Right now I've noticed an uncaught crash seemingly happening when the preview changes quickly so I'm focusing on trying to figure out what's causing that
To test it out you can use my own plugin which renders the query as the first line of the markdown so changing that changes the preview.

DavidGBrett:
The crash is not easy to reliably replicate but it seems that this is actually caused by a bug in how MdXaml handles image loading
`at MdXaml.ImageLoaderManager.OpenStreamAsync(Uri resourceUrl)`
Some unhandled http request failure

My test plugin included a markdown image link 
I only added that recently which explains why im only seeing it now.

The reason why this was crashing on my build was that for some reason in debug mode the top level exception handlers are disabled - unfortunately I would need to discuss with the team and fix that in a sperate PR so will have to leave this unaddressed here

It's less than ideal to leave it like this, but at least in release mode the crash shouldn't happen for normal users...

Flow.Launcher/ViewModel/MainViewModel.cs
DavidGBrett: I've refactored away this concern now - there's no need to remember the previous state anymore

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1150,6 +1193,12 @@ private void HideInternalPreview()
 
         public void ResetPreview()
         {
+            if (ShouldSuppressPreview(PreviewSelectedItem))
+            {
+                _ = SuppressPreviewAsync();
+                return;
+            }
```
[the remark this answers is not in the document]
DavidGBrett: This is fixed in commit f826c36

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1100,16 +1102,57 @@ when CanExternalPreviewSelectedResult(out var path):
...
+        }
+
         [RelayCommand]
         private void TogglePreview()
         {
+            if (ShouldSuppressPreview(PreviewSelectedItem))
+            {
+                _ = SuppressPreviewAsync();
+                return;
+            }
```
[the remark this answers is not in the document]
DavidGBrett: This is fixed in commit f826c36

Flow.Launcher/MainWindow.xaml
```diff
@@ -442,7 +443,9 @@
                         HorizontalAlignment="Center"
                         VerticalAlignment="Stretch"
                         Background="Transparent"
-                        ShowsPreview="True"
+                        FocusVisualStyle="{x:Null}"
+                        IsTabStop="False"
```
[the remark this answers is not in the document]
DavidGBrett: Not an issue - tab key is used for changing results not anything else

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
```diff
@@ -0,0 +1,245 @@
...
+    protected override void OnPropertyChanged(DependencyPropertyChangedEventArgs e)
+    {
+        base.OnPropertyChanged(e);
+
+        if (e.Property == MarkdownProperty)
+        {
+            UpdateDocumentPageWidth(ActualWidth);
+            _ = Dispatcher.BeginInvoke(ApplyMarkdownCompatibilityFixes, DispatcherPriority.Loaded);
+        }
+    }
```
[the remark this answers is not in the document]
DavidGBrett: I believe this was fixed in 4ffa212

Flow.Launcher/MainWindow.xaml
```diff
@@ -462,6 +465,15 @@
...
                             Visibility="{Binding ShowDefaultPreview}">
+                            <Border.Style>
+                                <Style TargetType="Border">
+                                    <Style.Triggers>
+                                        <DataTrigger Binding="{Binding IsMarkdownPreview}" Value="True">
+                                            <Setter Property="MaxHeight" Value="600" />
+                                        </DataTrigger>
+                                    </Style.Triggers>
+                                </Style>
+                            </Border.Style>
```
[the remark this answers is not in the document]
DavidGBrett: fixed in 6db5575f09226077fda7125067cd89d2b7de07ba

DavidGBrett (dismissed):
I've polished it up so I believe it should be good enough to merge now

A good few changes in the mean time, but I've updated the PR description so hopefully they should be clear.

If anyone else wants to review this remember to use the release build - dev build can crash as described above but that's not a concern for this PR I believe

Flow.Launcher/Resources/Controls/PreviewMarkdownScrollViewer.cs
```diff
@@ -0,0 +1,465 @@
...
+        // successfully parsed and opened the url link,
+        // so we can hide the window now
+        App.API.HideMainWindow();
+        e.Handled = true;
+    }
+
+    private static bool IsSafeScheme(string scheme)
+    {
+        var lowered = scheme?.ToLowerInvariant() ?? "";
+        return lowered is not "javascript" and not "vbscript";
```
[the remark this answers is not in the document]
DavidGBrett: I think this is fine, file: and data: aren't inherently dangerous just from a like I believe - Though I could be wrong
Since this always goes to the browser unusual custom protocols wouldn't work.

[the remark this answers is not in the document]
DavidGBrett: To be safe I applied this safety change in 265df1c

Flow.Launcher.Plugin/Result.cs
```diff
@@ -416,7 +434,52 @@ public record PreviewInfo
                 IsMedia = false,
                 PreviewDelegate = null,
                 FilePath = null,
+                ContentType = PreviewContentType.ImageWithText,
```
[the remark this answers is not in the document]
DavidGBrett: As I mentioned this is a pre-existing issue #4561
But I guess I could just solve it here, perhaps its becoming more of a problem now with this change and focus on previews

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -365,7 +344,7 @@ private void LoadHistory()
             {
                 SelectedResults = Results;
                 PreviewSelectedItem = Results.SelectedItem;
-                _ = UpdatePreviewAsync();
+                await UpdatePreviewAsync();
```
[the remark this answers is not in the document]
DavidGBrett: Should be fixed in 1d73a4a5ff20eac761f3a5467e6bd2ce0ac3208c

Flow.Launcher.Plugin/Result.cs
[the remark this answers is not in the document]
DavidGBrett: Should be solved in 9a7f95d4e62c6e2b900206ab5632442f3f9e1abd

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1026,13 +1005,59 @@ private static string VerifyOrSetDefaultHotkey(string hotkey, string defaultHotk
...
+                    _selectedItemFromQueryResults = true;
+                    PreviewSelectedItem = Results.SelectedItem;
+                    try
+                    {
+                        await UpdatePreviewAsync();
+                    }
+                    catch (Exception e)
+                    {
+                        App.API.LogError(ClassName,
+                            $"Error updating preview on result selection: {e.Message}");
```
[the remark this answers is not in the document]
DavidGBrett: Should be solved in 9ad461c amend

Flow.Launcher.Plugin/Result.cs
```diff
@@ -285,7 +285,17 @@ public string PluginDirectory
         /// <summary>
         /// Contains data used to populate the preview section of this result.
         /// </summary>
-        public PreviewInfo Preview { get; set; } = PreviewInfo.Default;
+        public PreviewInfo Preview { get; set; } = new PreviewInfo();
```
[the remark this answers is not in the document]
DavidGBrett: That's potentially a breaking API change so I will not do that in this PR - out of scope

Flow.Launcher/MainWindow.xaml
```diff
@@ -403,9 +404,9 @@
 
                 <Grid x:Name="ResultPreviewArea">
                     <Grid.ColumnDefinitions>
-                        <ColumnDefinition Width="*" MinWidth="80" />
+                        <ColumnDefinition Width="0.6*" MinWidth="80" />
                         <ColumnDefinition Width="Auto" />
-                        <ColumnDefinition Width="0.85*" MinWidth="244" />
+                        <ColumnDefinition Width="1.2*" MinWidth="320" />
                     </Grid.ColumnDefinitions>
```
jjw24, line 410: I started a fresh install of flow and turned on the preview via F1, I think we need to tweak the size of both view areas:
Image

 In comparison this prod version:
Image

Also happy if we just revert back to these widths.

Flow.Launcher/Languages/en.xaml
```diff
@@ -150,7 +150,7 @@
     <system:String x:Key="HelpUsTranslateFlow">Help us translate Flow</system:String>
 
     <system:String x:Key="AlwaysPreview">Always Preview</system:String>
-    <system:String x:Key="AlwaysPreviewToolTip">Always open preview panel when Flow activates. Press {0} to toggle preview.</system:String>
+    <system:String x:Key="AlwaysPreviewToolTip">Always open the preview panel. Press {0} to toggle. Some results may override this to always or never show their preview.</system:String>
```
```suggestion
    <system:String x:Key="AlwaysPreviewToolTip">Always open the preview panel. Press {0} to toggle. Per their own requirements, plugins may override this to always or never for their results.</system:String>
```

Flow.Launcher/Themes/BlurBlack Darker.xaml
```diff
@@ -137,6 +137,27 @@
...
+                    <Border
+                        Background="#FFFFFF"
+                        BorderBrush="Transparent"
+                        BorderThickness="0"
+                        CornerRadius="2"
+                        Opacity="0.5" />
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>
+    </Style>
```
jjw24, line 160: What happens to existing custom themes created by the community? If this change is not compatible with them, any way it can be made so?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1064,9 +1089,8 @@ public bool InternalPreviewVisible
 
         public int ResultAreaColumn { get; set; } = ResultAreaColumnPreviewShown;
 
-        // This is not a reliable indicator of whether external preview is visible due to the
-        // ability of manually closing/exiting the external preview program which, does not inform flow that
-        // preview is no longer available.
+        // Tracks whether Flow opened an external preview.
+        // External programs can close without notifying Flow, so this can be stale.
```
jjw24:
```suggestion
        // This can be stale if external preview program closes the preview or exit without notifying Flow.
```

README.md
```diff
@@ -161,9 +161,9 @@ Or download the [early access version](https://github.com/Flow-Launcher/Prerelea
 
 <img src="https://user-images.githubusercontent.com/6903107/207159213-662999d3-2c18-4256-b473-c417efca0069.png" width="400">
 
-- Use <kbd>F1</kbd> to toggle the preview panel.
-- Media files will be displayed as large images, otherwise a large icon and full path will be displayed.
-- Turn on preview permanently via Settings (Always Preview).
+- Use <kbd>F1</kbd> to show or hide previews (when they are optional)
```
jjw24:
```suggestion
- Use <kbd>F1</kbd> to show or hide previews
```

README.md
```diff
@@ -161,9 +161,9 @@ Or download the [early access version](https://github.com/Flow-Launcher/Prerelea
...
+- Results can preview media files as large images, show an icon and description, or use markdown with code highlighting.
+- Show optional previews by default via Settings (Always Show Optional Previews).
```
jjw24:
```suggestion
- Show previews by default via Settings (Always Preview).
```

README.md
```diff
@@ -306,7 +306,7 @@ Or download the [early access version](https://github.com/Flow-Launcher/Prerelea
 | <kbd>←</kbd>/<kbd>→</kbd>                                                 | Back to result / Open Context Menu              |
 | <kbd>Ctrl</kbd>+<kbd>O</kbd> , <kbd>Shift</kbd>+<kbd>Enter</kbd>          | Open Context Menu                               |
 | <kbd>Ctrl</kbd>+<kbd>Tab</kbd>                                            | Autocomplete                                    |
-| <kbd>F1</kbd>                                                             | Toggle Preview Panel (default and configurable) |
+| <kbd>F1</kbd>                                                             | Toggle optional previews                        |
```
jjw24:
```suggestion
| <kbd>F1</kbd>                                                             | Toggle previews                                 |
```

jjw24 (changes requested):

Flow.Launcher/Themes/BlurBlack Darker.xaml
DavidGBrett: The themes should override HorizontalScrollBarStyle etc - if nothing is set then it will fall back to the default - Base.xaml defines these that map directly to the base
```
<!--  Dynamic-resource fallbacks for horizontal scrollbar — themes can override these keys.  -->
    <Style
        x:Key="HorizontalThumbStyle"
        BasedOn="{StaticResource BaseHorizontalThumbStyle}"
        TargetType="{x:Type Thumb}" />
    <Style
        x:Key="HorizontalScrollBarStyle"
        BasedOn="{StaticResource BaseHorizontalScrollBarStyle}"
        TargetType="{x:Type ScrollBar}" />
```

This isn't a major problem since right now horizontal scrollbars only appear in the markdown preview codeblocks - so the default wouldn't look too out of place nor would they be common.

We should communicate to theme makers a suggestion to update their themes when this update comes out - but they wont be broken looking

Ideally I wanted to reuse the vertical scrollbar but I wasn't able to get that working - it did things like reverse the scroll direction or themes applying width when we should be applying height (for thickness) - its safest to let them theme it separately if they want to

Flow.Launcher/MainWindow.xaml
DavidGBrett: Just reverted this change in 91b220a8426978407ef0dfe5d2b8b555363d435a as its probably unnecessary.
The user can manually adjust this width either way anyways

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1026,13 +1005,59 @@ private static string VerifyOrSetDefaultHotkey(string hotkey, string defaultHotk
...
+
         #region Preview
 
         private static readonly int ResultAreaColumnPreviewShown = 1;
         private static readonly int ResultAreaColumnPreviewHidden = 3;
 
         private bool? _selectedItemFromQueryResults;
 
+        // User explicitly toggled the preview on or off via hotkey.
+        // null means its unset ( so will fall back to the global AlwaysPreview setting).
```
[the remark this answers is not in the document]
DavidGBrett: fixed in latest commit

DavidGBrett:
In 51eae90 I made some logical changes regarding the toggle button

The toggle now always can hide or show the current result's preview (if its a never or always result)
This avoids the scenario where the toggle button does nothing - frustrating users

What the never or always visibility types now ignore is the saved preference after toggling (ie turning **all** future previews on or off, not just the current one)
so every time the user navigates to one of those (Always or Never) they will initially behave as their type intends - but the user can then again toggle them off or on if they want.

Optional type still acts just as the original did before this pr change

Flow.Launcher.Plugin/Result.cs
```diff
@@ -364,7 +374,8 @@ public Result Clone()
                 PreviewPanel = PreviewPanel,
                 ProgressBar = ProgressBar,
                 ProgressBarColor = ProgressBarColor,
-                Preview = Preview,
+                Preview = Preview with { },
```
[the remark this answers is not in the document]
DavidGBrett: seems like a pre-existing issue no? Though I could fix it here even if its a bit out of scope

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1064,13 +1089,15 @@ public bool InternalPreviewVisible
...
-        // preview is no longer available.
+        // Tracks whether Flow opened an external preview.
+        // This can be stale if external preview program closes the preview or exit without notifying Flow.
         public bool ExternalPreviewVisible { get; private set; }
 
         private async Task ShowPreviewAsync()
         {
+            if (PreviewSelectedItem == null)
+                return;
+
```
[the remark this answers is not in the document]
DavidGBrett: If there is no selected item then we shouldn't show the panel
We never want an empty panel

Flow.Launcher.Plugin/Result.cs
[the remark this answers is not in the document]
DavidGBrett: applied in b51bc3d

DavidGBrett:
above review comment can be ignored - those legacy enum values only exist in this pr so no backwards compatibility is required

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1098,38 +1125,54 @@ when CanExternalPreviewSelectedResult(out var path):
...
-            ExternalPreviewVisible = true;
+            try
+            {
+                await PluginManager.OpenExternalPreviewAsync(path, sendFailToast).ConfigureAwait(false);
+                ExternalPreviewVisible = true;
+            }
+            catch (Exception e)
+            {
+                App.API.LogError(ClassName, $"Error opening external preview: {e}");
+            }
```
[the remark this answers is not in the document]
DavidGBrett: I reverted the added try catches to OpenExternalPreviewAsync etc
That and this problem I'm going to consider out of scope for this PR

These are pre-existing problems with External Previews
I don't want to get sidetracked dealing with all those issues for this pr

DavidGBrett:
Ignoring those above comments from coderabbit as they are out of scope for this PR

Flow.Launcher/ViewModel/MainViewModel.cs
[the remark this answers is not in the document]
DavidGBrett: Yes if you can do that and reference this comment
Including the necessity to add back logging due to some swallowing that - though maybe thats also a separate issue

Flow.Launcher/Themes/BlurBlack Darker.xaml
DavidGBrett: Of note is that we should probably update the theme builder on the website to reflect this
https://www.flowlauncher.com/theme-builder

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1026,13 +1007,59 @@ private static string VerifyOrSetDefaultHotkey(string hotkey, string defaultHotk
...
 
         private static readonly int ResultAreaColumnPreviewShown = 1;
         private static readonly int ResultAreaColumnPreviewHidden = 3;
 
         private bool? _selectedItemFromQueryResults;
 
+        // User explicitly toggled the preview on or off via hotkey.
+        // null means it's unset (so will fall back to the global AlwaysPreview setting).
+        private bool? _previewPreferenceFromToggle;
+
```
jjw24, line 1063: Comparing this to prod version, when Always Preview is on and you toggle flow from off to on, there is a very quick split second where you can see the preview switching on, leaving a short ghosting effect trail of result text prior to preview fully switched on. This regression maybe related to the preview on off behaviour logic?

jjw24 (changes requested):

Flow.Launcher/ViewModel/MainViewModel.cs
DavidGBrett: Should be fixed in 25754b196bdf8b76dd6fbadfdddbcd5fa1fc3941

Looks I was hiding it before because of some other flickering issue but it seems that was separately solved by a different rework (at least I'm not seeing any flickering with it gone now) and this slipped by.

No need for it now.

jjw24 (approved):

PR #4541, opened 2026-06-19
[FEATURE] Add opt-in Vim Mode to the search bar

namefailed:
Adds an **opt-in** Vim editing mode to the search bar. Toggle it in **Settings → General → "Enable Advanced Vim Mode"**. It is **off by default** — when disabled, the search bar behaves exactly as before.

### Modes
- **Insert** (default) — normal Flow Launcher behavior
- **Normal** — block caret; keys are interpreted as commands
- **Visual** / **Visual Line** — character-wise / whole-query selection

A small color-coded dot at the left of the box shows the current mode (hidden in Insert).

### Normal mode
- **Motions:** `h l w W b B e E 0 ^ $ %`, `f`/`F`/`t`/`T{char}` with `;`/`,` repeat, and `j`/`k` to move through the results
- **Editing:** `x X s S r{char} ~ dd cc D C Y p`, `gu`/`gU`, undo `u` / redo `Ctrl+R`
- **Operators + text objects:** `d c y gu` combined with `iw`/`aw`, `i"`/`a"`, `i'`/`a'`, `i(`/`a(`, `i[`/`a[`, `i{`/`a{` (e.g. `diw`, `ci"`, `ya(`)
- **Repeat & counts:** `.` repeats the last change; most motions/operators accept a count (`3w`, `5x`, `d3w`, `3f,`)
- **Mode switch:** `i I a A v V`

### Visual mode
Extend the selection with any motion, then operate on it: `x`/`d`, `y`, `c`/`s`, `r{char}`, `~`, `gu`/`gU`. `i`/`a` start a text object, `o` swaps the selection ends, `Esc` returns to Normal.

### Notes for reviewers
- Off by default (`EnableVimMode` defaults to `false`).
- `Ctrl`/`Alt` chords pass through untouched, so existing hotkeys are unaffected.
- The logic is split into a pure `VimMotionEngine` (caret/range math), a `VimEngine` mode state machine, and a `VimManager` integration layer. `VimManager` is `IDisposable` and is torn down in `MainWindow.Dispose`.
- Includes unit tests for the motion engine, operator ranges, text objects, and the mode state machine.
- **The project README is intentionally left untouched.** Feature documentation lives in `Flow.Launcher/VimMode/README.md`.


namefailed:
@cubic-dev-ai review

namefailed:
Thanks for the reviews. The earlier findings have been addressed in the latest commits — opt-in default (`EnableVimMode = false`), full UI restore when the setting is toggled off at runtime (IME re-enabled and caret brush restored via the settings `PropertyChanged` handler and the disabled branch in `HandlePreviewKeyDown`), Escape clearing pending state, yank/inclusive-motion handling, text objects, and counts.

On the mode indicator: it intentionally uses a small color-coded dot rather than a `NORMAL`/`VISUAL` text label, so it doesn't overlap the query text or change Flow Launcher's existing search-bar layout (non-Vim users see no difference). This is now documented in code and in the Vim mode README; happy to add a text label if you'd prefer one.

Could you re-review the current head?

@coderabbitai review
@cubic-dev-ai review

namefailed:
Latest commit addresses the open review findings:

**Fixed**
- **Insert-mode undo** — entering Insert via `i`/`I`/`a`/`A` now pushes an undo snapshot first, so `u` reverts the whole insert session. (Change-operators like `s`/`c{motion}` already snapshot via `SetText`, so they're unchanged — no double snapshot.)
- **Counted dot-repeat of `x`/`r`** — `r` now stores its length and `.` falls back to the stored count, so `3x` / `3rx` then `.` repeat the count instead of defaulting to `1`.
- **`ci"` / `di"` at end-of-query** — `TextObjectQuote` clamps the caret to the last index (WPF `CaretIndex` can equal `text.Length`), so a quoted region that ends the query still matches. New unit-test cases cover it.

**Not changed (already handled / intentional)**
- *Restore IME + caret when Vim mode is toggled off* — already handled: `OnSettingsPropertyChanged` switches to Insert (re-enabling IME and clearing the transparent `CaretBrush`), and the `!EnableVimMode` guard in `HandlePreviewKeyDown` does the same on the next keystroke.
- *Mode indicator shows no text* — intentional: the indicator is a small color-coded dot (documented in `Flow.Launcher/VimMode/README.md`) so it never overlaps the query text or alters the search-bar layout for users who don't enable Vim mode.

All 35 Vim unit tests pass; AppVeyor is green.

jjw24:
Hi there @namefailed, hope all is well. Thank you for putting forward the PR to add Vim mode to flow's search bar and for sorting out the AI review comments. This is an interesting feature addition.

I have created an [issue](https://github.com/Flow-Launcher/Flow.Launcher/issues/4569) to measure how much interest there is in this feature. This is the first time a Vim mode has been requested, and we want to see if there is enough interest to proceed with adding it.

In the meantime, let's keep this PR open. Feel free to continue tweaking the changes, and let's see if there is significant user interest in this feature before merging.

PR #4609, opened 2026-08-07
Feature: History context menu with delete & info item & Refactor context menu logic

### LINKED ISSUE #4626: BUG: Returning from history menu sets selected result to first one after brief delay
### Checks

- [x] I am using the prerelease version of Flow Launcher.

### Problem Description

When you toggle the history menu it takes a second and then selects the first result - meaning you lose your position in the results list.

### To Reproduce

1. Query and scroll down to a result that is not the first one
2. Press ctrl + H to open the history view/menu
3. Press ctrl + H again (or esc) and wait for a second or two

### Screenshots

_No response_

### Flow Launcher Version

2.1.3

### Windows Build Number

Version 10.0.26200.8973

### Error Log

_No response_

Jack251970:
image

Relates to https://github.com/Flow-Launcher/Flow.Launcher/pull/4042

---
## Summary by cubic
Adds a context menu for history items with Delete and Info, and refactors the context menu flow to be fully async and toggleable. Previously, history items had no context menu and Right Arrow only opened the menu for query results; now Right Arrow works for both, options are fuzzy-filtered by your query, window activation respects visibility, and returning restores your query, preview, and selection.

**Summary of changes**
- Replaced `LoadContextMenuCommand` with `ToggleContextMenuCommand`, so the same command opens and closes the menu.
- Esc closes the context menu and restores the previous results view, preview, query text, and reselects the targeted item, falling back to the prior index if the history list rebuilt.
- Right Arrow opens the context menu when a result or history item is selected and the caret is at the end.
- Context menu acts on the selected item, with history items handled via `ContextData`; returning restores the preview, and home results refresh after deletion.
- Context menu entries are filtered by the fuzzy-search algorithm using the current query.
- Flow is fully async: returning from the menu, changing query text, updating preview, and delete actions are awaited.
- Consolidated context menu state (source, query, selection) into a single `ResultsQuerySelection` record.
- Window activation checks visibility to avoid activating or focusing when the window is hidden.
- Added: Delete and History Info options for history items, `History.Remove(...)` to delete a single entry or all matching entries (LastOpened style) with storage persisted, a delete icon asset with `Constant.DeleteIcon`, and title-case localization strings for the history labels.
- Removed: the guard that blocked context menus for history results.
- Memory impact: minimal, one record for context menu state and a single icon asset.
- Security: no new inputs or elevated actions, only local history mutation.
- Tests: added `Flow.Launcher.Test/QueryHistoryTest.cs` for single-entry and multi-entry removal across matching results.

**Release Note**
You can open a menu on history items to delete entries or view info, with options filtered by your query and your previous view restored after closing.

<sup>Written for commit bed0e19c6d38566b54f7383339e43a9024201930. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Flow-Launcher/Flow.Launcher/pull/4609?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

# TEST

- Returning logic of context menu.
- Deletion function of history items.


Jack251970:
@codex Review it

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -413,23 +417,49 @@ private async Task LoadContextMenuAsync()
...
+            SelectedResults = ContextMenu;
+        }
+
+        private async Task ReturnFromContextMenuAsync()
+        {
+            var source = _contextMenuSource ?? Results;
+            var queryText = _queryTextBeforeContextMenu;
+
+            // Return to the previous results view and restore the query text
+            SelectedResults = source;
```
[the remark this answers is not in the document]
Jack251970: 👍

Jack251970:
@codex Review it

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -413,22 +418,82 @@ private async Task LoadContextMenuAsync()
...
+                    && result1.SubTitle == result2.SubTitle
+                    && result1.PluginID == result2.PluginID;
+            }
             else
             {
-                SelectedResults = Results;
-                PreviewSelectedItem = Results.SelectedItem;
-                await UpdatePreviewAsync();
+                return result1.RecordKey == result2.RecordKey
+                    && result1.PluginID == result2.PluginID;
```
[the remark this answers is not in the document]
Jack251970: No need to check query here since the query should be the same for one history context menu.

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -1867,6 +1958,51 @@ private Result ContextMenuTopMost(Result result)
...
+            return new Result
+            {
+                Title = Settings.HistoryStyle == HistoryStyle.Query ? Localize.queryHistory() : Localize.executedHistory(),
+                IcoPath = Constant.HistoryIcon,
+                Glyph = new GlyphInfo(FontFamily: "/Resources/#Segoe Fluent Icons", Glyph: "\uE81C"),
+                PluginDirectory = Constant.ProgramDirectory,
+                Action = _ => false,
+                OriginQuery = historyItem.OriginQuery
+            };
+        }
```
DavidGBrett, line 2004: What is the intention of this context menu result?
It looks to do nothing to me...
Perhaps we could change the name or add comments to make this clearer

Flow.Launcher/ViewModel/MainViewModel.cs
DavidGBrett, line 404: You already separated ReturnFromContextMenuAsync out from LoadContextMenuAsync so I think it would be cleaner to also make a separate method for entering the context menu, 
LoadContextMenuAsync can just hold the toggle logic - maybe renamed to ToggleContextMenuAsync for clarity

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -41,8 +41,13 @@ public partial class MainViewModel : BaseModel, ISavable, IDisposable, IResultUp
         private readonly ConcurrentDictionary<Guid, Query> _progressQueryDict = new(); // Used for QueryResultAsync
         private Query _updateQuery; // Used for ResultsUpdated
         private string _queryTextBeforeLeaveResults;
+        private string _queryTextBeforeContextMenu;
         private string _ignoredQueryText; // Used to ignore query text change when switching between context menu and query results
 
+        private ResultsViewModel _contextMenuSource;
+        private Result _contextMenuTarget;
+        private int _contextMenuTargetIndex = -1;
```
DavidGBrett, line 49: These seem very tightly coupled so perhaps we can make a new record type to hold these together.
That would make their relationship clearer and also make it easier to reuse it e.g. for returning from the history view - while testing this I found a bug #4626 that we may be able to fix by doing that.

The previous query text could also be stored here

ResultsQuerySelection or something like that
`{ ResultsViewModel source, string query, Result selection, int selectionIndex }`

Either way I think the names could be improved since _contextMenuTarget is the same selected result we use elsewhere right?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -413,22 +418,82 @@ private async Task LoadContextMenuAsync()
...
+            // Regular results need a valid plugin ID. History results use Flow's built-in context menu and
+            // can appear in either the dedicated history view or the home-page result list.
+            var selected = SelectedResults.SelectedItem?.Result;
+            if (selected == null) return;
+            var isHistoryResult = selected is LastOpenedHistoryResult
+                && selected.ContextData is LastOpenedHistoryResult;
+            if (!isHistoryResult && (!QueryResultsSelected() || string.IsNullOrEmpty(selected.PluginID))) return;
+
+            _contextMenuSource = SelectedResults;
+            _contextMenuTarget = selected;
```
DavidGBrett, line 438: Perhaps it would be cleaner to set _contextMenuTarget first and use that instead of a new selected variable?

Flow.Launcher/ViewModel/MainViewModel.cs
```diff
@@ -41,8 +41,13 @@ public partial class MainViewModel : BaseModel, ISavable, IDisposable, IResultUp
         private readonly ConcurrentDictionary<Guid, Query> _progressQueryDict = new(); // Used for QueryResultAsync
         private Query _updateQuery; // Used for ResultsUpdated
         private string _queryTextBeforeLeaveResults;
+        private string _queryTextBeforeContextMenu;
```
DavidGBrett, line 44: I was a bit confused at first why we needed both of these
I see that its because of the extra nesting when opening a context menu from the history menu right?

Some kind of stack might better match what we are doing but could be overkill for just two layers right now

Perhaps just add comments after the variables to explain why they are separately needed?

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > What is the intention of this context menu result? It looks to do nothing to me... Perhaps we could change the name or add comments to make this clearer

Here I want to add the result info items as the plugin info items of other items. I know this could be little useless, but I want to make this behaviour consistent with the items from other plugins.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > You already separated ReturnFromContextMenuAsync out from LoadContextMenuAsync so I think it would be cleaner to also make a separate method for entering the context menu, LoadContextMenuAsync can just hold the toggle logic - maybe renamed to ToggleContextMenuAsync for clarity

Good idea! Finished in https://github.com/Flow-Launcher/Flow.Launcher/pull/4609/changes/dc5a865b29888f45ebf4ed4a305cb89fc2f2682e

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > These seem very tightly coupled so perhaps we can make a new record type to hold these together. That would make their relationship clearer and also make it easier to reuse it e.g. for returning from the history view - while testing this I found a bug #4626 that we may be able to fix by doing that.
> 
> The previous query text could also be stored here
> 
> ResultsQuerySelection or something like that `{ ResultsViewModel source, string query, Result selection, int selectionIndex }`
> 
> Either way I think the names could be improved since _contextMenuTarget is the same selected result we use elsewhere right?

Thanks for your suggestion! I have added `ResultsQuerySelection` to record this in https://github.com/Flow-Launcher/Flow.Launcher/pull/4609/changes/1cc225635e1b24a7ae16b109930e0c2d37f10675.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > I was a bit confused at first why we needed both of these I see that its because of the extra nesting when opening a context menu from the history menu right?
> 
> Some kind of stack might better match what we are doing but could be overkill for just two layers right now
> 
> Perhaps just add comments after the variables to explain why they are separately needed?

Yeah! I have removed `_queryTextBeforeContextMenu` added `ResultsQuerySelection` for clearlity in https://github.com/Flow-Launcher/Flow.Launcher/commit/1cc225635e1b24a7ae16b109930e0c2d37f10675.

Flow.Launcher/ViewModel/MainViewModel.cs
Jack251970: > Perhaps it would be cleaner to set _contextMenuTarget first and use that instead of a new selected variable?

I have refactored it in https://github.com/Flow-Launcher/Flow.Launcher/pull/4609/changes/1cc225635e1b24a7ae16b109930e0c2d37f10675. And I think this comment is no longer applicable.

Jack251970:
@DavidGBrett Hi, thanks for your comments! I have polished my codes to meet your suggestions and they works for me. Could you please check if they meet your ideas and work correctly? Also please help me confirm if #4626 is resolved. I cannot reproduce this issue on my side.

DavidGBrett (approved):
Thank you for your work on this!
Looks clearer now and I can confirm it works!

That issue looks to be resolved, though I thought it was pre-existing - at least I can't replicate it either now

Flow.Launcher/ViewModel/MainViewModel.cs
DavidGBrett: Ok I see,
I added a comment above to make it's purpose a little clearer.

Issue #3812, opened 2025-07-09
BUG: log spam from COM exception


dcog989:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I am using the latest version of Flow Launcher.

- [ ] I am using the prerelease version of Flow Launcher.

### Problem Description

My logs are being spammed with:

```
Exception full name: System.Runtime.InteropServices.COMException
Error status: UNKNOWN
Class name: IShellLinkW
Calling method: retrieveTargetPath
Program path: C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Accessories\Windows Media Player Legacy.lnk
InnerException number: 1
Exception message: Error HRESULT E_FAIL has been returned from a call to a COM component.
Exception error type: HResult -2147467259
Exception thrown in called method: Void GetDescription(Windows.Win32.Foundation.PWSTR, Int32)
Possible interpretation of the error: Error caused likely due to trying to get the description of the program
Possible resolution: Not yet known
01:44:34.6510+01:00 - ERROR -  - ------------- END Flow.Launcher.Plugin.Program exception -------------
01:44:43.9120+01:00 - ERROR -  - ------------- BEGIN Flow.Launcher.Plugin.Program exception -------------
01:44:43.9120+01:00 - ERROR -  -
```

The AI gave me this update to ShellLinkHelper.cs as a fix:

```
using System;
using System.Runtime.InteropServices;
using System.Runtime.InteropServices.ComTypes;
using Flow.Launcher.Plugin.Program.Logger;
using Windows.Win32.Foundation;
using Windows.Win32.UI.Shell;
using Windows.Win32.Storage.FileSystem;

namespace Flow.Launcher.Plugin.Program.Programs
{
    public class ShellLinkHelper
    {
        // Reference : http://www.pinvoke.net/default.aspx/Interfaces.IShellLinkW
        [ComImport(), Guid("00021401-0000-0000-C000-000000000046")]
        public class ShellLink
        {
        }

        // To initialize the app description
        public string description = string.Empty;
        public string arguments = string.Empty;

        // Retrieve the target path using Shell Link
        public unsafe string retrieveTargetPath(string path)
        {
            var link = new ShellLink();
            const int STGM_READ = 0;
            ((IPersistFile)link).Load(path, STGM_READ);
            var hwnd = new HWND(IntPtr.Zero);
            // Use SLR_NO_UI to avoid showing any UI during resolution, like Problem with Shortcut dialogs
[...]

PR #4617, opened 2026-08-15
Lower program plugin shortcut description E_FAIL logs to debug level

### LINKED ISSUE #4616: BUG: Program plugin logs an ERROR block per CLSID shortcut on every index (retrieveDescription), even with EnableDescription off
### Checks

### Problem Description

`ShellLinkReader.retrieveDescription` throws `COMException` (`E_FAIL`, HResult `-2147467259`) for shortcuts whose target is a shell-namespace/CLSID item rather than a filesystem path. Each failure is written to the log as a 14-line `ERROR` block, and because it happens on every index pass the log fills with repeats of the same handful of shortcuts.

On my machine (Windows 11 build 26200.9168, Flow Launcher 2.1.3, x64) exactly two shortcuts trigger it, 240 times across my logs:

```
%APPDATA%\Microsoft\Windows\Start Menu\Programs\System Tools\Control Panel.lnk  ->  ::{5399E694-6CE5-4D6C-8FCE-1D8870FDCBA0}
%APPDATA%\Microsoft\Windows\Start Menu\Programs\System Tools\Run.lnk            ->  ::{2559A1F3-21D7-11D4-BDAF-00C04F60B9F0}
```

Both are valid, working shortcuts. They simply have no description and no filesystem target (`IShellLink::GetPath` returns an empty string for them), so `GetDescription` returns `E_FAIL`.

Logged block, repeated per shortcut per index pass:

```
------------- BEGIN Flow.Launcher.Plugin.Program exception -------------

Exception full name: System.Runtime.InteropServices.COMException
Error status: UNKNOWN
Class name: ShellLinkReader
Calling method: retrieveDescription
Program path: C:\Users\<user>\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\System Tools\Control Panel.lnk
InnerException number: 1
Exception message: Unexpected HRESULT has been returned from a call to a COM component.
Exception error type: HResult -2147467259
Exception thrown in called method: Void GetDescription(Windows.Win32.Foundation.PWSTR, Int32)
Possible interpretation of the error: Error caused likely due to trying to get the description of the program
Possible resolution: Not yet known
------------- END Flow.Launcher.Plugin.Program exception -------------
```

### Two things that look wrong

**1. An expected condition is logged at ERROR level.** The `catch` in `retrieveDescription` already carries a comment noting that a particular shortcut "always cause exception" — so this outcome is known and anticipated. A shortcut having no description is normal, not an error; `Debug`/`Info` (or a one-line message rather than the full exception block) seems more appropriate.

**2. The description is fetched even when the user has descriptions turned off.** My Program plugin settings contain `"EnableDescription": false`, but `ShellLinkReader.Read` calls it unconditionally:

```csharp
var target = retrieveTargetPath((IShellLinkW)link, path);
var description = retrieveDescription((IShellLinkW)link, path);
var arguments = retrieveArguments((IPropertyStore)link, path);
```

When `EnableDescription` is `false` the returned value is discarded downstream, so the COM call — and on these shortcuts its exception and log block — is pure overhead on every index. Skipping the call when the setting is off would avoid the cost entirely for users who have it disabled.

### Related

#3812 fixed this same log spam for `retrieveTargetPath`. The `retrieveDescription` path still has it.

DavidGBrett:
Resolves #4616 and also readdresses #3812

Added a LogDebug to the program plugin's ProgramLogger and used it specifically for E_FAIL hresults from description reads

There doesn't seem to be any way to distinguish between a shortcut missing a description and an actual failure here, but it is expected in many cases, so as a compromise I downgraded this log to debug level

I decided against removing the call entirely when EnableDescription is false as that would require reindexing when its toggled on.

---
## Summary by cubic
**Summary of changes**

Downgrades logging for expected shortcut description failures from error to debug. This reduces error log noise when Windows shortcuts lack a description.

- Changed behavior: When `IShellLinkW.GetDescription` returns `E_FAIL`, `ShellLinkReader.retrieveDescription` logs via `ProgramLogger.LogDebug` and returns an empty string. Previously it called `LogException` and produced error logs.
- New logic: Added `ProgramLogger.LogDebug` (synchronized) to emit debug messages with class, method, and path; added `E_FAIL` constant in `ShellLinkReader`.
- Removed logic: None. The description read remains even when description support is disabled to avoid forcing a reindex when the setting is toggled later.
- Memory impact: None material. The change only alters log level and adds a small debug-only code path.
- Security risks: None. This is a logging-only change; no changes to permissions, data handling, or external interfaces.
- Unit tests: No new tests added; behavior change is limited to logging level for `E_FAIL`.

**Release Note**

Reduces error log noise by treating missing shortcut descriptions as debug messages instead of errors.

<sup>Written for commit 229b586eace12fb8fc90dd616b0823a1b1d7deed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Flow-Launcher/Flow.Launcher/pull/4617?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Plugins/Flow.Launcher.Plugin.Program/Logger/ProgramLogger.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Logger/ProgramLogger.cs
@@ -86,6 +86,16 @@ namespace Flow.Launcher.Plugin.Program.Logger
             LogException(classname, callingMethodName, loadingProgramPath, interpretationMessage, e);
         }
 
+        /// <summary>
+        /// Logs a debug message
+        /// </summary>
+        [MethodImpl(MethodImplOptions.Synchronized)]
+        internal static void LogDebug(string classname, string callingMethodName, string loadingProgramPath, string message)
+        {
+            var logger = LogManager.GetLogger("");
+            logger.Debug($"{classname}.{callingMethodName}: {message} ({loadingProgramPath})");
+        }
+
         private static bool IsKnownWinProgramError(Exception e, string callingMethodName)
         {
             if (e.TargetSite?.Name == "GetDescription" && callingMethodName == "LnkProgram")
--- a/Plugins/Flow.Launcher.Plugin.Program/Programs/ShellLinkReader.cs
+++ b/Plugins/Flow.Launcher.Plugin.Program/Programs/ShellLinkReader.cs
@@ -15,6 +15,8 @@ namespace Flow.Launcher.Plugin.Program.Programs
 {
     public static class ShellLinkReader
     {
+        private const int E_FAIL = unchecked((int)0x80004005); // HResult returned by IShellLinkW.GetDescription
+
         // Retrieves the target path, arguments, and description from a shell link
         public static ShellLinkReadResult Read(string path)
         {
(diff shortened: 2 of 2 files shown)


Jack251970 (approved):
LGTM!

PR #4635, opened 2026-08-29
Fix theme reset window size

### LINKED ISSUE #4619: BUG: Theme reset doesn't actually reset Flow search bar size
### Checks

### Problem Description

Reset button below doesn't rest the width/height of Flow. Tested on Win10, Flow 2.1.3 using the Circle theme

### To Reproduce

1. ... Manually change size of search bar (setting for  Fixed Window Size needs to be off
2. ... Open Settings - Theme - Edit pen icon and hit Reset
3. ... Activate Flow and search bar size is the same and hasn't reset

Restarting Flow doesn't help

### Screenshots

Image

### Flow Launcher Version

2.1.3

### Windows Build Number

Microsoft Windows [Version 10.0.19045.7663]

pttydou:
Fixes #4619.

Theme reset now restores the persisted search window width and maximum result count to their shared defaults, allowing the existing property-change listeners to update the active window.

Tests:
- `dotnet build Flow.Launcher.sln -c Release --no-restore --verbosity quiet`
- `dotnet test Flow.Launcher.Test\Flow.Launcher.Test.csproj -c Release --no-build --no-restore --verbosity quiet`

---
## Summary by cubic
**Summary of changes**

Theme reset now restores the persisted search window width and maximum result count to their shared defaults and refreshes the active window even when the values are already at those defaults.

- Replaced inline default values in `Settings.WindowSize` and `Settings.MaxResultsToShow` with `DefaultWindowSize` and `DefaultMaxResultsToShow` constants.
- Added `Settings.ResetSearchWindowSize()`, which resets both settings to the defaults and fires `OnPropertyChanged` for any value that is already at its default.
- Changed `SettingsPaneThemeViewModel.Reset()` to call `Settings.ResetSearchWindowSize()`; previously the user's customized values were kept.
- No unit tests cover this change; the previously added theme reset unit test was removed.
- No meaningful memory usage impact; only constants and a method were added.
- No security risk introduced.

**Release Note**

Resetting a theme now also restores the search window size and result count to their default values.

<sup>Written for commit 01212ab5ba2229bcd36fe1a8382e4809354f5918. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Flow-Launcher/Flow.Launcher/pull/4635?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


VictoriousRaptor:
I suggest we remove the unit test in this PR. The change is trivial. Any idea?

VictoriousRaptor:
Looks like it's not working on my side. Does it work on your side? @pttydou

pttydou:
Thanks for checking, and sorry I haven't been able to investigate this yet. My computer currently only boots in Safe Mode, so I can't run the normal development and test environment reliably. Repairing it may take some time. Once the machine is working normally again, I'll reproduce the behavior on my side, revisit the test, and follow up here.

pttydou:
I removed the unit test as suggested and reran the full `Flow.Launcher.Test` suite: all 434 tests pass.

I also traced the production update path: resetting `WindowSize` and `MaxResultsToShow` raises the settings notifications consumed by `MainViewModel` and `ResultsViewModel`. I have not been able to reproduce the end-to-end UI mismatch from that code path, though. Could you confirm whether the window width, the result-list height/count, or both remain unchanged on your side, and whether **Fixed Window Size** is off when testing?

VictoriousRaptor:
> I removed the unit test as suggested and reran the full `Flow.Launcher.Test` suite: all 434 tests pass.
> 
> I also traced the production update path: resetting `WindowSize` and `MaxResultsToShow` raises the settings notifications consumed by `MainViewModel` and `ResultsViewModel`. I have not been able to reproduce the end-to-end UI mismatch from that code path, though. Could you confirm whether the window width, the result-list height/count, or both remain unchanged on your side, and whether **Fixed Window Size** is off when testing?

iirc fixed window size is on in my test. I just widen the window and hit reset, open flow again and the width is unchanged. the other options should be default. busy today so can't test it again.

pttydou:
Thanks, that detail exposed the missing edge case.

With **Fixed Window Size** enabled, the window's actual WPF width can differ from the stored `WindowSize`, while the stored value is already the default (`580`). The previous reset assigned `580` again, so no `PropertyChanged` notification was raised and the visible window kept its widened width when Flow was reopened.

I updated the reset path to explicitly re-notify both `WindowSize` and `MaxResultsToShow` when they are already at their defaults. This forces the active window and results view to reapply the default dimensions. I kept the unit test out of the PR as requested.

Validation on commit `01212ab5b`:
- clean Release build passes
- a local-only regression reproduces the no-notification case before the fix and passes after it
- 433/434 existing tests pass together; the remaining unrelated `ShellThumbnailFailure_Directory_ReturnsDefaultFolderImageAsync` test passes when rerun alone

No rush to retest today—thanks for providing the exact state.

